PR #2, opened 2022-06-21
DOC: Update KBS attestation protocol

jialez0:
1. Fixed formatting confusion caused by JSON indentation errors.
2. Modify the `symkey-length` filed in `Response` to `initialization-vector`, the key length can be specified in the 
   `algorithm` field, but in many symmetric encryption algorithms, initialization vector is required.
    
Initialization vector is the input to a cryptographic primitive being used to provide the initial state. For more infomation about initialization vector, please refer to https://en.wikipedia.org/wiki/Initialization_vector

Signed-off-by: Jiale Zhang <zhangjiale@linux.alibaba.com>


docs/kbs_attestation_protocol.md
```diff
@@ -183,7 +183,8 @@ If KBS verified the token or the attestation evidence successfully, it will retu
 	    /* Symmetric key and algorithm information for encrypting `output`. */
 	    "crypto-annotation": {
             "algorithm": "",
-            "symkey-length": "",
+			/* The symmetric encryption IV, if the algorithm used does not need IV, this field left blank */
+            "iv": "",
```
jodh-intel, line 187: Hi @jialez0 - Please can you provide further detail in the commit for this change. For example:

- What is "iv"? (What does it stand for? Is it https://en.wikipedia.org/wiki/Initialization_vector?)
- Is "iv" an exact replacement for "symkey-length"? If the URL above is correct, I think the answer is no since one is a length, and the other appears to be a blob of random data, like a salt?
- iff "iv" == "Initialization vector", why can't we call the member "initialization-vector" (or "initialization-vector-blog or similar). That would be much clearer and consistent with all the other members that are human-readable.
- Is there a canonical reference URL to explain this change (official doc(s) or an GitHub issue?

Also, please can you add an entry for "IV" to https://github.com/confidential-containers/documentation/wiki/Acronyms.

fitzthum (approved):
LGTM

docs/kbs_attestation_protocol.md
jialez0: Sorry for the confusing, I have changed the name `iv` to `initialization-vector` and we don't need to use the abbreviation anymore, That does cause some confusion.

sameo (approved):

Issue #7, opened 2022-08-23
Modify documentation to track "reference-kbs"


c3d:
@slp (@virtee) has begun a [reference implementation](https://github.com/slp/reference-kbs) of the KBS, which as far as I can tell has interesting extensions, notably a way to [register workloads](https://github.com/slp/reference-kbs/blob/main/src/main.rs#L147) which is quite useful in practice.

What is learned from this implementation should probably would need to be integrated in the documentation.

jialez0:
Looks great!  @slp 's word is a significant progress in the KBS project, maybe some of them can become the PR of KBS repository (It has been silent for a long time:))? We can jointly build KBS and make it more practical.

By the way, @sameo mentioned in #3  that will update the KBS protocol document, how is it going?

jialez0:
@slp @sameo @c3d I think workload ID can be used as a special type of measurement value name, which is compatible with the AS verification scheme we currently envision. It should not be passed in the Request, but should be passed in the `tee evidence` in the evidence payload. It has the same status as the original measurement value name of component granularity, and does not need to change our existing protocol to support it.

For convenience of explanation, we assume that for a certain attester/verifier driver, the format of `tee evidence` in the evidence payload is as follows:

```json
{
  "tee-pubkey": "$tee-pubkey",
  "tee-evidence": {
    "measurements": {},
    "tee-report": "$tee-report"
  }
}
```

- If the user chooses not to use workload ID, the evidence payload is as follows:

```json
{
  "tee-pubkey": "$tee-pubkey",
  "tee-evidence": {
    "measurements": {
      "kernel":"1123abc",
      "firmware":"22334bca",
      "agent":"2211cba",
    },
    "tee-report": "$tee-report"
  }
}
```

In this case, AS will obtain and cache all the latest trusted mesurement values of each component of the confidential containers from the software supply chain security architecture for CC (we are planning to build). At this time, the reference value list in AS is as follows:

```json
{
  "kernel":["1123abc", "1142bca"],
  "firmware":["22334bca", "4456bac", "3344ddf"],
  "agent":"["2211cba", "1122gfd"],
  ...
}
```

The user can configure AS to make different decisions according to the matching of measurement values of each component, so as to give different permissions to the verified object.

- If the user chooses to use workload ID, the evidence payload is as follows:

```json
{
  "tee-pubkey": "$tee-pubkey",
  "tee-evidence": {
    "measurements": {
      "$(workload_id)": "$(workload_hash)"
    },
[...]

PR #9, opened 2022-08-29
docs: Attestation protocol updates

sameo:
Signed-off-by: Samuel Ortiz <sameo@rivosinc.com>


sameo:
cc @slp

jialez0 (commented):
@sameo Thanks! A few comments...

docs/kbs_attestation_protocol.md
```diff
@@ -274,3 +274,126 @@ We define the error return payload as follows:
...
+   JSON payload. This endpoint is for the attester to initiate the attestation
+   protocol and authenticate itself against the KBS. The KBS reply HTTP response
+   is composed of:
+   a. A `Set-Cookie` header set to `kbs-session-id=<session>`. The KBS tracks
+   the attester and its attestation results with that cookie.
+   b. An attestation challenge for the attester to take. This is the content of
+   the response, set to a [KBS Challenge](#challenge) JSON payload.
+2. `/kbs/v0/attest` only accepts `POST` requests which body is a [KBS Attestation](#attestation)
+   JSON payload and the header contains a `Cookie` set to the value received in
+   step 2.a. This is how the attester replies to attestation challenge received
```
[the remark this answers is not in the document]
sameo: Thanks, fixed.

docs/kbs_attestation_protocol.md
```diff
@@ -274,3 +274,126 @@ We define the error return payload as follows:
...
+protected resources from the KBS.
+
+The KBS implementation keeps track of attestation results and binds them to a
+cookie identifier. During the second phase, the KBS can then decide if a
+specific resource could be released to a given attester, by mapping the cookie
+identifier the attester includes in its resource request message to its
+attestation results.
+
+To request a protected resource from the KBS, the attester sends a `GET` request
+to a resource specific endpoint. If the attester is allowed to access the
```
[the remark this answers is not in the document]
sameo: Yes, I added that section now.
Note that I'm planning to also add an OpenAPI definition to this PR. Stay tuned ;-)

sameo:
@slp @jialez0 @jiazhang0 The document update is ready to review. I'm planning to add a OpenAPI yaml document to that PR, it would be great to get your review on the HTTP integration part first.

docs/kbs_attestation_protocol.md
```diff
@@ -108,113 +146,303 @@ attestation challenge to KBC. The payload format is as follows:
...
 
 ```json
 {
     "evidence": {
-        /* The nonce in the Challenge; its hash needs to be included in HW-TEE evidence and signed by HW-TEE hardware. */
+        /*
+         * The nonce in the Challenge; its hash needs to be included in HW-TEE
+         * evidence and signed by HW-TEE hardware.
+         */
         "nonce": "",
```
jodh-intel, line 169: Is it worth changing lines like this to something like:

```suggestion
        "nonce": "$nonce",
```

... since currently if you don't read the comments and only look at the JSON, it suggests this field is / should be blank.

docs/kbs_attestation_protocol.md
```diff
@@ -108,113 +146,303 @@ attestation challenge to KBC. The payload format is as follows:
...
 
-The [JSON web token](https://jwt.io/) standard is adopted for the token. The format is as follows:
+Before being able to request a protected resource (a key, a token, etc), an
+attester must authenticate itself against the remote KBS. Only after succesfully
+authenticated itself, it can request those resources.
 
-(Note that the `/*...*/` comments are not valid in JSON web token, and must not be used in real token.)
+The authentication service is provided by the KBS through two endpoints:
+
+1. `/kbs/v0/auth` only accepts `POST` requests which body is a [KBS Request](#request)
```
jodh-intel, line 308:
```suggestion
1. `/kbs/v0/auth` only accepts `POST` requests whose body is a [KBS Request](#request)
```

docs/kbs_attestation_protocol.md
```diff
@@ -108,113 +146,303 @@ attestation challenge to KBC. The payload format is as follows:
...
+   JSON payload. This endpoint is for the attester to initiate the attestation
+   protocol and authenticate itself against the KBS. The KBS reply HTTP response
+   is composed of:
+   a. A `Set-Cookie` header set to `kbs-session-id=<session>`. The KBS tracks
+   the attester and its attestation results with that cookie.
+   b. An attestation challenge for the attester to take. This is the content of
+   the response, set to a [KBS Challenge](#challenge) JSON payload.
+2. `/kbs/v0/attest` only accepts `POST` requests which body is a [KBS Attestation](#attestation)
```
jodh-intel, line 316:
```suggestion
2. `/kbs/v0/attest` only accepts `POST` requests whose body is a [KBS Attestation](#attestation)
```

docs/kbs_attestation_protocol.md
```diff
@@ -108,113 +146,303 @@ attestation challenge to KBC. The payload format is as follows:
...
+   not.
+
+```
+┌──────────┐                                          ┌─────┐
+│ Attester │                                          │ KBS │
+└────┳─────┘                                          └──┳──┘
+     ┃                 ┌─────────────────┐               ┃
+     ┃─────────────────┼POST /kbs/v0/auth┼──────────────▶┃
+     ┃                 ├─────────────────┤               ┃
+     ┃                 │Header: N/A      │               ┃
```
jodh-intel, line 330: When you say "n/a" do you mean literally empty, or just nothing of interest to show here?
For example, are `Content-Length` and `Content-Type` required? What about a date? If specified, is that validated (what if the date was the epoch or a value far in the future)?

docs/kbs_attestation_protocol.md
```diff
@@ -108,113 +146,303 @@ attestation challenge to KBC. The payload format is as follows:
...
+   in step 1.b. The KBS replies to that request with an empty HTTP response
+   (no content), which HTTP status indicates if the attestation was succesful or
+   not.
+
+```
+┌──────────┐                                          ┌─────┐
+│ Attester │                                          │ KBS │
+└────┳─────┘                                          └──┳──┘
+     ┃                 ┌─────────────────┐               ┃
+     ┃─────────────────┼POST /kbs/v0/auth┼──────────────▶┃
```
jodh-intel, line 328: Micro-nit: The horizontal lines with arrow heads don't seem to render as designed as they seem to overflow slightly, making the vertical lines bulge. Maybe you need to insert some sort of unicode "half space" char on the other lines to make them line up?

docs/kbs_attestation_protocol.md
```diff
@@ -108,113 +146,303 @@ attestation challenge to KBC. The payload format is as follows:
...
+   the response, set to a [KBS Challenge](#challenge) JSON payload.
+2. `/kbs/v0/attest` only accepts `POST` requests which body is a [KBS Attestation](#attestation)
+   JSON payload and the header contains a `Cookie` set to the value received in
+   step 1.a. This is how the attester replies to attestation challenge received
+   in step 1.b. The KBS replies to that request with an empty HTTP response
+   (no content), which HTTP status indicates if the attestation was succesful or
+   not.
+
+```
+┌──────────┐                                          ┌─────┐
```
jodh-intel, line 324: I :heart:  these diagrams! :smile:

docs/kbs_attestation_protocol.md
```diff
@@ -108,113 +146,303 @@ attestation challenge to KBC. The payload format is as follows:
...
+    ┃            ┌─────────────────────────┐            ┃
+    ┃────────────┼GET /kbs/v0/<resource_id>┼───────────▶┃
+    ┃      ┌─────┴─────────────────────────┴────┐       ┃
+    ┃      │Header: "Cookie kbs-session-id=123" │       ┃
+    ┃      │Body: N/A                           │       ┃
+    ┃      └────────────────────────────────────┘       ┃
+    ┃                    ┌────────┐                     ┃
+    ┃◀───────────────────┼Response┼─────────────────────┃
+    ┃     ┌──────────────┴────────┴───────────────┐     ┃
+    ┃     │Header: "Set-Cookie kbs-session-id=123"│     ┃
```
jodh-intel, line 395: How does the KBS expire these session IDs to avoid having to cache them forever (aka a potential slow memory leak)? There doesn't appear to be a final call that the Attester makes. But even if there was that wouldn't be sufficient to deal with a buggy / nefarious attester. I'm assuming some sort of timeout, but where would that need to be specified / documented?

docs/kbs_attestation_protocol.md
```diff
@@ -244,50 +475,21 @@ The [JSON web token](https://jwt.io/) standard is adopted for the token. The for
...
-1. KBC sends `GET /kbs/key/key_id` request. The request payload is the `Request` message in the "RCAR" semantics.
-2. If no token is provided in the request payload or the token is invalid, KBS returns `HTTP/1.1 200 OK` response to KBC, 
-but the payload of the response is the `Challenge` message in the "RCAR" semantics. 
-If a valid token is provided in the request payload, KBS directly jumps to step 4.
-3. KBC sends `GET /kbs/key/key_id` request to KBS again, but the payload of this request is the `Attestation` message in the "RCAR" semantics.
-4. KBS returns an `HTTP/1.1 200 OK` response to KBC and the payload is the `Response` message in the "RCAR" semantics.
+The payload part is divided into official claim set and custom claim set. In the
+official claim set, we select three fields, `exp`, `iat`' and `iss`, which
+respectively declare the expiration time, issuing time and issuer (KBS URL
+address) of the token. The custom field in the payload contains the HW-TEE's
```
jodh-intel, line 484: I wonder if a table would make this clearer:

| field | Description | Type | Example |
|-|-|-|-|
| `exp` | Expiration time | seconds and nanoseconds since epoch? | ? |
| `iat` | Issuing time | seconds and nanoseconds since epoch? | ? |
| `iss` | Issuer | KBS URL | ? |

Aside: Are these fields validated? And if so how? I've mentioned this before in various other places, but how are the timestamps validated, for example, to ensure they have not been spoofed? Does the overally system rely on Secure NTP for example?

docs/kbs_attestation_protocol.md
```diff
@@ -244,50 +475,21 @@ The [JSON web token](https://jwt.io/) standard is adopted for the token. The for
...
+public key sent by KBC along with the attestation evidence, which is valid
+within the validity period of the token. When KBC uses the token to request
```
jodh-intel, line 486: What is a "reasonable" or normal validity period?

jodh-intel (commented):
Thanks @sameo - a few initial comments...

slp:
@sameo Thanks a lot for working on this. A couple of comments about the payloads:

- I think it'd be better to drop the inner object and just put the fields on the root object, as it makes parsing the payloads simpler and cleaner. As an example, `Request` would change from this:

```
{
	"request": {
		"version": "0.1.0",
		"tee": "",
		"extra-params": {}
	}
}
```

To this:

```
{
	"version": "0.1.0",
	"tee": "",
	"extra-params": {}
}
```

- `Request` needs a field to identify the workload running in the TEE. In `kbs-types`/`reference-kbs` I've used `workload_id`, but feel free to use the name you think fits better here. This field is needed to tell the Attestation Server against which stored launch measurement should try to match the contents of `tee_evidence` from the `Attestation` payload. This is convenient not only to avoid the need to check against all registered launch measurements, but also to enable per-workload resource unlocking (and some other tricks, such as one time tickets for workloads), as it's likely multiple workloads will share the same launch measurement (same number of vCPUs + same guest initial payload).

docs/kbs_attestation_protocol.md
slp: I think it's reasonable leaving that up to the implementation and out of the scope of the protocol. Managing the lifetime of the sessions is a extremely common thing in web development and there is plenty of literature about best practices.

jodh-intel: Maybe we could note explicitly that this is out of scope to avoid any confusion then? I would imagine that a very long expiration time could be problematic from a memory and security standpoint.

jialez0:
> * `Request` needs a field to identify the workload running in the TEE. In `kbs-types`/`reference-kbs` I've used `workload_id`, but feel free to use the name you think fits better here. This field is needed to tell the Attestation Server against which stored launch measurement should try to match the contents of `tee_evidence` from the `Attestation` payload. This is convenient not only to avoid the need to check against all registered launch measurements, but also to enable per-workload resource unlocking (and some other tricks, such as one time tickets for workloads), as it's likely multiple workloads will share the same launch measurement (same number of vCPUs + same guest initial payload).

@slp Does workload here refer to the user's container? I'm not sure whether our remote attestation system will provide services for applications in the user's workload in the future. At least so far, AA's functional scope is only to provide services for Kata components in Pod, including image-rs and ocicrypt-rs in Kata-agent. Therefore, in fact, our remote attestation system is pod granular. When the pod needs a key to decrypt the container image, or needs some confidential resources, such as the pod's security policy configuration file, it will send a request to KBS.

Or is the workload ID you mean here actually the pod ID? If so, how can we identify a pod in the k8s cluster?

slp:
> @slp Does workload here refer to the user's container? I'm not sure whether our remote attestation system will provide services for applications in the user's container in the future. At least so far, AA's functional scope is only to provide services for components in Pod, such as image-rs and ocicrypt-rs in Kata-agent. Therefore, in fact, our remote attestation system is pod granular. When the pod needs a key to decrypt the container image, or needs some confidential resources, such as the pod's security policy configuration file, it will send a request to KBS.
> 
> Or is the workload ID you mean here actually the pod ID? If so, how can we identify a pod in the k8s cluster?

@jialez0 The definition of `workload` varies from one use case to the other, and it's more related to the process in which a launch measurement is registered (which is not covered by the KBS protocol) than the Attestation Server itself. Ultimately, it's just a way to know which launch measurement should be used to attest a TEE.

I don't know much about Kata's specifics, but there are some challenges that apply to every Confidential Computing deployment model. At some point, before launching the TEE, something needs to generate the launch measurement for that particular TEE, based on the initial guest payload and the number of vCPUs, and register it with the Attestation Server. And we need a way to relate the launch measurement with the TEE, hence the need for `workload_id`.

docs/kbs_attestation_protocol.md
```diff
@@ -1,99 +1,137 @@
...
-2. **Challenge**: After receiving the request, KBS returns a challenge to KBC and asks KBC to 
-send evidence to prove that its environment (HW-TEE) is safe and reliable.
-3. **Attestation**: KBC sends the evidence and the HW-TEE generated, ephemeral public key to the KBS.
-4. **Response**: The `Response` payload includes the resource data requested when sending the initial `Request` to
-the specific KBS resource endpoint. The resource data is protected by the ephemeral public key
-passed via the previously sent Attestation payload. Within the valid time of the token, KBC can directly 
-request resources or services from KBS by virtue of the token without going through step 2 and step 3.
-
-## Request
+The purpose of the attestation between KBS and KBC is actually to confirm
```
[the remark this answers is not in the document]
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
```diff
@@ -1,99 +1,137 @@
...
+application layer semantics designed in this document. This is because HTTPS
+provides KBC with a means to authenticate KBS identity, which effectively avoids
+malicious attackers from hijacking KBS address to impersonate KBS and deceive
+KBC.
+In order to achieve this, the public key of KBS needs to be transmitted to KBC
+through an effective way, and the specific way of public key distribution is out
+of this document scope.
+In addition, it should be noted that the confidentiality protection provided by
+HTTPS alone is not enough to meet the security requirements of KBS protocol,
+therefore, as mentioned above, the key generated by HW-TEE needs to be used to
```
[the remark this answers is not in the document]
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
```diff
@@ -1,99 +1,137 @@
...
-If other requests have been made before this request, in the previous response KBS would return a 
-token to KBC. There are two cases for the contents of this field: 
-
-1. A token is provided, which means a previous attestation was done. If the token is valid, 
-the Challenge/Attestation part can be skipped. If it is not valid, KBS will send a Challenge.
-
-2. Left blank, which means the KBC is implictly expecting to go through attestation 
-and KBS will reply with a Challenge.
+Used to declare the type of HW-TEE platform where KBC is located, the valid
+values are `tdx`, `sgx` and `sev-snp`.
```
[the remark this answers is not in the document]
sameo: It's not too late, and your suggestion makes sense. Thanks.

docs/kbs_attestation_protocol.md
```diff
@@ -1,99 +1,137 @@
...
 
 - `extra-params`
 
-In the run-time attestation scenario (TDX, SEV-SNP, SGX), the `extra-params` are not used, so it is kept empty.
-However, for the attestation of some special HW-TEE platforms, this field may be used to transfer some specific information,
-for example, some attestations depend on Diffie–Hellman to build a secure channel and transfer secret messages 
-(Such as the SEV(-ES) pre-attestation).
+In the run-time attestation scenario (TDX, SEV-SNP, SGX), the `extra-params` are
+not used, so it is kept empty. However, for the attestation of some special
```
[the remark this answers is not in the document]
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
```diff
@@ -1,99 +1,137 @@
...
+channel and transfer secret messages (Such as the SEV(-ES) pre-attestation).
 
-## Challenge
+## `Challenge`
 
-After KBS receives the request, if the token is found to be empty or expired, KBS will return an 
-attestation challenge to KBC. The payload format is as follows:
+If the KBC does not own any KBS generated HTTP Cookie, or if the Cookie validity
+has expired, the KBC is not allowed to directly request resources. It must go
+through the attestation process and sends an initial `Request` payload to the
```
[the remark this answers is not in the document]
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
sameo: I mean that the header should not contain any kbs related cookie (as opposed  to the next ones).
I tried to fix it now, let me know if that looks better.

docs/kbs_attestation_protocol.md
sameo: Fixed it, thanks.

docs/kbs_attestation_protocol.md
sameo: https://monodraw.helftone.com/#introduction ;-)

slp:
Thanks @sameo, a couple of comments:

- The `Response` payload still has a nested object (thanks for dropping it from the other payloads).
- While re-reviewing the payloads, I've noticed that the fields `nonce` and `tee` from `Attestation` are probably redundant. The first one is provided by the server to the client in `Challenge`, while the second is provided by the client to the guest in `Request`. This doesn't do much harm, but it's a bit confusing since it may look like they can be different from the ones exchanged before.

docs/kbs_attestation_protocol.md
```diff
@@ -1,99 +1,137 @@
...
+         * [Token format](#token-format)
+
 # KBS attestation Protocol
 
-The Key Broker Service attestation protocol defines communication between a Key Broker Client in a confidential guest 
-and a trusted Key Broker Service. The protocol uses the simple, universal, and extensible 
-"Request-Challenge-Attestation-Response" (RCAR) method to facilitate guest attestation and secret injection.
+The Key Broker Service attestation protocol defines communication between a
+Key Broker Client (KBC) in a confidential guest and a trusted Key Broker
+Service  (KBS). The protocol uses the simple, universal, and extensible
```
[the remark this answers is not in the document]
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
```diff
@@ -1,99 +1,137 @@
...
+   key.
+4. **Response**: Upon succesfull attestation, the KBC can request resources by
+   sending HTTP `GET` requests to resource specific endpoints. The KBS responds
+   with a `Response` payload that includes the requested resource data. The
+   resource data is protected by the ephemeral public key passed via the
+   previously sent `Attestation` payload. Within the valid time of the HTTP
+   Cookie generated by the KBS during step 2, the KBC can directly
+   request resources or services from KBS, without going through steps 2 and 3.
+
+
```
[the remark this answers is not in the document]
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
```diff
@@ -108,113 +146,303 @@ attestation challenge to KBC. The payload format is as follows:
 
 - `nonce`
 
-The fresh number passed to KBC. KBC needs to place it in the evidence sent to KBS in the next 
-step to prevent replay attacks.
+The fresh number passed to KBC. KBC needs to place it in the evidence sent to
+the KBS in the next step to prevent replay attacks.
```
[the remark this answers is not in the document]
sameo: Not to my knowledge, no.

sameo:
Thanks @slp for the review.

> Thanks @sameo, a couple of comments:
> 
>     * The `Response` payload still has a nested object (thanks for dropping it from the other payloads).

Sorry, fixed now.

>     * While re-reviewing the payloads, I've noticed that the fields `nonce` and `tee` from `Attestation` are probably redundant. The first one is provided by the server to the client in `Challenge`, while the second is provided by the client to the guest in `Request`. This doesn't do much harm, but it's a bit confusing since it may look like they can be different from the ones exchanged before.

I agree it could be confusing. I removed them.

Before I add the OpenAPI description to that PR, we need to resolve the open around the `workload_id` proposal. From what I see from the `kbs-reference` implementation, something registers a workload against the attestation service and maps its launch measurement to a workload ID. Is that correct?
I propose to make that field optional, as  other attestation flows may run the evidence verification from some of the attestation parameters like e.g. the guest parameters (SVN, firmware versions, number of vcpus, etc).

slp:
> Before I add the OpenAPI description to that PR, we need to resolve the open around the `workload_id` proposal. From what I see from the `kbs-reference` implementation, something registers a workload against the attestation service and maps its launch measurement to a workload ID. Is that correct?

That's correct. The `/register_workload` router isn't part of the protocol, but it comes very handy for testing the attestation workflow.

> I propose to make that field optional, as other attestation flows may run the evidence verification from some of the attestation parameters like e.g. the guest parameters (SVN, firmware versions, number of vcpus, etc).

The field `workload_id` is basically a way of synthesizing those guest parameters (SVN, firmware versions, number of vcpus...) into a single string, which can be a random UUIDv4 (or a formatted string with the parameters, or something else). Otherwise, the protocol would need to have an object that could hold a potentially arbitrary number of guest parameters.

I'm fine with making it optional, but I think it's worth discussing how the protocol would be used when it's missing to make it more solid.

jialez0:
> * While re-reviewing the payloads, I've noticed that the fields `nonce` and `tee` from `Attestation` are probably redundant. The first one is provided by the server to the client in `Challenge`, while the second is provided by the client to the guest in `Request`. This doesn't do much harm, but it's a bit confusing since it may look like they can be different from the ones exchanged before.

@slp @sameo I agree these two fields seems a little redundant here and remove them can make our kbs protocol more concise. But if we do this removing, let us remember that AS needs `nonce` and `tee` fileds to verify the Evidence, so KBS should manage these two fields and pass them together with `tee-pubkey` and `tee-evidence` to AS in a new structure.

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,402 @@
...
+
+The KBS protocol currently supports two kinds of resources for an attester to
+request: keys and tokens.
+
+### Key Resource
+
+Keys are generic protected resources that an authenticated attester can get by
+sending a `GET` HTTP request to the `/kbs/v0/key/<key_id>` endpoint.
+
+The `<key_id>` resource name is defined by the attestation service. There should
```
jialez0, line 381: There may be some confusion here. The attestation service should only be responsible for the verification of the evidence in the `Attestation` step, and the resource name should be defined by KBS itself.

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,402 @@
...
-            "initialization-vector": "",
-            /* The symmetric key used to encrypt `output`, which is encrypted by HW-TEE's public key */
-            "enc-symkey": ""
-        },
-        /* The token issued by KBS to KBC. If there is a token in the request, this field will be left blank. */
-        "token": "",
-    }
+    /*
+     * The output of the KBS service API.
+     * It must be encrypted with the ephemeral symmetric key generated by the KBC.
```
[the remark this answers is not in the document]
sameo: Thanks, I already fixed that one. I am going to send and updated version soon.

docs/kbs_attestation_protocol.md
sameo: Thanks, I'll fix this as this could be quite confusing.

sameo:
> > Before I add the OpenAPI description to that PR, we need to resolve the open around the `workload_id` proposal. From what I see from the `kbs-reference` implementation, something registers a workload against the attestation service and maps its launch measurement to a workload ID. Is that correct?
> 
> That's correct. The `/register_workload` router isn't part of the protocol, but it comes very handy for testing the attestation workflow.
> 
> > I propose to make that field optional, as other attestation flows may run the evidence verification from some of the attestation parameters like e.g. the guest parameters (SVN, firmware versions, number of vcpus, etc).
> 
> The field `workload_id` is basically a way of synthesizing those guest parameters (SVN, firmware versions, number of vcpus...) into a single string, which can be a random UUIDv4 (or a formatted string with the parameters, or something else). Otherwise, the protocol would need to have an object that could hold a potentially arbitrary number of guest parameters.
> 
> I'm fine with making it optional, but I think it's worth discussing how the protocol would be used when it's missing to make it more solid.

Don't get me wrong, I like the idea and I think it's a simple way to bind a set of measurements/reference values to a workload or a set of workloads. But how is that different from hashing the "unstable" parts  of the attestation evidence, where "unstable" here means everything that is workload specific (code measurement, launch params, num of vcpus, SVNs, etc)? That hash would represent the workload itself, and could otoh be generated by the attestation service.

Adding a `workload_id` to the `Request` payload (even if it's optional) means we need to define how the guest must generate it, and in all cases it implies some a-priori knowledge from the guest about how the measurements were generated and registered in the attestation service. You don't want to add a `workload_id` that wouldn't match exactly how it was registered against a set of measurements that were generated when e.g. creating the guest artifacts.
When not including a `workload_id`, the guest does not need that a-priori knowledge and it is up to the verifier to match the evidence against a set of measurement (And this could be done for example by hashing a subset of the attestation evidence, when receiving it on the attestation service side).

docs/kbs_attestation_protocol.md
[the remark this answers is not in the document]
sameo: Fixed now.

docs/kbs_attestation_protocol.md
sameo: Fixed, thanks.

sameo:
@jialez0 @Xynnn007 @slp @jodh-intel  I updated the document, addressing a bunch of comments from @jodh-intel and a few from others.

One noticeable change is the proposal to use JSON Web standards for our public key formats (JWK), our Response payload (JWE) and our additional error information (a.k.a. `Problem Details for HTTP APIs`), as suggested by @thomas-fossati. Please have a look and let us know if it makes sense.

jialez0 (approved):
LGTM. Now our protocol has finally become concise, standardized and beautiful, I like it! Thanks a lot @sameo .

In addition, I have some new views on the proposal of workload ID, but I think it should go beyond the scope of this PR. we can continue to discuss it in #7 .

slp:
> In addition, I have some new views on the proposal of workload ID, but I think it may go beyond the scope of this PR. we can continue to discuss it in #7 . @slp

@jialez0 Your proposal sounds interesting, but shouldn't the `measurements` object be part of the KBS protocol?

jialez0:
> @jialez0 Your proposal sounds interesting, but shouldn't the `measurements` object be part of the KBS protocol?

@slp If we put the `measurement` field in the KBS protocol, it can be placed in the Evidence payload, as follows:

```json
{
  "tee-pubkey":"$tee-pubkey",
  "measurements": {},
  "tee-evidence":"$tee-evidence"
}
```

But at the same time, it also means that on the attester side, we need to have a general mechanism for generating and obtaining measurement values, but I am not sure whether different HW-TEE types have different mechanisms for generating and recording measurement values (for example, as far as I know, for Intel TDX, the generation of kernel measurement value is completed by virtual firmware and recorded in the Eventlog under the TCG standard). However, for me personally, it seems great to have a common mechanism for generating and obtaining metrics.

What do you think? @sameo

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
+verifiable.
+
+In this document, HW-TEE attestation process is the semantics of the application
+layer, which is defined as a simple, universal and extensible
+"Request-Challenge-Attestation-Response" (RACR) protocol. The temporary
+asymmetric key generated by HW-TEE is used to encrypt the response output
+payload, and the standard [HTTP Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies)
+mechanism is used to avoid the performance problems caused by multiple
+attestation: it's fairly typical to request for several resources, and then send
+from the same KBC multiple Requests. Having to do the whole attestation dance
```
jodh-intel, line 23: This might be slightly clearer?:

```suggestion
attestation: it's fairly typical for the KBC to require several resources, resulting in multiple Requests to the KBS. Having to do the whole attestation dance
```

jialez0:
> But at the same time, it also means that on the attester side, we need to have a general mechanism for generating and obtaining measurement values, but I am not sure whether different HW-TEE types have different mechanisms for generating and recording measurement values (for example, as far as I know, for Intel TDX, the generation of kernel measurement value is completed by virtual firmware and recorded in the Eventlog under the TCG standard). However, for me personally, it seems great to have a common mechanism for generating and obtaining measurement values.

Or, we can add an interface to each HW-TEE's attester diver to obtain measurements, so that each HW-TEE's attester diver has two interfaces, one for obtaining `tee-evidence`, which is often called report or quote; the other is used to obtain measurements. That sounds like a good idea, too.

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
+   the KBS also sends a session identifier to the KBC, as an HTTP Cookie. This
+   session identifier can be used to skip steps 2 and 3 after a successful
+   attestation.
+3. **Attestation**: The KBC replies to the attestation challenge from step 2
+   with an attestation evidence, in order to prove that its environment (HW-TEE)
+   is safe and reliable. The KBC sends an `Attestation` payload to the KBS, that
+   contains the attestation evidence and the HW-TEE generated, ephemeral public
+   key.
+4. **Response**: Upon succesfull attestation, the KBC can request resources by
+   sending HTTP `GET` requests to resource specific endpoints. The KBS responds
```
jodh-intel, line 60: Could you change the order of these sentences to make this step clearer? I think it would be clearer to state first that the KBS will return a successful `Response` payload iff the KBC's `Attestation` payload was successfully validated.

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
+    /*
+     * Type of HW-TEE platforms where KBC is located,
+     * e.g. "intel-tdx", "amd-sev-snp", etc.
+     */
+    "tee": "$tee",
+    /* Reserved fields to support some special requests sent by HW-TEE. */
+    "extra-params": {}
 }
 ```
+<sub>Note that the `/*...*/` comments are not valid in JSON, and must not be
```
jodh-intel, line 86: You could use a ["note" as we do in Kata](https://github.com/kata-containers/kata-containers/blob/main/docs/Documentation-Requirements.md#notes) to make this stand out (and avoid the need for embedded HTML tags). Same comment for the other instances in this doc.

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
+Diffie–Hellman key exchange protocol to first build a secure channel and
+transfer secret messages (Such as AMD SEV(-ES) pre-attestation).
 
-- `token`
+## `Challenge`
 
-If other requests have been made before this request, in the previous response KBS would return a 
-token to KBC. There are two cases for the contents of this field: 
+If the KBC does not own any KBS generated HTTP Cookie, or if the Cookie validity
+has expired, the KBC is not allowed to directly request resources. It must go
```
jodh-intel, line 112: > the KBC is not allowed to directly request resources.

Is this worth clarifying the meaning of "allowed" here? Presumably if the KBC attempts to use an expired cookie, the KBS will return an error?

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
+After KBC receives the attestation challenge, an ephemeral asymmetric key pair
+is generated in HW-TEE. The private key is stored in HW-TEE. The public key and
+its description information are exported and placed in the `tee-pubkey` field
+and sent to the KBS together with the attestation evidence. The hash of the
+`tee-pubkey` field must be included in the custom field of HW-TEE evidence and
+signed by HW-TEE hardware. This public key is valid until the next time the KBC
+receives an attestation challenge from the KBS.
 
-- `extra-params`
+The `tee-pubkey` follows the  [JWK](https://www.rfc-editor.org/rfc/rfc7517)
```
jodh-intel, line 173:
```suggestion
The `tee-pubkey` follows the  [JSON Web Key (JWK)](https://www.rfc-editor.org/rfc/rfc7517)
```

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
+    "kty": "$key_type",
+    "alg": "$key_algorithm"
+    "k": "public_key"
 }
 ```
 
-- `output`
+# HTTP Integration
+
+KBS uses the HTTP transport protocol to exchange the above described
```
jodh-intel, line 265: This is rather confusing as the introduction says we use HTTPS (line 27).

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
+"Request-Challenge-Attestation-Response" (RACR) protocol. The temporary
+asymmetric key generated by HW-TEE is used to encrypt the response output
+payload, and the standard [HTTP Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies)
+mechanism is used to avoid the performance problems caused by multiple
+attestation: it's fairly typical to request for several resources, and then send
+from the same KBC multiple Requests. Having to do the whole attestation dance
+would be time consuming and have a latency impact on the guest.
+
+In order to ensure the ease of use and security completeness of KBS, we will use
+[HTTPS](#http-integration) as the default transport protocol to carry the
```
jodh-intel, line 27: > the default transport protocol

What are the alternatives to the HTTP default? How would they be configured?

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
+to a resource specific endpoint. If the attester is allowed to access the
+resource, the KBS will respond to the `GET` request with an HTTP response which
+content is set to a [KBS Response](#response) JSON payload:
+
+```
+┌──────────┐                                         ┌─────┐
+│ Attester │                                         │ KBS │
+└───┳──────┘                                         └──┳──┘
+    ┃            ┌─────────────────────────┐            ┃
+    ┃────────────│GET /kbs/v0/<resource_id>│───────────▶┃
```
jodh-intel, line 361: What if `<resource_id>`  is `auth` or `attest`? Could that be confusing to clients or KBS programmers, and hence potentially dangerous from a security pov? I see that this endpoint requires `GET` and those "privileged" commands are using `POST`, but would it be safer if this endpoint was different to the "privileged" one (maybe `/kbs/v0/request/<resource_id>` or siimilar?)

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
+### Token Resource
 
-If the attestation evidence sent by KBC is verified before returning the response, KBS will issue 
-a token to KBC in this field and sign it with KBS's root key. KBS will maintain the corresponding 
-relationship between each issued token and its original attestation result. KBC can use this token 
-to request resources or services from KBS without Challenge and Attestation within the validity period of 
-the token. If there is already a token in the KBC's request, no new token will be issued and this 
-field will be left blank.
+Authenticated attesters can also request a token from the KBS, from the
+`/kbs/v0/token` endpoint. Attesters can use tokens to request additional
```
jodh-intel, line 409: Is this accessible via `GET`?

jodh-intel (commented):
Thanks @sameo - a few comments...

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
-3. **Attestation**: KBC sends the evidence and the HW-TEE generated, ephemeral public key to the KBS.
-4. **Response**: The `Response` payload includes the resource data requested when sending the initial `Request` to
-the specific KBS resource endpoint. The resource data is protected by the ephemeral public key
-passed via the previously sent Attestation payload. Within the valid time of the token, KBC can directly 
-request resources or services from KBS by virtue of the token without going through step 2 and step 3.
-
-## Request
+The purpose of the attestation between KBS and KBC is to confirm whether **the
+platform where the KBC is located is in the expected security state**, i.e. if
+it runs in a real and harmless HW-TEE, and the software stack is measured and
```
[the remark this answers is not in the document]
sameo: I meant trustworthy. Fixed now, thanks.

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
+verifiable.
+
+In this document, HW-TEE attestation process is the semantics of the application
```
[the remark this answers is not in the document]
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
+In order to achieve this, the public key of KBS needs to be transmitted to KBC
+through an effective way, and the specific way of public key distribution is out
+of this document scope.
+
+It should be noted that the confidentiality protection provided by HTTPS alone
+is not enough to meet the security requirements of the KBS protocol.
+Therefore, as mentioned above, the key generated by HW-TEE needs to be used to
+encrypt and protect the confidential data.
+
+# `RCAR` semantics
```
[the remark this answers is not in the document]
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,418 @@
...
-
-## Request
+The purpose of the attestation between KBS and KBC is to confirm whether **the
+platform where the KBC is located is in the expected security state**, i.e. if
+it runs in a real and harmless HW-TEE, and the software stack is measured and
+verifiable.
+
+In this document, HW-TEE attestation process is the semantics of the application
+layer, which is defined as a simple, universal and extensible
+"Request-Challenge-Attestation-Response" (RACR) protocol. The temporary
```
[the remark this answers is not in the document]
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
sameo: Nice, suggestion applied.

docs/kbs_attestation_protocol.md
sameo: I tried to make it clearer now. Let me know if it's not.

docs/kbs_attestation_protocol.md
sameo: Yes. Clarified now.

docs/kbs_attestation_protocol.md
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
sameo: Fixed, thanks.

docs/kbs_attestation_protocol.md
sameo: There are no alternatives :)

docs/kbs_attestation_protocol.md
sameo: I added a `resources/` prefix. And it can either be `resources/key/keyid` or `resources/token/`

docs/kbs.yaml
```diff
@@ -0,0 +1,267 @@
...
+              Set-Cookie:
+                name: kbs-session-id
+                schema:
+                  type: string
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ResponseHeader'
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ResponsePayload'
```
thomas-fossati, line 93: This looks like it's defining two response bodies.  I don't think this is what you mean :-)  Also, syntactically, I guess this would not validate against the OpenAPI schema.  BTW, have you tried this against one of the [validation tools](https://openapi.tools/#description-validators)?  As a recommendation, I'd hook one in (e.g., via Makefile) to check both as you develop and as part of the CI.

docs/kbs.yaml
```diff
@@ -0,0 +1,267 @@
...
+              Set-Cookie:
+                name: kbs-session-id
+                schema:
+                  type: string
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TokenHeader'
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TokenPayload'
```
thomas-fossati, line 123: ditto

thomas-fossati (commented):
The protocol flow looks good.  I am not sure the openapi spec is fully correct though.

docs/kbs.yaml
```diff
@@ -0,0 +1,267 @@
+openapi: 3.0.1
+info:
+  title: Key Broker Service (KBS) API
+  description: HTTP based API for describing the communication between a a
```
jodh-intel, line 4:
```suggestion
  description: HTTP based API for describing the communication between a
```

docs/kbs.yaml
```diff
@@ -0,0 +1,267 @@
...
+               Key Broker Client in a confidential guest and a trusted
+               Key Broker Service.
+  license:
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.1.0
```
jodh-intel, line 10: I wonder if it's worth moving `version` up above `license:` to avoid any confusion with it being `license.version` ;)

docs/kbs.yaml
```diff
@@ -0,0 +1,267 @@
+openapi: 3.0.1
+info:
+  title: Key Broker Service (KBS) API
+  description: HTTP based API for describing the communication between a a
+               Key Broker Client in a confidential guest and a trusted
```
jodh-intel, line 5: For parity with KBS:

```suggestion
               Key Broker Client (KBC) in a confidential guest and a trusted
```

docs/kbs.yaml
```diff
@@ -0,0 +1,267 @@
...
+    ResponseHeader:
+      required:
+        - alg
+        - enc
+      type: object
+      properties:
+        alg:
+          type: string
+        enc:
+          type: string
```
jodh-intel, line 190: Does OpenAPI allow you to add a description for `alg` and `enc` as they are brutally terse ;)

docs/kbs.yaml
```diff
@@ -0,0 +1,267 @@
...
+        Javascript Object Signing and Encryption (JOSE) header.
+
+    ResponsePayload:
+      required:
+        - ciphertext
+        - encrypted_key
+      properties:
+        ciphertext:
+          type: string
+        iv:
```
jodh-intel, line 201: What is this? Can it also have a description?

docs/kbs.yaml
```diff
@@ -0,0 +1,267 @@
...
+          type: string
+      description:
+        Javascript Object Signing and Encryption (JOSE) header.
+
+    TokenPayload:
+      required:
+        - ext
+        - iat
+        - iss
+        - tee_pubkey
```
jodh-intel, line 227: Could these be documented too?

docs/kbs.yaml
```diff
@@ -0,0 +1,267 @@
...
+          type: string
+      description:
+        A Problem Details for HTTP APIs (https://www.rfc-editor.org/rfc/rfc7807)
+        formatted payload.
+
+    PublicKey:
+      required:
+        - kty
+        - alg
+        - k
```
jodh-intel, line 258: Ditto ;)

docs/kbs.yaml
sameo: @thomas-fossati I wonder what would be the proper way to represent a JWT or a JWE with an OpenAPI schema.

Using the Compact Serialization scheme, iiuc it would be something like:

```yaml
Response:
     required:
          - jwe
        type: object
        properties:
          jwe: string
          description: A JWE serialized with Compact Serialization
```

A less compact option would be to use the Flattened Serialization syntax, which would make it look like that:

```yaml
Response:
      required:
        - unprotected
        - ciphertext
        - encrypted_key
      type: object
      properties:
        unprotected:
          type: string
        ciphertext:
          type: string
        encrypted_key:
          type: string
      description: JWE serialized with the Flattened Serialization format
```

AFAICT there are no OpenAPI standard ways of defining a JWE or JWS.

sameo: @thomas-fossati Thanks for the offline advice (https://json-schema.org/draft/2020-12/json-schema-validation.html#section-8.6).
I think the OpenAPI 3.1 description looks good now, and it's validated with the redocly linter.

docs/kbs.yaml
sameo: Fixed, thanks.

docs/kbs.yaml
sameo: Fixed as well, thanks.

docs/kbs.yaml
sameo: Description fixed now.

docs/kbs.yaml
sameo: They are documented now.

sameo:
@jodh-intel Thanks a bunch for the thorough review. I think most of your comments are addressed now.

docs/kbs.yaml
```diff
@@ -0,0 +1,297 @@
...
+      required:
+        - version
+        - tee
+      type: object
+      properties:
+        version:
+          type: string
+        tee:
+          type: string
+          enum: [intel-tdx, intel-sgx, amd-sev]
```
jodh-intel, line 146: Nit: Unsorted list.

docs/kbs.yaml
thomas-fossati: 👍 awesome job!

docs/kbs.yaml
sameo: Fixed, thanks.

docs/kbs.yaml
```diff
@@ -0,0 +1,297 @@
...
+        - kty
+        - alg
+        - k
+      properties:
+        kty:
+          type: string
+          description: Key type
+        alg:
+          type: string
+          description: Key Algorithn
```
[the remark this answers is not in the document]
sameo: Fixed as well.

docs/kbs.yaml
```diff
@@ -0,0 +1,297 @@
...
+      properties:
+        ciphertext:
+          type: string
+          description: The encrypted reponse data.
+        encrypted_key:
+          type: string
+          description: The encrypted key used for encrypting the response data.
+        if:
+          type: string
+          description: Encryption algorithm initialisation vector.
```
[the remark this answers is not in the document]
sameo: Good catch, fixed now.

sameo:
@slp @jialez0 I think we're good to go with that PR. Could you please have one last look?

slp:
@sameo If you don't mind waiting a bit, I'd like to update kbs-types/reference-kbs to see how the protocol looks like. I'll try to find some time to do it this week.

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,412 @@
...
+     */
+    "ciphertext": "$response_output",
+    "iv": "$initialization_vector",
+    /*
+     * The symmetric key used to encrypt `ciphertext`, which is encrypted by
+     * the HW-TEE's public key.
+     */
+    "encrypted_key": "$enc_kbs_symkey"
 }
 ```
```
slp, line 209: JSON lint tells me this isn't valid JSON. Is this intended to be an array? Can the JOSE header object be a field within the root object?

docs/kbs_attestation_protocol.md
```diff
@@ -234,43 +426,57 @@ The [JSON web token](https://jwt.io/) standard is adopted for the token. The for
...
-    }
+    /*
+     * The HW-TEE's public key sent by KBC in the attestation evidence, which is
+     * valid within the validity period of the token.
+     * It follows the JWK format.
+     */
+    "tee-pubkey": $pubkey
 }.
 [Signature]
 ```
```
slp, line 437: Ditto.

sameo:
> @sameo If you don't mind waiting a bit, I'd like to update kbs-types/reference-kbs to see how the protocol looks like. I'll try to find some time to do it this week.

Sure thing. I'm actually happy to not merge this PR until you can verify it's implementable from `reference-kbs`.

docs/kbs_attestation_protocol.md
sameo: @slp Yes, the example is incorrectly formatted and the intention is for it to be an array, yes. So it should really look like that:

```json
[{
	"alg": "$alg",
	"enc": "$enc"
}, {
	"ciphertext": "$response_output",
	"iv": "$initialization_vector",
	"encrypted_key": "$enc_kbs_symkey"
}]
```

docs/kbs_attestation_protocol.md
```diff
@@ -1,220 +1,412 @@
...
-        /* Extra parameters to support some special HW-TEE attestation. */
-        "extra-params": {}
-    }
+    /*
+     * A JWK-formatted public key, generated by the KBC running in the HW-TEE.
+     * It is valid until the next time an attestation is required. Its hash must
+     * be included in the HW-TEE evidence and signed by the HW-TEE hardware.
+     */
+    "tee-pubkey": $pubkey
+
```
jialez0, line 147: As we discussed earlier, perhaps we need to add a `measurements` field (or another synonymous name) to `Attestation` json payload? Just like this:

```json
{
    "tee-pubkey": $pubkey,
    "measurements": {
        $name: $value,
        ...
    },
    "tee-evidence": {}
}
```

jialez0: cc @slp @sameo

docs/kbs_attestation_protocol.md
slp: I'm not familiar with the JOSE specs. Is it required to be found in this form? As ordering in an array isn't guaranteed, consuming this payload would require to peek into the object to find out its format.

Would it be possible for the JOSE header to be an object in named field? Something like:

```json
{
	"ciphertext": "$response_output",
	"iv": "$initialization_vector",
	"encrypted_key": "$enc_kbs_symkey",
	"jose_params": {
		"alg": "$alg",
		"enc": "$enc"
	}
}
```

Xynnn007: > I'm not familiar with the JOSE specs. Is it required to be found in this form? As ordering in an array isn't guaranteed, consuming this payload would require to peek into the object to find out its format.
> 
> Would it be possible for the JOSE header to be an object in named field? Something like:
> 
> ```json
> {
> 	"ciphertext": "$response_output",
> 	"iv": "$initialization_vector",
> 	"encrypted_key": "$enc_kbs_symkey",
> 	"jose_params": {
> 		"alg": "$alg",
> 		"enc": "$enc"
> 	}
> }
> ```

AFAIK, the architecture of a JOSE object should look like
```rust
BASE64URL_ENCODE({"alg": "$alg","enc": "$enc"}).
BASE64URL_ENCODE($enc_kbs_symkey).
BASE64URL_ENCODE($initialization_vector).
BASE64URL_ENCODE($response_output).
BASE64URL_ENCODE($authentication_tag)
```

Here `$authentication_tag` is extra result of specific symmetric encryption (s.t. AEAD) besides cipher text. Algorithms like `AES/GCM` will produce such authentication data. And care about each adjacent two base64url-encoded part are connected by a dot `.`

I think the json format shown in the document may be for understanding the fields of the JWE used here. Am I right? @sameo 

Reference to [JOSE Cookbook](https://datatracker.ietf.org/doc/html/rfc7520#section-5.1.5) and [JWE Spec](https://www.rfc-editor.org/rfc/rfc7516#section-2).

sameo: Sorry for the delayed answer, post LPC has been busy for me...

@Xynnn007 you're partly right, although what you're describing is the [compact JWE serialization format](https://www.rfc-editor.org/rfc/rfc7516#section-7.1).
But I think we want to use the [flattened JWE JSON serialization format](https://www.rfc-editor.org/rfc/rfc7516#section-7.2.2). 

With that format, our `Response` JSON-formatted payload would then look like that:

```json
{
    "protected": "$jose_header",
    "encrypted_key": "$encrypted_key",
    "iv": "$iv",
    "ciphertext": "$ciphertext",
    "tag": "$tag"
}
```

Where:

```shell
jose_header = BASE64URL_ENCODE({"alg": "$alg","enc": "$enc"})
encrypted_key = BASE64URL_ENCODE($enc_kbs_symkey)
iv = BASE64URL_ENCODE($initialization_vector)
ciphertext = BASE64URL_ENCODE($response_output)

# tag is optional and depend on the encryption algorithm.
tag = BASE64URL_ENCODE($authentication_tag) 
```

I find it a little convoluted, but this really is what the spec defines. FWIW there are some on going, recent work on a rust crate for JWE and JWK: https://github.com/RustCrypto/JOSE

If we're fine with that definition, I'll update that PR accordingly.

docs/kbs_attestation_protocol.md
slp: @sameo Just to be sure, the KBS is expected to send the token pre-encrypted, doesn't it? That is, three base64-URL strings (header, payload and signature) separated by dots. Is that right? If so, perhaps the docs should explicitly state that?

sameo: That's correct @slp. I expanded the token section quite a bit to describe this. I also updated the openAPI description accordingly. Please have a look.

docs/kbs_attestation_protocol.md
sameo: Thanks @slp. I updated the document and the openAPI description as well. I think we *may* be good now :) PTAL.

slp (approved):
I've just updated `kbs-types` and `reference-kbs`, and I think the protocol, in its current state, fits quite well with web frameworks. Thanks @sameo !

sameo:
@jiazhang0 Could you PTAL?

jiazhang0 (approved):

PR #13, opened 2022-10-20
docs: Add a README and reference the OpenAPI document

sameo:


sameo:
cc @slp

slp:
If we're moving `reference-kbs` here, it may be worth waiting a bit to avoid having to amend the link. Perhaps is also worth linking `kbs-types`?

sameo:
@jialez0 Fixed and removed the reference-kbs link for now.

jialez0 (approved):
LGTM.

PR #21, opened 2022-12-13
Document: Optimized the path of resources.

jialez0:
KBS should not only support the acquisition of keys, but also support the acquisition of multiple confidential resources. This commit revises the path of resource location and describes resources in a more unified format.

cc @sameo @Xynnn007 

Refer to: https://github.com/confidential-containers/documentation/issues/85

### DIFF
--- a/docs/kbs.yaml
+++ b/docs/kbs.yaml
@@ -71,19 +71,37 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorInformation'
 
-  /resources/key/<key_id>:
+  /resource/{repository}/{type}/{tag}:
     get:
-      operationId: getKey
-      summary: Get a key from the Key Broker Service.
+      operationId: getResource
+      summary: Get a secret resource from the Key Broker Service.
       parameters:
         - in: cookie
           name: kbs-session-id
           schema:
             type: string
+        - name: repository
+          in: path
+          description: A parent path of resource, can be empty to use the default repository.
+          schema:
+            type: string
+          required: false
+        - name: type
+          in: path
+          description: Resource type name
+          schema:
+            type: string
+          required: true
+        - name: tag
+          in: path
+          description: Resource instance tag
+          schema:
+            type: string
+          required: true
       responses:
         200:
           description: >-
-            The KBS reponse including the key.
+            The KBS reponse including the resource.
             The KBS session ID is returned in a cookie named `kbs-session-id`.
           headers:
             Set-Cookie:
@@ -97,10 +115,10 @@ paths:
         401:
           description: Missing or invalid session ID
         403:
-          description: The KBC is not allowed to get that key
+          description: The KBC is not allowed to get that resource
         404:
-          description: The requested key does not exist
-  /resources/token:
+          description: The requested resource does not exist
+  /resource/token/{tag}:
     get:
       operationId: getToken
       summary: Get an external resource token from the Key Broker Service.
@@ -109,6 +127,12 @@ paths:
           name: kbs-session-id
           schema:
             type: string
+        - name: tag
+          in: path
+          description: Can be left blank to use default.
+          schema:
+            type: string
+          required: false
       responses:
         200:
           description: >-
@@ -128,7 +152,7 @@ paths:
         403:
           description: The KBC is not allowed to get a token.
         404:
-          description: The KBS does not provide token resources.
+          description: The KBS does not provide token resource.
 
 components:
   schemas:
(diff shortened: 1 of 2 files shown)


docs/kbs.yaml
```diff
@@ -71,19 +71,37 @@ paths:
...
+          schema:
+            type: string
+          required: false
+        - name: type
+          in: path
+          description: Resource type name
+          schema:
+            type: string
+          required: true
+        - name: type
```
sameo, line 95: This should be `name: tag`.

docs/kbs.yaml
```diff
@@ -71,19 +71,37 @@ paths:
...
+      operationId: getResource
+      summary: Get a secret resource from the Key Broker Service.
       parameters:
         - in: cookie
           name: kbs-session-id
           schema:
             type: string
+        - name: repository
+          in: path
+          description: A parent path of resource, can be empty to use the default repository.
```
sameo, line 85: I am not sure what this is for. Could you give an example here?

docs/kbs.yaml
```diff
@@ -71,19 +71,37 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorInformation'
 
-  /resources/key/<key_id>:
+  /resources/{repository}/{type}/{tag}:
```
sameo, line 74: Maybe we should rename `resources` to `resource`?

sameo, line 74: Also, could we use that more generic resource definition for tokens as well:

`GET /resource/token/token_tag`

docs/kbs.yaml
jialez0: This is similar to the concept of container image repository, which is used to facilitate users to manage different resource groups. Its name should be completely set by users.

For example, if the user has two TEE Pod instances, the `repository` field can be used to manage the resources that need to be provided to the two TEE Pod instances:
```
/kbs/v0/resource/tee-instance-abcde123/key/<key-id>
/kbs/v0/resource/tee-instance-edcba321/key/<key-id>
```

jialez0: And, if KBS needs to support multiple tenants in the future, repository can also be the name of the tenant, for example:

```
/kbs/v0/resource/Alice/key/<key-id>
```

docs/kbs.yaml
jialez0: OK, I think so.

docs/kbs.yaml
jialez0: Thanks! Fixed now.

docs/kbs_attestation_protocol.md
```diff
@@ -383,20 +383,22 @@ A request for protected resource can fail for three reasons:
 3. The requested resource does not exist. The KBS implementation sends an HTTP
    response with a 404 (`Not Found`) status code.
 
-The KBS protocol currently supports two kinds of resources for an attester to
-request: keys and tokens.
+### Resource Path
```
sameo, line 386: Could we call that one `### Secret Resource`

docs/kbs_attestation_protocol.md
```diff
@@ -383,20 +383,22 @@ A request for protected resource can fail for three reasons:
...
 
-### Key Resource
+KBS uses the following path format to locate resources:
 
-Keys are generic protected resources that an authenticated attester can get by
-sending a `GET` HTTP request to the `/kbs/v0/resources/key/<key_id>` endpoint.
+```
+/kbs/v0/resources/<repository>/<type>/<tag>
```
sameo, line 391: You need to fix the path here as well.

docs/kbs_attestation_protocol.md
```diff
@@ -383,20 +383,22 @@ A request for protected resource can fail for three reasons:
...
+```
+
+Here,
```
sameo, line 394:
```suggestion
Where the URL path parameters are:
```

docs/kbs_attestation_protocol.md
```diff
@@ -383,20 +383,22 @@ A request for protected resource can fail for three reasons:
...
 
-The `<key_id>` resource name is defined by the KBS. There must be a different
-`<key_id>` name for each generic secret or resource that the KBS intends to
-deliver to authenticated attesters.
+- `<repository>`: A parent path, can be empty to use the default repository.
+- `<type>`: To distinguish different resource types.
+- `<tag>`: To distinguish different resource instances.
```
sameo, line 399: It would be really useful to add a similar comment than https://github.com/confidential-containers/kbs/pull/21/files#r1048155490 here.

docs/kbs_attestation_protocol.md
jialez0: Sure, updated now.

sameo (approved):

PR #22, opened 2022-12-16
docs: Adds an acknowledgements section to the protocol

sameo:
Many folks have put significant efforts on defining that protocol, let's acknowledge that.

Signed-off-by: Samuel Ortiz <sameo@rivosinc.com>


fitzthum (approved):
Good thinking. This seems pretty comprehensive but we can always update in the future if need be.

jialez0 (approved):
Great.

PR #24, opened 2023-01-09
api_server: Implement the `/resource` endpoint

jialez0:
This PR uses local file as the resource repository. I'm not sure whether we need to use database. If we use database to store resources, it may mean that we need to add API endpoints to let users upload and modify resources in the database. @sameo @fitzthum

### DIFF
--- a/src/api_server/src/http.rs
+++ b/src/api_server/src/http.rs
@@ -9,6 +9,7 @@ use std::sync::Arc;
 use strum_macros::EnumString;
 use tokio::sync::Mutex;
 
+use crate::resource::{secret_resource, Repository, ResourceDesc};
 use crate::session::{Session, SessionMap, KBS_SESSION_ID};
 
 const ERROR_TYPE_PREFIX: &str = "https://github.com/confidential-containers/kbs/errors/";
@@ -19,6 +20,7 @@ pub enum ErrorInformationType {
     FailedAuthentication,
     InvalidCookie,
     MissingCookie,
+    UnAuthenticatedCookie,
     VerificationFailed,
 }
 
@@ -114,4 +116,93 @@ pub(crate) async fn attest(
     };
 
     HttpResponse::Unauthorized().json(error_info)
+}
+
+/// GET /resource/{repository}/{type}/{tag}
+/// GET /resource/{type}/{tag}
+pub(crate) async fn resource(
+    request: HttpRequest,
+    repository: web::Data<Arc<dyn Repository + Send + Sync>>,
+    map: web::Data<SessionMap<'_>>,
+) -> HttpResponse {
+    let cookie = match request.cookie(KBS_SESSION_ID) {
+        None => {
+            log::error!("Missing KBS cookie");
+            return HttpResponse::Unauthorized()
+                .json(kbs_error_info(ErrorInformationType::MissingCookie, ""));
+        }
+        Some(c) => c,
+    };
+
+    let session_map = map.sessions.read().await;
+    let locked_session = match session_map.get(cookie.value()) {
+        None => {
+            log::error!("Invalid KBS cookie {}", cookie.value());
+            return HttpResponse::Unauthorized().json(kbs_error_info(
+                ErrorInformationType::InvalidCookie,
+                cookie.value(),
+            ));
+        }
+        Some(ls) => ls,
+    };
+
+    let session = locked_session.lock().await;
+
+    log::info!("Cookie {} request to get resource", session.id());
+
+    if !session.is_authenticated() {
+        log::error!("UnAuthenticated KBS cookie {}", cookie.value());
+        return HttpResponse::Unauthorized().json(kbs_error_info(
+            ErrorInformationType::UnAuthenticatedCookie,
+            cookie.value(),
+        ));
+    }
+
+    if session.is_expired() {
+        log::error!("Expired KBS cookie {}", cookie.value());
+        return HttpResponse::Unauthorized().json(kbs_error_info(
+            ErrorInformationType::ExpiredCookie,
+            cookie.value(),
+        ));
+    }
+
+    let resource_description = ResourceDesc {
+        repository_name: request
+            .match_info()
+            .get("repository")
+            .unwrap_or("default")
+            .to_string(),
+        resource_type: request.match_info().get("type").unwrap().to_string(),
+        resource_tag: request.match_info().get("tag").unwrap().to_string(),
+    };
+
+    if resource_description.resource_type == "token" {
+        // TODO: Distribute attestation token (Passport).
+        return HttpResponse::NotFound()
+            .message_body(BoxBody::new(
+                "Token resource is unsupported now".to_string(),
+            ))
+            .unwrap();
+    }
+
+    log::info!("Resource description: {:?}", &resource_description);
+
+    if session.tee_public_key().is_none() {
+        return HttpResponse::InternalServerError()
+            .message_body(BoxBody::new("TEE Pubkey not found"))
+            .unwrap();
+    }
+
+    match secret_resource(
+        session.tee_public_key().unwrap(),
+        repository.get_ref(),
+        resource_description,
+    ) {
+        Ok(response) => HttpResponse::Ok()
+            .content_type("application/json")
+            .body(serde_json::to_string(&response).unwrap()),
+        Err(e) => HttpResponse::InternalServerError()
+            .message_body(BoxBody::new(format!("Get Resource failed: {e}")))
+            .unwrap(),
+    }
 }
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -19,13 +19,18 @@ extern crate uuid;
 use actix_web::{middleware, web, App, HttpServer};
 use anyhow::Result;
 use attestation_service::AttestationService;
+use config::Config;
 use semver::{BuildMetadata, Prerelease, Version, VersionReq};
 use std::net::SocketAddr;
 use std::sync::Arc;
 
 use crate::session::SessionMap;
 
+/// KBS config
+pub mod config;
+
 mod http;
+mod resource;
 mod session;
 
 static KBS_PREFIX: &str = "/kbs";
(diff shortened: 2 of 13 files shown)


src/api_server/src/http.rs
```diff
@@ -115,3 +118,99 @@ pub(crate) async fn attest(
...
+
+            log::info!("Cookie {} request to get resource", session.id());
+
+            if !session.is_expired() {
+                if session.is_authenticated() {
+                    let mut resource_path =
+                        PathBuf::from(repo_dir_path.into_inner().to_str().unwrap_or_default());
+
+                    let resource_repo = request.match_info().get("repository").unwrap_or("default");
+                    let resource_type = request.match_info().get("type").unwrap();
```
Xynnn007, line 141: Could `unwrap()` here cause the server crash?

src/api_server/src/http.rs
```diff
@@ -115,3 +118,99 @@ pub(crate) async fn attest(
...
+                        resource_tag
+                    );
+
+                    // Resource file path:
+                    // `/opt/confidential-containers/kbs/repository/{repository}/{type}/{tag}`
+                    let ref_resource_path =
+                        format!("{resource_repo}/{resource_type}/{resource_tag}");
+                    resource_path.push(ref_resource_path);
+
+                    match ::std::fs::read(&resource_path) {
```
Xynnn007, line 166: Would it be a better choice to define a trait probably like
```rust
trait XXX {
    async fn index_resource(path: &str) -> Result<Vec<u8>>;
}
```

And impl the local fs as this PR, to leave extensibility for other potential storage backends

src/api_server/src/http.rs
```diff
@@ -115,3 +118,99 @@ pub(crate) async fn attest(
...
+    repo_dir_path: web::Data<PathBuf>,
+    map: web::Data<SessionMap<'_>>,
+) -> HttpResponse {
+    let error_info = if let Some(cookie) = request.cookie(KBS_SESSION_ID) {
+        if let Some(locked_session) = map.sessions.read().await.get(cookie.value()) {
+            let session = locked_session.lock().await;
+
+            log::info!("Cookie {} request to get resource", session.id());
+
+            if !session.is_expired() {
```
Xynnn007, line 135: Shall we change this into `if session.is_expired()` to avoid the nested condition statement blocks?

src/api_server/src/http.rs
```diff
@@ -115,3 +118,99 @@ pub(crate) async fn attest(
...
+                                    .message_body(BoxBody::new("TEE Pubkey not found"))
+                                    .unwrap();
+                            }
+                        },
+                        Err(e) => {
+                            return HttpResponse::NotFound()
+                                .message_body(BoxBody::new(format!(
+                                    "Read resource file failed: {e}"
+                                )))
+                                .unwrap();
```
Xynnn007, line 195: Could the `unwrap()` here cause the server crash?

src/api_server/src/http.rs
```diff
@@ -115,3 +118,99 @@ pub(crate) async fn attest(
...
+            let session = locked_session.lock().await;
+
+            log::info!("Cookie {} request to get resource", session.id());
+
+            if !session.is_expired() {
+                if session.is_authenticated() {
+                    let mut resource_path =
+                        PathBuf::from(repo_dir_path.into_inner().to_str().unwrap_or_default());
+
+                    let resource_repo = request.match_info().get("repository").unwrap_or("default");
```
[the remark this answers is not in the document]
jialez0: This is the default repository name :-)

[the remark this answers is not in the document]
Xynnn007: Yea, sorry for not noticing that KBS protocol takes a default repository `"default"`

jialez0:
@Xynnn007 Thank you for your review and comments. These are some very good suggestions, but I think the modifications needed to implement them are beyond the scope of this PR. Let's remember them and optimize them in the subsequent PR!

src/api_server/src/http.rs
sameo: That's definitely a good idea, but I'd propose to leave that for a separate PR?

src/api_server/src/crypto.rs
```diff
@@ -0,0 +1,59 @@
...
+    aead::{Aead, KeyInit, OsRng},
+    Aes256Gcm, Nonce,
+};
+use anyhow::{anyhow, Result};
+use kbs_types::{Response, TeePubKey};
+use rsa::pkcs8::DecodePublicKey;
+use rsa::{PaddingScheme, PublicKey, RsaPublicKey};
+use serde_json::json;
+
+const RSA_KEY_TYPE: &str = "RSA1_5";
```
sameo, line 16: According to the [JWA RFC](https://www.rfc-editor.org/rfc/rfc7518.html#page-28), `kty` should be `RSA`.

src/api_server/src/crypto.rs
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) 2022 by Alibaba.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
```
sameo, line 4: Could we call that file `resource.rs`? And then make it focus on generating resources:

- `fn secret_resource(tee_pub_key: TeePubKey, resource_byte: Vec<u8>) -> Result<Response>`
- `fn token_resource(tee_pub_key: TeePubKey) -> Result<Token>`

The fact that we're using crypto to generate those resources is an implementation detail imho.

src/api_server/src/http.rs
```diff
@@ -162,11 +163,37 @@ pub(crate) async fn resource(
                         format!("{resource_repo}/{resource_type}/{resource_tag}");
                     resource_path.push(ref_resource_path);
 
-                    if let Ok(resource_byte) = ::std::fs::read(&resource_path) {
-                        // TODO
-                        return HttpResponse::NotFound().finish();
-                    } else {
-                        return HttpResponse::NotFound().finish();
+                    match ::std::fs::read(&resource_path) {
+                        Ok(resource_byte) => match session.tee_public_key() {
```
sameo, line 167: Adding a couple of `Session` methods: `secret_resource()` and `token_resource()` that would call into the `resource.rs` API and return an HTTP response would make the code below more compact and easier to read.

src/api_server/src/http.rs
Xynnn007: > That's definitely a good idea, but I'd propose to leave that for a separate PR?

Agreed!

jialez0:
@sameo @Xynnn007 Thanks for reviewing! Now this PR is updated and your comments have been resolved, please take a look again :-)

src/api_server/src/http.rs
```diff
@@ -115,3 +117,84 @@ pub(crate) async fn attest(
...
 }
+
+/// GET /resource/{repository}/{type}/{tag}
+/// GET /resource/{type}/{tag}
+pub(crate) async fn resource(
+    request: HttpRequest,
+    repository: web::Data<Arc<dyn Repository + Send + Sync>>,
+    map: web::Data<SessionMap<'_>>,
+) -> HttpResponse {
+    let error_info = if let Some(cookie) = request.cookie(KBS_SESSION_ID) {
```
Xynnn007, line 128: Shall we change the logic here to something like following to reduce the nested conditions further?
```rust
let cookie = match request.cookie(KBS_SESSION_ID) {
    None => {
        log::error!("Missing KBS cookie");
        return HttpResponse::Unauthorized().json(kbs_error_info(ErrorInformationType::MissingCookie, ""));
    }
    Some(c) => c,
};

let locked_session = match map.sessions.read().await.get(cookie.value()) {
    None => {
        log::error!("Invalid KBS cookie {}", cookie.value());
        return HttpResponse::Unauthorized().json(kbs_error_info(ErrorInformationType::InvalidCookie, cookie.value()));
    }
    Some(ls) => ls,
};

...
```

src/api_server/src/lib.rs
```diff
@@ -82,14 +90,27 @@ impl ApiServer {
...
+            .config
+            .repository_type
+            .to_repository(&self.config.repository_description)?;
+
         HttpServer::new(move || {
             App::new()
                 .wrap(middleware::Logger::default())
                 .app_data(web::Data::clone(&sessions))
                 .app_data(web::Data::clone(&attestation_service))
+                .app_data(web::Data::new(repository.clone()))
```
Xynnn007, line 103: As the ownership of `repository` is transferred to this function, we might delete the `clone` here to reduce the memory usage.

jialez0: If we don't use clone here:

``` 
error[E0507]: cannot move out of `repository`, a captured variable in an `Fn` closure
   --> src/api_server/src/lib.rs:103:42
    |
93  |         let repository = self
    |             ---------- captured outer variable
...
98  |         HttpServer::new(move || {
    |                         ------- captured by this `Fn` closure
...
103 |                 .app_data(web::Data::new(repository))
    |                                          ^^^^^^^^^^ move occurs because `repository` has type `Arc<dyn Repository + Send + Sync>`, which does not implement the `Copy` trait

For more information about this error, try `rustc --explain E0507`.
```

Xynnn007: Sorry, you're right. I made a mistake, :-)

sameo:
> @sameo @Xynnn007 Thanks for reviewing! Now this PR is updated and your comments have been resolved, please take a look again :-)

Thanks a lot @jialez0, this looks pretty good now.

sameo (approved):
LGTM. Let's merge it when @Xynnn007 is fine with it as well.

jialez0:
@sameo @Xynnn007 Added a Client cmdline tool for test :-)

jialez0:
@sameo I am trying to debug the complete process of the whole path of cc-kbc ->kbs ->AS, using sample TEE. This may take some time. In addition, I encountered some problems when using sample TEE for testing. The root of the problem is that the `Tee` enumeration type in the `kbs-types` crate does not contain "sample". I have sent a [PR to virtee/kbs-types](https://github.com/virtee/kbs-types/pull/13), and I will temporarily use the `kbs-types` in the downstream version before that PR is merged...

Xynnn007 (approved):
Thanks, LGTM, great!

jialez0:
@sameo Now this PR can be merged!

PR #30, opened 2023-02-02
Feat resource registration

Xynnn007:
This PR adds the resource registration api. Refer to the implementation of an image registry, use http POST seems a good choice.

However, as confidential resource insertion is different from retrievement, it should ensure the caller is either the owner of the KBS or the owner of the repository. For test, This pr wraps all registration-related code into a feature `unsafe`, meaning this is only for test. 

Close https://github.com/confidential-containers/kbs/pull/30

### DIFF
--- a/src/api_server/Cargo.toml
+++ b/src/api_server/Cargo.toml
@@ -14,6 +14,7 @@ no-verifier = ["attestation-service/in-toto", "attestation-service/rvps-server"]
 actix-web = { version = "4", features = ["rustls"] }
 aes-gcm = "0.10.1"
 anyhow.workspace = true
+async-trait.workspace = true
 attestation-service.workspace = true
 base64.workspace = true
 env_logger.workspace = true
--- a/src/api_server/Cargo.toml
+++ b/src/api_server/Cargo.toml
@@ -12,12 +12,14 @@ no-verifier = ["attestation-service/in-toto", "attestation-service/rvps-server"]
 
 [dependencies]
 actix-web = { version = "4", features = ["rustls"] }
+actix-web-httpauth = "0.8.0"
 aes-gcm = "0.10.1"
 anyhow.workspace = true
 async-trait.workspace = true
 attestation-service.workspace = true
 base64.workspace = true
 env_logger.workspace = true
+jwt-simple = "0.11.4"
 lazy_static = "1.4.0"
 log.workspace = true
 kbs-types = { git = "https://github.com/virtee/kbs-types.git", rev = "13b9378f" }
--- a/src/api_server/Cargo.toml
+++ b/src/api_server/Cargo.toml
@@ -32,3 +32,6 @@ strum = "0.24.1"
 strum_macros = "0.24.1"
 tokio.workspace = true
 uuid = { version = "1.2.2", features = ["serde", "v4"] }
+
+[dev-dependencies]
+tempfile.dev-workspace = true
--- a/src/api_server/src/http.rs
+++ b/src/api_server/src/http.rs
@@ -202,7 +202,9 @@ pub(crate) async fn resource(
         session.tee_public_key().unwrap(),
         repository.get_ref(),
         resource_description,
-    ) {
+    )
+    .await
+    {
         Ok(response) => HttpResponse::Ok()
             .content_type("application/json")
             .body(serde_json::to_string(&response).unwrap()),
--- a/src/api_server/src/http.rs
+++ b/src/api_server/src/http.rs
@@ -141,7 +141,7 @@ pub(crate) async fn attest(
 
 /// GET /resource/{repository}/{type}/{tag}
 /// GET /resource/{type}/{tag}
-pub(crate) async fn resource(
+pub(crate) async fn get_resource(
     request: HttpRequest,
     repository: web::Data<Arc<dyn Repository + Send + Sync>>,
     map: web::Data<SessionMap<'_>>,
--- a/src/api_server/src/http.rs
+++ b/src/api_server/src/http.rs
@@ -4,11 +4,13 @@
 
 use actix_web::{body::BoxBody, web, HttpRequest, HttpResponse};
 use attestation_service::AttestationService;
+use jwt_simple::prelude::Ed25519PublicKey;
 use kbs_types::{Attestation, Challenge, ErrorInformation, Request};
 use std::sync::Arc;
 use strum_macros::EnumString;
 use tokio::sync::{Mutex, RwLock};
 
+use crate::auth::validate_auth;
 use crate::resource::{get_secret_resource, set_secret_resource, Repository, ResourceDesc};
 use crate::session::{Session, SessionMap, KBS_SESSION_ID};
 
@@ -22,6 +24,7 @@ pub enum ErrorInformationType {
     MissingCookie,
     UnAuthenticatedCookie,
     VerificationFailed,
+    JWTVerificationFailed,
 }
 
 fn kbs_error_info(error_type: ErrorInformationType, detail: &str) -> ErrorInformation {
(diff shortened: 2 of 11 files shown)


Xynnn007:
This way seems a little weird that the API is open to all people. Another choice is to also open an gRPC server only listening to localhost requests to handle registration requests, in this way to avoid unexpected access from the anonymous remotely. I'm not sure about this, plz share your views and let me do some change.

Xynnn007:
cc @sameo @jialez0

jialez0:
Maybe we can add a very simple signup and login API, for example:

```rust
type UserMap = HashMap<String, String>;
struct UserInfo {
    username: String,
    password: String,
}

/// POST /signup
pub(crate) async fn signup(
    user_info: web::Json<UserInfo>,
    users_list: web::Data<UserMap>,
    timeout: web::Data<i64>,
) -> HttpResponse {}

/// POST /login
pub(crate) async fn signup(
    user_info: web::Json<UserInfo>,
    users: web::Data<UserMap>,
    sessions: web::Data<SessionMap<'_>>,
    timeout: web::Data<i64>,
) -> HttpResponse {}
```

If users want to register resources, they must first log in and get cookie, and then post their resources with cookie.

Xynnn007:
> @Xynnn007 Maybe we can add a simple signup and login API, for example:
> 
> ```rust
> type UserMap = HashMap<String, String>;
> struct UserInfo {
>     username: String,
>     password: String,
> }
> 
> /// POST /signup
> pub(crate) async fn signup(
>     user_info: web::Json<UserInfo>,
>     users_list: web::Data<UserMap>,
>     timeout: web::Data<i64>,
> ) -> HttpResponse {}
> 
> /// POST /login
> pub(crate) async fn login(
>     user_info: web::Json<UserInfo>,
>     users: web::Data<UserMap>,
>     sessions: web::Data<SessionMap<'_>>,
>     timeout: web::Data<i64>,
> ) -> HttpResponse {}
> ```
> 
> If users want to register resources, they must first log in to get cookie, and then post their resources with cookie.
> 
> Each user can only modify the resources in the repository with username as the name. For example, user `{username: Alice, password: 123456}` can only post resources like `/kbs/v0/resource/Alice/<type>/<tag>`.

That's a good idea at my first impression. In my understanding user cookie is a kind of user token to show the identity of the user. Maybe we should talk about the user token here furthur, about how it can co-work with KBS/Pod/Tenant identity machinism to let only specific pod/user can access specific repo (like `policy.json`, etc)

src/api_server/Cargo.toml
```diff
@@ -25,3 +26,9 @@ strum = "0.24.1"
 strum_macros = "0.24.1"
 tokio.workspace = true
 uuid = { version = "1.2.2", features = ["serde", "v4"] }
+
+[features]
+default = []
+
+# Functionalities under this feature are not safe for the trust model
+experimental = []
```
jialez0, line 34: It is better to use a clearer and clear feature name like: `regiester-resource`

src/api_server/src/http.rs
```diff
@@ -198,14 +194,47 @@ pub(crate) async fn resource(
...
 }
+
+/// POST /resource/{repository}/{type}/{tag}
+/// POST /resource/{type}/{tag}
+/// TODO: This API surely needs identification check, only the owner of this repo
+/// or the owner of this KBS can call. The identification mechinism remains
+/// undesigned and unimplemented. Now it can be called by anyone, which is very
+/// dangerous and only for test.
+#[cfg(feature = "experimental")]
+pub(crate) async fn registry_resource(
```
jialez0, line 218: `registry_resource` -> `register_resource`

src/api_server/src/resource/mod.rs
```diff
@@ -24,51 +26,83 @@ const RSA_ALGORITHM: &str = "RSA1_5";
...
 /// Interface of a `Repository`.
+#[async_trait::async_trait]
 pub trait Repository {
     /// Read secret resource from repository.
-    fn read_secret_resource(&self, resource_desc: ResourceDesc) -> Result<Vec<u8>>;
+    async fn read_secret_resource(&self, resource_desc: ResourceDesc<'_>) -> Result<Vec<u8>>;
+
+    /// Registry secret resource into repository. All data from the `reader` will be read
+    /// and written to where `resource_desc` points to logically.
+    async fn registry_secret_resource<R: AsyncBufRead + Send + Unpin>(
```
jialez0, line 36: `registry_secret_resource` -> `register_secret_resource`

src/api_server/src/lib.rs
```diff
@@ -90,10 +91,10 @@ impl ApiServer {
         let sessions = web::Data::new(SessionMap::new());
         let http_timeout = self.http_timeout;
 
-        let repository = self
-            .config
-            .repository_type
-            .to_repository(&self.config.repository_description)?;
+        let repository = RepositoryType::from_name(
```
jialez0, line 94: Why can't we directly store the instanced repository instead of `RepositoryType`?  I really think the original plan is better.

sameo:
@Xynnn007 Can we hold on this PR until we get HTTPS (#41) support in?
Then with HTTPS enabled by default, we can start thinking about authenticated APIs, and I suggest we use JWT for that purpose. Then we can add an `experimental` or `insecure` feature gating the ability to use any random JWT to provision a secret resource.

Xynnn007:
Sorry to make the `async` in a single commit seems a hard operation. Let me rewrite this PR and reorganize into little commits

Xynnn007:
Hi @sameo @jialez0 , as this PR seems huge enough, I'd rather another PR to add a `insecure` feature to switch JWK verification and the `user_public_key` launch parameter. How do you like?

jialez0 (approved):
LGTM.

src/api_server/src/http.rs
```diff
@@ -198,14 +201,52 @@ pub(crate) async fn resource(
...
     }
 }
+
+/// POST /resource/{repository}/{type}/{tag}
+/// POST /resource/{type}/{tag}
+///
+/// TODO: This API surely needs identification check, only the owner of this repo
+/// or the owner of this KBS can call. The identification mechinism remains
+/// undesigned and unimplemented. Now it can be called by anyone, which is very
+/// dangerous and only for test.
```
sameo, line 224:
```suggestion
/// TODO: Although this endpoint is authenticated through a JSON Web Token (JWT),
/// only identified users should be able to get a JWT and access it.
/// At the moment user identification is not supported, and the KBS CLI `--user-public-key`
/// defines the authorized user for that endpoint. In other words, any JWT signed with the
/// user's private key will be authenticated.
/// JWT generation and user identification is unimplemented for now, and thus this endpoint
/// is insecure and is only meant for testing purposes.
```

src/api_server/src/http.rs
```diff
@@ -198,14 +201,52 @@ pub(crate) async fn resource(
...
+            .get("repository")
+            .unwrap_or("default")
+            .to_string(),
+        resource_type: request.match_info().get("type").unwrap().to_string(),
+        resource_tag: request.match_info().get("tag").unwrap().to_string(),
+    };
+
+    match set_secret_resource(&repository, resource_description, data.as_ref()).await {
+        Ok(_) => HttpResponse::Ok().content_type("application/json").body(""),
+        Err(e) => internal!(format!("Registry Resource failed: {e}")),
```
sameo, line 250:
```suggestion
        Err(e) => internal!(format!("Resource registration failed: {e}")),
```

src/kbs/src/main.rs
```diff
@@ -45,6 +45,10 @@ struct Cli {
     /// KBS config file path.
     #[arg(default_value_t = String::default(), short, long)]
     config: String,
+
+    /// User's public key to verify JWT
```
sameo, line 49:
```suggestion
    /// Public key used to authenticate the resource registration endpoint token (JWT).
    /// Only JWTs signed with the corresponding private keys will be authenticated.
```

src/kbs/src/main.rs
```diff
@@ -45,6 +45,10 @@ struct Cli {
...
+    #[arg(long)]
+    user_public_key: PathBuf,
```
sameo, line 51:
```suggestion
    auth_public_key: PathBuf,
```

sameo (commented):
Looks pretty good to me, thanks for quickly putting that together @Xynnn007.

I only have a few, wording/syntax nits.

PR #32, opened 2023-02-22
Update Cargo.lock and update Client Tool dependency.

jialez0:
cc @sameo


sameo:
@jialez0 Why do we need to entirely remove `Cargo.lock`?

jialez0:
> @jialez0 Why do we need to entirely remove `Cargo.lock`?

@sameo This is my mistake. Since the product of compiling KBS is an executable binary, `Cargo.lock` is still needed.

Now this PR is ready for review, you can have a look:-)

sameo (approved):

PR #34, opened 2023-02-28
Add Dockerfile to build and deploy KBS expediently

jialez0:
Resolve: #33 

cc @sameo

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,47 @@
+FROM rust:1.67-slim
+
+WORKDIR /usr/src/kbs
+COPY . .
+
+# Install Build Dependencies
+RUN apt-get update && apt-get install apt-utils
+RUN apt-get install -y \
+clang \
+cmake \
+curl \
+gnupg \
+libclang-dev \
+libprotobuf-dev \
+libssl-dev \
+make \
+pkg-config \
+protobuf-compiler \
+wget
+RUN wget https://go.dev/dl/go1.20.1.linux-amd64.tar.gz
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM ubuntu:latest
```
sameo, line 1: 2 things:

1. We may want to stick to a given tag
2. We may want to use the rust image instead, so that we can run a regular `cargo build` and `cargo install`. Those images are debian based, so it should work with all the apt based instructions here. See the 1.67.1-slim-buster tag at https://registry.hub.docker.com/_/rust/ for example

README.md
```diff
@@ -37,6 +37,20 @@ A custom, [JSON-formatted configuration file](src/config.rs) can be used:
...
+Build Container image:
+
+```shell
+DOCKER_BUILDKIT=1 docker build -t kbs:latest . -f Dockerfile
+```
+
+Deploy KBS:
+
+```shell
+docker run -it --net host --name=kbs kbs:latest kbs -s <IP>:<PORT>
```
sameo, line 51: Would:

```
docker run -it --name =kbs --ip=<IP> -p <PORT>:<PORT> kbs:latest kbs -s <IP>:<PORT>
```

work as well and avoid using host networking?

sameo: Would `-p 8081:8081` help then?

sameo (approved):
Very nice, thanks a lot @jialez0 !

PR #35, opened 2023-03-01
Optimize Dockerfile and fix some nits.

jialez0:
1. Fix `sed` expression to avoid Error `sed: -e expression #1, char 35: unknown option to 's'` (refer to [this](https://netjunky.net/tag/linux/page/2/))
2. Separate the building environment and the running environment, to reduce the container image size.


jialez0:
cc @sameo

sameo (approved):

PR #40, opened 2023-03-09
Update Client Tool to adapt KBS resource URI.

jialez0:
Updated the Client Tool to use the [KBS resource URI](https://github.com/confidential-containers/attestation-agent/blob/main/docs/KBS_URI.md) format defined by the Attention Agent.


jialez0:
cc @sameo

sameo (approved):

PR #50, opened 2023-03-16
Add insecure API support

Xynnn007:


src/api_server/src/http.rs
```diff
@@ -232,9 +232,10 @@ pub(crate) async fn set_resource(
     request: HttpRequest,
     data: web::Bytes,
     user_pub_key: web::Data<Ed25519PublicKey>,
+    insecure: web::Data<bool>,
     repository: web::Data<Arc<RwLock<dyn Repository + Send + Sync>>>,
 ) -> HttpResponse {
-    if let Err(e) = validate_auth(&request, user_pub_key.get_ref()) {
+    if let Err(e) = validate_auth(&request, user_pub_key.get_ref(), *insecure.get_ref()) {
```
sameo, line 238: If we're in insecure, we dont even need to try and validate the request.

What about:
```rust
 if !*insecure.get_ref() {
      if let Err(e) = validate_auth(&request, user_pub_key.get_ref()) {
      }
}
```

Xynnn007: yes, thanks. Fixed it

Xynnn007:
I added two commits, which changes `user_public_key` into a optional parameter.

sameo (approved):

jialez0 (approved):

PR #59, opened 2023-03-27
docker-compose: add default rersources

Xynnn007:
when `simple-signing` is enabled, by default the file `default/sigstore-config/test` will be retrieved

when `security_validate`, by default the file
`default/security-policy/test` will be retrieved

Those two default files are for quick start

Fixes: #58

### DIFF
--- a/data/kbs-storage/default/sigstore-config/test
+++ b/data/kbs-storage/default/sigstore-config/test
@@ -1,0 +1,4 @@
+docker:
+    quay.io/kata-containers/confidential-containers:
+        sigstore: file:///etc/containers/quay_verification/signatures
+        sigstore-staging: file:///etc/containers/quay_verification/signatures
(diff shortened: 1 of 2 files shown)


sameo (approved):

data/kbs-storage/default/sigstore-config/test
```diff
@@ -0,0 +1,4 @@
+docker:
+    quay.io/kata-containers/confidential-containers:
+        sigstore: file:///etc/containers/quay_verification/signatures
+        sigstore-staging: file:///etc/containers/quay_verification/signatures
```
fitzthum, line 4: Are these paths populated via attestation-agent as well or do they need to be pre-provisioned to initrd?

Xynnn007: These paths will not be created at the runtime.

If `simple-signing` is actually used (in `policy.json`), these paths and the related signature files need to be pre-provisioned to initrd. (As we only support to look for signature files from local filesystem now for `simple-signing`)

If `simple-signing` is not used. These paths will not be accessed during the runtime, so we do not need to create them or pre-provision to  initrd.

jialez0 (approved):

PR #61, opened 2023-03-27
Use kbs-types v0.2.0

### LINKED ISSUE #60: Update kbs-types crate to v0.2.0
At the moment we reference the git repo of kbs-types instead of the published crate. This is seemingly due to a breaking change in v0.2.0, which instead of a pem string puts modulus and exponent in the TeePubKey. 

relates to: https://github.com/confidential-containers/attestation-agent/issues/169

mkulke:
Similar to [this PR](https://github.com/confidential-containers/attestation-agent/pull/170) on attestation-agent, we need to adjust `TeePubKey` handling to reflect the updated struct on virtee/kbs-types. The fix needs to be coupled with the respective PRs:

* [ ] https://github.com/confidential-containers/attestation-agent/pull/170
* [ ] https://github.com/confidential-containers/attestation-service/pull/78

Note: This should only be merged once the AA/AS fixes are merged. The last commit is for testing the build, using git revisions pointing to the other 2 PRs.

fixes #60

### DIFF
--- a/docs/kbs.yaml
+++ b/docs/kbs.yaml
@@ -326,21 +326,21 @@ components:
 
     PublicKey:
       required:
-        - kty
         - alg
-        - k
+        - k-mod
+        - k-exp
       properties:
-        kty:
-          type: string
-          description: Key type
         alg:
           type: string
           description: Key Algorithm
-        k:
+        k-mod:
           type: string
-          description: Key data
+          description: Key modulus
+        k-exp:
+          type: string
+          description: Key exponent
       description: >-
-        A JSON Web Key (https://www.rfc-editor.org/rfc/rfc7517) formatted object.
+        A JSON Web Key (https://www.rfc-editor.org/rfc/rfc7517) formatted RSA Public Key.
 
     ErrorInformation:
       required:
--- a/docs/kbs_attestation_protocol.md
+++ b/docs/kbs_attestation_protocol.md
@@ -247,11 +247,14 @@ The [`Attestation`](#attestation) and the [Token Resource](#token-resource)
 `tee-pubkey` field contains a public key from a HW-TEE generated asymmetric
 key pair.
 
+This field follows the [JSON Web Key](https://www.rfc-editor.org/rfc/rfc7517)
+format.
+
 ``` json
 {
     "alg": "$key_algorithm",
-    "k_mod": "$pubkey modulus",
-    "k_exp": "$pubkey exponent"
+    "k-mod": "$pubkey modulus",
+    "k-exp": "$pubkey exponent"
 }
 ```
 
--- a/docs/kbs_attestation_protocol.md
+++ b/docs/kbs_attestation_protocol.md
@@ -246,14 +246,12 @@ in `alg`.
 The [`Attestation`](#attestation) and the [Token Resource](#token-resource)
 `tee-pubkey` field contains a public key from a HW-TEE generated asymmetric
 key pair.
-This field follows the [JSON Web Key](https://www.rfc-editor.org/rfc/rfc7517)
-format:
 
 ``` json
 {
-    "kty": "$key_type",
     "alg": "$key_algorithm",
-    "k": "public_key"
+    "k_mod": "$pubkey modulus",
+    "k_exp": "$pubkey exponent"
 }
 ```
 
--- a/src/api_server/src/resource/mod.rs
+++ b/src/api_server/src/resource/mod.rs
@@ -10,7 +10,6 @@ use aes_gcm::{
 use anyhow::{anyhow, Result};
 use kbs_types::{Response, TeePubKey};
 use local_fs::{LocalFs, LocalFsRepoDesc};
-use num_traits::Num;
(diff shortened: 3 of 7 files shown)


src/api_server/src/resource/mod.rs
```diff
@@ -106,8 +105,13 @@ pub(crate) async fn get_secret_resource(
         .encrypt(nonce, resource_byte.as_slice())
         .map_err(|e| anyhow!("AES encrypt Resource payload failed: {:?}", e))?;
 
-    let rsa_pub_key = RsaPublicKey::from_public_key_pem(&tee_pub_key.k)
-        .map_err(|e| anyhow!("Parse TEE pubkey PEM failed: {:?}", e))?;
+    let n = BigUint::from_str_radix(&tee_pub_key.k_mod, 10)
+        .map_err(|e| anyhow!("Parse TEE pubkey modulus failed: {:?}", e))?;
+    let e = BigUint::from_str_radix(&tee_pub_key.k_exp, 10)
+        .map_err(|e| anyhow!("Parse TEE pubkey exponent failed: {:?}", e))?;
```
sameo, line 111: Both k_mod and k_exp are `base64` formatted so I think you want to decode that first.

Also, I'm trying to push https://github.com/virtee/kbs-types/pull/19 to make it use JWA compliant JSON payloads, if we decide to drop the more generic `k` field to describe a key.

mkulke: Ah that makes sense. It wasn't documented on the types what the encoding is supposed to be, afaict. Will update this on and the AA PR accordingly. 

the n/e renaming should be transparent to this code, right?

sameo: > Ah that makes sense. It wasn't documented on the types what the encoding is supposed to be, afaict. Will update this on and the AA PR accordingly.
> 

Talking about documentation, would you mind updating the kbs_protocol as well (when/if the pending PR gets merged) at https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#public-key ?

> the n/e renaming should be transparent to this code, right?

It should. I did not change the structure type, only the JSON serialization format.

sameo: > > Talking about documentation, would you mind updating the kbs_protocol as well (when/if the pending PR gets merged) at https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#public-key ?
> 
> yup, will do. We'd need a kbs-types release for that too, right?

Thanks for the doc fix.
We'd need a new `kbs-types` release, provided the PR is accepted. The new release will include the `AzSnpVtpm` type as well, so would you prefer to wait for that release or to just go ahead with this set of PRs?

sameo: > > Thanks for the doc fix. We'd need a new `kbs-types` release, provided the PR is accepted. The new release will include the `AzSnpVtpm` type as well, so would you prefer to wait for that release or to just go ahead with this set of PRs?
> 
> Not sure about the kbs-types release cadence. I'd recommend to go ahead with this set of fixes, limiting the scope to an update to kbs-types#v0.2.0 and then deal with `n/e` and `AzSnpVtpm` individually. There is already a lot of untangling and cognitive gymnastics required to update the kbs-types dep in concert across three repos.

Fine with me. Moving to 0.3.0 should be less painful as hopefully there won't be any API breaking changes.

docs/kbs_attestation_protocol.md
```diff
@@ -246,14 +246,12 @@ in `alg`.
...
-This field follows the [JSON Web Key](https://www.rfc-editor.org/rfc/rfc7517)
-format:
 
 ``` json
 {
-    "kty": "$key_type",
     "alg": "$key_algorithm",
-    "k": "public_key"
+    "k_mod": "$pubkey modulus",
+    "k_exp": "$pubkey exponent"
```
sameo, line 254: Apologies for asking for another change, but could you also please fix https://github.com/confidential-containers/kbs/blob/main/docs/kbs.yaml#L327 accordingly?

mkulke: Oh, well spotted, thx. Also there is a typo in the above docs, it's serialized with a hypen (`k-mod`). I changed the openapi definitions to `TeePubKey` and adjusted the description, since it's no longer supposed to be a JWK I think. 

The n/e change should be easier to apply, since we know now what needs to be changed

docs/kbs.yaml
```diff
@@ -322,25 +322,25 @@ components:
...
+        k-mod:
+          type: string
+          description: Key modulus
+        k-exp:
           type: string
-          description: Key data
+          description: Key exponent
       description: >-
-        A JSON Web Key (https://www.rfc-editor.org/rfc/rfc7517) formatted object.
+        An RSA Public Key for the TEE
```
sameo, line 343: The payload follows the JWK format, although we only support RSA keys:

```suggestion
        A JSON Web Key (https://www.rfc-editor.org/rfc/rfc7517) formatted RSA Public Key.
```

sameo (approved):

Xynnn007 (approved):

PR #62, opened 2023-03-28
Add openssl support for s390x

tnakaike:
This PR provides the openssl option for the KBS https server because rustls does not work on s390x.

For selection of rustls and openssl, `#[cfg(not(target_arch = "s390x"))]` and `#[cfg(arget_arch = "s390x")]` are used. One curious thing is that `[target.'cfg(not(target_arch = "s390x"))'.dependencies]` does not work in src/api_server/Cargo.toml. So, we still need to comment out the dependencies manually to build kbs on s390x.

I appreciate if anyone give an advice to fix the above issue.

### DIFF
--- a/src/api_server/Cargo.toml
+++ b/src/api_server/Cargo.toml
@@ -7,19 +7,21 @@ documentation.workspace = true
 edition.workspace = true
 
 [features]
-default = ["native-as"]
+default = ["native-as", "rustls"]
 native-as = ["attestation-service/default"]
 native-as-no-verifier = ["attestation-service/rvps-native"]
 grpc-as = ["tonic", "tonic-build", "prost"]
+rustls = ["actix-web/rustls", "dep:rustls", "dep:rustls-pemfile"]
+openssl = ["actix-web/openssl", "dep:openssl"]
 
 [dependencies]
-actix-web = { version = "4", features = ["rustls"] }
+actix-web = "4"
 actix-web-httpauth = "0.8.0"
 aes-gcm = "0.10.1"
 anyhow.workspace = true
 async-trait.workspace = true
-as-types = { git = "https://github.com/confidential-containers/attestation-service.git", rev = "57c55db"}
-attestation-service = { git = "https://github.com/confidential-containers/attestation-service.git", default-features = false, rev = "57c55db", optional = true}
+as-types = { git = "https://github.com/confidential-containers/attestation-service.git", rev = "eb9c069"}
+attestation-service = { git = "https://github.com/confidential-containers/attestation-service.git", default-features = false, rev = "eb9c069", optional = true}
 base64.workspace = true
 cfg-if = "1.0.0"
 env_logger.workspace = true
@@ -30,8 +32,8 @@ log.workspace = true
 prost = { version = "0.11", optional = true }
 rand = "0.8.5"
 rsa = "0.7.2"
-rustls = "0.20.8"
-rustls-pemfile = "1.0.2"
+rustls = { version = "0.20.8", optional = true }
+rustls-pemfile = { version = "1.0.2", optional = true }
 semver = "1.0.16"
 serde = { version = "1.0", features = ["derive"] }
 serde_json.workspace = true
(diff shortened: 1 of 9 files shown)


src/api_server/src/lib.rs
```diff
@@ -146,6 +154,25 @@ impl ApiServer {
...
+            .clone()
+            .ok_or_else(|| anyhow!("Missing certificate"))?;
+
+        let key_file = self.private_key
+            .clone()
+            .ok_or_else(|| anyhow!("Missing private key"))?;
+
+        let mut builder = SslAcceptor::mozilla_modern(SslMethod::tls()).map_err(|e| anyhow!(e.to_string()))?;
+        builder.set_private_key_file(key_file, SslFiletype::PEM).map_err(|e| anyhow!(e.to_string()))?;
+        builder.set_certificate_chain_file(cert_file).map_err(|e| anyhow!(e.to_string()))?;
```
Xynnn007, line 171: I'm not sure about this. `anyhow` gives a simple way to convert errors using `?`, could you try
```
let mut builder = SslAcceptor::mozilla_modern(SslMethod::tls())?;
builder.set_private_key_file(key_file, SslFiletype::PEM)?;
builder.set_certificate_chain_file(cert_file)?;
```

Xynnn007 (commented):
Overall, LGTM! Great work and thanks! We will disable `in-toto` by default in another PR and the features here could be deleted.

src/api_server/src/lib.rs
```diff
@@ -202,9 +229,13 @@ impl ApiServer {
...
 
         if !self.insecure {
-            let tls_config = self.tls_config()?;
-            http_server
-                .bind_rustls(&self.sockets[..], tls_config)?
+            #[cfg(not(target_arch = "s390x"))]
+            let tls_server = http_server.bind_rustls(&self.sockets[..], self.tls_config()?)?;
+
+            #[cfg(target_arch = "s390x")]
+            let tls_server = http_server.bind_openssl(&self.sockets[..], self.tls_config()?)?;
```
sameo, line 236: Could we please use the `cfg_if` crate here? https://crates.io/crates/cfg-if

tnakaike:
Thank you for your comments. I updated the commit to reflect your suggestion. Is there is any further problems, please let me know.

src/api_server/Cargo.toml
```diff
@@ -41,6 +39,15 @@ tokio.workspace = true
 tonic = { version = "0.8", optional = true }
 uuid = { version = "1.2.2", features = ["serde", "v4"] }
 
+[target.'cfg(not(target_arch = "s390x"))'.dependencies]
```
sameo, line 42: I think it would be more generic to define the openssl vs rustls independently from the architecture. If someone wants to build kbs with `openssl` they should build with a `--features openssl` option instead. The default should be rustls imho, but openssl would be a build time option that is not architecture dependent.

@Xynnn007 @jialez0 WDYT?

Xynnn007: Yes. I agree. As `rustls`-`ring` does not work on `s390x`, and `openssl` can. It will be helpful for folks working on that arch. 

We need to impl this from the AS side.

tnakaike: @sameo @Xynnn007 @jialez0 I see. Let me add two features. Are the following feature names suitable?

- rustls
- openssl

Xynnn007: It is perfect to me!

tnakaike:
I added rustls and openssl features instead of architecture comparison. Further, I updated the revision of attestation service so that [my type error](https://github.com/confidential-containers/attestation-service/commit/f1ae7d2b4877b3fc71d91364107fc11b4add19f3) fix is included. Please review my code.

Thanks,

sameo (approved):
Thanks for the PR @tnakaike, this LGTM.

sameo:
FWIW I'm ok with that PR, but we should follow up with a CI addition to build with openssl.

tnakaike:
@sameo Thank you so much. It looks like there are some errors in CI. Let me fix them.

tnakaike:
I fixed `src/kbs/Cargo.toml` and changed the workflow for openssl CI. If you agree with the change in the workflow, please approve it. If not, please let me know. I will revert the change. Thanks,

tnakaike:
@jialez0 I fixed the CI issues. I hope that this is final. I realized that feature names are commonly used by `src/api_server` and `src/kbs`. So, we need to define the same features for them.

Issue #20, opened 2022-12-02
Support Token distribution and verification.


jialez0:
This is an enhancement and future work.

At present, we have designed this enhanced function in the KBS protocol: it can support clients to request [Token resources](https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#token-resource), and KBS will sign and distribute a Token representing the identity of the client after the success of the Attestation. If the Token is brought with the subsequent resource Request and verified by KBS, there is no need to do Attestation again. 

In addition, we should further extend the function of KBS to verify the token. In addition to verifying the token signed by KBS itself, we can also verify the token signed by the trusted third party Attestation Service trusted by users (for example, from Azure Attestation-Service, etc.).

It should be noted that we first need to implement a Session-Cookie based KBS. After its basic functions are available, the function described in this issue will be added as an additional option.

jialez0:
It should be noted that **our top priority is still to build a session based KBS**. After the basic functions are available, we can start to add the functions described in this issue.

jialez0:
It should be noted that we first need to implement a session based KBS. After its basic functions are available, the function described in this issue will be added as an additional option.

PR #74, opened 2023-04-13
Feat Attestation Token distribution.

jialez0:
Fixed: #20


Xynnn007:
cc @surajssd

docs/kbs_attestation_protocol.md
```diff
@@ -413,18 +413,15 @@ A POST request with the content of resource to `/kbs/v0/resource/<repository>/<t
 
 ### Token Resource
 
-Authenticated attesters can also request a token from the KBS.
+Authenticated attesters can also request an attestation token from the KBS.
```
sameo, line 416: What we're providing through that endpoint are attestation results, I think:

```suggestion
Authenticated attesters can also request an attestation result token from the KBS.
```

We probably should also update the section title from `Token Resource` to `Attestation Result`.

docs/kbs_attestation_protocol.md
```diff
@@ -413,18 +413,15 @@ A POST request with the content of resource to `/kbs/v0/resource/<repository>/<t
...
 Attesters can use tokens to request additional resources from external (i.e. not the KBS) services.
```
sameo, line 417:
```suggestion
Attesters can use the attestation result token to request additional resources from external services, a.k.a. relying parties. 

```

docs/kbs_attestation_protocol.md
```diff
@@ -413,18 +413,15 @@ A POST request with the content of resource to `/kbs/v0/resource/<repository>/<t
...
 
 ```
-/kbs/v0/resource/token/{tag}
+/kbs/v0/attestation-token
 ```
 
-Where the URL path parameter `tag` specify some special token requirements,
-it can be left blank and use default token.
-
 #### Token Format
```
sameo, line 425:
```suggestion
#### Attestation Result Format
```

docs/kbs_attestation_protocol.md
```diff
@@ -413,18 +413,15 @@ A POST request with the content of resource to `/kbs/v0/resource/<repository>/<t
...
 
 The token follows the [JSON web token](https://jwt.io/) standard and format.
```
sameo, line 427:
```suggestion
The provided attestation result follows the [JSON web token](https://jwt.io/) standard and format.
```

docs/kbs_attestation_protocol.md
```diff
@@ -438,92 +435,7 @@ eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.
...
-#### HTTP Response
-
-The token is included in the `/kbs/v0/resource/token/` `GET` HTTP response
-body, as JSON content:
-
-``` json-with-comments
-{
-    "token": "$serialized_token"
-}
-```
```
sameo, line 512: I suggest that we keep that part.

docs/kbs_attestation_protocol.md
```diff
@@ -438,92 +435,7 @@ eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.
...
-
-let jwt_claims_string = format!(r#"{{"exp": {}, "iat": {}, "iss": {}, "tee-pubkey": {}}}"#, exp, iat, iss, pubkey);
-let jwt_claims = base64_url::encode(&jwt_claims_string);
-
-let jws_signature_input = format!("{}.{}", jwt_header, jwt_claims);
-let jws_signature = base64_url::encode(rs256_key_pair.sign(&jws_signature_input));
-
-let serialized_token = format!("{}.{}.{}", jwt_header, jwt_claims, jwt_signature);
-```
+Where the custom claims in the payload is the attestation result from the Attention Service.
```
sameo, line 438:
```suggestion
The `JWT` payload is the set of attestation result custom Claims, as generated by the Attestation Service.
```

Also, shouldn't we require for the generated JWT to include those registered claims:

``` json
{
    "exp": 1568187398,
    "iat": 1568158598,
    "iss": "https://xxxxxx",
    "tee-pubkey": $pubkey
}
```
```

**Review by sameo — COMMENTED** [review]
Thanks a lot for this work @jialez0 , this is going to allow us to support the passport model and close a gap with KBS.

File: src/api_server/src/token/mod.rs [hunk 7]
```diff
@@ -0,0 +1,41 @@
...
+use anyhow::*;
+use serde::Deserialize;
+use serde_json::Value;
+use std::sync::Arc;
+use strum_macros::EnumString;
+use tokio::sync::RwLock;
+
+mod simple;
+
+pub trait TokenBroker {
```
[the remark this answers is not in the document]
jialez0: The use of a trait object is to enable us to implement more specialized token distribution modules, which can support other token formats (such as CWT), use external keys, and possibly implement token signing and distribution in TEE (such as token distribution modules in Intel SGX Enclave).

src/api_server/src/http/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+mod user_api;
+
+/// RESTful APIs that require Attestation and are protected by KBS Attestation Protocol
+pub use confidential_api::*;
+
+/// RESTful APIs that require user authentication
+pub use user_api::*;
+
+/// RESTful APIs that is public
+pub use public_api::*;
```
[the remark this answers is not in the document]
jialez0: I divide them into confidential, user, and public because calling some APIs of KBS need to go through the `R-C-A-R` four step attestation protocol, calling some APIs of KBS need to carry authentication tokens that represent the user's identity, and calling some APIs of KBS do not require any additional authentication. Therefore, organizing them according to this classification would make the code more maintainable. What do you think?

docs/kbs_attestation_protocol.md
jialez0: OK, I'll keep them :)

docs/kbs_attestation_protocol.md
jialez0: Yes, I agree. In the future, we can implement both the attestation result token and the pubkey registration token here, and I will keep the description in the document. Of course, this PR only implements the former.

src/api_server/src/token/mod.rs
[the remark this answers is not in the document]
sameo: Since the attestation service has a well defined attestation results formats: https://github.com/confidential-containers/attestation-service#attestation-result-format, I think it's safe to say that it's the KBS responsibility to extract claims from it and format them in different formats. So I agree with you, and a KBS deployment can decide which `AttestationTokenBroker` implementation it wants to use to format and distribute the attestation results.

src/api_server/src/token/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+    fn issue_attestation_token(&self, custom_claims: Value) -> Result<String>;
+
+    /// Verify an signed attestation token
+    fn verify_attestation_token(&self, token: String) -> Result<()>;
+
+    /// Get the x509 certificate
+    /// of the public key to verify attestation token.
+    /// This certificate can be self-signed.
+    /// Return certificate in DER format.
+    fn get_x509_certificate(&self) -> Result<Vec<u8>>;
```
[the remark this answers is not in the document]
sameo: Suggestion updated ;-)

src/api_server/src/http/mod.rs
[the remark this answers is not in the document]
sameo: I think splitting the API implementation into logical pieces makes a lot of sense and would help maintaining the code. And that made me think about how we'd want to split it, hence my proposal ;-)
And the distinction between RCAR related APIs and the rest of the API makes sense to me as well. But RCAR covers both `/auth` and `/attest`, which is what I tried to put under the `protocol.rs` file (although that one could be called `rcar.rs` as well).

The confidential/user/public split that you're proposing is based on credentials needed (although for example `/auth` is a public API that does not require authentication) rather than usage of the over all KBS interface. So I was proposing a usage based split instead: One part for the core KBS protocol (`{protocol|core|rcar}.rs`), that KBS users must go through if they want to be able to use the second part of the API (`attestation.rs`) in order to fetch any resource/item (secret or not) out of the KBS. 
Then we have part of our interface who is meant for administration/configuring the KBS itself (and that implicitly must be used by authenticated users), which is the `config.rs` part of my proposal.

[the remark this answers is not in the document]
jialez0: @sameo Good suggestion. Now we have three types of APIs: `core.rs`, `attestation.rs`, and `config.rs`. I think this is good and the division is clear. Thanks!

In addition, do we also need `public.rs` to organize public APIs that do not require attestation or user authentication? For example, getting cert chain.

src/api_server/src/token/simple.rs
```diff
@@ -0,0 +1,46 @@
...
+            .verify_token::<Value>(&token, None)?;
+        Ok(())
+    }
+
+    fn get_x509_certificate(&self) -> Result<Vec<u8>> {
+        let rcgen_key_pair = rcgen::KeyPair::from_pem(&self.key_pair.to_pem()?)?;
+        let mut cert_params = CertificateParams::new(vec!["CoCo-Attestation-Service".to_string()]);
+        cert_params.key_pair = Some(rcgen_key_pair);
+        let cert = Certificate::from_params(cert_params)?.serialize_der()?;
+        Ok(cert)
```
sameo, line 44: We may want to cache that certificat and avoid having to regenerate the same one for each call?

src/api_server/src/http/mod.rs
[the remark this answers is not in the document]
sameo: > @sameo Good suggestion. Now we have three types of APIs: `core.rs`, `attestation.rs`, and `config.rs`. I think this is good and the division is clear. Thanks!
> 

👍 

> In addition, do we also need `public.rs` to organize public APIs that do not require attestation or user authentication? For example, getting cert chain.

Yes, that makes sense to me.

jialez0:
@sameo Thanks for reviewing! Now the code is update and you can take a look again:-)

sameo (commented):
@jialez0 The PR looks good now, thanks a lot.
I had some further comments on the documentation part and the missing OpenAPI changes. I sent a PR instead for you at https://github.com/jialez0/kbs/pull/1
Please review it and feel free to squash it into your PR. Once that's done we could go ahead and merge, unless @Xynnn007 or @surajssd have further comments.

sameo (approved):

fitzthum:
Haven't had a chance to really review, but I'm glad we're adding this feature. Will be very useful for more complex workloads. Including the public key of the TEE in the token will prevent replay, right?

src/api_server/src/token/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+
+mod simple;
+
+pub trait AttestationTokenBroker {
+    /// Issue an signed attestation token with custom claims.
+    /// Return base64 encoded Token.
+    fn issue(&self, custom_claims: Value) -> Result<String>;
+
+    /// Verify an signed attestation token
+    fn verify(&self, token: String) -> Result<()>;
```
Xynnn007, line 20: The `verify` function here seems to be called by a different entity from KBS. The code of that entity will import this code? If so, do we need to make the `token` a separate crate?

src/api_server/src/session.rs
```diff
@@ -104,13 +113,63 @@ impl<'a> Session<'a> {
...
     pub fn set_tee_public_key(&mut self, key: TeePubKey) {
         self.tee_pub_key = Some(key)
     }
+
+    pub fn to_jwe(&self, payload_data: Vec<u8>) -> Result<Response> {
+        let tee_pub_key = self
+            .tee_public_key()
+            .ok_or_else(|| anyhow!("No TEE public Key"))?;
+        if tee_pub_key.alg != *RSA_ALGORITHM {
+            return Err(anyhow!("Unsupported TEE Pub Key type or algorithm"));
```
Xynnn007, line 133: Please change this into `bail!()` macro to make it dry
```suggestion
            bail!("Unsupported TEE Pub Key type or algorithm");
```

Xynnn007 (commented):
@jialez0 Thanks a lot for the great huge work! Also some comments...

sameo:
> Haven't had a chance to really review, but I'm glad we're adding this feature. Will be very useful for more complex workloads. Including the public key of the TEE in the token will prevent replay, right?

Correct. And yes, this is going to be useful for many workloads, including e.g. the ML ones that were presented last Thursday.

src/api_server/src/token/mod.rs
jepio, line 20: For `verify` to be useful to callers, it needs to return the claims from the token, and in particular the TEE pubkey (as that is what should be used to wrap secrets).

src/api_server/src/http/attestation.rs
```diff
@@ -0,0 +1,177 @@
...
+        format!("no attestation results generated")
+    );
+    let token_claims = bail_error_internal!(serde_json::to_value(&attestation_results));
+
+    let token = match token_broker.read().await.issue(token_claims) {
+        Ok(token) => token,
+        Err(e) => internal!(format!("Issue Attestation Token failed: {e}")),
+    };
+
+    match session.to_jwe(token.into_bytes()) {
```
jepio, line 96: Why would you encrypt the attestation token? The token can only be used by the holder of the tee-pubkey anyway, so there is no need to further encrypt it.

jepio (commented):
I like the general direction this is headed, I suggest some changes for better compatibility with other attestation services and i think we need to think about the big picture a bit.

To implement the passport model we'll need to have:
1. an actual "attestation-service" that acts as a RATS verifier. This could be implemented as a version of KBS which only exposes `/attest` and `/token-certificate-chain` endpoints, but giving it a proper name would be helpful for communication.
2.  a KBS that does not perform attestation itself and instead calls `token.verify()` acting as the RATS relying party.

In that case the relying-party-KBS could be stateless (no session/cookie), and the attestation token would be passed together with the `/resource/...` calls. The TEE would communicate separately with an attestation service and (1+) KBS.

Does this direction make sense?

src/api_server/src/session.rs
```diff
@@ -104,13 +113,63 @@ impl<'a> Session<'a> {
...
+            .map_err(|e| anyhow!("RSA encrypt sym key failed: {:?}", e))?;
+
+        let protected_header = json!(
+        {
+           "alg": RSA_ALGORITHM.to_string(),
+           "enc": AES_GCM_256_ALGORITHM.to_string(),
+        });
+
+        Ok(Response {
+            protected: serde_json::to_string(&protected_header)?,
```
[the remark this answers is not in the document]
jialez0: I don't have much experience in this area, will the header here be unsafe? Specifically, how should it be encrypted?

Additionally, this is not the scope of this PR, we can create a new PR to fix it after this PR is merged.

@sameo What do you think?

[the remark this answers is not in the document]
jepio: The JWE spec defines that the "protected" field should be integrity protected. This is "additional authenticated data" in AES-GCM, the resulting tag goes in the "tag" field. It does not need to be encrypted.

Otherwise we the field should be called "unprotected".

I'm OK if we address this later. This will be a breaking change.

jialez0:
@jepio Thanks for reviewing, now this PR is update and resolved your comments. Please review it again, thanks.

jialez0:
> * an actual "attestation-service" that acts as a RATS verifier. This could be implemented as a version of KBS which only exposes `/attest` and `/token-certificate-chain` endpoints, but giving it a proper name would be helpful for communication.
> * a KBS that does not perform attestation itself and instead calls `token.verify()` acting as the RATS relying party.

I agree with you, we do need a clear RATS verifier to issue attestation token and a clear KBS as RATS rely party that does not perform attestation itself. This means we may need to reorganize and split our existing code at the repository level, and carefully rename them. I think this is a necessary task, and we need to start considering it now. @sameo

src/api/src/token/simple.rs
```diff
@@ -0,0 +1,81 @@
...
+
+        let jwks = json!({
+            "kty": "EC",
+            "crv": "P-256",
+            "x": base64::encode_config(ecdsa_x, base64::URL_SAFE_NO_PAD),
+            "y": base64::encode_config(ecdsa_y, base64::URL_SAFE_NO_PAD),
+            "use": "sig",
+            "kid": self.key_pair.key_id(),
+            "x5c": vec![base64::encode_config(self.cert.clone(), base64::URL_SAFE_NO_PAD)],
+        });
```
jepio, line 77: sorry, one more ask. that's a jwk, a jwks (jwk set) needs one more json wrapper (haven't compiled this):
```
let jwk = ...;
let jwks = json!({
  "keys": [jwk],
});
```
https://www.rfc-editor.org/rfc/rfc7517#appendix-A.1

src/api/src/http/attest.rs
```diff
@@ -0,0 +1,142 @@
...
+                .issue(token_claims, *timeout.into_inner() as usize)
+            {
+                Ok(token) => token,
+                Err(e) => internal!(format!("Issue Attestation Token failed: {e}")),
+            };
+            HttpResponse::Ok()
+                .cookie(session.cookie())
+                .content_type("application/json")
+                .body(bail_error_internal!(serde_json::to_string(&json!({
+                    "attestation-token": token,
```
jepio, line 137: Can we rename to `token` instead of `attestation-token`?
That's how Amber names it: https://github.com/confidential-containers/kbs/pull/77/files#diff-9721abc057e81f3838be6c05283374cccdc44cf55fe09c6f8d3bbc9eaf9f75a0R28-R30.

jepio (approved):
LGTM, but with 2 small comments around schema remaining.

src/api/src/http/attest.rs
jialez0: Sure, it doesn't matter.

src/api/src/token/simple.rs
jialez0: Fixed.

Xynnn007 (approved):

PR #79, opened 2023-04-26
Improve support for non CoCo attestation services

sameo:
Although KBS only supports the CoCo attestation service, it is designed to support any external AS.

This PR brings no functional changes but rearrange and rename code and features to more cleanly support external ASes.
As a consequence, the CoCo AS support is now moved under `api/src/attestation/coco`


src/api/src/attestation/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+/// Attestation Service
+#[derive(Clone)]
+pub struct AttestationService(pub Arc<Mutex<dyn Attest>>);
+
+impl AttestationService {
+    /// Create and initialize AttestionService
+    pub async fn new(kbs_config: &Config) -> Result<Self> {
+        let attestation_service: Arc<Mutex<dyn Attest>> = {
+            cfg_if::cfg_if! {
+                if #[cfg(any(feature = "coco-as-builtin", feature = "coco-as-builtin-no-verifier"))] {
```
Xynnn007, line 39: An idea hits me: as CoCo-AS and Amber can both support remote connection, we could integration both of the client code and let the configuration to determine the runtime behavior (which one to use) rather than compiling time. In this way an user would not need to worry about the complex different compile features? 

This way the binary would not be much bigger because only client code are included.

How do you think of this?

src/api/Cargo.toml
```diff
@@ -7,10 +7,11 @@ documentation.workspace = true
...
 
 [features]
-default = ["native-as", "rustls"]
-native-as = ["attestation-service/default"]
-native-as-no-verifier = ["attestation-service/rvps-native"]
-grpc-as = ["tonic", "tonic-build", "prost"]
+default = ["coco-as-builtin", "rustls"]
+coco-as = []
+coco-as-builtin = ["coco-as", "attestation-service/default"]
+coco-as-builtin-no-verifier = ["coco-as", "attestation-service/rvps-native"]
```
Xynnn007, line 13: A question not related to this PR. Why do we need a "no-verifier" feature? If it is for test, we could implement a mock-ed object of trait `Attest`

cc @jialez0

Xynnn007 (approved):
LGTM. Thanks! Only some comments not related to this PR.

Makefile
```diff
@@ -2,29 +2,29 @@
...
-	cargo build --no-default-features --features native-as-no-verifier,rustls
+.PHONY: kbs-coco-as-no-verifier
+kbs-coco-as-no-verifier:
+	cargo build --no-default-features --features coco-as-builtin-no-verifier,rustls
 
-.PHONY: kbs-native-as-openssl
-kbs-native-as-openssl:
-	cargo build --no-default-features --features native-as,openssl
+.PHONY: kbs-coco-as-openssl
+kbs-coco-as-openssl:
```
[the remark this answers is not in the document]
sameo: I was intentionally trying to keep the Makefile target short, and since builtin is the default I though we could keep it shorter by not adding `builtin`.
But I really don't have any strong preference and can add `builtin` for consistency sake, if you prefer it this way.

src/api/src/attestation/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+use kbs_types::Tee;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+#[cfg(feature = "coco-as")]
+mod coco;
+
+/// Interface for Attestation Services.
+///
+/// Attestation Service implementations should implement that interface.
```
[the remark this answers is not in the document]
sameo: Thanks, fixed now.

src/api/src/attestation/mod.rs
sameo: I think this is a very good idea. Should we wait for the Amber PR to be merged and then do that change?

src/api/Cargo.toml
sameo: The `no-verifier` is really for testing purposes, so that one can build the project without pulling all the tdx/sgx dependencies.
I agree a mocked implementation of the Attest trait would make more sense, I like that.

src/api/src/attestation/mod.rs
Xynnn007: Great, agreed!

src/api/Cargo.toml
Xynnn007: Great. I'll work on this then

Makefile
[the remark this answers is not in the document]
Xynnn007: I'm also not strong with one choice. Let me do some refactor after Amber is merged. As now "built-in" is clearly for CoCo-Attestation-Service, we might use a more short feature name `"kbs-builtin-openssl"` and `"kbs-builtin"`

sameo:
@jialez0 Could you PTAL?

jialez0:
> @jialez0 Could you PTAL?

OK, I'll look at this PR :-)

jialez0 (approved):
LGTM.

PR #81, opened 2023-05-06
Add endpoint for set attestation policy

jialez0:


docs/kbs.yaml
```diff
@@ -259,6 +270,15 @@ components:
...
             Encryption algorithm used to encrypt the response payload.
 
+    AttestationPolicy:
+      required:
+        - policy
+      properties:
+        policy:
+          type: string
+          description: >-
+            Base64 encoded attestation verification policy.
```
sameo, line 280: Which policy format would you recommend? OPA? We should document which ones KBS supports.

docs/kbs_attestation_protocol.md
```diff
@@ -288,9 +288,8 @@ The authentication service is provided by the KBS through two endpoints:
 2. `/kbs/v0/attest` only accepts `POST` requests whose body is a [KBS Attestation](#attestation)
    JSON payload and the header contains a `Cookie` set to the value received in
    step 1.i. This is how the attester replies to attestation challenge received
-   in step 1.ii. The KBS replies to that request with an empty HTTP response
-   (no content), which HTTP status indicates if the attestation was successful or
-   not.
+   in step 1.ii. If the attestation was successful, the KBS replies to that request with an 
+   [Attestation Results Token](#attestation-results-token).
```
sameo, line 292: If I understand correctly, this is not related to the attestation policy setting. If that's so, could we have a separate PR for this fix, please?

jialez0: OK, I'll do this work.

docs/kbs.yaml
jialez0: I have changed the input here now to specify the type of policy engine and the id of the policy :)

docs/kbs.yaml
```diff
@@ -259,6 +270,25 @@ components:
...
+      required:
+        - type
+        - policy_id
+        - policy
+      properties:
+        type:
+          type: string
+          description: >-
+            Policy Engine type, for example: "opa"
+        policy_id:
```
sameo, line 283: So the id here is used to index possibly multiple policy entries in the `policy` payload below?

docs/kbs_attestation_protocol.md
```diff
@@ -474,6 +474,29 @@ When the KBC uses this token to request resources or services from a relying
 party service API, then the symmetric key used to encrypt the output payload can
 be wrapped with the provided `tee-pubkey`.
 
+### Set Attestation Policy
+User of KBS can set attestation verification policy with posting following endpoint:
```
sameo, line 478:
```suggestion
User of KBS can set an attestation verification policy through the following endpoint:
```

docs/kbs.yaml
jialez0: The ID here is the ID of the policy payload below. Once setting operation can only set a single policy, but there may be multiple policies in the policy engine at the same time. Different policies can be selected according to the situation during each evaluation, as mentioned here: https://github.com/confidential-containers/attestation-service/pull/92#issuecomment-1541543936

docs/kbs_attestation_protocol.md
```diff
@@ -474,6 +474,29 @@ When the KBC uses this token to request resources or services from a relying
...
+    "policy_id": "default",
+    "policy": <base64encoded policy>
+}
+```
+
+Where `type` specified the policy format, `policy_id` specified the ID of the policy
+and `policy` is the base64 encoded policy content.
+
+POST to this endpoint needs user identity authentication with carrying JWT (JSON Web Token)
+signed by the user's private key in the Request.
```
sameo, line 498: So the attestation policy is global to the whole KBS? Are we expecting the policy to describe e.g. different policies for different resources?

sameo (commented):
LGTM. Only a few lexical nits, that I figured you could fix since you will need to fix the CI anyway.

docs/kbs_attestation_protocol.md
jialez0: The policy here is the policy for verifying evidence, and in the RATS model, it should be the policy of the Verifier. And the policy you mentioned should be a policy of the rely party, which distributes different confidential resources based on the attestation results.

I think it's best for us to separate these two policies, although we haven't implemented the latter yet.

docs/kbs.yaml
sameo: Ah, I did not see you changed the AS API. That makes sense to me, this is really forwarding the verifier policy to the AS, and maps the policy to an ID.

docs/kbs_attestation_protocol.md
sameo: Ok, that makes sense to me. Thanks for the clarification.

sameo (approved):

PR #83, opened 2023-05-12
Update KBS client Tools to test RESTful APIs

jialez0:


tools/client/README.md
```diff
@@ -11,12 +10,7 @@ For help:
...
 ```
 
-Request resource from KBS:
-
-```shell
-./client --kbs-uri https://127.0.0.1:8080 --resource-path /<repository>/<type>/<tag>
-```
-
+If you want use this client to test KBS APIs that need attestation,
 Make sure this client run inside a real TEE which [Attestation Agent](https://github.com/confidential-containers/attestation-agent) supported (otherwise attestation will failed).
```
sameo, line 14:
```
If you want use this client to test KBS APIs that need attestation, make sure this client run inside an [Attestation Agent](https://github.com/confidential-containers/attestation-agent) supported TEE, otherwise attestation will fail.
```

sameo (commented):
Just a minor nit, otherwise LGTM.

sameo (approved):

PR #89, opened 2023-05-22
github: Enable dependabot

sameo:
For automatic tracking of dependencies and version updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 1
+    allow:
+      - dependency-type: direct
+      - dependency-type: indirect
+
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
LGTM, thanks!

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: cargo
```
Xynnn007, line 3: Do we need an indent here? As https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#example-of-a-basic-setup-for-three-package-managers shows, the whole `package-ecosystem` object should be one of the list of values ​​of `updates`.

sameo: I'm not sure if it is mandatory, but it is more readable for sure. Fixed now.

sameo:
With github-actions updates now as well.

PR #103, opened 2023-05-31
tools/client: move to CoCo AA, use single threaded runtime

mythi:


fitzthum (approved):
LGTM.

fitzthum:
CI is failing because of https://github.com/confidential-containers/kbs/issues/102

tools/client/Cargo.toml
```diff
@@ -13,8 +13,8 @@ base64.workspace = true
 clap = { version = "4.0.29", features = ["derive"] }
 env_logger.workspace = true
 jwt-simple = "0.11.4"
-kbs_protocol = { git = "https://github.com/jialez0/attestation-agent", branch = "token" }
+kbs_protocol = { git = "https://github.com/confidential-containers/attestation-agent", rev = "aa1d3c5" }
```
Xynnn007, line 16: This dependency relationship brings a potential problem:

If we want the code dependencies in each release to be self-consistent, here the `rev` of `attestation-agent` should be a release tag.

However there is a dilemma:
- release of `attestation-agent` depends on `KBS`: CI of guest stack should use a KBS/AS image.
- release of `KBS` depends on `attestation-agent`: Here

There might be two different solutions to resolve
1. move this tool to attestation-agent repo from here
2. we first cut releases for guest stack using a self-built image in CI, after that update the `rev` here and cut release for AS/KBS and finally replace the image in CI. Also we can think about @surajssd 's idea as https://github.com/confidential-containers/kbs/issues/96#issuecomment-1564680358

Xynnn007:
@mythi Please rebase with the latest code. Ideally it could help pass the CI.

tools/client/Cargo.toml
mythi: While playing with the tool, I also thought about option 1 which would probably be the simplest "fix". I guess the caveat to that is that then there's no "sample client" in this repo.

PR #112, opened 2023-06-12
build: add missing -y flags to apt install in dockerfile

katexochen:
There were `-y` flags missing in two commands, causing the build to fail. I've also removed a surplus `apt-get update` and improved formatting of installed packages.
Adding `DEBIAN_FRONTEND=noninteractive` to env for a cleaner log without frontend warnings.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,8 +3,10 @@ FROM rust:1.67-slim as builder
 WORKDIR /usr/src/kbs
 COPY . .
 
+ENV DEBIAN_FRONTEND noninteractive
+
 # Install Build Dependencies
-RUN apt-get update && apt-get install apt-utils
+RUN apt-get update && apt-get install -y apt-utils
 RUN apt-get install -y \
     clang \
     cmake \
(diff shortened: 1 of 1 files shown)


sameo (approved):

jepio (approved):

docker/Dockerfile
```diff
@@ -23,28 +25,37 @@ RUN tar -C /usr/local -xzf go1.20.1.linux-amd64.tar.gz
...
+    clang \
+    curl \
+    gnupg
 
 # Install TDX Runtime Dependencies
 RUN curl -L https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | apt-key add -
 RUN echo 'deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu focal main' | tee /etc/apt/sources.list.d/intel-sgx.list
 RUN apt-get update
-RUN apt-get install -y libtdx-attest libsgx-dcap-default-qpl libsgx-dcap-quote-verify tpm2-tools
+RUN apt-get install -y \
```
mythi, line 54: it might be leaner to also install without the recommended dependencies:
```suggestion
RUN apt-get install -y --no-install-recommends \
```

fitzthum (approved):

katexochen:
> One comment regarding DCO: can you signoff the commits with a proper e-mail instead of the github generated alias?

@jepio is that required by DCO? I couldn't find information about what qualifies an email to be used for it. Never had problems in CNCF context before with that signoff, and I would prefer not to maintain another email that only receives spam.

BTW there is no DCO document in this repo.

jepio:
> @jepio is that required by DCO? I couldn't find information about what qualifies an email to be used for it. Never had problems in CNCF context before with that signoff, and I would prefer not to maintain another email that only receives spam.

I assumed this was necessary, but I can't find any clarity from CNCF regarding using proper emails. If no one else thinks this is in issue, then discard my comment.

sameo:
> > @jepio is that required by DCO? I couldn't find information about what qualifies an email to be used for it. Never had problems in CNCF context before with that signoff, and I would prefer not to maintain another email that only receives spam.
> 
> I assumed this was necessary, but I can't find any clarity from CNCF regarding using proper emails. If no one else thinks this is in issue, then discard my comment.

I also prefer proper emails, but CNCF only requires the contributor to use it's [real name](https://github.com/cncf/foundation/blob/main/dco-guidelines.md), whatever that means.
Paul's SOB fills that requirement imo.

PR #126, opened 2023-07-11
ci: don't fail fast on matrix runs

katexochen:
This way, other matrix runs won't be canceled if one of them fails.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,3 +1,5 @@
+name: rust tests
+
 on:
   push:
     branches: [ "main" ]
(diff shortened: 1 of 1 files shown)


fitzthum (commented):
What's the thinking here? Why do we want to continue running?

katexochen:
You'll get better feedback. In cases where only one of the matrix elements causes a failure, you are ensured the others work. In case of flaky tests, you can rerun only the faild one, without risking to run into a flake failure of another test that would have succeeded before. As these are short running tests, there is also no reason to be concerned of the resource consumption in case all tests run until failure.

fitzthum (approved):
Seems reasonable. Ironically one of the tests is failing.

katexochen:
It was flaky at that time, that's the reason I wrote the patch. :wink:

PR #130, opened 2023-07-14
Feature: Add access resource using the KBS-provisioned Token

### LINKED ISSUE #129: Feature | Access resource with the KBS-provisioned token
Now after the handshake of the RCAR protocol, KBS will response a token to the requester. This token should have access to the KBS' resources. (This avoids to re-establish a RCAR session again after the former session breaks in the timeout of the token)

Now we are using the same session where RCAR handshake happens to access resource. This issue is to normalize this behavior. That means the resource access should be spilit into two:
1. With RCAR protocol, the client get a short-liveness token (currently ~10min)
2. the client then access resource with this token, or with the same session that RCAR uses. KBS should support both of the ways.

cc @sameo @jialez0 Please check the rationale of this, thanks! And I'd like to push this both kbs and AA side.

Xynnn007:
This PR mainly aims to add a new feature that a requester can bring the KBS-provisioned token to access the resources. The rationale behind this is the token is always short-lived, which could help decrease the risk of token leakage.

This PR also refactor the error handling of the `http` module a little to use the question mark `?` for better error handling.

Close #129

### DIFF
--- a/docs/kbs.yaml
+++ b/docs/kbs.yaml
@@ -95,6 +95,7 @@ paths:
           name: kbs-session-id
           schema:
             type: string
+          required: false
         - name: repository
           in: path
           description: A parent path of resource, can be empty to use the default repository.
@@ -113,6 +114,16 @@ paths:
           schema:
             type: string
           required: true
+      requestBody:
+        required: false
+        description: >-
+          A KBS provisioned token that includes the Tee public key.
+          KBS will then use the Tee public key to perform JSON Web Encryption
+          for the resource in a response if the token is valid.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AttestationToken'
       responses:
         200:
           description: >-
--- a/docs/kbs_attestation_protocol.md
+++ b/docs/kbs_attestation_protocol.md
@@ -340,6 +340,14 @@ protocol, where attesters authenticate themselves against a KBS implementation.
 This leads to the second phase, during which authenticated attesters can request
 protected resources from the KBS.
 
+Attesters can access protected resources by providing the KBS with either the HTTP
+cookie or the Attestation Results token returned as part of the Attestation HTTP
+response. The former is usually short-lived and is typically used for one-time,
+synchronous resources request. For longer KBS sessions, it is more convenient to
+use the latter.    
+
+### HTTP Cookie Authentication
+
 The KBS implementation keeps track of attestation results and binds them to a
 cookie identifier. During the second phase, the KBS can then decide if a
 specific resource could be released to a given attester, by mapping the cookie
@@ -370,7 +378,7 @@ content is set to a [KBS Response](#response) JSON payload:
     ┃                                                   ┃
     ┃                                                   ┃
     ┃                                                   ┃
-    ▽        Figure 2: KBS Resource Request Phase       ▽
+    ▽   Figure 2: KBS Resource Request Phase - Session  ▽
 ```
 
 A request for protected resource can fail for three reasons:
@@ -378,6 +386,51 @@ A request for protected resource can fail for three reasons:
 1. The requester is not authenticated and thus can not provide the right cookie
    identifier. The KBS implementation sends an HTTP response with a 401
    (`Unauthorized`) status code.
+2. The attester is authenticated but requests a resource that it's not allowed
+   to receive. The KBS implementation sends an HTTP response with a 403
+   (`Forbidden`) status code.
+3. The requested resource does not exist. The KBS implementation sends an HTTP
+   response with a 404 (`Not Found`) status code.
+
+### Attestation Results Token Authentication
+
+As part of the Authentication process, the requester generates a TEE asymmetric key 
+pair. Upon successful Attestation, the KBS provisions a token that contains the
+Attestation Results and endorsements for the TEE public key.
+
+To request a protected resource from the KBS, the attester sends a `GET` request
+to a resource specific endpoint with the token as the bearer authorization. If the
+token passes verification, the KBS will respond to the `GET` request with an HTTP
+response whose content is set to a [KBS Response](#response) JSON payload. The
+symetric key inside the [KBS Response](#response) JSON payload will be wrapped by
+the public key inside the token.
+
+```
+┌──────────┐                                         ┌─────┐
+│ Attester │                                         │ KBS │
+└───┳──────┘                                         └──┳──┘
+    ┃            ┌─────────────────────────┐            ┃
+    ┃────────────│GET /kbs/v0/<resource_id>│───────────▶┃
+    ┃      ┌─────┴─────────────────────────┴────┐       ┃
+    ┃      │Header: Authorization Bearer        │       ┃
+    ┃      │Body: N/A                           │       ┃
+    ┃      └────────────────────────────────────┘       ┃
+    ┃                    ┌────────┐                     ┃
+    ┃◀───────────────────│Response│─────────────────────┃
+    ┃     ┌──────────────└────────┘───────────────┐     ┃
+    ┃     │Header: No Cookie                      │     ┃
+    ┃     │Body: Response                         │     ┃
+    ┃     └───────────────────────────────────────┘     ┃
+    ┃                                                   ┃
+    ┃                                                   ┃
+    ┃                                                   ┃
+    ▽    Figure 3: KBS Resource Request Phase - Token   ▽
+```
+
+A request for protected resource can fail for three reasons:
+
+1. The requester is not authenticated and thus can not provide the valid token.
+   The KBS implementation sends an HTTP response with a 401 (`Unauthorized`) status code.
 2. The attester is authenticated but requests a resource that it's not allowed
    to receive. The KBS implementation sends an HTTP response with a 403
    (`Forbidden`) status code.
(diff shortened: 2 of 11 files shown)


Xynnn007:
About the error handling, there are still a lot of `anyhow` crate used in KBS. KBS protocol document said that we should return different HTTP status code due to different error. I did not dive into this, but let me put an issue to record this.

docs/kbs.yaml
```diff
@@ -113,6 +114,16 @@ paths:
...
+      requestBody:
+        required: false
+        description: >-
+          A KBS provisioned token that includes the Tee public key.
+          KBS will then use the Tee public key to perform JSON Web Encryption
+          for the resource in a response if the token is valid.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/AttestationToken'
```
jialez0, line 126: I think we'd better put the token in the `bear_auth` header of HTTP Req. (Same as what we do in the `set-resource` API)

Xynnn007: Good point. Fixed.

src/api/src/http/resource.rs
```diff
@@ -59,27 +53,135 @@ pub(crate) async fn get_resource(
...
+    let bearer = Authorization::<Bearer>::parse(request)
+        .map_err(|e| Error::InvalidRequest(format!("parse Authorization header failed: {e}")))?
+        .into_scheme();
+
+    let token = bearer.token().to_string();
+
+    let claims = token_broker
+        .read()
+        .await
+        .verify(token)
```
jialez0, line 120: We'd better verify token here locally instead of call `token_broker` to do it, because the issuer may not be this KBS itself. I think we  can set a "get_token_verification_pubkey" API, it can help to get pubkey from  token broker or remote token issuer, which can be specified by config parameter.

Xynnn007: Currently we only suppose this token is provisioned by the same KBS just then, for the token is short-lived ~10min. Do you mean a long-lived token?

jialez0: This is not related to the lifecycle of the token. According to my understanding, in the confidential data hub design, the issuer of the token will not be the same entity as the consumer of the token most of the time, even if it is a token with a short lifecycle, it does not matter.

Xynnn007: Do you mean a scenario like this:
1. Client gets a token provisioned by an attestation service rather than CoCo. For example Amber
2. Client wants to use this token to request resource in KBS

jialez0: > Do you mean a scenario like this:
> 
> 1. Client gets a token provisioned by an attestation service rather than CoCo. For example Amber
> 2. Client wants to use this token to request resource in KBS

Exactly. The token issuer in 1 may not only be Amber, but also CoCo AS deployed by trusted parties (specifically used to issue tokens).

Xynnn007: Ok. I think this should be carefully handled. I think there are two different tokens in the context:
I. In KBS: Signed(claims, pubkey)
II. In Amber: Signed(claims)

We should firstly ensure this token is with the proper token (type I), with which we can encrypt the dataflow of the confidential resource. If the token is type II, we should think about how we send back the confidential resource.

Also, the basic logic of a certificated token is:
1. Verify the endorsement and claims due to the user defined policy
3. A token is provisioned

Also, if tokens provisioned by other CoCo-KBS is to be supported, we should recheck the claims via giving it to the backend CoCo-AS against the policy of the current CoCo-KBS/AS. 

My idea is that we can start with a simple scenario, s.t.
1. This token is provisioned by the same KBS.

To accept a token from another KBS, we should
1. impl new API for AS to receive different signed claims rather than a raw evidence to verify against a policy.
2. Let KBS leverage this new API to forward the signed claims of the token (or directly the token)

To accept a token from another attestation service like Amber, we should at least
1. Let Amber to support a token with pubkey
Or 2. Think of a way where a verified token can be derived into a tee key pair to fit in the KBS protocol to encrypt the dataflow
Or 3. Do some modifications upon the current KBS protocol

jialez0: OK, I agree start with the simple scenario that token provisioned by the same KBS. 

However, two points need to be emphasized:

1. Verify that the claims in the token should not be in AS, but should introduce a new policy engine in KBS, which is consistent with the RATS passport model. The policy here is the policy of rely party verifying the attachment results.
2. After this PR is merged, I will transfer the logic of the Token broker from KBS to AS. In other words, AS no longer returns a results claims, but directly returns the token. Therefore, I may modify the token's verification logic here to: obtain the token verification public key from AS's new API and then have KBS verify the token.

docs/kbs_attestation_protocol.md
```diff
@@ -340,6 +340,8 @@ protocol, where attesters authenticate themselves against a KBS implementation.
 This leads to the second phase, during which authenticated attesters can request
 protected resources from the KBS.
```
sameo, line 342: I would add the following paragraph to introduce the two authentication scheme:

```
Attesters can access protected resources by providing the KBS with either the HTTP cookie or the
Attestation Results token returned as part of the Attestation HTTP response. The former is usually
short-lived and is typically used for one-time, synchronous resources request. For longer KBS
sessions, it is more convenient to use the latter.    
```

docs/kbs_attestation_protocol.md
```diff
@@ -340,6 +340,8 @@ protocol, where attesters authenticate themselves against a KBS implementation.
...
 
+### Using cookie identifier
+
```
sameo, line 344:
```
### HTTP Cookie Authentication
```

docs/kbs_attestation_protocol.md
```diff
@@ -384,6 +386,50 @@ A request for protected resource can fail for three reasons:
 3. The requested resource does not exist. The KBS implementation sends an HTTP
    response with a 404 (`Not Found`) status code.
 
+### Using KBS-Provisioned Token
```
sameo, line 389:
```
### Attestation Results Token Authentication
```

docs/kbs_attestation_protocol.md
```diff
@@ -384,6 +386,50 @@ A request for protected resource can fail for three reasons:
...
+
+After a successful Authentication process, a tee asymmetric keypair will be generated
+by the requester and KBS will provision a token that contains attestation results and
+endorsement for the tee public key.
+
```
sameo, line 394:
```markdown

As part of the Authentication process, the requester generates a TEE asymmetric key 
pair. Upon successful Attestation, the KBS provisions a token that contains the
Attestation Results and endorsements for the TEE public key.
```

docs/kbs_attestation_protocol.md
```diff
@@ -384,6 +386,50 @@ A request for protected resource can fail for three reasons:
...
+To request a protected resource from the KBS, the attester sends a `GET` request
+to a resource specific endpoint with the token as the bearer authorization. If the
+token passes verification, the KBS will respond to the `GET` request with an HTTP
+response which content is set to a [KBS Response](#response) JSON payload, whose
+encryption key is the one inside the token:
```
sameo, line 399: Isn't the KBS response encrypted with a symetric key, that is wrapped with the attestation results TEE key?

src/api/Cargo.toml
```diff
@@ -33,7 +33,6 @@ jwt-simple = "0.11.4"
 kbs-types = "0.4"
 lazy_static = "1.4.0"
 log.workspace = true
-mime.workspace = true
```
sameo: I'm not sure this is related to this commit?

src/api/src/http/error.rs
```diff
@@ -95,8 +93,8 @@ impl ResponseError for Error {
         // error cannot happen.
         // A test covering all the possible error types are given to ensure this.
         let body = serde_json::to_string(&info).expect("serialize error response failed");
-        
-        // Due to the definition of KBS attestation protocol, 
+
+        // Due to the definition of KBS attestation protocol, we set the http code.
```
sameo, line 103: I think that belongs to the HTTP error handling improvement commit (9cccf7ab4b3dca86fa2172a26bb1da1c85bb299f)

docs/kbs_attestation_protocol.md
sameo: > The former is usually
short-lived and is typically used for one-time, synchronous resources request. For longer KBS
sessions, it is more convenient to use the latter.

This may not be accurate at the moment, but I believe this is the intention: Being able to request resources with the token, regardless of the HTTP session status.

Xynnn007: Thanks for the suggestion. It looks more comprehensible!

docs/kbs_attestation_protocol.md
Xynnn007: Yes, thanks. I added some more details about the declaration.

src/api/Cargo.toml
Xynnn007: Oh, let me fix this. It is not used.

src/api/src/http/error.rs
Xynnn007: Oh yes! Have rebased. Thanks!

Xynnn007:
Hi @sameo @jialez0 , thanks for the suggestions and comments. I've updated the PR. PTAL

jialez0 (approved):
LGTM. Thanks @Xynnn007

sameo (approved):

Xynnn007:
Hi @sameo sorry for not being clear. Could we revert this PR now? As it is during code freeze, I'm not sure whether this change would break other release components. Maybe we can re-merge this after the release?

Xynnn007:
Emmm, let's retrigger the tdx CI and see if any new errors occur. If not, we do not need to revert. As revert is a dirty operaiton.

PR #139, opened 2023-07-28
Optimized the logic for token issuance and verification.

jialez0:
1. Move the token issuance process down to Attestation Service completion, which is in line with the responsibility division and RATS model. Related to: https://github.com/confidential-containers/attestation-service/pull/121
2. Add some new compiling features to support the functional division of KBS in two deployment forms: background checking model and passport model.
3. Add an optional feature: fine-grained validation of the Claims of Attention Results Token using a policy engine to determine whether the requester has access rights to specific resource.
4. Update scripts and documents, add a quick start guide.
5. Update client tool to support get resource with token. (Rely on the new `kbs_protocol` in https://github.com/confidential-containers/guest-components/pull/304)

@Xynnn007 @sameo @bpradipt @mkulke @mythi If you have any comments on these modifications, please let me know.


Makefile
```diff
@@ -1,54 +1,52 @@
...
-	cargo build --no-default-features --features coco-as-grpc,openssl
+.PHONY: background-check-kbs
+background-check-kbs:
+	cargo build --release --no-default-features --features $(AS_FEATURE),resource,$(HTTPS_CRYPTO),$(POLICY_ENGINE)
 
-.PHONY: kbs-coco-as-no-verifier-openssl
-kbs-coco-as-no-verifier-openssl:
-	cargo build --no-default-features --features coco-as-builtin-no-verifier,openssl
+.PHONY: passport-abs
+passport-abs:
```
fitzthum, line 22: I wonder if it is really worth distinguishing between the KBS and the ABS. It looks like they are basically the same thing. I know that the RATS diagrams show the passport model making a direct connection between Attester and Verifier, but we aren't going to do that either way, so why not just say that we always have the KBS in front of the AS? I think introducing this new piece might confuse people. Instead we can just say that if you are using the passport model you can build the KBS without support for resources and then you can use another KBS that does have support for resources.

fitzthum (commented):
Haven't had time to look at the code yet. Just a few comments.

One thing overall is that I would like us to settle on some more formal definitions of background check vs passport and how it relates to our project specifically.

I think we've been making a slightly different distinction between two modes. Currently we have one mode where the HW evidence is only the initial launch measurement. This is the 2-stage attestation. MSFT have proposed another mode where we measure the policy and bind the hw measurement to the workload. In some ways these two modes map onto background check vs passport, but not quite. 

When we use a two-stage attestation, the hw evidence alone does not define the workload. Currently we use the RATS background check model for this. The KBS asks the AS to validate the hw evidence, and then the KBS injects secrets into the guest that define the workload identity.

What we've noticed lately is that it would be useful to be able to provide a workload with some kind of "passport" that reflects the workload claims. In the two-stage approach we can't use the HW evidence as a passport, because it only includes the launch measurement. Instead, this passport would have to come from the KBS, which knows the full set of workload claims. Even though we've referred to this as a passport, it wouldn't work with the RATS passport model, because we need to get it from the KBS, not from the AS (assuming the launch measurement is truly generic).

On the other hand, if we use the policy-based approach with the host data, the launch measurement would now reflect the entire workload. Here we would already have a "passport" even if we never connect to a verifier. This evidence could be sent to an enlightened KMS directly to retrieve secrets.

So we have different meanings of the passport/background check idea. In the RATS model one of the crucial differences is whether or not you connect directly to the AS. For CoCo we probably care more about what exactly the evidence measures. The relationship between these two things is a little murky. 

What I would tentatively propose, is that we talk about a passport as some kind of certficate that the workload can use to prove its validity. I don't think this is mutually exclusive with a background check. The KBS can generate a passport after doing a background check.

Does this make any sense to people? cc: @Xynnn007 @mkulke 

I don't think I've really arrived at a definition yet, but I think it would be good if we all got on the same page.

jialez0:
> So we have different meanings of the passport/background check idea. In the RATS model one of the crucial differences is whether or not you connect directly to the AS. For CoCo we probably care more about what exactly the evidence measures. We've also just introduced a distinction between tokens and evidence. I think we need to make it much more clear what the relationship between all these things is.
> 
Thanks a lot for the deeply explain @fitzthum .

I agree with your idea of a strict and formal definition of passport models and background-check models in CoCo, as they require more consideration and discussion.
This PR provides two different ways to deploy and use KBS, perhaps these two more accurate names should be: KBS's integrated attestation mode and KBS's separated attestation mode. If we use the above two names to define the deployment model in this PR instead of the background-check and passport of RATS, do you think it is appropriate?

Xynnn007:
> Haven't had time to look at the code yet. Just a few comments.
> 
> One thing overall is that I would like us to settle on some more formal definitions of background check vs passport and how it relates to our project specifically.
> 
> I think we've been making a slightly different distinction between two modes. Currently we have one mode where the HW evidence is only the initial launch measurement. This is the 2-stage attestation. MSFT have proposed another mode where we measure the policy and bind the hw measurement to the workload. In some ways these two modes map onto background check vs passport, but not quite.
> 
> When we use a two-stage attestation, the hw evidence alone does not define the workload. Currently we use the RATS background check model for this. The KBS asks the AS to validate the hw evidence, and then the KBS injects secrets into the guest that define the workload identity.
> 
> What we've noticed lately is that it would be useful to be able to provide a workload with some kind of "passport" that reflects the workload claims. In the two-stage approach we can't use the HW evidence as a passport, because it only includes the launch measurement. Instead, this passport would have to come from the KBS, which knows the full set of workload claims. Even though we've referred to this as a passport, it wouldn't work with the RATS passport model, because we need to get it from the KBS, not from the AS (assuming the launch measurement is truly generic).
> 
> On the other hand, if we use the policy-based approach with the host data, the launch measurement would now reflect the entire workload. Here we would already have a "passport" even if we never connect to a verifier. This evidence could be sent to an enlightened KMS directly to retrieve secrets. Maybe we shouldn't refer to this as a passport, but as a workload certificate (or something).
> 
> So we have different meanings of the passport/background check idea. In the RATS model one of the crucial differences is whether or not you connect directly to the AS. For CoCo we probably care more about what exactly the evidence measures. We've also just introduced a distinction between tokens and evidence. I think we need to make it much more clear what the relationship between all these things is.
> 
> Does this make any sense? cc: @Xynnn007 @mkulke

Great to hear about this. We team once had a discussion with Intel and found that the attestation (bind policy content) + policy check is what we want. I think we are very glad to be involved in the discussion. wdyt @arronwy : )?There are some open thoughts:
1. To leverage RA+policy, we should design a general enough mechanism by which we can inject any data from the k8s, and this data will be inject into VM by VMM using HOSTDATA mechanism, and then the integrity be checked by remote attestation. This mechanism should have an abstract enough api at VMM side, e.g. `set_hostdata()` or something to leverage `mr_owner_config` for tdx, `host_data` for SNP, etc. This mechanism also helps to inject a `Pod id` field which might help to reduce the risk of evidence-factory.
2. We should discuss the start-up order and analyze the security & rationality. After bringing OPA in, there are many components to be launched inside a pod: `attestation-agent`, `confidential-datahub`, `opa`, `kata-agent`. There might be some time during which OPA is starting to work while the remote attestation does not happen.
3. About the KBS protocol. I am going to start an issue for some discussion and share some point of views, where I hope you and @jialez0  can share some thoughts.

mkulke:
> So we have different meanings of the passport/background check idea. In the RATS model one of the crucial differences is whether or not you connect directly to the AS. For CoCo we probably care more about what exactly the evidence measures. We've also just introduced a distinction between tokens and evidence. I think we need to make it much more clear what the relationship between all these things is.
> 
> Does this make any sense? cc: @Xynnn007 @mkulke

Thank you Tobin, that's a succinct description of my confusion 😅

mkulke:
> I agree with your idea of a strict and formal definition of passport models and background-check models in CoCo, as they require more consideration and discussion. This PR provides two different ways to deploy and use KBS, perhaps these two more accurate names should be: **KBS's integrated attestation mode** and **KBS's separated attestation mode**. If we use the above two names to define the deployment model in this PR instead of the background-check and passport of RATS, do you think it is appropriate?

I think that's a great idea. 

The notions behind the terms **Integrated/seperated attestation** are not entirely obvious to me (2-stage attestation, attestation-with-workload-policy, ...), maybe there are more descriptive names?

jialez0:
> What we've noticed lately is that it would be useful to be able to provide a workload with some kind of "passport" that reflects the workload claims. In the two-stage approach we can't use the HW evidence as a passport, because it only includes the launch measurement. Instead, this passport would have to come from the KBS, which knows the full set of workload claims. Even though we've referred to this as a passport, it wouldn't work with the RATS passport model, because we need to get it from the KBS, not from the AS (assuming the launch measurement is truly generic).

I don't think we need to be constrained to obtain this so-called passport from KBS or AS. RATS defines a set of models without defining specific names and implementations. In our existing architecture, if we use one KBS to distribute tokens and another KBS to verify this token, then the former can fully correspond to the generalized AS of RATS. These are just some discrepancies in the name and do not affect the model.

> On the other hand, if we use the policy-based approach with the host data, the launch measurement would now reflect the entire workload. Here we would already have a "passport" even if we never connect to a verifier. This evidence could be sent to an enlightened KMS directly to retrieve secrets. Maybe we shouldn't refer to this as a passport, but as a workload certificate (or something).

The mechanism you mentioned essentially integrates the validation logic of CoCo-AS into the guest and passes in the policy from the host data. The token obtained after verification can be directly used as a passport to retrieve resources from the resource server. This has the same effect as obtaining tokens remotely. But I don't think this mechanism is what this PR needs to discuss. What this PR does is divide the usage of remote KBS and AS, distinguishing between two modes: KBS integrating AS to verification evidence and KBS verifying AS tokens.

jialez0:
> So we have different meanings of the passport/background check idea. In the RATS model one of the crucial differences is whether or not you connect directly to the AS.

As I said, RATS defines a model rather than an implementation, and we don't have to be bound by the specific name of our project. In fact, when we have a KBS that is only responsible for issuing tokens, it is the AS in the RATS model.

README.md
```diff
@@ -1,159 +1,118 @@
...
-
-## Attestation
+KBS has two deploy model, which are consistent with [RATS](https://www.ietf.org/archive/id/draft-ietf-rats-architecture-22.html)
+- Background Check Model: KBS integrate AS to verify TEE evidence, then distribute resource data.
+- Passport Model: ABS (Attestation Broker Service) integrate AS to verify TEE evidence and distribute token,
+KBS verify the token then distribute resource data.
 
-By default, KBS relies on the [Confidential Containers Attestation-Service (AS)](https://github.com/confidential-containers/attestation-service)
-to verify the TEE Evidence.
+## Background Check Mode
```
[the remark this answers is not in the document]
jialez0: Here we use "mode" to distinguish from "model". "mode" are a specific usage of KBS, and "model" are abstract architectures defined by RATS.

src/api/src/policy_engine/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+
+#[cfg(feature = "opa")]
+mod opa;
+
+const DEFAULT_POLICY_PATH: &str = "/opa/confidential-containers/kbs/policy.rego";
+
+#[async_trait]
+pub trait PolicyEngineInterface: Send + Sync {
+    async fn evaluate(&self, resource_path: String, input_claims: String)
+        -> Result<(bool, String)>;
```
mkulke, line 22: Documentation on the return type would be good

src/api/src/policy_engine/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+
+/// Resource policy engine interface
+#[async_trait]
+pub trait PolicyEngineInterface: Send + Sync {
+    /// Determine whether there is access to a specific path resource based on the input claims.
+    async fn evaluate(&self, resource_path: String, input_claims: String)
+        -> Result<(bool, String)>;
+
+    /// Set policy (Base64 encode)
+    async fn set_policy(&mut self, policy: String) -> Result<()>;
```
mkulke, line 25: I wouldn't recommend anyhow results on a public trait

src/api/src/policy_engine/opa/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+impl PolicyEngineInterface for Opa {
+    async fn evaluate(
+        &self,
+        resource_path: String,
+        input_claims: String,
+    ) -> Result<(bool, String)> {
+        let policy = tokio::fs::read_to_string(
+            self.policy_path
+                .to_str()
+                .ok_or_else(|| anyhow!("Miss Policy Path"))?,
```
mkulke, line 58:
```suggestion
                .ok_or_else(|| anyhow!("Missing Policy Path"))?,
```

src/api/src/policy_engine/opa/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+        )
+        .await
+        .map_err(|e| anyhow!("Read OPA policy file failed: {:?}", e))?;
```
mkulke, line 61:
```suggestion
        .map_err(|e| anyhow!("Reading OPA policy file failed: {:?}", e))?;
```

src/api/src/policy_engine/opa/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+        let decision_str: &CStr = unsafe { CStr::from_ptr(decision_buf) };
+        let res = decision_str.to_str()?.to_string();
+        log::debug!("Evaluated: {}", res);
+        if res.starts_with("Error::") {
+            return Err(anyhow!(res));
+        }
+
+        let res_kv: Value = serde_json::from_str(&res)?;
+
+        Ok((res_kv["allow"].as_bool().unwrap_or(false), res))
```
mkulke, line 92: Shouldn't we return an Error if `allow` isn't a bool value?

src/api/src/token/coco.rs
```diff
@@ -0,0 +1,85 @@
...
+        if let Some(exp) = claims_value["exp"].as_i64() {
+            if exp < now {
+                bail!("token expired");
+            }
+        }
+        if let Some(nbf) = claims_value["nbf"].as_i64() {
+            if nbf > now {
+                bail!("before validity");
+            }
+        }
```
mkulke, line 39: Do we always want to have expiration set on tokens?

```suggestion
        let now = time::OffsetDateTime::now_utc().unix_timestamp();
        let Some(exp) = claims_value["exp"].as_i64() else {
            bail!("token expiration unset");
        }   
        if exp < now {
             bail!("token expired");
        }
        if let Some(nbf) = claims_value["nbf"].as_i64() else {
            if nbf > now {
                bail!("before validity");
            }
        }
```

src/api/src/token/coco.rs
```diff
@@ -0,0 +1,85 @@
...
+    kty: Option<String>,
+    alg: String,
+    n: String,
+    e: String,
+}
+
+fn rs384_verify(payload: &[u8], signature: &[u8], jwk: &str) -> Result<()> {
+    let jwk = serde_json::from_str::<RsaJWK>(jwk)?;
+    if jwk.alg != *"RS384" {
+        bail!("Unmatch RSA JWK alg");
```
mkulke, line 72:
```suggestion
        bail!("Unmatched RSA JWK alg");
```

tools/client/src/lib.rs
```diff
@@ -90,6 +149,50 @@ pub async fn set_attestation_policy(
...
+        .json::<ResourcePolicyData>(&post_input)
+        .send()
+        .await?;
+
+    match res.status() {
+        reqwest::StatusCode::OK => Ok(()),
+        _ => {
+            bail!("Request Failed, Response: {:?}", res.text().await?)
+        }
+    }
```
mkulke, line 193:
```suggestion
    if res.status() != reqwest::StatusCode::OK {
       bail!("Request Failed, Response: {:?}", res.text().await?);
    }
    Ok(())
```

fitzthum:
Ok, I think what is most important is that we make sure we can support a feature where we can get certificates of identity. This isn't directly related to the RATS passport or background check model. Really it seems like the passport / background check is just an implementation detail, so I think this PR should be fine. I will try to review the changes tomorrow.

fitzthum (commented):
Ok one more little question but the code looks really solid.

jialez0:
@fitzthum To make clear the difference between the two policies in AS and KBS, we need to first review completely what we have verified in the whole process of the currently implemented authentication model:

In the implementation of KBS and AS, first, the client in TEE collects HW evidence and sends it to KBS (through RCAR), and then KBS forwards HW evidence to AS for verification.
Our AS has verified the following three items:
1. Hardware signature, specific to TEE hardware type and vendor.
2. Compare the general content of HW evidence with the reference value required by the vendor (like Intel, AMD…).
3. Compare the general content of HW evidence with the customized reference value required by the AS owner according to the AS policy.
The first two steps are completed by the verifier plug-in in AS, and the third step is completed by the policy engine in AS. The “general content of HW evidence” refers to the content that is only related to the TEE platform itself, such as the hardware version number.
We must note that the AS three-step verification does not involve any content about workload identity and KBS resource access permissions (and the AS will never perceive these). All verifications of AS only show that the hardware and software versions of TEE and the measurement values of the common system modules on the platform meet the requirements of TEE suppliers and users.
Therefore, we can see that the role of policies in AS can be summarized as: ensure that HW TEE software and hardware versions and measurements meet the customized requirements of owner. An analogy: make sure that the material of the safe deposit box provided by the bank meets the customer's own requirements.

After completing the above authentication, AS only needs to tell KBS: I think the TEE platform represented by the HW evidence you gave me can be trusted (or not). This is a boolean value, and there is no other choice (it is confused to say that a TEE is between trustworthy and untrustworthy). In addition, the AS needs to provide a result token, which contains the parsed HW evidence content, and the token will be sent to the TEE to indicate that the credibility of the TEE is confirmed by a specific AS (that is why the token needs to carry the AS signature)

So far, we have completed the first phase of authentication. The token issued by the AS for workload indicates that we have confirmed that the TEE environment where workload is located is secure. Let's come to the second stage: authenticate the identity of workload and grant it access to resources.

First, please note that some fields in HW evidence can be set by workload owner (such as HOST DATA), which allows us to set the identity of workload in HW evidence. However, as we said, AS can never recognize these non universal evidence fields. What AS can do is to parse them and include them in the result token.
In the second stage of authentication, workload shows the result token signed by AS to KBS and tells KBS the resource index it wants to obtain.
Our KBS verifies the AS token as follows:
1. The AS signature on the token, which is specific to the type of token issued by the specific AS.
2. Check the workload specific fields in the AS token (such as HOST DATA) according to the policy made by the KBS owner, and determine whether the workload has access to the resource it wants to obtain.
Therefore, we can see that the role of policies in KBS can be summarized as: ensure that a specific workload can only access KBS resources that it has access to. An analogy: make sure that specific safe deposit box contain items arranged by customers themselves.

After completing the second phase of authentication above, if the authentication passes, KBS will return the resource that workload wants to obtain. This resource may have two possibilities:
1. Secret data (has been implemented now): e.g. decryption key. This is the main use of KBS at present
2. Identity certificate (has not been implemented now): a certificate that can represent the identity and security of workload. Workload can use this certificate to access a wider range of resource servers, such as KMS. This is not implemented by KBS at present, which requires more discussion and design.

The above is the whole process. I hope these can answer your questions.

cc @Xynnn007 @mkulke

jialez0:
@fitzthum BTW, I have some explanations about why we should divide the policy into AS policy and KBS policy (or why we should divide the authentication into two stages) instead of putting them in the same policy for only one stage authentication:

At present, I am responsible for adopting CoCo's KBS and AS for productization in [Alibaba Cloud](https://us.alibabacloud.com/?utm_key=se_1012047735&utm_content=se_1012047735) and [OpenAnolis](https://openanolis.cn/?lang=en). We found that in the deployment requirements of some scenarios, the verification of the security of the TEE platform itself and the authorization of resource access need to be completed by two different owner entities. Some KBS owners do not want to install AS on their own platforms, but they want some other entities they trust to help them verify the security of the TEE platform itself, the KBS deployed by themselves is only responsible for verifying some specific content related to workload and executing resource authorization.

On the other hand, considering the description of the authentication model in the RATS architecture of the IETF, both the Verifier and the Reply Party should have their own policies. The former’s policy is responsible for verifying the original evidence from the attester and generating attestation results, while the latter’s policy is responsible for verifying attestation results. This is consistent with the implementation in this PR.

The following is the original description in the RATS architecture:

### Passport Model

> In this model, an Attester conveys Evidence to a **Verifier that compares the Evidence against its appraisal policy**. The Verifier then gives back an Attestation Result that the Attester treats as opaque data.

> The Attester does not consume the Attestation Result, but it might cache it. The Attester can then present the Attestation Result (and possibly additional Claims) to a Relying Party, **which then compares this information against its own appraisal policy**. The Attester may also present the same Attestation Result to other Relying Parties.

The original link is: https://datatracker.ietf.org/doc/html/rfc9334#section-5.1

### Background-Check Model

> In this model, an Attester conveys Evidence to a Relying Party, which treats it as opaque and simply forwards it on to a Verifier. **The Verifier compares the Evidence against its appraisal policy** and returns an Attestation Result to the Relying Party. **The Relying Party then compares the Attestation Result against its own appraisal policy**.

The original link is: https://datatracker.ietf.org/doc/html/rfc9334#section-5.2

fitzthum:
@jialez0 Yes, I understand the model and I think that it will work. I think it's also a fine place to start, but I'm not sure it's going to be optimal in the long term.

One thing to think about is that this model requires resource policy to be platform specific. I'm not sure this is ideal.

I suggested an alternate model where the output of the AS policy validation would be the input to the KBS policy validation. In this model the AS could evaluate the claims and return not just true/false, but a wide range out outputs. For instance the AS could return level1, level2, level3 etc. to reflect different levels of strictness. Then the KBS policy could require a certain level for a certain secret. In this approach the resource policy is decoupled from the platform. The AS could define level2 for many several different platforms. I think this would really simplify coming up with resource policies, especially for resources that are deployed on multiple platforms.

That said, I think the use case you mention is valid and should be supported. I think we could easily allow both if we just include the output of the AS validation as part of the token along with the claims. Then if people want to write a KBS policy that parses all the claims again they can, but they can also just use the output from the AS. Does that seem reasonable?

jialez0:
> I suggested an alternate model where the output of the AS policy validation would be the input to the KBS policy validation. In this model the AS could evaluate the claims and return not just true/false, but a wide range out outputs. For instance the AS could return level1, level2, level3 etc. to reflect different levels of strictness. Then the KBS policy could require a certain level for a certain secret. In this approach the resource policy is decoupled from the platform. The AS could define level2 for many several different platforms. I think this would really simplify coming up with resource policies, especially for resources that are deployed on multiple platforms.

@fitzthum Okay, I think this is reasonable, and I will implement it in this PR and AS PR.

fitzthum (approved):

PR #140, opened 2023-07-28
Add docker-compose dependencies

johananl:
Right now the services are racing when running `docker-compose up`. Add depends_on clauses to the services to fix the race. This also allows running e.g. `docker-compose up kbs` which would start just KBS and its dependencies.

cc: @Xynnn007


fitzthum (approved):
LGTM. Thanks

PR #141, opened 2023-07-28
Add keys to gitignore

johananl:
We specify generating config/private.key and config/public.pub in docs/cluster.md. Gitignore these two files to avoid accidentally committing them.


fitzthum (approved):
LGTM

PR #148, opened 2023-08-18
Fix the passport-resource-kbs Make target name

johananl:
The `.PHONY` directive for `passport-resource-kbs` spells "passport-kbs" and doesn't match the target name.


jepio (approved):

fitzthum (approved):
Seems good

PR #152, opened 2023-09-01
docker: update key-broker-service Dockerfile

mythi:
Improvements and fixes:

* openssl-sys fails to build without certain .pm files so the intermediate fix is to install perl.
* reduce build steps and move "COPY" later in the build to allow caches to be used more efficiently (previously, a single modification to the repo files forced 1G of apt-get to be re-installed).
* move away from deprecated apt-key.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,54 +1,56 @@
 FROM rust:slim as builder
 
-WORKDIR /usr/src/kbs
-COPY . .
-
 ENV DEBIAN_FRONTEND noninteractive
 
-# Install Build Dependencies
-RUN apt-get update && apt-get install -y apt-utils
-RUN apt-get install -y \
-    clang \
-    cmake \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -1,54 +1,56 @@
...
 ARG KBS_FEATURES=coco-as-builtin,rustls,resource,opa
 RUN cargo install --path src/kbs --no-default-features --features ${KBS_FEATURES}
 
 FROM debian:stable-slim
 
-RUN apt-get update && apt-get install -y apt-utils
-RUN apt-get install -y \
+RUN apt-get update && \
+    apt-get install -y \
     clang \
```
mythi, line 46: question: why is `clang` needed here? it makes the final image pretty big

mythi: `After this operation, 978 MB of additional disk space will be used.`

PR #155, opened 2023-09-04
Fix security issue: avoid directory interleaving vulnerabilities

jialez0:
Add legality checking of resource path to avoid reading files in parent directory.

cc @Xynnn007

### DIFF
--- a/src/api/src/resource/mod.rs
+++ b/src/api/src/resource/mod.rs
@@ -35,6 +35,19 @@ pub struct ResourceDesc {
     pub resource_tag: String,
 }
 
+impl ResourceDesc {
+    pub fn is_valid(&self) -> bool {
+        if &self.repository_name == "."
+            || &self.repository_name == ".."
+            || &self.resource_type == "."
+            || &self.resource_type == ".."
+        {
+            return false;
+        }
+        true
+    }
+}
+
 #[derive(Clone, Debug, Deserialize, EnumString)]
 #[serde(tag = "type")]
 pub enum RepositoryConfig {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -404,7 +404,7 @@ dependencies = [
  "actix-web-httpauth",
  "aes-gcm",
  "anyhow",
- "as-types 0.1.0 (git+https://github.com/jialez0/attestation-service.git?branch=token)",
+ "as-types",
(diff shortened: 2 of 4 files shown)


jiazhang0 (approved):

docker/Dockerfile
```diff
@@ -1,37 +1,21 @@
-FROM rust:slim as builder
+FROM rust:latest as builder
 
 ENV DEBIAN_FRONTEND noninteractive
 
-RUN apt-get update && \
```
mythi, line 5: These `Dockerfile` changes are not related to the error or are they? They seem to revert my recent changes but it's not clear what's the error?

jialez0: Yes, I am studying the real reasons for CI errors, and I will remove the modifications here.

src/api/src/resource/mod.rs
```diff
@@ -35,6 +35,18 @@ pub struct ResourceDesc {
     pub resource_tag: String,
 }
 
+impl ResourceDesc {
+    pub fn check_legality(&self) -> Result<()> {
```
sameo, line 39:
```suggestion
    pub fn is_valid(&self) -> Bool {
```

A terminology nit: legality refers to the actual law (as in judges and lawyers :) ). And I think returning a boolean seems more natural?

src/api/src/resource/mod.rs
```diff
@@ -35,6 +35,18 @@ pub struct ResourceDesc {
...
+        if &self.repository_name == "." || &self.repository_name == ".." {
+            bail!(anyhow!("Illegal resource repository name"));
```
sameo, line 41: Similar to the previous comment:

```suggestion
            bail!(anyhow!("Invalid resource repository name"));
```

src/api/src/resource/mod.rs
```diff
@@ -35,6 +35,18 @@ pub struct ResourceDesc {
...
+        }
+        if &self.resource_type == "." || &self.resource_type == ".." {
```
fitzthum, line 43: I assume we have already parsed out any slashes by this point?

jialez0: es, the Actix library will help us complete the parsing of slashes when parsing the path information of URL.

PR #159, opened 2023-09-14
api_server: add SGX support to Amber AS

mythi:


mythi:
the SGX part tested with Amber, @Lu-Biao to help check TDX

Lu-Biao:
It works well with TDX. @mythi

src/api/src/attestation/amber/mod.rs
```diff
@@ -49,18 +50,18 @@ pub struct Amber {
...
-            bail!("Only implement for tdx now");
+        if tee != Tee::Tdx && tee != Tee::Sgx {
+            bail!("Amber: TEE {tee:?} is not supported.");
         }
         // get quote
         let attestation = serde_json::from_str::<Attestation>(attestation)
             .map_err(|e| anyhow!("Deserialize Attestation failed: {:?}", e))?;
-        let tdx_evidence = serde_json::from_str::<TdxEvidence>(&attestation.tee_evidence)
-            .map_err(|e| anyhow!("Deserialize TDX Evidence failed: {:?}", e))?;
+        let evidence = serde_json::from_str::<AmberTeeEvidence>(&attestation.tee_evidence)
```
Xynnn007, line 59: We'd better raise an `info` or `warn` here to tell the user that `eventlog` is ignored.

Xynnn007 (approved):
The patch looks good to me, while we need continue the talk about the architecture of KBS.

src/api/src/attestation/amber/mod.rs
mythi: With amber, `event_log` needs to be added as a separate field in the appraisal request but it's currently not done so there's no change in functionality with this PR.

mythi:
> The patch looks good to me, while we need continue the talk about the architecture of KBS.

can we get it for the release?

jialez0 (approved):

PR #168, opened 2023-10-17
Fixed e2e bgcheck + passport tests

mkulke:
- Bins are installed locally
- Reduced noisy output
- Use as much local resources as possible (AS/RVPS still uses a hardcoded /opt/coco location, issue [here](https://github.com/confidential-containers/attestation-service/issues/151))
- Reorganized the e2e test, with pid files and orderly service startup/shutdown


Xynnn007 (approved):
Thanks for fixing this!

PR #169, opened 2023-10-18
Add support az-tdx-vtpm tee

mkulke:
This is a tracking issue for adding the `az-tdx-vtpm` tee that is required to add support for Azure's TDX CVMs to CoCo.

# Related PRs

- [x] https://github.com/kinvolk/azure-cvm-tooling/pull/39
- [x] https://github.com/confidential-containers/guest-components/pull/375
- [x] https://github.com/virtee/kbs-types/pull/29

# e2e Test

The test performs a key release via remote-attestation using the `az-tdx-vtpm` tee. Build and run this PR on an Azure TDX CVM.

## Prepare

The steps assume Ubuntu 22.04:

```bash
curl -fsSL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo apt-key add -
echo 'deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu focal main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list
sudo apt-get update
sudo apt-get install -y \
    libsgx-dcap-default-qpl \
    libsgx-dcap-quote-verify \
    libsgx-dcap-quote-verify-dev \
    libsgx-urts \
    libssl-dev \
    libtdx-attest \
    libtdx-attest-dev \
    libtss2-dev \
    make \
    clang \
    protobuf-compiler \
    pkg-config

# The default golang version is 1.18 and we need at least 1.20 so install it like this.
curl -o /tmp/pullgo.sh https://raw.githubusercontent.com/surajssd/dotfiles/master/local-bin/pullgo.sh
chmod +x /tmp/pullgo.sh
/tmp/pullgo.sh 1.21.2

cat <<EOF | sudo tee /etc/sgx_default_qcnl.conf
{
  "pccs_url": "https://global.acccache.azure.net/sgx/certification/v3/",
  "use_secure_cert": true,
  "collateral_service": "https://api.trustedservices.intel.com/sgx/certification/v4/",
  "pccs_api_version": "3.1",
  "retry_times": 6,
  "retry_delay": 5,
  "local_pck_url": "http://169.254.169.254/metadata/THIM/sgx/certification/v3/",
  "pck_cache_expire_hours": 24,
  "verify_collateral_cache_expire_hours": 24,
  "custom_request_options": {
    "get_cert": {
      "headers": {
        "metadata": "true"
      },
      "params": {
        "api-version": "2021-07-22-preview"
      }
    }
  }
}
EOF
```

## Run tests

Install rust + golang

```bash
cd kbs/test
sudo apt-get update && sudo apt-get install -y build-essential
sudo make install-dependencies
make bins
sudo make e2e-test
```


attestation-service/verifier/src/az_tdx_vtpm/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// 2. Attestation nonce matches TPM Quote nonce
+    /// 3. TD Quote is genuine
+    /// 4. TD Report's report_data field matches hashed HCL variable data
+    async fn evaluate(
+        &self,
+        evidence: &[u8],
+        expected_report_data: &ReportData,
+        expected_init_data_hash: &InitDataHash,
+    ) -> Result<TeeEvidenceParsedClaim> {
+        let ReportData::Value(expected_report_data) = expected_report_data else {
```
Xynnn007, line 41:
```suggestion
        let expected_report_data = regularize_data(expected_report_data, 64, "REPORT_DATA", "Azure TDX vTPM");
        let ReportData::Value(expected_report_data) = expected_report_data else {
```

Please add a sentence before this to prevent from unexpected failure when the given `expected_report_data` is not 64 bytes long. See https://github.com/confidential-containers/kbs/pull/274/files#diff-257710d85707521628ac32349008e86e91a8fe19e49aaa4b16384e125abebe60R62-R63

attestation-service/verifier/src/az_tdx_vtpm/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+        Ok(claim)
+    }
+}
+
+fn verify_report_data(hcl_report: &HclReport, td_quote: &TdQuote) -> Result<()> {
+    let var_data_hash = hcl_report.var_data_sha256();
+    if var_data_hash != td_quote.report_body.report_data[..32] {
+        bail!("TDX Quote report data mismatch");
+    }
+    debug!("Report data verification completed successfully.");
```
Xynnn007, line 70: I love the `debug` check point here

attestation-service/verifier/src/tdx/mod.rs
```diff
@@ -10,9 +10,9 @@ use eventlog::{CcEventLog, Rtmr};
 use quote::{ecdsa_quote_verification, parse_tdx_quote};
 use serde::{Deserialize, Serialize};
 
-mod claims;
+pub(crate) mod claims;
```
Xynnn007, line 13: Just a nit: do we need this?

Xynnn007 (approved):
Thanks for the PR Magnus! There is a only problem that might cause unexpected behavior

attestation-service/verifier/src/tdx/mod.rs
mkulke: I think so, since we reference that (sub-)module from the az-tdx-vtpm sibling module of the tdx module (to share code)

```rust
use super::tdx::claims::generate_parsed_claim;
```

attestation-service/verifier/src/az_tdx_vtpm/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        let hcl_report = HclReport::new(evidence.hcl_report)?;
+        verify_tpm_quote(&evidence.tpm_quote, &hcl_report, &expected_report_data)?;
+
+        ecdsa_quote_verification(&evidence.td_quote).await?;
+        let td_quote = parse_tdx_quote(&evidence.td_quote)?;
+
+        verify_report_data(&hcl_report, &td_quote)?;
+
+        let claim = generate_parsed_claim(td_quote, None)?;
```
fitzthum, line 62: I think the TDX claims parsing has been updated to include an i[nit_data field](https://github.com/confidential-containers/kbs/blob/5210b8651bff63abf5294e01ca0f30744bc3714e/attestation-service/verifier/src/tdx/claims.rs#L128).

Is this something that we want to include for the AZ case or should we remove that key like we were talking about for SNP.

fitzthum (approved):
Looks good. See one comment

attestation-service/verifier/src/az_tdx_vtpm/mod.rs
fitzthum: Ok, you're saying that since this attester will (only?) be used with peer pods we can point the init_data field to a PCR, which will have been used to measure the init_data?

mkulke: yes

mkulke: (independent of peerpods, a user wouldn't be able to seed init_data into HW reports when using those specific Azure TEE's. Dynamic aspects like report_data and init_data would have to be covered via TPM)

fitzthum: Yeah at first I was a little concerned that this wouldn't be generic enough and would be weird for people not using peer pods, but I guess nobody will be able to use the init data anyway. So you might as well give them init data from some PCR and hope that people find the documentation for it.

mkulke: I wouldn't know another way to implement measured config-data - at least for the type that is used to bootstrap confidentiality, so things like aa_kbc_params, kbs cert.

attestation-service/verifier/src/az_tdx_vtpm/mod.rs
Xynnn007: Oh. I once thought that vtpm is also padded to 64bytes.

mkulke: 👍 I think it's limited to a maximum of 64, but you can also supply less and the quote will only contain that

attestation-service/verifier/src/az_tdx_vtpm/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+        Ok(claim)
+    }
+}
+
+fn verify_report_data(hcl_report: &HclReport, td_quote: &TdQuote) -> Result<()> {
```
fitzthum, line 67: Would it be better to call this `verify_hcl_report`. When I hear `report_data` I think of the `expected_report_data` which actually seems to be verified in the function below.

PR #171, opened 2023-10-19
Make as-types a workspace dependency

mkulke:
We need the client and the api server to be in sync, so it makes sense to share the as-types dependency among both.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,7 +32,7 @@ dependencies = [
  "actix-utils",
  "ahash 0.8.3",
  "base64 0.21.4",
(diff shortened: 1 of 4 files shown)


Xynnn007:
Would we apply https://github.com/confidential-containers/confidential-containers/issues/169 and then we can tidy the overall workspace?

Xynnn007:
@mkulke I will delete the crate in https://github.com/confidential-containers/kbs/pull/216/commits/4fab016707f799756385602baf2d6a0b63299688 

Could this issue be closed?

PR #163, opened 2023-09-20
prep for: let AS to initiate nonce/challenge

mythi:
ref: #162 / 1.

PR #227, opened 2023-11-17
attestation: verifier: tdx: Allow equals in kernel param values

jodh-intel:
Allow kernel parameter values to contain equal signs, for example:

```
module_foo=bar=baz,wibble_setting=2
```

Fixes: #163.

### DIFF
--- a/attestation-service/attestation-service/src/verifier/tdx/claims.rs
+++ b/attestation-service/attestation-service/src/verifier/tdx/claims.rs
@@ -208,14 +208,12 @@ fn parse_kernel_parameters(kernel_parameters: &[u8]) -> Result<Map<String, Value
             if item.is_empty() {
                 return None;
             }
-            let it: Vec<&str> = item.split('=').collect();
-            match it.len() {
-                1 => Some((it[0].to_owned(), Value::Null)),
-                2 => Some((it[0].to_owned(), Value::String(it[1].to_owned()))),
-                _ => {
-                    warn!("Illegal parameter: {item}");
-                    None
-                }
+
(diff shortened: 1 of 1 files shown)


jodh-intel:
Replaces https://github.com/confidential-containers/attestation-service/pull/164.

attestation-service/attestation-service/src/verifier/tdx/claims.rs
```diff
@@ -275,4 +272,229 @@ mod tests {
 
         assert_json_eq!(expected, claims);
     }
+
+    #[test]
+    fn test_parse_kernel_parameters() {
```
Xynnn007, line 277: I like the detailed test cases very much, thanks for this! Could you use https://crates.io/crates/rstest crate to make it more rusty?

fitzthum (approved):
Nice. Thanks for adding so many test cases. See Ding's comment about `rstest`.

attestation-service/attestation-service/src/verifier/tdx/claims.rs
jodh-intel: @Xynnn007 - Branch updated.

mythi:
The fixes link looks wrong

fitzthum:
> The fixes link looks wrong

True. Probably because this PR was originally opened against the AS repo. We don't actually require the fixes thing in this repo.

fitzthum:
> I mentioned it because I'm not able to re-open to continue the work. Are you able to do that? 

Ah, yes. I can reopen. Sorry about that.

PR #229, opened 2023-11-21
attestation-service: Fix report signature validation in SNP verifier

mkulke:
I noticed some issues with the signature verification in the SNP verifier code:

- openssl's ecdsa verify() will produce a bool, not a result. hence we need to consume the fn output.
- verify() requires a sha384 (SEV-SNP ABI spec, ch10) encoded report body, not the raw bytes for a successful validation

I added test cases for both successful and failed signature verification

cc @fitzthum @larrydewey PTAL


fitzthum (approved):
🙏

Thanks Magnus

PR #230, opened 2023-11-21
attestation-service: reuse SNP verifier logic in az-snp-vtpm

mkulke:
Draft for now b/c it's based on an unrelated PR to the SNP verifier

The SNP verifier functions have been made crate-public and adjusted slightly, so we can call them from the `az-snp-vtpm` verifier.


attestation-service/attestation-service/src/verifier/az_snp_vtpm/mod.rs
```diff
@@ -187,29 +130,4 @@ mod tests {
         let wrong_hcl_data: HclData = wrong_report.as_slice().try_into().unwrap();
         verify_quote(&quote, &wrong_hcl_data, nonce).unwrap_err();
     }
-
-    #[test]
-    fn test_parse_evidence() {
```
fitzthum, line 192: Might be nice to add a test like this to the other verifier at some point.

fitzthum (approved):
LGTM

Issue #231, opened 2023-11-21
Attestation-Service: replacing anyhow with explicit error types


mkulke:
In a library crate it's better to specify expected error types explicitly rather than relying on generic anyhow errors. [thiserror ](https://docs.rs/thiserror/latest/thiserror/) is a crate by the same author, which allows doing this in a convenient fashion.

kartikjoshi21:
@Xynnn007 please assign this one to me. Thanks

PR #233, opened 2023-11-22
attestation-service: fix checks for VCEK signature

mkulke:
Similar to #230 the verify function here is actually not checking the result of the verification.

There is a fail-test for the vcek signature verification, but it fails at parsing the DER cert. We should get rid of anyhow here and use proper error types (#231) which we can then assert in such tests.

- Fixed the verification logic
- Added test-case for toggled ark-ask
- Fixed fail-test to produce a legit VCEK
- ~failed parsing of bundled milan certs hould not produce runtime errors, but panic~

cc @fitzthum @larrydewey

### DIFF
--- a/attestation-service/attestation-service/src/verifier/snp/mod.rs
+++ b/attestation-service/attestation-service/src/verifier/snp/mod.rs
@@ -11,7 +11,7 @@ use openssl::{
     ecdsa,
     pkey::{PKey, Public},
     sha::sha384,
-    x509,
+    x509::{self, X509},
 };
 use serde_json::json;
 use sev::firmware::guest::AttestationReport;
@@ -104,7 +104,8 @@ pub(crate) fn verify_report_signature(
     cert_chain: &[CertTableEntry],
 ) -> Result<()> {
     // check cert chain
-    let vcek = verify_cert_chain(cert_chain)?;
+    let (ask, ark) = load_milan_cert_chain()?;
+    let vcek = verify_cert_chain(cert_chain, ask, ark)?;
 
     // OpenSSL bindings do not expose custom extensions
     // Parse the vcek using x509_parser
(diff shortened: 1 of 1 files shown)


fitzthum (approved):
LGTM.

attestation-service/attestation-service/src/verifier/snp/mod.rs
```diff
@@ -148,36 +149,36 @@ pub(crate) fn verify_report_signature(
...
 }
 
-pub fn load_milan_cert_chain() -> Result<(x509::X509, x509::X509)> {
-    let certs = x509::X509::stack_from_pem(include_bytes!("milan_ask_ark.pem"))?;
+fn load_milan_cert_chain() -> (X509, X509) {
+    let certs =
+        X509::stack_from_pem(include_bytes!("milan_ask_ark.pem")).expect("Malformed Milan ASK/ARK");
     if certs.len() != 2 {
-        bail!("Malformed Milan ASK/ARK");
+        panic!("Malformed Milan ASK/ARK");
```
Xynnn007, line 156: I am a little worried about `panic!` and `expect()` here. As attestation-service might be a service that serves different attestation requests. panic here would leave no logs to the upper caller. wdyt?

mkulke: hmm, ideally this wouldn't even compile (I'd make this const if that would be feasible). We're bundling vendor certificates in this verifier. If we fail to load them then something is seriously off. But I understand the concern that it might affect other code, if for some reason under some unforeseen circumstances this fails.

However, I think the right place to bail, if we fail to load those certs, is at verifier initialization. WDYT?

https://github.com/confidential-containers/kbs/blob/54820fd0d8b31725dbbd6cf8b9e50fa85ba7ea93/attestation-service/attestation-service/src/verifier/mod.rs#L50

Xynnn007: Sounds good to me!

jepio (approved):
lgtm

attestation-service/attestation-service/src/verifier/snp/mod.rs
mkulke: pushed. PTAL

Xynnn007 (approved):
LGTM, thanks

PR #236, opened 2023-11-23
attestation-agent: fail fast on broken AMD certs

mkulke:
Currently the milan vendor certs from AMD are bundled and parsed each time there is verification in process attestation. While this should never fail, we cannot `panic!` as it would impact healthy TEE implementations.

As an alternative we can fail fast on init of the verifier, indicating a problem at an earlier point. We also don't need to process the raw bytes multiple times, the vendor certs can be cached once parsed.

This impacts both SNP verifiers (`snp` + `az-snp-vtpm`)


Xynnn007 (approved):
LGTM, thanks!

Xynnn007:
Needs a rebase

Issue #217, opened 2023-11-12
Refactoring KBS


Xynnn007:
After merging AS code into KBS, it is a good time to refactor KBS codes.

Before v0.8.0, we use KBS to return an attestation token from CoCo AS that includes a tee pubkey inside as a claim. This will make the things coupled, because it requires that any token provisioned by an attestation service behind KBS should have a claim named `tee-pubkey`. However, except CoCo-AS, other attestation services like Amber does not have such a claim.

To let KBS to work with different kinds of attestation services, it is a good practice to decouple KBS protocol semantics from the attestation token. That is to say, the attestation service only helps to check the base TCB of the TEE, while other information like tee pubkey, the nonce, etc are authentication semantics higher than attestation. KBS should leverage the returned attestation token from the attestation behind to check if the step `Attestation` succeeds, and then decide to provision a token/credential of itself. This new token is endorsed by KBS, showing the TEE client has passed the KBS protocol and mainly used for the TEE to retrieve the resources of KBS and other privileged operations.

Xynnn007:
cc @jialez0 @fitzthum @mythi

mythi:
> To let KBS to work with different kinds of attestation services, it is a good practice to decouple KBS protocol semantics from the attestation token.

I don't think the KBS protocol is coupled with the AS today.  The challenges are mostly with the current implementation and I've summarized the gaps in #162.

You mentioned `tee-pubkey` claim. I've checked Intel Trust Authority and Azure Attestation. Both have a similar JWT token claim about "TEE held data". Intel Trust Authority based flow can be made to work just doing:

```
diff --git a/src/api/src/http/resource.rs b/src/api/src/http/resource.rs
index 26ac847..4b4f7de 100644
--- a/src/api/src/http/resource.rs
+++ b/src/api/src/http/resource.rs
@@ -47,7 +47,7 @@ pub(crate) async fn get_resource(
     })?;
 
     let pkey_value = claims
-        .get("tee-pubkey")
+        .get("attester_runtime_data")
         .ok_or(Error::AttestationClaimsParseFailed(String::from(
             "No `tee-pubkey` in the attestation claims",
         )))?;
```

> while other information like tee pubkey, the nonce, etc are authentication semantics higher than attestation.

We must also have a nonce originated from the attestation service so that each quote it verifies can be considered "fresh".

Xynnn007:
> I don't think the KBS protocol is coupled with the AS today.

please see the call stack here:

[...]

PR #240, opened 2023-11-24
[Attestation Service] Change the API of CoCo-AS

### LINKED ISSUE #177: Refactoring the Attestation Service

Xynnn007:
This PR will refactor the CoCo-AS furthermore, which will help to let CoCo-AS run as a seperate attestation-service without any coupling with KBS.

The API of CoCo AS will be definitely not related to KBS protocol. In this PR, CoCo-AS gRPC service will have the following API 
```proto
message AttestationRequest {
    // TEE enum. Specify the evidence type
    Tee tee = 1;

    // Base64 encoded evidence. The alphabet is URL_SAFE_NO_PAD.
    // defined in https://datatracker.ietf.org/doc/html/rfc4648#section-5
    string evidence = 2;

    // Runtime Data used to check the binding relationship with report data in
    // Evidence
    oneof runtime_data {
        // Base64 encoded runtime data slice. The whole string will be base64
        // decoded. The result one will then be accumulated into a digest which
        // is used as the expected runtime data to check against the one inside
        // evidence.
        //
        // The alphabet is URL_SAFE_NO_PAD.
        // defined in https://datatracker.ietf.org/doc/html/rfc4648#section-5
        string raw_runtime_data = 3;

        // Runtime data in a JSON map. CoCoAS will rearrange each layer of the
        // data JSON object in dictionary order by key, then serialize and output
        // it into a compact string, and perform hash calculation on the whole
        // to check against the one inside evidence.
        //
        // After the verification, the structured runtime data field will be included
        // inside the token claims.
        string structured_runtime_data = 4;
    }

    // Init Data used to check the binding relationship with init data in
    // Evidence
    oneof init_data {
        // Base64 encoded init data slice. The whole string will be base64
        // decoded. The result one will then be accumulated into a digest which
        // is used as the expected init data to check against the one inside
        // evidence.
        //
        // The alphabet is URL_SAFE_NO_PAD.
        // defined in https://datatracker.ietf.org/doc/html/rfc4648#section-5
        string raw_init_data = 5;

        // Init data in a JSON map. CoCoAS will rearrange each layer of the
        // data JSON object in dictionary order by key, then serialize and output
        // it into a compact string, and perform hash calculation on the whole
        // to check against the one inside evidence.
        // 
        // After the verification, the structured init data field will be included
        // inside the token claims.
        string structured_init_data = 6;
    }

    // Hash algorithm used to calculate runtime data. Currently can be "sha256",
    // "sha384" or "sha512". If not specified, "sha384" will be selected.
    string runtime_data_hash_algorithm = 7;

    // Hash algorithm used to calculate init data. Currently can be "sha256",
    // "sha384" or "sha512". If not specified, "sha384" will be selected.
    string init_data_hash_algorithm = 8;

    // List of IDs of the policy used to check evidence. If not provided,
    // a "default" one will be used.
    repeated string policy_ids = 9;
}
```

Close #177
Related to #217

cc @jialez0


kbs/src/api/src/attestation/coco/grpc.rs
```diff
@@ -85,10 +89,27 @@ impl Attest for Grpc {
...
+        let customized_claims = serde_json::to_string(&pubkey_claim)?;
+
         let req = tonic::Request::new(AttestationRequest {
-            tee: to_grpc_tee(tee) as i32,
-            nonce: String::from(nonce),
-            evidence: String::from(attestation),
+            tee: to_grpc_tee(tee).into(),
+            evidence: STANDARD.encode(attestation.tee_evidence),
+            runtime_data,
+            init_data: vec![],
```
Xynnn007, line 109: cc @mkulke You might be interested in this -- This is the KBS side code that does not leverage the ability of the initdata verafication in this PR.

jialez0 (approved):
LGTM. Thanks! @Xynnn007

mythi:
> `message AttestationRequest {`

we may also need `nonce`?

Xynnn007:
> > `message AttestationRequest {`
> 
> we may also need `nonce`?

Nonce is something that will be bound to the report data. This API now gives a general parameter `rubtime`

Xynnn007:
> > `message AttestationRequest {`
> 
> we may also need `nonce`?

Nonce is something that will be bound to the report data. This API now gives a general parameter `runtime_data` to check the binding between any report data and the evidence.

Xynnn007:
> This API now gives a general parameter runtime_data to check the binding between any report data and the evidence.\n\nAFAIUI, the AS cannot generate \"expected report data\" based on runtime_data (Tee pubkey) only if it does not know the nonce the attester used to generate the quote/report data.

You are right. 

This api is now for user who knows the materials to calculate the report data.

We can add an extra API, which will give a nonce to the caller to get evidence. Then the user can call this attestation API.

attestation-service/attestation-service/src/lib.rs
```diff
@@ -72,44 +103,50 @@ impl AttestationService {
...
+    /// AS will not do any extra check upon this parameter.
+    #[allow(clippy::too_many_arguments)]
+    pub async fn evaluate(
+        &self,
+        evidence: Vec<u8>,
+        tee: Tee,
+        runtime_data: Vec<Vec<u8>>,
+        init_data: Vec<Vec<u8>>,
+        hash_algorithm: HashAlgorithm,
+        policy_ids: Vec<String>,
```
jepio, line 131: why multiple `policy_ids`? is it worth the added complexity on the verifier side (the order of policy evaluation can change). can we stick to one policy_id.

attestation-service/attestation-service/src/lib.rs
```diff
@@ -72,44 +103,50 @@ impl AttestationService {
...
-            .context("Failed to deserialize Attestation")?;
+    /// Issue an attestation results token which contain TCB status and TEE public key. Input parameters:
+    /// - `evidence`: TEE evidence bytes. This might not be the raw hardware evidence bytes. Definitions
+    /// are in `verifier` crate.
+    /// - `tee`: concrete TEE type
+    /// - `runtime_data`: plaintexts of the runtime_data. These data will be hased using the specified algorithm
+    /// to calculate a digest. Then the digest will be used to match the `report_data` field of the evidence.
+    /// - `init_data`: plaintexts of the init_data. These data will be hased using the specified algorithm
+    /// to calculate a initdata hash. Then the initdata hash will be used to match the `init_data_hash` field of the
+    /// evidence.
```
jepio, line 115: we need to have a better way to differentiate between "no runtime data" and "empty runtime data". and same with initdata. Hash of "" is not all zeros should be valid and is different than all zeros.

jepio:
I'm still reviewing, but the attestation requester should not be passing `policy_id` to evaluate against, that is insecure. The policy in RATS is supplied by the verifier owner, which is different then the attester.

attestation-service/attestation-service/src/lib.rs
```diff
@@ -122,18 +159,30 @@ impl AttestationService {
...
-            .evaluate(reference_data_map, tcb_json, None)
+            .evaluate(reference_data_map, tcb_json, policy_ids.clone())
             .await
             .map_err(|e| anyhow!("Policy Engine evaluation failed: {e}"))?;
 
+        let evaluation_reports: Vec<_> = evaluation_report
+            .into_iter()
+            .map(|(k, v)| {
+                json!({
+                    "policy-id": k,
```
jepio, line 172: i see that policy-id is a `filename` i don't think that is future proof or secure.

security:
think of the relying party that wants to check that the token was checked against a certain policy. filenames can be changed.

future-proof:
filenames are tied to file abstractions, policies could be hosted elsewhere.

Can we store the hash of the policy in the attestation result instead? That is stable and unique.

attestation-service/bin/grpc-as/src/server.rs
```diff
@@ -86,17 +92,52 @@ impl AttestationService for Arc<RwLock<AttestationServer>> {
...
+
+        let init_data: Vec<Vec<u8>> = request
+            .init_data
+            .iter()
+            .map(|data_item| {
+                STANDARD
+                    .decode(data_item)
+                    .map_err(|e| Status::aborted(format!("base64 decode init data: {e}")))
+            })
+            .collect::<std::result::Result<Vec<Vec<u8>>, Status>>()?;
```
jepio, line 121: Why array of base64 strings? I think these should just be base64 strings. We also need to document the fact that parameters are base64 encoded and document which base64 encoding is used in README.md. In remote APIs it is most commonly `URL_SAFE_NO_PAD`, not `STANDARD`.

attestation-service/attestation-service/src/lib.rs
```diff
@@ -18,14 +18,45 @@ pub use kbs_types::{Attestation, Tee};
...
 
 use crate::utils::flatten_claims;
 
+#[derive(EnumString, AsRefStr)]
+pub enum HashAlgorithm {
+    #[strum(ascii_case_insensitive)]
+    Sha384,
+
+    #[strum(ascii_case_insensitive)]
+    Sha512,
```
jepio, line 36: can you add sha256 (SNP host data is only 256 bits)

jepio (changes requested):
I think this is a step in the right direction but needs some more thought.

It's not secure to mix claims attested through runtime data (like tee-pubkey, nonce) with "custom claims" which are unattested and user controlled. So we really need to impose some structure on runtime data (like a json dictionary) that can be included in the attested token separately from any user supplied data.

Custom claims should instead be a "nonce" supplied by the user to be included in the attestation token.

Xynnn007:
@jepio `policy_id` in the api would help the caller to apply different use cases.

The key point is that the returned attestation token will have information about the policy ids that are used to check against the evidence. Current AS only has a fixed policy and cannot choose a specific one from API.

Doea this make sense?

jepio:
> @jepio `policy_id` in the api would help the caller to apply different use cases.
> 
> The key point is that the returned attestation token will have information about the policy ids that are used to check against the evidence. Current AS only has a fixed policy and cannot choose a specific one from API.
> 
> Doea this make sense?

I see. I now see that ITA also allows the user the option to pass the policy in a similar way. Why don't we do it same way as well:
- policy id parameter is optional and then default policy is applied
- policy id is hash/uuid (i think hashes are a better option as they are unique and unmodifiable), not name

attestation-service/attestation-service/src/lib.rs
Xynnn007: Nice idea. Let me add the digest of the policy too. BTW, the name of the policy might help the consumers to have some (though probably not trusted as you mentioned) contexts of the policy, so maybe we can keep it there?

Xynnn007:
> > @jepio `policy_id` in the api would help the caller to apply different use cases.
> > The key point is that the returned attestation token will have information about the policy ids that are used to check against the evidence. Current AS only has a fixed policy and cannot choose a specific one from API.
> > Doea this make sense?
> 
> I see. I now see that ITA also allows the user the option to pass the policy in a similar way. Why don't we do it same way as well:
> 
> * policy id parameter is optional and then default policy is applied
> * policy id is hash/uuid (i think hashes are a better option as they are unique and unmodifiable), not name

This is a great point. I think this is related to how policies are indexed. Currently we only have OPA as policy engine and the regos are stored in filesystem so we directly retrieve them with the file name.

In this PR, I added the policy digest into the result JWT as you suggested. However how to specify which policy should be used might be another topic, as it is related to how it is stored. We can promote this in another PR. wdyt?

Xynnn007:
> It's not secure to mix claims attested through runtime data (like tee-pubkey, nonce) with "custom claims" which are unattested and user controlled. So we really need to impose some structure on runtime data (like a json dictionary) that can be included in the attested token separately from any user supplied data.


It is a good idea that we should have a runtime data format/spec like undergoing initdata. This could help us to leverage report_data field in a more standard way.

But for custormized claims, my idea is that this would give ability to the user to embed some extra context information. Whether to use such claims is determined by the consumer of the JWT. In CoCo-AS/KBS model, the attestation requester and JWT consumer both are KBS, so it is secure for KBS itself to leverage the pub-key fields inside the claims.


> Custom claims should instead be a "nonce" supplied by the user to be included in the attestation token.


From the perspective of opposing server replay attack, this is worth doing.

attestation-service/protos/attestation.proto
```diff
@@ -10,13 +10,53 @@ enum Tee {
...
+
+    // Base64 encoded runtime data slices. Every single string will
+    // be base64 decoded and all the result slices will then be concentrated
+    // to get a compound one. This compound one will then be accuulated into
+    // a digest which is used as the expected runtime data to check against
+    // the one inside evidence.
+    //
+    // The alphabet is URL_SAFE_NO_PAD.
+    // defined in https://datatracker.ietf.org/doc/html/rfc4648#section-5
+    repeated string runtime_data = 3;
```
mythi, line 32: this seems partly redundant because the api already has `evidence` which is according to the KBS protocol spec ("The KBS does not parse or analyze the attestation evidence, it forwards it to the Attestation-Service for verification."). The part that is missing is `nonce` so it could be a dedicated entry here (instead of `runtime_data`).

mythi: Sorry,I realized I mixed things so my earlier comment can be ignored.

attestation-service/attestation-service/src/lib.rs
jepio: in that case i'd prefer uuid + hash than name and hash

attestation-service/bin/grpc-as/example.token.json
```diff
@@ -0,0 +1,47 @@
+{
+    "customized_claims": {
+        "test_key": "test_value"
+    },
+    "evaluation-reports": [
+        {
+            "evaluation-result": "{\"allow\":true}",
```
jepio, line 7: does the evaluation result have to be a string? can't it be:
```json
"evaluation-result": {
  "allow": true
}
```

attestation-service/attestation-service/src/lib.rs
jepio: i think so

attestation-service/bin/grpc-as/src/server.rs
jepio: are there optional parameters in grpc? i care more about how this works for a JSON http API, where we can have can have:
```json
"runtime_data": {}
```
or
```json
"runtime_data": { "data": "" }
```
or skip passing runtime_data.

jepio:
> > > @jepio `policy_id` in the api would help the caller to apply different use cases.
> > > The key point is that the returned attestation token will have information about the policy ids that are used to check against the evidence. Current AS only has a fixed policy and cannot choose a specific one from API.
> > > Doea this make sense?
> > 
> > 
> > I see. I now see that ITA also allows the user the option to pass the policy in a similar way. Why don't we do it same way as well:
> > 
> > * policy id parameter is optional and then default policy is applied
> > * policy id is hash/uuid (i think hashes are a better option as they are unique and unmodifiable), not name
> 
> This is a great point. I think this is related to how policies are indexed. Currently we only have OPA as policy engine and the regos are stored in filesystem so we directly retrieve them with the file name.
> 
> In this PR, I added the policy digest into the result JWT as you suggested to ensure the integrity of policy recorded in the JWT. However how to specify which policy should be used might be another topic, as it is related to how it is stored. We can promote this in another PR. wdyt?

I think we need to address this before. We can't change the token format every week because without stability we won't see adoption. We should not be leaking the fact that policies are stored in filesystem into our token API - in a cloud deployment of CoCo AS the backend storage will be blob storage. So I think we need policy management APIs like ITA (https://docs.trustauthority.intel.com/main/restapi/restapi-policy-management.html). When a user registers a policy they need to either pass a UUID or the API returns a UUID. And then this UUID would be passed with the attestation request. WDYT?

jepio:
> > It's not secure to mix claims attested through runtime data (like tee-pubkey, nonce) with "custom claims" which are unattested and user controlled. So we really need to impose some structure on runtime data (like a json dictionary) that can be included in the attested token separately from any user supplied data.
> 
> It is a good idea that we should have a runtime data format/spec like undergoing initdata. This could help us to leverage report_data field in a more standard way. A json map would help both protect the integrity and embed into extra claims.
> 
> But for custormized claims, my idea is that this would give ability to the user to embed some extra context information without integrity protection. Whether to use such claims is determined by the consumer of the JWT. In CoCo-AS/KBS model, the attestation requester and JWT consumer both are KBS, so it is secure for KBS itself to leverage the pub-key fields inside the claims.

No, it's not secure. Anyone can request a CoCo-AS token  passing an arbitrary public key in "custom claims", also one not stored in a TEE and then pass that token to CoCo KBS. So "runtime-claims" (attested through a REPORTDATA mechanism) need to be separated from arbitrary "custom claims" in the token. Anything else can't be trusted by consumers.

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -63,51 +63,74 @@ impl PolicyEngine for OPA {
...
+                .map(|a| a.unwrap_or(true))
+                .unwrap_or(true);
 
-        // If a clear approval opinion is given in the evaluation report,
-        // the rejection information will be reflected in the evaluation failure return value.
-        let res_kv: Value = serde_json::from_str(&res)?;
-        if let Some(allow) = res_kv["allow"].as_bool() {
             if !allow {
-                bail!("Untrusted TEE evidence")
+                bail!("TEE evidence does not pass policy {policy_id}");
```
jepio, line 131: does the policy evaluation return a more verbose error when evaluation fails? something like "field X didn't match condition Y".

if so, then let's return that.

attestation-service/attestation-service/src/lib.rs
```diff
@@ -122,18 +171,29 @@ impl AttestationService {
...
             .await
             .map_err(|e| anyhow!("Policy Engine evaluation failed: {e}"))?;
 
+        let policies: Vec<_> = evaluation_report
+            .into_iter()
+            .map(|(k, v)| {
+                json!({
+                    "policy-id": k,
+                    "policy-hash": v.0,
+                    "evaluation-result": v.1,
```
jepio, line 186: if this is always guaranteed to be `{"allow" : true }` then we don't need this field any longer

attestation-service/bin/grpc-as/example.token.json
Xynnn007: This is the raw string output by OPA when using the `default_policy.rego`. I am afraid that we cannot assume what the result will look like. It is decided by the rego.

jepio:
we can specify what it needs to look like to be valid. and besides: the code already parses the OPA output and checks that it matches this structure.
________________________________

attestation-service/bin/grpc-as/example.token.json
jepio: we can specify what it needs to look like to be valid. and besides: the code already parses the OPA output and checks that it matches this structure.

attestation-service/bin/grpc-as/src/server.rs
jepio: if we stick to an array it would be more like:
* no "runtime_data" or `"runtime_data": []` - we don't check
* `"runtime_data": [""]` we check `hash(b"")`.

attestation-service/bin/grpc-as/example.token.json
Xynnn007: Let me parse the output directly as a JSON object. I think the document of policies can be added in another PR that also can promote the policy storage.

attestation-service/attestation-service/src/lib.rs
Xynnn007: No. as mentioned, it is determined by the rego.

Xynnn007:
@jepio 
1. about policy management. I agree with your ideas. I think we can implement them in another PR rather than this one. Currently `(name, hash)` pair is equal to `(uuid, hash)` pair, so after merging this PR we can still impl the policy part without changing the `attestation` api. In the token, it is always called `policy-id`.
2. Custormized claims. Makes sense. It requires that we should check the integrity of runtime data with the one in the evidence.
3. Let me added a separate commit to check whether `protoc` in GH runner supports `optional` in proto3.

Xynnn007:
@jepio Updated the PR. Added a separate commit where I use `oneof` key word to support both `raw` init/runtime data match and `structured` init/runtime data match. PTAL if it is good.

attestation-service/attestation-service/src/lib.rs
```diff
@@ -72,44 +122,54 @@ impl AttestationService {
...
+    /// - `init_data`: These data field will be used to check against the counterpart inside the evidence.
+    /// The concrete way of checking is decide by the enum type. If this parameter is set `None`, the comparation
+    /// will not be performed.
+    /// - `hash_algorithm`: The hash algorithm that is used to calculate the digest of `runtime_data` and
+    /// `init_data`.
+    /// - `policy_ids`: The policy ids that used to check this evidence. Any check fails against a policy will
+    /// not cause this function to return error. The result check against every policy will be included inside
+    /// the finally Token returned by CoCo-AS.
+    /// - `customized_claims`: user defined JSON Map object that will be included in the result token.
+    /// AS will not do any extra check upon this parameter.
```
fitzthum, line 142: Looks like this field as removed? We should update the comments.

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -63,51 +63,76 @@ impl PolicyEngine for OPA {
...
+                .map(|a| a.unwrap_or(true))
+                .unwrap_or(true);
 
-        // If a clear approval opinion is given in the evaluation report,
-        // the rejection information will be reflected in the evaluation failure return value.
-        let res_kv: Value = serde_json::from_str(&res)?;
-        if let Some(allow) = res_kv["allow"].as_bool() {
             if !allow {
-                bail!("Untrusted TEE evidence")
+                bail!("TEE evidence does not pass policy {policy_id}, reason: {policy_res}");
```
fitzthum, line 131: I'm not sure this is the best way to handle this. It might be better to return a token that says that a policy has not passed evaluation. It's a bit hard to know. Maybe that would increase the complexity of the KBS.

If we stick with this approach we should make sure to document it well for clients who are trying to write policies.

fitzthum (commented):
This looks good. 

A few questions: 

Do you think it would be useful if we had some kind of default policy? This could be set per-AS and evaluated on every request. It might help simplify the process of provisioning policies for simple use cases.

How does reference data fit into this? It looks like reference values aren't set via the attestation request or reported in the attestation results. The same policy could have very different results with different reference values.

attestation-service/attestation-service/src/lib.rs
Xynnn007: Thanks for this. Let me fix

attestation-service/attestation-service/src/lib.rs
```diff
@@ -72,44 +112,54 @@ impl AttestationService {
...
+    #[allow(clippy::too_many_arguments)]
+    pub async fn evaluate(
+        &self,
+        evidence: Vec<u8>,
+        tee: Tee,
+        runtime_data: Vec<Vec<u8>>,
+        init_data: Vec<Vec<u8>>,
+        hash_algorithm: HashAlgorithm,
```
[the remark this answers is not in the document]
Xynnn007: Sounds a good idea. Let me add the parameter.

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
Xynnn007: In the initial implementation I embed the policy validation results into the result token. Suggested by @jepio  https://github.com/confidential-containers/kbs/pull/240#discussion_r1411795084   I think it is right that if any policy fails to check, the verifiecation should fail.

attestation-service/attestation-service/src/lib.rs
```diff
@@ -72,44 +112,54 @@ impl AttestationService {
...
     /// Evaluate Attestation Evidence.
-    /// Issue an attestation results token which contain TCB status and TEE public key.
-    pub async fn evaluate(&self, tee: Tee, nonce: &str, attestation: &str) -> Result<String> {
-        let attestation = serde_json::from_str::<Attestation>(attestation)
-            .context("Failed to deserialize Attestation")?;
+    /// Issue an attestation results token which contain TCB status and TEE public key. Input parameters:
+    /// - `evidence`: TEE evidence bytes. This might not be the raw hardware evidence bytes. Definitions
+    /// are in `verifier` crate.
+    /// - `tee`: concrete TEE type
+    /// - `runtime_data`: plaintexts of the runtime_data. These data will be hased using the specified algorithm
```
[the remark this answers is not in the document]
Xynnn007: Thanks for this. BTW in the last commit of this PR I overwrote the logic here. If the last commit makes sense, I will do a rebase to fix the typo

Xynnn007:
> How does reference data fit into this? It looks like reference values aren't set via the attestation request or reported in the attestation results. The same policy could have very different results with different reference values.

Reference data seems a design challlenge. Reasons

1. We currently tightly depend on the OPA policy to check the reference data. That is to say, if the policy does not do the check, the reference data actually do not work.
2. @thomas-fossati once raised a good point that compound reference values are needed but still not supported by the corrent RVPS.

But overall, I agree with you that attestation token should have the reference data information inside. So let just temporarily ignore how 1 & 2 are handled, and just try to fix the token format, s.t. another field containing reference data information that was used when doing verification. wdyt?

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
Xynnn007: So there are two options for now
1. Provision token only if all policies pass. less flexible for policy but easy to use token.
2. Provision token regardless of any policy enforcement result. More flexible for policy but needs extra logic to use token.

right?

Now I am ok with both.

wdyt? @fitzthum @jepio

Xynnn007:
Rebased and update the PR with the new API.

@jepio @fitzthum @jialez0 a decision left: whether the token can be provisioned without passing all OPA policies

@mythi Added the KBS side check for https://github.com/confidential-containers/guest-components/issues/366 [here](https://github.com/confidential-containers/kbs/pull/240/files#diff-16118c763a8e72d853ec0c7c5d6e8355496c41d1b3da27bf0e4c432a2ea158a8R30) I will put a PR on `guest-components` side then.

Xynnn007:
Container build CI is not related to this PR. Will be resolved in a separate PR.

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
fitzthum: Yeah I think either one is fine. The first option sort of implies that the AS policies are going to return binary values, which isn't necessarily the case. The second option will likely increase the complexity of the KBS policies. I guess we can pick one, merge it, and see how we feel about it down the road.

fitzthum (approved):

jialez0 (approved):

mythi (approved):
LGTM

jepio (approved):
lgtm. I'm ok with either of the approaches (always release token or only release token when all policies pass) until we find some problem.

PR #241, opened 2023-11-27
Cca: Get the evidence from EAR (EAT Attesation Result)

chendave:
CCA validation by the Verasion has some overlapping with the RVPS, the similar validation has been done by the Verasion already. So the validation of evidence in CoCo is not needed in theory.

The parse of the CCA token here is to align with other verifier, e.g. TDX, and to support RVPS if that is the case of future planning.

NOTE that each of the value we get from EAR is base64 encoded hex value. So, the reference value that are registered for RVPS are expected to be base64 encoded hex value as well.

[1] https://datatracker.ietf.org/doc/draft-ietf-rats-ar4si/

### DIFF
--- a/attestation-service/verifier/Cargo.toml
+++ b/attestation-service/verifier/Cargo.toml
@@ -11,7 +11,7 @@ sgx-verifier = [ "scroll", "sgx-dcap-quoteverify-rs" ]
 az-snp-vtpm-verifier = [ "az-snp-vtpm", "sev", "snp-verifier" ]
 snp-verifier = [ "asn1-rs", "openssl", "sev", "x509-parser" ]
 csv-verifier = [ "openssl", "csv-rs", "codicon" ]
-cca-verifier = [ "cbor-diag", "veraison-apiclient" ]
+cca-verifier = [ "ear", "veraison-apiclient" ]
 
 [dependencies]
 anyhow.workspace = true
@@ -21,7 +21,6 @@ az-snp-vtpm = { version = "0.4", default-features = false, features = ["verifier
 base64 = "0.21"
 bincode = "1.3.3"
 byteorder = "1"
-cbor-diag = { version = "0.1.11", optional = true }
 cfg-if = "1.0.0"
 codicon = { version = "3.0", optional = true }
 # TODO: change it to "0.1", once released.
@@ -40,7 +39,7 @@ sev = { version = "1.2.0", features = ["openssl", "snp"], optional = true }
 sgx-dcap-quoteverify-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", tag = "DCAP_1.16", optional = true }
 strum.workspace = true
 veraison-apiclient = { git = "https://github.com/chendave/rust-apiclient", branch = "token", optional = true }
-ear = { git = "https://github.com/veraison/rust-ear", rev = "cc6ea53" }
+ear = { git = "https://github.com/veraison/rust-ear", rev = "cc6ea53", optional = true }
 x509-parser = { version = "0.14.0", optional = true }
 
 [build-dependencies]
--- a/attestation-service/verifier/src/cca/mod.rs
+++ b/attestation-service/verifier/src/cca/mod.rs
@@ -10,7 +10,7 @@ use base64::Engine;
 use core::result::Result::Ok;
 use ear::{Ear, RawValue};
 use jsonwebtoken::{self as jwt};
-use log::{error, info, warn};
+use log::{debug, error, info, warn};
 use serde::{Deserialize, Serialize};
 use std::{collections::BTreeMap, str};
 use veraison_apiclient::*;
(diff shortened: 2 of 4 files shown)


attestation-service/verifier/Cargo.toml
```diff
@@ -11,7 +11,7 @@ sgx-verifier = [ "scroll", "sgx-dcap-quoteverify-rs" ]
 az-snp-vtpm-verifier = [ "az-snp-vtpm", "sev", "snp-verifier" ]
 snp-verifier = [ "asn1-rs", "openssl", "sev", "x509-parser" ]
 csv-verifier = [ "openssl", "csv-rs", "codicon" ]
-cca-verifier = [ "cbor-diag", "veraison-apiclient" ]
+cca-verifier = [ "veraison-apiclient" ]
```
Xynnn007, line 14: not related to this PR. Could you also add `ear` dependency as optional under cca-verifier feature?

attestation-service/verifier/src/cca/mod.rs
```diff
@@ -69,12 +86,12 @@ impl Verifier for CCA {
 
         match expected_report_data.len() {
             0..=63 => {
-                warn!("The input report_data of CCA is shorter than 64 bytes, will be padded with '\\0'.");
+                log::warn!("The input report_data of CCA is shorter than 64 bytes, will be padded with '\\0'.");
```
Xynnn007, line 89: We might already have import `warn!` macro, thus do not need this?

attestation-service/verifier/src/cca/mod.rs
```diff
@@ -69,12 +86,12 @@ impl Verifier for CCA {
...
                 expected_report_data.resize(64, b'\0');
             }
             64 => {}
             _ => {
-                warn!("The input report_data of CCA is longer than 64 bytes, will be truncated to 64 bytes.");
+                log::warn!("The input report_data of CCA is longer than 64 bytes, will be truncated to 64 bytes.");
```
Xynnn007, line 94: We might already have import warn! macro, thus do not need this?

attestation-service/verifier/src/cca/mod.rs
```diff
@@ -118,6 +135,8 @@ impl Verifier for CCA {
             .context("decrypt the ear with the decoding key")?;
 
         let ear_nonce = plain_ear.nonce.context("get nonce from ear")?;
+        let evidence = &plain_ear.submods["CCA_SSD_PLATFORM"].annotated_evidence;
```
Xynnn007, line 138: Is submods expected to have a field named "CCA_SSD_PLATFORM"? BTreeMap implementation shows that if we use `[...]`, the key should exist, or the process would panic https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#2328

Xynnn007 (approved):
Thanks @chendave Only some small nits

attestation-service/verifier/src/cca/mod.rs
chendave: I removed the import of the `warn` from log, will revert it back if you prefer to import the macro.

attestation-service/verifier/src/cca/mod.rs
Xynnn007: Ok. I get your point that veraison is expected to have such a field. However, from the perspective of AS we cannot expect it will happen as a malicious/wrong remote veraison might do unexpected things. In such case, the caller of `verifier` should know why the function fails to execute rather than unexpected panic. That is to say we'd better use an error handling here.  wdyt?

chendave: make sense, will update.

attestation-service/verifier/src/cca/mod.rs
```diff
@@ -130,156 +154,104 @@ impl Verifier for CCA {
             warn!("CCA currently does not support parse `cca_realm_personalization_value`. Init data hash check skipped.");
```
Xynnn007, line 154: Can we support to verify this item after this PR?

attestation-service/verifier/Cargo.toml
chendave: fixed.

attestation-service/verifier/src/cca/mod.rs
chendave: yes, I will push one additional commit to fix it.

attestation-service/verifier/src/cca/mod.rs
chendave: fixed.

attestation-service/verifier/src/cca/mod.rs
chendave: pls see: https://github.com/confidential-containers/kbs/pull/241/commits/90fb4b61e3e69c0cbbf993832808c080676b9da9

fitzthum (approved):
LGTM

chendave:
Can we merge this? or anything else I can do?

PR #243, opened 2023-11-28
CSV Verifier: Update Evidence format

jialez0:
1. Remove unnecessary Base64 encoding from the Evidence structure and replace it with native structure content. This is because in the parameters of the AS top-level API, we assume that Evidence has already undergone Base64 encoding once, and the structure here serves as an internal interface parameter, so there is no need to repeat Base64 encoding.

2. Added the Serial number field in the Evidence structure, serial number is to identify the HYGON chip ID. This field improves the completeness of CSV evidence, allowing for different release logic to be set for different HYGON chips when setting AS policies. This modification comes from the needs of scene users.

3. Added the content of report data in the parsed Claims.

cc @Xynnn007 @BaoshunFang

### DIFF
--- a/attestation-service/verifier/src/csv/mod.rs
+++ b/attestation-service/verifier/src/csv/mod.rs
@@ -26,8 +26,9 @@ struct CertificateChain {
 
 #[derive(Serialize, Deserialize)]
 struct CsvEvidence {
-    attestation_report: String,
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
lgtm

fitzthum (approved):
LGTM

Issue #87, opened 2023-05-17
 Amber currently has no public document.


Lu-Biao:
There is already a patch(https://github.com/confidential-containers/kbs/pull/77) to support Amber as the authentication service for KBS. But Amber has no public document for describing the details of the Amber API. After the Amber document is published, the Amber document URL should be added to the KBS document.

PR #244, opened 2023-11-28
Rename Amber to Intel Trust Authority

mythi:
Amber is officially called Intel Trust Authority and the code name is not to be used anymore.

Fixes: #87


fitzthum (approved):
LGTM. It might be nice to put a note in the docs that mentions that it used to be called Amber. Also, in some places I wonder if we could just call it `trust-authority` rather than `intel-trust-authority`. Both of these might be non-starters so I guess it is fine either way.

Xynnn007 (approved):

PR #245, opened 2023-11-28
attestation-service: bump az-snp-vtpm verifier

mkulke:
Updated the crate for az-snp-vtpm TEE. Since the API evolved the verification code was adjusted slightly and the test coverage has been increased.

### DIFF
--- a/attestation-service/verifier/src/az_snp_vtpm/mod.rs
+++ b/attestation-service/verifier/src/az_snp_vtpm/mod.rs
@@ -12,11 +12,12 @@ use crate::snp::{
 use anyhow::{bail, Context, Result};
 use async_trait::async_trait;
 use az_snp_vtpm::certs::Vcek;
-use az_snp_vtpm::hcl::HclData;
-use az_snp_vtpm::vtpm::{Quote, VerifyVTpmQuote};
-use log::warn;
+use az_snp_vtpm::hcl::HclReport;
+use az_snp_vtpm::report::AttestationReport;
+use az_snp_vtpm::vtpm::Quote;
+use log::{debug, warn};
+use openssl::pkey::PKey;
 use serde::{Deserialize, Serialize};
-use sev::firmware::guest::AttestationReport;
 use sev::firmware::host::{CertTableEntry, CertType};
 
 const HCL_VMPL_VALUE: u32 = 0;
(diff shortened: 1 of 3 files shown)


attestation-service/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -106,49 +123,96 @@ fn verify_snp_report(
...
-        let message = include_bytes!("../../test_data/az-vtpm-quote-msg.bin").to_vec();
-        let quote = Quote { signature, message };
-        let report = include_bytes!("../../test_data/az-hcl-data.bin");
-        let hcl_data: HclData = report.as_slice().try_into().unwrap();
+        let quote = Quote {
+            signature: SIGNATURE.to_vec(),
+            message: MESSAGE.to_vec(),
+        };
+        let hcl_report = HclReport::new(REPORT.to_vec()).unwrap();
         let nonce = "challenge".as_bytes();
```
fitzthum, line 176: So this value is baked into the `message` that is loaded from binary above? Maybe this should also be a constant? Might be clearer to call it `report_data` rather than nonce. Not really a big deal either way.

attestation-service/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -42,6 +43,13 @@ impl AzSnpVtpm {
     }
 }
 
+// The verifier performs the following verification steps:
```
fitzthum, line 46: Slightly confusing that these don't seem to happen in this order.

fitzthum (approved):
A few minor comments but looks good

attestation-service/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -42,6 +43,13 @@ impl AzSnpVtpm {
...
+// 1. SNP Report is genuine
+// 2. SNP Report has been issued in VMPL 0
+// 3. TPM Quote has been signed by AK included in the HCL variable data
+// 4. Attestation nonce matches TPM Quote nonce
+// 5. SNP report's report_data field matches hashed HCL variable data
```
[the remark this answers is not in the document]
mkulke: yup that makes sense 👍

Xynnn007 (approved):
lgtm

Issue #228, opened 2023-11-21
[AS] Return generic report_data & init_data field in Attestation Claim


Xynnn007:
Inspired by https://github.com/confidential-containers/kbs/pull/216#discussion_r1399837721 we should also embed two more claims `init_data` and `report_data` to the Attestation-Claims of each verifier handler. This would do great help to the consumer of the attestation claims without needing to understand the TEE specific information.

cc @fitzthum @jialez0

Xynnn007:
Tasklist
- [ ] Sample
- [ ] SGX
- [ ] TDX
- [ ] SNP
- [ ] Azure-vTPM-SNP
- [ ] CSV
- [ ] CCA

Xynnn007:
> Second, is there any way in the source to make sure that developers don't forget to add the right claims to the map? I'm not really aware of a way to do this.

This is surely a problem. Might be two ways.

1. Mandatory one enforced by compiler: A well defined structure as return type.
2. non-mandatory one: prompt developers by adding more comments above the definition of `verifier`.

Also would like to get viewpoints from @mkulke

mkulke:
If we pass through init_data_hash and report_data to the verifiers `evaluate()` fn, can't we add those to the claim-set on the caller site after a successful verification?

regarding init-data on vTPM based attestation mechanisms (this will also apply to the az-tdx-vtpm TEE):

If we consider measured initialisation data to be a key property of cc_kbc then we there should be support for that in all verifiers, I think.

For SNP/TDX the measurement would be included as part of the signed report/quote in HOSTDATA/MRCONFIGID fields. vTPM based attesters would also need to provide a measurement as part of the evidence. It doesn't really matter how init-data would be injected/retrieved in the TEE, but it needs to be measured somehow into the vTPM (which is linked to a SNP/TDX attestation report/quote).

consequently, we need to establish trust in the component that performs the init-data measurement.

For example: If we assume init-data is provided via a CSP's user-data/IMDS facility (doesn't really matter, could be disk or a hardcoded file) then this component could be a script in userland, which itself is covered by measured boot. Extending the respective measurement with a hash of the provided init-data would yield `expected_init_data_hash`.

It's a lot more ceremony than providing an expected hash at guest-launch and having it embedded in a signed report, but I wouldn't know another way. In CAA there are many permutations of TEEs and Infrastructure, but I think this reasoning applies to all of those. (cc @stevenhorsman)

PR #253, opened 2023-12-05
Update SNP Verifier with report and init claims

fitzthum:
Also, move the SNP testing data into the test_data directory.

@mkulke This seems like it's going to have somewhat incorrect implications for the AZ Verifier (since we don't want to use the init data from the report for that). Not sure what the best way to reconcile is. 

@Xynnn007 It seems like we don't specify what format the report_data and init_data should be returned in the claims map. I am assuming we want base64.

I am planning to also add to the parsed claims doc once https://github.com/confidential-containers/kbs/pull/248 is merged.


Xynnn007:
Now SGX and TDX are using hex as the claims. Maybe we should have a standard for all platforms? All using base64 or hex? It would help the consumer without worrying about the encoding of the claims.

Xynnn007:
btw #248 is mergwd

fitzthum:
> Now SGX and TDX are using hex as the claims. Maybe we should have a standard for all platforms? All using base64 or hex? It would help the consumer without worrying about the encoding of the claims.

Yeah I think it has to be standardized or we aren't going to be able to move the checks out of individual verifiers.

mkulke:
If the claims are generic for each TEE, should we add them individually in the TEE-specific parts of a the verification logic? can we just add them after the verification has passed?

https://github.com/confidential-containers/kbs/blob/eebaa977e0a23febb5318994cb0e5ba9daa7f8d5/attestation-service/attestation-service/src/lib.rs#L102-L116

(IMO, it probably would be cleaner to have them added in the `verifier` crate, so there would be `evidence -> claim_set` logic, be but we're already messing with the claim (flattening) in the `attestation-service` snippet above, so that might be ok)

mkulke:
I recommend decoupling test cleanups into a discrete PR. We should really attempt to reduce the number of drive-by fixes. Non-controversial but necessary quality improvements could be merged quickly, and we can focus our review-time on parts which require more scrutiny. untangling unrelated parts from change-sets is a cognitive overhead.

mkulke:
@fitzthum 

> As of today we could add the expected init_data and report_data values to the claims map after the evidence has been verified, but I think the next step is to remove verification of those fields from the verifiers.

For az-{snp,tdx}-vtpm TEEs we would have to make changes to support this, at the moment nonce/report_data verification is coupled with signature validation in an encapsulated library fn, like this:

image

but in principle the verification can be split, I think. What would be the projected timeframe for such a change?

> Instead, the verifier will just return the values as part of the claims map and the AS will make sure they are correct. I think this approach will make the interface a bit cleaner and help decouple verification from evaluation. One downside is that it will expose the report_data which is not necessary and might be confusing.

Is there a reason why report_data should be part of a claims map? It's a property of a concrete RCAR exchange and hence not useful for a consumer of a claims map, is it?

Evidence extraction logic for report_data would still need to be implemented individually for each TEE, even if the verification is moved to upper layers. I would still recommend to have this close to the verifier implementation, maybe we can change the trait signatures, so a given TEE's evidence needs to implement `report_data()`:

```Rust
trait ExtractReportData {
    fn report_data(&self) -> &[u8];
}

type ClaimsMap = (); // dummy

trait Verifier {
    type Evidence: ExtractReportData;
    
    fn evaluate(&self, evidence: &Self::Evidence) -> Result<ClaimsMap>;
}


struct MyTee;
type MyEvidence = [u8;64]; // dummy

impl ExtractReportData for MyEvidence {
    fn report_data(&self) -> &[u8] {
        todo!();
    }
}

impl Verifier for MyTee {
    type Evidence = MyEvidence;
    
    fn evaluate(&self, evidence: &MyEvidence) -> Result<ClaimsMap> {
       todo!(); 
    }
}
```

> I am assuming [init-data] is consumed by azure to provision the vTPM.

I'm not sure I understand. Init-data is a CoCo concept, how would that be related to provisioning of a vTPM?

fitzthum:
> Can you add this change to bottom of the az-snp-vtpm verifier logic, so we unset the respective keys from the json map?

If we drop both of these the AS isn't going to be able to verify that the report_data is correct. I guess we could drop the init_data from the AZ SNP claims. The check in the AS for the init_data can maybe be optional?

mkulke:
> > Can you add this change to bottom of the az-snp-vtpm verifier logic, so we unset the respective keys from the json map?
> 
> If we drop both of these the AS isn't going to be able to verify that the report_data is correct. I guess we could drop the init_data from the AZ SNP claims. The check in the AS for the init_data can maybe be optional?

re report_data:

as long as the signature to `verifier::evaluate()` doesn't change and we still receive `expected_report_data` as argument, we can just pass it to the claim after verification and verify it again in upper layers:

```
async fn evaluate(
      &self,
      evidence: &[u8],
====> expected_report_data: &ReportData,
)
```

If the signature changes, and we don't receive `expected_report_data` any more, then there will be breakage, b/c atm we cannot extract the nonce from the evidence.

re init_data:

at the moment it's still not entirely clear to me how we want to curate init_data in attestation-agent on a peerpod. If we can ignore it for the time being until we know more, that would be fine w/ me.

fitzthum:
> If the signature changes, and we don't receive expected_report_data any more, then there will be breakage, b/c atm we cannot extract the nonce from the evidence.

If there is no way to extract the report data from the evidence, then https://github.com/confidential-containers/kbs/issues/228 might not really be feasible.

mkulke:
> If there is no way to extract the report data from the evidence, then #228 might not really be feasible.

Ah, I just meant it won‘t work automagically ootb, we need to make a crate release + bump a dep in this repo before we can use the respective code to do that. in principle #228 shouldn‘t be blocked, just needs prepwork

fitzthum:
> Ah, I just meant it won‘t work automagically ootb, we need to make a crate release + bump a dep in this repo before we can use the respective code to do that. in principle https://github.com/confidential-containers/kbs/issues/228 shouldn‘t be blocked, just needs prepwork

Ok, in that case let me add the code you suggested and once we are ready we can tweak things to work better for #228 

btw I think in the long term we can just have the AS remove the report_data from the claims map before it does the policy evaluation.

fitzthum:
Ok, I forget what was happening with this PR. I think we ran into some confusion regarding how to handle `init_data` and `report_data`. I have dropped the commit that adds these to the SNP claims map. Now this PR just moves the test files and adds a description of the SNP claims to the doc. PTAL @mkulke @Xynnn007

attestation-service/verifier/src/snp/mod.rs
```diff
@@ -258,6 +258,11 @@ mod tests {
     use super::*;
     use openssl::nid::Nid;
 
+    const VCEK: &[u8; 1360] = include_bytes!("../../test_data/snp/test-vcek.der");
```
Xynnn007, line 261: Might not related to the PR. But if we can have an agreement that the binary values should be all showed as hex or base64? see Line 250 in this file (I cannot add comment to line out of this PR)

attestation-service/docs/parsed_claims.md
```diff
@@ -92,3 +92,27 @@ The claim inherit the fields from the SEV-SNP claim with and additional `tpm` hi
...
+- `policy_abi_major`: Minimum ABI major version allowed for guest
+- `policy_abi_minor`: Minimum ABI minor version allowed for guest
+- `policy_debug_allowed`: Whether SNP debug features are allowed for guest
+- `policy_migrate_ma`: Whether migration agent can be connected to guest
+- `policy_single_socket`: Whether guest can be activated only on one socket
+- `policy_smt_allowed`: Whether guest can run on a system with SMT enabled
+- `reported_tcb_bootloader`: Reported SVN of ASP bootloader
+- `reported_tcb_microcode`: Reported microcode version
+- `reported_tcb_snp`: Reported SVN of SNP Firmware
+- `reported_tcb_tee`: Reported SVN of ASP OS
```
Xynnn007, line 110: all claims above would be add a prefix `snp.` due to the function https://github.com/confidential-containers/trustee/blob/main/attestation-service/attestation-service/src/utils.rs#L39

Xynnn007 (commented):
Thanks for this. The PR brings me to old times.. ; P

attestation-service/verifier/src/snp/mod.rs
fitzthum: Yeah let's take this in a separate issue but I think it is probably a good idea.

Xynnn007: Ok. See https://github.com/confidential-containers/trustee/issues/372. And I think other fields of SNP should also do the same, such as https://github.com/confidential-containers/trustee/blob/main/attestation-service/verifier/src/snp/mod.rs#L240

mkulke (approved):
lgtm

fitzthum:
Rebased to pick up the vlek change. I think the comments are accounted for so I merged before I forget about this PR again.

PR #255, opened 2023-12-07
attestation: verifier: tdx: Rework TdShimPlatformConfigInfo try_from

jodh-intel:
Rework `TdShimPlatformConfigInfo` to use types for its fields to
simplify the logic in it's `try_from()` function.

Added a unit test to check the expected behaviour.


jialez0 (approved):
LGTM.

PR #258, opened 2023-12-11
KBS: Optimize performance and memory usage

### LINKED ISSUE #256: KBS: Abondon the Mutex to promote the concurrency performance
Currently in KBS we are using a global `Mutex` to wrap the `AttestationService` object to ensure the thread safety.

https://github.com/confidential-containers/kbs/blob/main/kbs/src/api/src/attestation/mod.rs#L39-L41

However, the `Mutex` would bring bottle neck when concurrent client requests come together. That means, if N attestation requests come, only at most one can be processed at the same time, which will sacrifice the concurrency performance.

FWIW, we should abondon the global `Mutex` here. In stead, every attestation request to KBS will arise a tokio co-routine to do the communication with backend attestation service.

Xynnn007:
Currently we are using a global Mutex to protect the only one attestation service client from unsafe thread sync & send. This brings performance bottle neck.

This commit will create a temporary client object to handle the current attestation request, which can break the mutex.

Fixes #256

### DIFF
--- a/kbs/src/api/src/lib.rs
+++ b/kbs/src/api/src/lib.rs
@@ -233,10 +233,22 @@ impl ApiServer {
         );
 
         #[cfg(feature = "as")]
-        let attestation_service = web::Data::new(self.attestation_service.clone());
+        let (attestation_service, sessions) = {
+            let attestation_service = web::Data::new(self.attestation_service.clone());
+            let sessions = web::Data::new(SessionMap::new());
+            let sessions_clone = sessions.clone();
 
-        #[cfg(feature = "as")]
-        let sessions = web::Data::new(SessionMap::new());
+            tokio::spawn(async move {
+                loop {
+                    tokio::time::sleep(std::time::Duration::from_secs(60)).await;
+                    sessions_clone
+                        .sessions
+                        .retain_async(|_, v| !v.is_expired())
+                        .await;
+                }
+            });
+            (attestation_service, sessions)
+        };
 
         let http_timeout = self.http_timeout;
 
--- a/kbs/src/api/src/http/attest.rs
+++ b/kbs/src/api/src/http/attest.rs
@@ -57,10 +57,12 @@ pub(crate) async fn attest(
         (session.tee(), session.nonce().to_string())
     };
 
+    let attestation_str = serde_json::to_string(&attestation)
+        .map_err(|e| Error::AttestationFailed(format!("serialize attestation failed : {e:?}")))?;
     let token = attestation_service
-        .verify(tee, &nonce, &serde_json::to_string(&attestation).unwrap())
+        .verify(tee, &nonce, &attestation_str)
         .await
-        .map_err(|e| Error::AttestationFailed(e.to_string()))?;
+        .map_err(|e| Error::AttestationFailed(format!("{e:?}")))?;
 
     let claims_b64 = token
         .split('.')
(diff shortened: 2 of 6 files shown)


kbs/src/api/src/http/config.rs
```diff
@@ -25,9 +25,6 @@ pub(crate) async fn attestation_policy(
     }
 
     attestation_service
-        .0
-        .lock()
-        .await
```
fitzthum, line 30: Any concerns about not locking here given that this will change the state of the AS?

fitzthum (approved):
I think this is good but see one comment

kbs/src/api/src/http/config.rs
Xynnn007: Thanks for pointing out this. IMO, the lock/synchronization mechanism should be token cared of on the server side, like in CoCo-AS we use [a read-write lock](https://github.com/confidential-containers/kbs/blob/1cdd24d/attestation-service/bin/grpc-as/src/server.rs#L227) to handle the requests.

The client side code just call the API and the server can response in some order. That means, if users want to use a new policy to do attestation. In their code should call `set_policy` of KBS and wait for the successful return. Then call attestation API.

Among current three implementations of AS client, `built-in-coco-as` seems to have some trouble for synchronization. This is because the policy management is now based on the filesystem and we do not use a explicit lock/mutex which would make trouble. Let me add a Mutex on the client side as a workaround. In future we can fix this in the `attestation-service` further.

Xynnn007:
We are doing some tests to check if this works.

Xynnn007:
With much help from @Lu-Biao , we found some bottle necks:

**Throughput**
1. KBS uses session to save the context of an RCAR handshake protocol. This session is currently saved globally using a Mutex-modified HashMap, which has become the current concurrency performance bottleneck.
2. KBS uses a single gRPC client globally to connect to gRPC-CoCo-AS, and all requests will be processed serially.
3. A Mutex is used for the AS object during KBS programming. In fact, it is only mutually exclusive when modifying the policy operation.

**Memory usage**
1. Currently, the global Hashmap that manages sessions only records new entries. Old expired session records are actually useless, but they will not be eliminated, resulting in a one-way increase in memory usage.

We did some optimizations

1. Replace the HashMap that manages sessions with a concurrently safe HashMap, abandon Mutex, and improve concurrent throughput.
2. Create a new gRPC client connection pool mechanism. On the one hand, it ensures that attestation requests can be processed concurrently and connections are reused. On the other hand, it prevents too many connections from being created at the same time and exhausting the temporary ports at the bottom of the operating system.
3. Encapsulate concurrency security inside the Attestation object and change Mutex to read-write lock RwLock
4. Add a coroutine (green thread in Tokio) to regularly release expired sessions in the session map

Xynnn007:
The following are the test results before and after optimization

image
Throughput without the patch
image
Throughput with the patch

Throughput increased by ~5x

Memory usage will never increase endlessly.

image
Memory usage without the patch ( ~7min)
image
Memory usage with the patch ( ~1hour)

fitzthum (approved):
This is great. I really like the new SessionStatus enum (even though the variants share some of the same fields) and the macro for its fields. Ofc the results are great too. 

I have one minor comment but generally seems ready.

kbs/src/api/src/lib.rs
```diff
@@ -238,6 +238,19 @@ impl ApiServer {
         #[cfg(feature = "as")]
         let sessions = web::Data::new(SessionMap::new());
 
+        #[cfg(feature = "as")]
+        let sessions_clone = sessions.clone();
+
+        #[cfg(feature = "as")]
```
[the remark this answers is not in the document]
Xynnn007: You are right. Let me fix this.

Issue #223, opened 2023-11-14
Enable e2e tests for real TEEs


mkulke:
We would like to enable e2e tests on the repo for real TEE's. At moment we only use the dummy TEE to test the integration of attester + verifier components in general. We'd like to extend the coverage to the integration with real TEEs.

To do this we need to run the workflows on self-hosted runners or external CI systems. For Azure-based TEE's (`az-{snp,tdx}-vtpm`) we can use GARM, which we use in other kata + coco circumstances already. To make this work we need to provide a token scoped to the repository to CoCo's Garm instance and register a webhook secret on the kbs repo.

Xynnn007:
To achieve this goal, two stages in my mind
1. generate the evidence inside the guest. https://github.com/confidential-containers/guest-components/pull/418 would help to provide such a tool
2. deliver the evidence to CoCo-AS. #264 can be re-used.

mkulke:
> To achieve this goal, two stages in my mind
> 
> 1. generate the evidence inside the guest. [attester: add evidence_getter binary guest-components#418](https://github.com/confidential-containers/guest-components/pull/418) would help to provide such a tool
> 2. deliver the evidence to CoCo-AS. [Added e2e test for CoCo-AS using SNP evidence #264](https://github.com/confidential-containers/kbs/pull/264) can be re-used.

If we use a pre-generated fixture as evidence, wouldn't this just be a unit-test for a validator?

What I had in mind was simply running https://github.com/confidential-containers/kbs/blob/main/.github/workflows/kbs-e2e.yaml on a TEE-capable runner. If the test runs e.g. on a TDX worker it should work ootb (`AA_SAMPLE_ATTESTER_TEST` needs to be unset).

Xynnn007:
> If we use a pre-generated fixture as evidence, wouldn't this just be a unit-test for a validator?

Well, I misunderstood the idea. I thought there are two different e2e test
1. for CoCo-AS. including evidence generating on real TEE, API calling and get the attestation token
2. for KBS. including performing RCAR on real TEE, and get the resource

My last comment is for 1. Anyway, we both think that the action of obtaining evidence needs to be dynamically executed on the real TEE.

mkulke:
Ah, I see. I was looking at #264. I saw fixtures in the `tests/e2e` folder and assumed this was the idea. But if we want to run both AS and KBS e2e tests on real TEEs then this issue is the right one, it's more about the repo configuration and the workers we need.

[...]

PR #264, opened 2023-12-21
Added e2e test for CoCo-AS using SNP evidence

Xynnn007:
Depends on #262 

Use a fixed ~~SGX~~ SNP evidence to test gRPC and RESTful CoCoAS to get token


Xynnn007:
The error https://github.com/confidential-containers/kbs/actions/runs/7284163707/job/19849143732?pr=264 here is because PCCS cannot be connected (0xe019). I am not sure whether there is a accessable PCCS that is compatible with DCAP lib provided by Intel on GH Runner. Or we should use a self-hosted machine to run this test?

Hi @mkulke @jepio Could you give some comments?

mkulke:
There should be a PCCS available in azure, but I doubt that works ootb on github runners. If that's the case, I would suggest a self-hosted runner. There is already an [open issue](https://github.com/confidential-containers/kbs/issues/223) for that. We could use the coco azure subscription to spawn ephemeral self-hosted runners that have access to the PCCS. Maybe we can look into that sometime early next year. We should be able to reuse the existing Garm setup from the infra repo.


https://learn.microsoft.com/en-us/azure/security/fundamentals/trusted-hardware-identity-management#how-do-i-use-intel-qpl-with-trusted-hardware-identity-management

Xynnn007:
@mkulke Ok. I think the original aim of #223 is to generate the evidence in real TEE and do the verification. 
This PR only works for the second part, s.t. verify the evidence as a basic test case.

Let me try to change this PR into another platform test case like SNP which does not rely on the GH Runner env.

After this we can keep on #223

Xynnn007:
Updated the code for easily support real TEE in future.

fitzthum (approved):
LGTM.

In the future maybe we can extend it to evaluate some policy as well. 

Also, if we use a self-hosted runner it shouldn't be too difficult to get some real evidence to use.

PR #265, opened 2023-12-21
Support X.509 Certificate in Attestation Token.

jialez0:
cc @Xynnn007 @Lu-Biao @peterzcst @fitzthum


attestation-service/attestation-service/src/token/simple.rs
```diff
@@ -56,8 +68,16 @@ impl AttestationTokenBroker for SimpleAttestationTokenBroker {
...
 
+        let id: String = thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(10)
+            .map(char::from)
+            .collect();
+
         let mut claims = json!({
-            "iss": ISSUER_NAME,
+            "iss": self.config.clone().issuer_name.unwrap_or(ISSUER_NAME.to_string()),
```
Xynnn007, line 78: Maybe we can set `ISSUER_NAME` to `config.issuer_name` in `default()`

attestation-service/attestation-service/src/token/simple.rs
```diff
@@ -87,20 +107,58 @@ impl AttestationTokenBroker for SimpleAttestationTokenBroker {
...
+            e: URL_SAFE_NO_PAD.encode(e),
+            x5u: None,
+            x5c: vec![],
+        };
+
+        if let Some(signer) = &self.config.signer {
+            if let Some(cert_url) = signer.cert_url.clone() {
+                jwk.x5u = Some(cert_url);
+            }
+            if let Some(cert_path) = signer.cert_path.clone() {
```
Xynnn007, line 126: Maybe we can store the cert string inside the `SimpleAttestationTokenBroker` object to avoid read the filesystem everytime needs to export pubkey

kbs/src/api/src/token/coco.rs
```diff
@@ -41,12 +68,18 @@ impl AttestationTokenVerifier for CoCoAttestationTokenVerifier {
 
         let jwk_value = claims_value["jwk"].as_object().ok_or_else(|| anyhow!("CoCo Attestation Token Claims must contain public key (JWK format) to verify signature"))?;
         let jwk = serde_json::to_string(&jwk_value)?;
+        let rsa_jwk = serde_json::from_str::<RsaJWK>(&jwk)?;
```
Xynnn007, line 71: Do we have a unit test that can read a RsaJWK that exported from CoCoAS? The definitions are a little bit different https://github.com/confidential-containers/kbs/pull/265/files#diff-0c923b04d229848729ac1682c38c345239a277ef23fc743c92d3f7e1ba9e675fR113-R120 https://github.com/confidential-containers/kbs/pull/265/files#diff-cdb725680d5fff6ed34b3eea443b0ea1db1135e30aecd9e1eb20b1ec80aee2a7L64-L69

attestation-service/attestation-service/src/token/simple.rs
```diff
@@ -39,10 +47,12 @@ impl SimpleAttestationTokenBroker {
 
 impl SimpleAttestationTokenBroker {
     fn rs384_sign(&self, payload: &[u8]) -> Result<Vec<u8>> {
-        let mut rng = rand::thread_rng();
-        let signing_key = SigningKey::<Sha384>::new(self.private_key.clone());
-        let signature = signing_key.sign_with_rng(&mut rng, payload);
-        Ok(signature.to_bytes().to_vec())
+        let rsa_pkey = PKey::from_rsa(self.private_key.clone())?;
```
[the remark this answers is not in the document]
jialez0: This is because when AS generates the JWK in Token, it needs to get the components (n & e) of RSA key. If we store `Pkey` in Structure, every time we try to get n & e components, we will need transform it to RSA Key.

kbs/src/api/src/token/coco.rs
jialez0: Good suggestion, I'll add this unit test.

kbs/src/api/src/token/coco.rs
```diff
@@ -55,6 +90,74 @@ impl AttestationTokenVerifier for CoCoAttestationTokenVerifier {
...
+                            }
+                        }
+                    } else {
+                        bail!("Missing certificate in Attestation Token JWK");
+                    }
+                }
+            };
+
+            // Check certificate is valid and trustworthy
+            let mut untrusted_stack = Stack::<X509>::new()?;
```
fitzthum, line 139: Can't we push each cert onto this stack as part of the previous step? (rather than making a Vec<X509> and then iterating through it and cloning here)

kbs/docs/config.md
```diff
@@ -32,6 +31,19 @@ section:
...
+The following properties can be set under the `attestation_token_config` section.
+
+>This section is available only when the `resource` feature is enabled.
+
+| Property                   | Type          | Description                                         | Required | Default   |
+|----------------------------|---------------|-----------------------------------------------------|----------|-----------|
+| `attestation_token_config` | String        | Attestation token broker type. Valid values: `CoCo` | Yes      | -         |
+| `trust_certs_paths`        | String Array  | Trusted root certificates file paths (PEM format).  | No       | -         |
+
+If `trust_certs_paths` is set, KBS will forcibly check the validity of the Attestation Token signature public key certificate.
```
fitzthum, line 45: What if it isn't set? It seems like that will cause all requests to a remote AS to fail?

kbs/src/api/src/token/coco.rs
```diff
@@ -55,6 +90,74 @@ impl AttestationTokenVerifier for CoCoAttestationTokenVerifier {
...
+                                    let trimmed =
+                                        format!("{}\n-----END CERTIFICATE-----", pem.trim());
+                                    if !trimmed.starts_with("-----BEGIN CERTIFICATE-----") {
+                                        continue;
+                                    }
+                                    let cert = X509::from_pem(trimmed.as_bytes())
+                                        .map_err(|_| anyhow!("Invalid PEM certificate chain"))?;
+                                    chain.push(cert);
+                                }
+                                chain
```
fitzthum, line 123: Might be nice to have a unit test for this that tries to download some reference certs or read them from a file.

fitzthum (commented):
Nice job. I think this is a good feature. I made a few notes.

A couple of other things:

Might be good to update the quickstart guide or some of the dockerfiles to make this change easier on users, especially since I think things might break if a user does not supply some trusted certs to the KBS. It would be good to clarify the default behavior.

I guess this does not support updating the certs at runtime. That is probably ok.

kbs/docs/config.md
jialez0: If not set this field, KBS will skip the verification of the certificate. I have add this explain to `config.md`

jialez0:
@fitzthum Thanks for reviewing, now this PR is updated.

kbs/src/api/src/token/coco.rs
jialez0: I have put the generation of `Vec<X509>` cert chain of `x5u` in a help function `download_cert_chain` now, so here iterating it and clone them (skip first cert) to Stack is OK for now.

kbs/quickstart.md
```diff
@@ -179,4 +179,39 @@ Where `/path/to/policy` should be replaced by the real path to your policy file.
 
 Resource policy also needs to be the `rego` syntax defined by [Open Policy Agent](https://www.openpolicyagent.org/).
 
-You can read the notes of [default resource policy file](./src/api/src/policy_engine/opa/default_policy.rego) for more details of resource policy.
\ No newline at end of file
+You can read the notes of [default resource policy file](./src/api/src/policy_engine/opa/default_policy.rego) for more details of resource policy.
+
+## Attestation Token Certificate
+
+You can configure certificate of the signing key of Attestation Token (JWT) by config files.
```
fitzthum, line 186: Can you mention that this is not required if you are using the Builtin AS (assuming that is true).

fitzthum (commented):
One more comment. Also, are we doing anything to verify requests made from the KBS to the AS? In particular when the AS sends a set policy request to the AS, the AS should probably check if the request is genuine. Do we support this?

kbs/quickstart.md
jialez0: I'll add some explain about Builtin AS, thanks.

jialez0:
> One more comment. Also, are we doing anything to verify requests made from the KBS to the AS? In particular when the AS sends a set policy request to the AS, the AS should probably check if the request is genuine. Do we support this?

@fitzthum When we use KBS with Builtin-AS or gRPC-AS, we have assumed that AS actually runs in the same trusted security domain as KBS and is controlled by the same entity (such as tenants). So under this premise, the interfaces provided by AS are all available for direct invocation without verification now.

We are currently building a scenario which using KBS with RESTful-AS (although we do not fully support it yet). In this scenario, AS and KBS may run on different nodes and be controlled by different entities, which AS may require some additional validation of request from KBS. This is the enhancement point we need to consider next.

fitzthum (approved):
I think this is fine to merge, but lack of validation of requests (particularly SetPolicy) to the AS seems like potentially a big problem.

attestation-service/attestation-service/src/token/simple.rs
```diff
@@ -5,42 +5,82 @@
...
     config: AttestationTokenConfig,
+    cert_url: Option<String>,
+    cert_chain: Option<Vec<X509>>,
 }
 
 impl SimpleAttestationTokenBroker {
     pub fn new(config: AttestationTokenConfig) -> Result<Self> {
-        let mut rng = rand::thread_rng();
-        let private_key = RsaPrivateKey::new(&mut rng, RSA_KEY_BITS)?;
+        let (private_key, cert_url, cert_chain) = match config.signer.clone() {
```
Xynnn007, line 34:
```suggestion
        let (private_key, cert_url, cert_chain) = match &config.signer {
```

We might not need this `clone()` but only a reference

attestation-service/attestation-service/src/token/simple.rs
```diff
@@ -5,42 +5,82 @@
...
+                        for pem in pem_cert_chain.split("-----END CERTIFICATE-----") {
+                            let trimmed = format!("{}\n-----END CERTIFICATE-----", pem.trim());
+                            if !trimmed.starts_with("-----BEGIN CERTIFICATE-----") {
+                                continue;
+                            }
+                            let cert = X509::from_pem(trimmed.as_bytes())
+                                .map_err(|_| anyhow!("Invalid PEM certificate chain"))?;
+                            chain.push(cert);
+                        }
+                        Some(chain)
```
Xynnn007, line 53: could you try to use `and_then` to prevent the pattern matching of `None => None`?

attestation-service/attestation-service/src/token/simple.rs
```diff
@@ -5,42 +5,82 @@
...
+                    None => None,
+                };
+
+                (
+                    Rsa::private_key_from_pem(&pem_data)?,
+                    signer.cert_url.clone(),
+                    cert_chain,
+                )
+            }
+            None => (Rsa::generate(RSA_KEY_BITS)?, None, None),
```
Xynnn007, line 64: Do we need to add `info!` here to tell the user that there is no signer configuration given, thus we create an ephemeral key and without CA pubkey cert?

attestation-service/attestation-service/src/token/simple.rs
```diff
@@ -56,8 +96,16 @@ impl AttestationTokenBroker for SimpleAttestationTokenBroker {
...
 
+        let id: String = thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(10)
+            .map(char::from)
+            .collect();
+
         let mut claims = json!({
-            "iss": ISSUER_NAME,
+            "iss": self.config.clone().issuer_name,
```
Xynnn007, line 106:
```suggestion
            "iss": self.config.issuer_name.clone(),
```

Xynnn007 (approved):
Thanks. Only some nits.

attestation-service/attestation-service/src/token/simple.rs
jialez0: Here use `match` may be better, because we need return Error description when read cert failed or verify cert failed, if we use `and_then`, report these error detail will be difficult.

PR #272, opened 2024-01-02
ci: build grpc kbs every merge to main

Xynnn007:
Now we only build kbs-built-in-coco-as every merging to main. We also need kbs-grpc-coco-as. In this way we can test whether the current KBS cluster (kbs-grpc-coco-as, coco-as, rvps and coco-keyprovider) works well together.

cc @mythi


mythi (approved):
thanks!

jialez0 (approved):

PR #274, opened 2024-01-03
AS/Verifier: fix the report/init data comparation

Xynnn007:
We used to compare the runtime/init data without caring about the length of each other. This commit resize the input expected data by trim or append '\0' to the equal length with the actual value from the one in evidence.

In this way we can avoid unexpected failures.

The error was firstly caught in enclave-cc's CI https://github.com/confidential-containers/enclave-cc/actions/runs/7299054486/job/20112097547?pr=312

### DIFF
--- a/attestation-service/verifier/src/az_snp_vtpm/mod.rs
+++ b/attestation-service/verifier/src/az_snp_vtpm/mod.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use crate::{InitDataHash, ReportData};
+use crate::{regularize_data, InitDataHash, ReportData};
 
 use super::{TeeEvidenceParsedClaim, Verifier};
 use crate::snp::{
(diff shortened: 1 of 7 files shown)


attestation-service/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -61,6 +61,19 @@ impl Verifier for AzSnpVtpm {
             bail!("unexpected empty report data");
         };
 
+        let mut expected_report_data = expected_report_data.to_vec();
+        match expected_report_data.len() {
```
jialez0, line 65: This `match` structure appear many times, maybe put it in a helper function like `regularize_report_data(expected_len: usize)` is better?

Xynnn007: nice idea.  Let me do this

jialez0 (approved):
LGTM.

PR #275, opened 2024-01-03
KBS/perf: promote the concurrency performance of KBS

### LINKED ISSUE #256: KBS: Abondon the Mutex to promote the concurrency performance

Lu-Biao:
Currently we are using a global Mutex to protect the only one attestation service client from unsafe thread sync & send. This brings performance bottle neck.

This commit brings some optimization to promote the performance and stability.

1. Abondon the global Mutex of the attestation service and change the API definition of Attest to non-mut. This would let the developers to handle the concurrency safe inside the concrete attestation-service inside. In this way, we prevent to lock the whole process logic.
2. Bring in a gRPC client pool to grpc-coco-as mode. This will help to avoid errors that caused by runing out all the temporaty ports provided by OS.
3. Replace the Mutex of session map with a concurrency-safe HashMap to avoid bottle neck.

Fixes #256

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -417,6 +417,7 @@ dependencies = [
  "kbs-types",
  "lazy_static",
  "log",
+ "mobc",
  "openssl",
  "prost",
  "rand",
@@ -425,6 +426,7 @@ dependencies = [
  "rstest",
  "rustls 0.20.9",
  "rustls-pemfile",
+ "scc",
  "semver 1.0.20",
  "serde",
  "serde_json",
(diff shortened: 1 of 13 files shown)


Xynnn007 (approved):
LGTM. Let me put another part upon this

PR #277, opened 2024-01-04
Fix RVPS binary building & push image every merge to main

Xynnn007:
This PR fixes some bugs for RVPS
1. Avoid duplicated launch parameter flag to prevent the process from panic
2. Fix the Dockerfile LABEL and error information for RVPS.
3. Publish the RVPS image every merge to main so that we can test the KBS cluster in time.

### DIFF
--- a/attestation-service/rvps/src/bin/rvps.rs
+++ b/attestation-service/rvps/src/bin/rvps.rs
@@ -19,7 +19,7 @@ const DEFAULT_STORAGE: &str = "LocalFs";
 #[command(author, version, about, long_about = None)]
 pub struct Cli {
     /// Underlying storage engine that RVPS uses.
-    #[arg(short, long, default_value = DEFAULT_STORAGE)]
+    #[arg(short = 'c', long, default_value = DEFAULT_STORAGE)]
     pub storage: String,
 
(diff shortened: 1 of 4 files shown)


attestation-service/rvps/src/bin/rvps.rs
```diff
@@ -19,7 +19,7 @@ const DEFAULT_STORAGE: &str = "LocalFs";
 #[command(author, version, about, long_about = None)]
 pub struct Cli {
     /// Underlying storage engine that RVPS uses.
-    #[arg(short, long, default_value = DEFAULT_STORAGE)]
+    #[arg(short = 'c', long, default_value = DEFAULT_STORAGE)]
```
jialez0, line 22: Why use `-c` for short instead of `-s`?

jialez0 (approved):
LGTM. Just a little question.

attestation-service/rvps/src/bin/rvps.rs
Xynnn007: By default both `storage` and `socket` will use the same short flag `-s` atomatically, which will make the rvps process panic when launched. This is why a different short flag is defined here.

PR #283, opened 2024-01-10
Add end-to-end test with docker compose and sample attester

fitzthum:
In order to integrate the KBS/AS into the Kata CI, we need an easy way for people to deploy it. I think the docker compose approach might be best, but it had some problems and wasn't tested. This should fix that.

~~The test itself is not yet done. PR is still WIP.~~


kbs/src/api/src/http/resource.rs
```diff
@@ -103,13 +103,17 @@ pub(crate) async fn get_resource(
             resource_description.resource_type,
             resource_description.resource_tag
         );
-        policy_engine
+        let (policy_valid, _extra_policy_output) = policy_engine
```
Xynnn007, line 106: Maybe `allow` could be a better name? As it is actually the evaluation result.

Surely, we need to emphasize in a document that the policy must have a value named `allow`. Or the evaluation will always return false.

attestation-service/attestation-service/src/lib.rs
```diff
@@ -204,6 +205,7 @@ impl AttestationService {
             .collect();
 
         let token_claims = json!({
+            "tee": to_variant_name(&tee)?,
```
Xynnn007, line 208: nit: https://github.com/confidential-containers/kbs/blob/main/attestation-service/docs/example.token.json should also be updated.

Xynnn007 (commented):
Thanks for the nice feature. Some nits at first glance. Hope that will not break your work

Cargo.lock
```diff
@@ -2921,18 +2943,18 @@ checksum = "ff011a302c396a5197692431fc1948019154afc178baf7d8e37367442a4601cf"
 
 [[package]]
 name = "openssl-src"
-version = "300.1.6+3.1.4"
```
Xynnn007, line 2924: The build error https://github.com/confidential-containers/kbs/actions/runs/7483975835/job/20370068894?pr=283#step:7:849 is caused by chaing the `version` and `checksum` here. Although `Cargo` will automatically update the rev here, but we'd better keep it as-is to make the csv building pass.

At the meantime, we are getting touch with csv colleagues to fix the upstream.

fitzthum: Thanks for the tip!

docker-compose.yml
```diff
@@ -0,0 +1,73 @@
+version: '3.2'
```
Xynnn007, line 1: Once we move the docker-compose to the top directory, we might need add some contents to the top [README](https://github.com/confidential-containers/kbs/blob/main/README.md) to help the users who is not familiar with the project to use this? like a link to https://github.com/confidential-containers/kbs/blob/main/kbs/docs/cluster.md

wdyt?

Xynnn007 (approved):
Thanks for the great e2e test.

docker-compose.yml
fitzthum: I have a number of changes to the docs planned, but I was thinking I would do that in a follow-up PR, especially so we can get the fix in this PR merged quickly. Does that work for you?

Xynnn007: It works

fitzthum: See https://github.com/confidential-containers/kbs/pull/287

PR #284, opened 2024-01-10
e2e-test: enable real TEE on self-hosted runners

mkulke:
## Why

We want to be able to test PRs in a real TEE on self-hosted runners. By
default the tests are executed in the fork from which the PR is crated.
The fork does not have the self-hosted runners which provide the TEE. We
need to run them in the scope of the target repository in a safe fashion
that prevents untrusted PRs from running arbitrary code on self-hosted
runners.

fixes: #223

## How

This change is modelled after simular workflows in kata and CAA repos.

The matrix-based kbs-e2e workflow has been turned into a reusable
workflow that is being called by specific TEE workflows. In this PR that
would be the "sample" and "az-snp-vtpm" TEE.

For a real TEE we follow the following ceremony:
1) verify whether the PR is authorized to run (a maintainer needs to set
   a label on the PR).
2) checkout the main branch from the target
3) rebase the changes from the source atop the main branch (using a
   script copied verbatim from CAA)
4) archive the resulting HEAD and upload it as artifact
5) call the re-usable workflow with the required runners param
6) extract the artifact
7) run the test

For the sample TEE steps 1-3 are replaced by a simple checkout

## Caveats

This yaml spaghetti is very hard to test. So while this passes actionlint,
it might need several iterations to always work.

The reusable workflow is targetting main and we cannot point it to a
fork, so no workflow run will successfully complete in this PR.

### DIFF
--- a/kbs/hack/ci-helper.sh
+++ b/kbs/hack/ci-helper.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+TARGET_BRANCH=${TARGET_BRANCH:-main}
+
+function rebase_atop_of_the_latest_target_branch() {
+        if [ -n "${TARGET_BRANCH}" ]; then
+                echo "Rebasing atop of the latest ${TARGET_BRANCH}"
+                # Recover from any previous rebase left halfway
+                git rebase --abort 2> /dev/null || true
+                if ! git rebase "origin/${TARGET_BRANCH}"; then
+                        # if GITHUB_WORKSPACE is defined and an architecture is not equal to x86_64
+                        # (mostly self-hosted runners), then remove the repository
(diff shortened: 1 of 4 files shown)


mkulke:
ok we can't merge this yet, since the e2e test is scheduled on the PR's fork (`mkulke/kbs`) which is not what we want. we want it to run on a self-hosted runner in the scope of this repo, but only
- after someone added an test-e2e label (@wainersm can you provide pointers on which yaml snippet to copypaste for that? afair think we need some additional rebase step)
- and on pushes to main (i.e. after merge)

see this: https://github.com/confidential-containers/kbs/issues/223#issuecomment-1867436941

let's do those changes to the workflow in the scope of this PR. sample e2e can still run anywhere

kbs/hack/ci-helper.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2023 Intel Corporation
```
Xynnn007, line 3: Is this `Intel` mark intended?

Xynnn007 (commented):
Thanks for the PR. We are excited to see the CI on real TEE.

kbs/hack/ci-helper.sh
mkulke: it's copied from CAA, which copied it from kata afaik, so I left the header in

PR #291, opened 2024-01-15
e2e: reference kbs-e2e.yaml worfklows locally

mkulke:
Context: https://github.com/confidential-containers/kbs/actions/runs/7523718849

The attempt to find the workflow failed because there is a typo (yml => yaml). But we should also reference a workflow in the same repository locally as described here:

https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow

Drive-by fix: artifact-download@v4 puts the artifacts in an `./artifact` subfolder.


Xynnn007 (approved):
Hey Magnus, thanks for the very quick action!

PR #295, opened 2024-01-19
az-snp-vtpm-verifier: remove report_data padding

mkulke:
A TPM quote's qualifyingData (nonce) is not padded to 64 bytes, it's of dynamic size and will be the same size as the passed-in nonce when requesting a Quote. Hence verification will fail using the padded report_data.

Also split up Quote verification into signature and nonce verification, so we can later move verification to upper layers (w/ report_data being part of the claim)


Xynnn007 (approved):

fitzthum (approved):
LGTM

PR #307, opened 2024-01-30
KBS /resource: support CoCo-AS token without public key in it

jialez0:


kbs/src/api/src/lib.rs
```diff
@@ -343,3 +347,8 @@ impl ApiServer {
         }
     }
 }
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct HttpsEnabledFlag {
+    enabled: bool,
+}
```
mkulke, line 354:
```suggestion
pub(crate) struct HttpsFlag {
    enabled: bool,
}
```

kbs/src/api/src/http/resource.rs
```diff
@@ -123,13 +99,24 @@ pub(crate) async fn get_resource(
...
-    let jwe = jwe(pubkey, resource_byte)?;
-
-    let res = serde_json::to_string(&jwe).map_err(|e| Error::JWEFailed(e.to_string()))?;
-
-    Ok(HttpResponse::Ok()
-        .content_type("application/json")
-        .body(res))
+    // If public key is included in claims, use JWE to wrap resource byte,
+    // else check if HTTPS is enabled.
+    match get_pkey(claims) {
```
mkulke, line 104: If I understand correctly we want to support requests without public key for KBS deployments without HTTPs. In this case there would never be a be pkey, right? Does it still make sense to attempt to fetch it or should we rather start the with the HttpsEnabled == false check and return the unwrapped resource? (Instead of swallowing the get_pkey() error).

jialez0: As you mentioned, this PR hopes to support requests without pubkey in KBS with HTTPS enabled. However, "HTTPS enabled" is a minimum security measure. Even if HTTPS is enabled, pubkeys may still exist. In this case, the returned resource data is protected by both HTTPS and JWE. Only when the pubkey cannot be found, data will be returned when HTTPS is enabled. If HTTPS is not enabled, it is completely unsafe, and KBS will refuse to return data.

mkulke: I see. thanks for clarifying.  So, it looks like having no pubkey in the claims is a legitimate code path, i.e. the signature is actually something like: `fn get_pkey() => Option<Result<TeePubKey>>`. A result should capture unexpected states. An alternative would be to match on a concrete error type:

```rust
Err(err) => {
   if let err = Error:NoPubkeyInClaims && https.enabled {
   ....
   }
   Err(Error::InsecureResponse)
}
```

Failing to parse (` TeePubKey::deserialize(pkey_value)`) should probably result in an error and not in a fallback to https, because this shouldn't happen in any case.

If `customized_claims` is specified as always having `runtime_data.tee-pubkey` we probably want to error out, too.

kbs/src/api/src/http/resource.rs
```diff
@@ -243,3 +230,32 @@ pub(crate) fn jwe(tee_pub_key: TeePubKey, payload_data: Vec<u8>) -> Result<Respo
         tag: "".to_string(),
     })
 }
+
+pub(crate) fn get_pkey(claims: Value) -> Result<TeePubKey> {
+    let pkey_value = claims
+        .get("customized_claims")
```
mkulke, line 236: Can we define/reuse a type `CustomizedClaim` instead of manually performing the parsing by stepping through properties?

kbs/src/api/src/http/resource.rs
```diff
@@ -123,13 +99,24 @@ pub(crate) async fn get_resource(
...
+            Ok(HttpResponse::Ok()
+                .content_type("application/json")
+                .body(res))
+        }
+        Err(_) => {
+            if !https_enabled.enabled {
+                return Err(Error::InsecureResponse);
+            }
+            let res = URL_SAFE_NO_PAD.encode(resource_byte);
+            Ok(HttpResponse::Ok().body(res))
```
fitzthum, line 117: I am confused by this PR. Isn't it fundamental that the secure channel is bound to the hw evidence? In other words, doesn't the KBS want to know that when it encrypts a value using a public key, it can only be decrypted inside the enclave that has the corresponding private key? I'm not sure it's valid to support secret delivery without this. What use case are you looking at?

jialez0: I am trying to get CDH to support calling AA to obtain a token from AS, and then using this token to access KBS to obtain resources.

jialez0: Here is the [PR](https://github.com/confidential-containers/guest-components/pull/465) for AA to add runtime data parameter. If we accept that, maybe this PR can be closed.

PR #308, opened 2024-01-31
ci: install libssl-dev for e2e on self-hosted runners

mkulke:
blocked-by: #309 

The self-hosted runners do not ship with openssl headers, so we need to install them explicitly.


fitzthum (approved):
Ok

Xynnn007 (approved):
lgtm

PR #309, opened 2024-01-31
ci: fetch the head of a PR in kbs TEE runs

mkulke:
There is a flaw in the checkout logic of the PR, which will always default to `main`, but it should pick the head of the PRs repo.

The e2e tests were not authorized to run on pushes to main, which has been fixed, too.


fitzthum (approved):
I got a bit confused by the JS syntax but this seems good.

PR #317, opened 2024-02-07
docs: Improve RVPS document

GabyCT:
This PR tries to improve the RVPS document, first by putting a pre-requisite to avoid the failure when building the RVPS and by correcting a command that will fail otherwise when installing the RVPS.


fitzthum (approved):
LGTM

PR #319, opened 2024-02-09
Release: Update KBS for v0.8.2 release

portersrc:
Preparing KBS for v0.8.2 following discussion from https://github.com/confidential-containers/kbs/issues/312


fitzthum (approved):
This implies two changes to the release process.

First, we're getting rid of the release workflows. Note that these workflows did not run any additional tests, they just built the images and pushed them. Instead, we can just tag the latest staged images as the release images (assuming we're happy with them). In the future we should add a script to do this. The advantage here is that it eliminates the possibility of the release workflows breaking. This has happened a couple of times in the past and it really slows things down. If the release workflows break, we have to cut another release to fix them.

Second, the docker-compose file is going to remain pointing at the head. This deployment will build all the containers from source. On the other hand the Kubernetes config will always pointed to the images corresponding to the latest release. This is a compromise between supporting developers who want to build from the latest source and consumers who want something stable.

These changes were discussed on https://github.com/confidential-containers/kbs/issues/312 and on Slack, but this release falls in a window where some maintainers are on vacation. These changes are not set in stone. If this doesn't turn out to be an improvement on the existing process we can revert for the next release.

PR #323, opened 2024-02-12
ci: add az-tdx-vtpm workflow for e2e tests

mkulke:
This adds a target for the az-tdx-vtpm TEE.

~TDX verifiers need an SGX quoting environment, for this we need to startup a local certificate caching service (PCCS) and point the quoting config to it.~

~The PCCS itself will need an API key to be able to query certificates from Intel's servers. If the apikey secret is set as a param to the callable workflow, it will install PCCS locally.~

We can use intel's collaterals service directly in the dcap configuration, however we need to add the local cert cache, so the endpoint is not hit by too many requests from the CI.

The workflow itself will not be executed in this PR (only after it has been merged) to test whether it works I ran the following on an Azure TDX VM Ubuntu 22.04 vm.

```bash
sudo apt-get update
sudo apt-get install make -y
git clone https://github.com/mkulke/kbs.git -b mkulke/add-az-tdx-vtpm-e2e-workflow
cd kbs/kbs/test/
sudo make install-dependencies
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
source ~/.cargo/env
wget https://go.dev/dl/go1.22.0.linux-amd64.tar.gz
sudo rm -rf /usr/local/go && sudo tar -C /usr/local -xzf go1.22.0.linux-amd64.tar.gz
rm go1.22.0.linux-amd64.tar.gz
export PATH=$PATH:/usr/local/go/bin
make bins
make generate-attestation-token-signer
sudo make e2e-test
```

### DIFF
--- a/.github/workflows/kbs-e2e-az-tdx-vtpm.yaml
+++ b/.github/workflows/kbs-e2e-az-tdx-vtpm.yaml
@@ -1,4 +1,4 @@
-name: KBS e2e with az-snp-vtpm TEE
+name: KBS e2e with az-tdx-vtpm TEE
 
 on:
   push:
@@ -60,5 +60,5 @@ jobs:
     - checkout-and-rebase
     uses: ./.github/workflows/kbs-e2e.yaml
     with:
-      runs-on: '["self-hosted","azure-cvm"]'
+      runs-on: '["self-hosted","azure-cvm-tdx"]'
       tarball: kbs.tar.gz
--- a/.github/workflows/kbs-e2e.yaml
+++ b/.github/workflows/kbs-e2e.yaml
@@ -49,6 +49,12 @@ jobs:
           target/
         key: rust-${{ hashFiles('./Cargo.lock') }}
 
+    - name: Set up SGX/TDX certificates cache
+      uses: actions/cache@v4
+      with:
+        path: /root/.dcap-qcnl
+        key: ${{ runner.os }}-dcap-qcnl
+
     - name: Install dependencies
       working-directory: kbs/test
       run: |
--- a/kbs/test/Makefile
+++ b/kbs/test/Makefile
@@ -1,6 +1,8 @@
 OS := $(shell lsb_release -si)
 RELEASE := $(shell lsb_release -sr)
 SGX_REPO_URL := https://download.01.org/intel-sgx/sgx_repo/ubuntu
+SGX_COLLATERAL_URL := https://api.trustedservices.intel.com/sgx/certification/v4/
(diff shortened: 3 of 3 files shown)


.github/workflows/kbs-e2e.yaml
```diff
@@ -49,6 +49,12 @@ jobs:
           target/
         key: rust-${{ hashFiles('./Cargo.lock') }}
 
+    - name: Set up SGX/TDX certificates cache
+      uses: actions/cache@v4
+      with:
+        path: /root/.dcap-qcnl
```
mythi, line 55: I don't know if this breaks but I was wondering if this is more future proof?
```suggestion
        path: $HOME/.dcap-qcnl
```

mkulke: do you think the path statement in that github actions works with envs?

mythi: I'm not sure. I'm fine with `/root` as well

mythi (approved):
LGTM

fitzthum (approved):
Nice

.github/workflows/kbs-e2e-az-tdx-vtpm.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - 'main'
+
+jobs:
+  authorize:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'test_e2e')
+    steps:
+    - run: "true"
+
+  checkout-and-rebase:
```
wainersm, line 32: Hi @mkulke, always good to see code well-documented!

Perhaps you want to run this checkout-and-rebase job if `authorize` , otherwise it will execute on pull requests without the `test_e2e` label (i.e. waste of resources).

.github/workflows/kbs-e2e.yaml
wainersm: @mkulke sorry, for my ignorance, who/how /root/.dcap-qcnl was generated? Is it  manually set on the `self-hosted` runner? All in all, I didn't understand why it is cached.

mythi: @wainersm it's a cache for the SGX/TDX TCB data written by `libsgx-dcap-quote-verify`.The reasons for why are explained in #329

mkulke: ah sorry, the PR description was not updated.

PR #324, opened 2024-02-12
kbs: improvements to quickstart and misc

wainersm:
I'm trying to catch up with latest changes of KBS project by following the [quickstart](https://github.com/confidential-containers/kbs/blob/main/kbs/quickstart.md). From within a fresh Ubuntu 22.04 VM (non TEE), I could run the  instructions for the KBS on background check mode but not before fixing some instructions. I took that opportunity to have some improvements to the doc.

### DIFF
--- a/kbs/quickstart.md
+++ b/kbs/quickstart.md
@@ -85,7 +85,7 @@ EOF
 Use `kbs-client` to upload resource data to KBS storage:
 
 ```shell
-kbs-client --url http://127.0.0.1:50000 config --auth-private-key config/private.key set-resource --resource-file test/dummy_data --path default/test/dummy
+kbs-client --url http://127.0.0.1:8080 config --auth-private-key config/private.key set-resource --resource-file test/dummy_data --path default/test/dummy
 ```
 
 Here we assigned a custom path for this resource: `default/test/dummy`.
@@ -94,7 +94,7 @@ Here we assigned a custom path for this resource: `default/test/dummy`.
 
 Run following command to get resource data from KBS:
 ```shell
-kbs-client --url http://127.0.0.1:50000 get-resource --path default/test/dummy
+kbs-client --url http://127.0.0.1:8080 get-resource --path default/test/dummy
 ```
 
 If you run the client outside of a TEE, the sample attester will be used.
@@ -103,7 +103,7 @@ To test the KBS with sample evidence, you'll need to update the resource policy
 to something more permissive.
 This can be done with a command such as
 ```shell
-./kbs-client --url http://127.0.0.1:50000 config --auth-private-key config/private.key  set-resource-policy --policy-file allow_all.rego
+./kbs-client --url http://127.0.0.1:8080 config --auth-private-key config/private.key  set-resource-policy --policy-file allow_all.rego
 ```
 
 ## Passport Mode
--- a/kbs/quickstart.md
+++ b/kbs/quickstart.md
@@ -103,7 +103,7 @@ To test the KBS with sample evidence, you'll need to update the resource policy
 to something more permissive.
 This can be done with a command such as
 ```shell
-./kbs-client --url http://127.0.0.1:8080 config --auth-private-key config/private.key  set-resource-policy --policy-file allow_all.rego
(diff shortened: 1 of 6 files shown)


kbs/tools/client/README.md
```diff
@@ -13,7 +13,29 @@ For more sophisticated attestation clients, please refer to [guest components](h
...
+
+Build the client binary with support to the default features as:
+
+```shell
+make -C ../../ cli
+```
+
+As an alternative, you can build it with the `sample_only` feature which allow
+to mock attestation, thus, you will be able to perform some operations (e.g.
+get resources) even by running outside of an TEE:
```
fitzthum, line 29: To be clear, the sample attester is always enabled and you can always use the `kbs-client` outside of a TEE. This flag means that no other attesters are available so you cannot do a real TEE attestation, but you don't have to worry about the dependencies that some of the hw attesters have.

Btw if we build the kbs-client in the CI we probably want to use this feature (or add a new one that disables the resource endpoint completely).

kbs/quickstart.md
```diff
@@ -43,12 +54,25 @@ and pass it to the KBS server through startup parameters.
...
+```
+
+Build and install the kbs-client:
+```shell
+make cli
+sudo make install-cli
+```
+
+If you want to run the kbs-client outside of TEE then you should compile it
+with the `sample_only` feature to mock the attestation:
```
fitzthum, line 67: See below.

kbs/sample_policies/README.md
```diff
@@ -0,0 +1,6 @@
+This directory contain sample policy files to configure the policy engine
+of the KBS. You can use those files to write your own policies.
```
fitzthum, line 2: nit: `contain` -> `contains`, `those` -> `these`.

fitzthum (commented):
Nice. I made a couple minor comments.

kbs/tools/client/README.md
Xynnn007: If I understand correctly, we do not need `CLI_FEATURES` flag. And we can always build the client-tool with feature `default`. This will support both TEE and non-TEE cases. Right?

fitzthum: The default features support all the use cases on all the platforms. For the tests in the CoCo CI, we only use the client for configuration. So we will use the `sample_only` feature there so that we don't have to install the SGX dependencies.

So yes, no need for any additional feature, although maybe at some point we could introduce features to enable/disable the config and resource endpoints separately. Not a high priority.

wainersm: hi @Xynnn007 @fitzthum , sorry, I misunderstood it completely. I built the `default` feature and run kbs-client on a non-TEE environment, and I was getting a attester related error (unfortunately I didn't save the logs). Then I built the client with `sample_only` feature and it magically worked out. Thus, I assumed (wrongly) `sample_only` feature should be used on non-TEE environments.

Now I updated my main branch and I no longer see the error with kbs-client built with `default`...probably I was hitting a bug solved recently.

I will re-phrase the sentences.

wainersm:
Updated with following changes:
 - the allow_all.rego policy file was missing a empty line at end of the file. I noticed that if it doesn't have empty lines at first and last line of the file then rego engineer simply fails to parse it. Is it expected?
 - Update the instruction to install rust. The packaged rustc on Ubuntu 22.04 no longer builds the kbs (some recent change probably)
 - Re-phrased the explanation of `sample_only` feature of client. Looks better now, @fitzthum @Xynnn007 ?

fitzthum:
> the allow_all.rego policy file was missing a empty line at end of the file. I noticed that if it doesn't have empty lines at first and last line of the file then rego engineer simply fails to parse it. Is it expected?

The Rego engine is really finicky about this sort of thing. In theory I don't think this should be required, but for now it is.

fitzthum (approved):
LGTM.

Do you want to add in a deny_all policy as well before we merge? Could come in handy.

Xynnn007 (approved):
lgtm. the CI error could be fixed by #331

wainersm:
> LGTM.
> 
> Do you want to add in a deny_all policy as well before we merge? Could come in handy.

Good idea. Let me do it.

PR #325, opened 2024-02-14
ci: set certs/key as makefile deps in e2e test

mkulke:
token-cert-chain.pem, ca-cert.pem and token-key.pem are specified in the configfiles. Without them a launch of kbs or resource-kbs will fail. Hence they are added as dependencies in the Makefile.


fitzthum (approved):
LGTM

PR #326, opened 2024-02-15
Verifier: Refactor errors in cca module

kartikjoshi21:
Fixes: #231


attestation-service/verifier/Cargo.toml
```diff
@@ -16,6 +16,7 @@ cca-verifier = [ "ear", "veraison-apiclient" ]
 
 [dependencies]
 anyhow.workspace = true
+thiserror.workspace = true
```
Xynnn007, line 19: Can you put this dep in dictionary order?

attestation-service/verifier/Cargo.toml
```diff
@@ -30,6 +31,7 @@ csv-rs = { git = "https://github.com/openanolis/csv-rs", rev = "b74aa8c", option
 eventlog-rs = { version = "0.1.3", optional = true }
 hex.workspace = true
 jsonwebtoken = "8"
+jsonwebkey = "0.3.5"
```
Xynnn007, line 34: ditto

attestation-service/verifier/src/cca/mod.rs
```diff
@@ -60,6 +61,44 @@ struct CcaEvidence {
...
+    #[error("get realm evidence from the cca evidence map")]
+    GetRealmEvidence,
+    #[error("CAA Verifier error: {0}")]
+    CAAVerifier(String),
+    #[error("verification error")]
+    Verification(#[from] veraison_apiclient::Error),
+    #[error("Failed to discover the verification endpoint details")]
+    VerificationEndpoint(),
+    #[error("anyhow error: {0}")]
+    Anyhow(#[from] anyhow::Error),
```
Xynnn007, line 99: I am worried about this kind of error. Other types have determined semantics while this not. Could you try to classify the source of this error as one of the above and use `map_err`?

attestation-service/verifier/src/cca/mod.rs
```diff
@@ -60,6 +61,44 @@ struct CcaEvidence {
...
+    #[error("get the decoding key from the pem public key: {0}")]
+    GetDecodingKey(String),
+    #[error("decrypt the ear with the decoding key: {0}")]
+    DecryptDecodingKey(String),
+    #[error("decode nonce byte from ear: {0}")]
+    DecodeNonceByte(String),
+    #[error("report data is different from that in ear's session nonce")]
+    ReportDataMismatch,
+    #[error("no entry found for CCA_SSD_PLATFORM")]
+    NoEntryFound,
```
Xynnn007, line 81: Could we rename this error type? `NoEntryFound` seems generic.

attestation-service/verifier/src/cca/mod.rs
```diff
@@ -80,13 +119,13 @@ impl Verifier for CCA {
         expected_init_data_hash: &InitDataHash,
     ) -> Result<TeeEvidenceParsedClaim> {
         let ReportData::Value(expected_report_data) = expected_report_data else {
-            bail!("CCA verifier must provide report data field!");
+            return Err(CCAError::ReportDataMissing.into());
```
Xynnn007, line 122: I notice that severval `.into()` are used in this function to convert `Result<_, CCAError>` to `Result<_, anyhow::Error>`.

One way to avoid this is to have an inner non-pub function `cca_evaluate() -> Result<_, CCAError>` to be wrapped by this `evaluate()`. In `cca_evaluate()` can delete all the `.into()`s. And in `evaluate()` just call `cca_evaluate()` to finish the error conversion.

Another way is to add another overall error type `VerifierError` under `verifier`. One enum is named `CCA`. And change the signature of `evaluate(..)->anyhow::Result<TeeEvidenceParsedClaim>` to `evaluate(..)->Result<TeeEvidenceParsedClaim, VerifierError>`. In this function different `CCAError`s will finally be mapped into `VerifierError::CCA`.

It's up to you : )

attestation-service/verifier/src/cca/mod.rs
```diff
@@ -60,6 +61,44 @@ struct CcaEvidence {
...
+    #[error("no entry found for CCA_SSD_PLATFORM")]
+    NoEntryFound,
+    #[error("init data hash is different from that in CCA token")]
+    InitDataHashMismatch,
+    #[error("Failed to decode base64: {0}")]
+    FailedtoDecodeBase64(String),
+    #[error("get platform evidence from the cca evidence map")]
+    GetPlatformEvidence,
+    #[error("Serde Json Error")]
+    SerdeJson(#[from] serde_json::Error),
```
Xynnn007, line 89: Could we add some extra information to this kind of error? I mean if this error is raised, the user will only see `Serde Json Error` and he will not know what is serded

Xynnn007 (commented):
Thanks for the great work of replacing anyhow to thiserror. I love this. Let's make it better

attestation-service/verifier/src/cca/mod.rs
kartikjoshi21: Adding error types with source as serde::json error and whole context for this.

thomas-fossati:
Hi @kartikjoshi21, thanks for this.

AFAICS, your fork is not up-to-date with upstream: in particular, the files you are touching have moved to `deps/verifier/src/cca`.  You may want to rebase it.

PR #327, opened 2024-02-15
attestation-service: Refactor errors in attestation module

kartikjoshi21:
Replace anyhow error crate with thiserror crate
Fixes: #231

### DIFF
--- a/attestation-service/attestation-service/src/bin/grpc/mod.rs
+++ b/attestation-service/attestation-service/src/bin/grpc/mod.rs
@@ -1,13 +1,15 @@
-use anyhow::{anyhow, Result};
 use attestation_service::policy_engine::SetPolicyInput;
 use attestation_service::HashAlgorithm;
-use attestation_service::{config::Config, AttestationService as Service, Tee};
+use attestation_service::{
+    config::Config, config::ConfigError, AttestationService as Service, ServiceError, Tee,
+};
 use base64::engine::general_purpose::URL_SAFE_NO_PAD;
 use base64::Engine;
 use log::{debug, info};
 use std::net::SocketAddr;
 use std::path::Path;
 use std::sync::Arc;
+use thiserror::Error;
 use tokio::sync::RwLock;
 use tonic::transport::Server;
 use tonic::{Request, Response, Status};
@@ -40,15 +42,24 @@ fn to_kbs_tee(tee: GrpcTee) -> Tee {
     }
 }
 
+#[derive(Error, Debug)]
+pub enum GrpcError {
+    #[error("Read AS config file failed: {0}")]
+    Config(#[source] ConfigError),
+    #[error("Creating attestation service failed: {0}")]
+    Service(#[from] ServiceError),
+    #[error("tonic transport error: {0}")]
+    TonicTransport(#[from] tonic::transport::Error),
+}
+
 pub struct AttestationServer {
     attestation_service: Service,
 }
 
 impl AttestationServer {
-    pub async fn new(config_path: Option<String>) -> Result<Self> {
+    pub async fn new(config_path: Option<String>) -> Result<Self, GrpcError> {
         let config = match config_path {
-            Some(path) => Config::try_from(Path::new(&path))
-                .map_err(|e| anyhow!("Read AS config file failed: {:?}", e))?,
+            Some(path) => Config::try_from(Path::new(&path)).map_err(GrpcError::Config)?,
             None => Config::default(),
         };
 
@@ -222,7 +233,7 @@ impl ReferenceValueProviderService for Arc<RwLock<AttestationServer>> {
     }
 }
 
-pub async fn start(socket: SocketAddr, config_path: Option<String>) -> Result<()> {
+pub async fn start(socket: SocketAddr, config_path: Option<String>) -> Result<(), GrpcError> {
     info!("Listen socket: {}", &socket);
 
     let attestation_server = Arc::new(RwLock::new(AttestationServer::new(config_path).await?));
--- a/attestation-service/attestation-service/src/bin/restful-as.rs
+++ b/attestation-service/attestation-service/src/bin/restful-as.rs
@@ -1,8 +1,8 @@
 use std::{net::SocketAddr, path::Path, sync::Arc};
 
 use actix_web::{web, App, HttpServer};
-use anyhow::{anyhow, Context, Result};
-use attestation_service::{config::Config, AttestationService};
+use anyhow::Result;
+use attestation_service::{config::Config, config::ConfigError, AttestationService, ServiceError};
 use clap::{arg, command, Parser};
 use log::info;
 use openssl::{
@@ -10,6 +10,7 @@ use openssl::{
     ssl::{SslAcceptor, SslMethod},
 };
 use strum::{AsRefStr, EnumString};
+use thiserror::Error;
 use tokio::sync::RwLock;
 
 use crate::restful::{attestation, set_policy};
@@ -49,8 +50,30 @@ enum WebApi {
     Policy,
 }
 
+#[derive(Error, Debug)]
+pub enum RestfulError {
+    #[error("Creating service failed: {0}")]
+    Service(#[from] ServiceError),
+    #[error("Failed to read AS config file: {0}")]
+    Config(#[from] ConfigError),
+    #[error("Openssl errorstack: {0}")]
+    Openssl(#[from] openssl::error::ErrorStack),
+    #[error("failed to read HTTPS private key: {0}")]
+    ReadHttpsKey(#[source] std::io::Error),
+    #[error("failed to get HTTPS private key from pem: {0}")]
+    ReadHttpsKeyFromPem(#[source] openssl::error::ErrorStack),
+    #[error("set private key failed: {0}")]
+    SetPrivateKey(#[source] openssl::error::ErrorStack),
+    #[error("set HTTPS public key cert: {0}")]
+    SetHttpsCert(#[source] openssl::error::ErrorStack),
+    #[error("io error: {0}")]
+    IO(#[from] std::io::Error),
+    #[error(transparent)]
+    Anyhow(#[from] anyhow::Error),
+}
+
 #[actix_web::main]
-async fn main() -> Result<()> {
+async fn main() -> Result<(), RestfulError> {
     env_logger::init_from_env(env_logger::Env::new().default_filter_or("info"));
 
     let cli = Cli::parse();
@@ -58,8 +81,7 @@ async fn main() -> Result<()> {
     let config = match cli.config_file {
         Some(path) => {
             info!("Using config file {path}");
-            Config::try_from(Path::new(&path))
-                .map_err(|e| anyhow!("Read AS config file failed: {:?}", e))?
+            Config::try_from(Path::new(&path))?
         }
         None => {
             info!("No confile path provided, use default one.");
@@ -83,16 +105,16 @@ async fn main() -> Result<()> {
 
             let prikey = tokio::fs::read(prikey)
                 .await
-                .context("read HTTPS private key")?;
+                .map_err(RestfulError::ReadHttpsKey)?;
             let prikey =
-                PKey::private_key_from_pem(&prikey).context("read HTTPS private key from pem")?;
+                PKey::private_key_from_pem(&prikey).map_err(RestfulError::ReadHttpsKeyFromPem)?;
 
             builder
                 .set_private_key(&prikey)
-                .context("set private key failed")?;
+                .map_err(RestfulError::SetPrivateKey)?;
             builder
                 .set_certificate_chain_file(pubkey_cert)
-                .context("set HTTPS public key cert")?;
+                .map_err(RestfulError::SetHttpsCert)?;
             log::info!("starting HTTPS server at https://{}", cli.socket);
             server.bind_openssl(cli.socket, builder)?.run()
         }
@@ -105,6 +127,5 @@ async fn main() -> Result<()> {
     };
 
     server.await?;
-
     Ok(())
 }
--- a/attestation-service/attestation-service/src/config.rs
+++ b/attestation-service/attestation-service/src/config.rs
@@ -1,10 +1,10 @@
 use crate::rvps::RvpsConfig;
 use crate::token::{AttestationTokenBrokerType, AttestationTokenConfig};
 
-use anyhow::{anyhow, Result};
 use serde::Deserialize;
 use std::fs::File;
 use std::path::{Path, PathBuf};
+use thiserror::Error;
 
 /// Environment macro for Attestation Service work dir.
 const AS_WORK_DIR: &str = "AS_WORK_DIR";
@@ -29,6 +29,18 @@ pub struct Config {
 
     /// The Attestation Result Token Broker Config
     pub attestation_token_config: AttestationTokenConfig,
+}
+
+#[derive(Error, Debug)]
+pub enum ConfigError {
+    #[error("io error: {0}")]
+    IO(#[from] std::io::Error),
+    #[error("failed to parse AS config file: {0}")]
+    FileParse(#[source] std::io::Error),
+    #[error("failed to parse AS config file: {0}")]
+    JsonFileParse(#[source] serde_json::Error),
+    #[error("Illegal format of the content of the configuration file: {0}")]
+    SerdeJson(#[from] serde_json::Error),
 }
 
 impl Default for Config {
@@ -63,12 +75,9 @@ impl TryFrom<&Path> for Config {
     ///            "duration_min": 5
     ///        }
     ///    }
-    type Error = anyhow::Error;
-    fn try_from(config_path: &Path) -> Result<Self, Self::Error> {
-        let file = File::open(config_path)
-            .map_err(|e| anyhow!("failed to open AS config file {}", e.to_string()))?;
-
-        serde_json::from_reader::<File, Config>(file)
-            .map_err(|e| anyhow!("failed to parse AS config file {}", e.to_string()))
+    type Error = ConfigError;
+    fn try_from(config_path: &Path) -> Result<Self, ConfigError> {
+        let file = File::open(config_path)?;
+        serde_json::from_reader::<File, Config>(file).map_err(ConfigError::JsonFileParse)
     }
 }
--- a/attestation-service/attestation-service/src/lib.rs
+++ b/attestation-service/attestation-service/src/lib.rs
@@ -17,12 +17,13 @@ use config::Config;
 pub use kbs_types::{Attestation, Tee};
 use log::debug;
 use policy_engine::{PolicyEngine, PolicyEngineType, SetPolicyInput};
-use rvps::RvpsApi;
+use rvps::{RvpsApi, RvpsError};
 use serde_json::{json, Value};
 use serde_variant::to_variant_name;
 use sha2::{Digest, Sha256, Sha384, Sha512};
 use std::{collections::HashMap, str::FromStr};
 use strum::{AsRefStr, EnumString};
+use thiserror::Error;
 use tokio::fs;
 use verifier::{InitDataHash, ReportData};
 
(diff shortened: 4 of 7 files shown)


attestation-service/attestation-service/src/rvps/mod.rs
```diff
@@ -55,10 +57,28 @@ impl Default for RvpsConfig {
     }
 }
 
+#[derive(Error, Debug)]
+pub enum RvpsError {
+    #[error("Failed creating rvps service: {0}")]
+    CreateRvps(String),
+    #[error("feature `rvps-grpc` or `rvps-builtin` not enabled: {0}")]
+    FeatureNotEnabled(String),
```
mkulke, line 65: Why does this have a String, there is parameter if it's called, correct?

attestation-service/attestation-service/src/rvps/mod.rs
```diff
@@ -55,10 +57,28 @@ impl Default for RvpsConfig {
...
+    #[error("Failed to register grpc agent: {0}")]
+    AgentRegistration(String),
```
mkulke, line 67: Where is AgentRegistration being used?

mkulke (changes requested):

attestation-service/attestation-service/src/lib.rs
```diff
@@ -77,6 +78,24 @@ pub enum Data {
...
+    #[error("Policy Engine {0} is not supported")]
+    UnsupportedPolicyEngine(String),
+    #[error("Create rvps failed.")]
+    Rvps(#[from] RvpsError),
+    #[error("Set Policy failed {0}")]
+    SetPolicyFailed(String),
+    #[error("Starting service failed: {0}")]
+    StartService(String),
+    #[error("Anyhow error: {0}")]
+    Anyhow(#[from] anyhow::Error)
```
Xynnn007, line 96: `anyhow` is too generic. Could you try to classify the source of this error as one of the above and use `map_err`?

A suggestion is to firstly define a set of meanful error types with determined semantics. Then Let's go to use them by `map_err`. The definition should not be too generic, otherwise the user will be confused if an error is raised.

wdyt?

attestation-service/attestation-service/src/rvps/builtin.rs
```diff
@@ -9,16 +10,21 @@ pub struct Rvps {
...
-        let core = Core::new(store_type)?;
-        Ok(Self { core })
+    pub fn new(store_type: &str) -> Result<Self, RvpsError> {
+        Core::new(store_type)
+            .map(|core| Self::from_core(core))
+            .map_err(|error| RvpsError::CreateRvps(error.to_string()))
+    }
+
+    fn from_core(core: Core) -> Self {
+        Self { core }
```
mkulke, line 20: why create a from_core() fn, we shouldn't  need to perform code changes if we adjust the error types, it should mostly propagate fine with `..?;`, no need for `map_err()` in most cases.

I don't think we should convert anyhow errors (`Core::new()` returns an anyhow error) and wrap them in ours. If we want to do not start "at the bottom", but do this later we can add:

```
    #[error("Error")]
    Anyhow(#[from] anyhow::Error),
```

to the error enum and the code should keep working.

attestation-service/attestation-service/src/rvps/mod.rs
```diff
@@ -55,10 +56,24 @@ impl Default for RvpsConfig {
...
+    #[error("Failed creating rvps service: {0}")]
+    CreateRvps(String),
+    #[error("feature `rvps-grpc` or `rvps-builtin` should be enabled")]
+    FeatureNotEnabled,
+    #[error("Serde Json Error")]
+    SerdeJson(#[from] serde_json::Error),
+    #[error("Returned status: {0}")]
+    Status(#[from] tonic::Status),
+    #[error("tonic transport error: {0}")]
+    TonicTransport(#[from] tonic::transport::Error),
```
mkulke, line 70: I don't think those `...: {0}` will do anything, there's no embedded string, see thiserror docs:

image

If you wrap an error you don't need string templating, thiserror will do this.

attestation-service/attestation-service/src/rvps/grpc.rs
```diff
@@ -37,8 +37,7 @@ impl RvpsApi for Agent {
             .lock()
             .await
             .register_reference_value(req)
-            .await
-            .context("register failed")?;
```
mkulke, line 41: it's still an anyhow error, why remove the context?

attestation-service/attestation-service/src/rvps/builtin.rs
```diff
@@ -9,16 +10,21 @@ pub struct Rvps {
...
     }
 }
 
 #[async_trait]
 impl RvpsApi for Rvps {
     async fn verify_and_extract(&mut self, message: &str) -> Result<()> {
-        self.core.verify_and_extract(message).await
+        Ok(self.core.verify_and_extract(message).await?)
```
mkulke, line 27:
```suggestion
        self.core.verify_and_extract(message).await?;
        Ok(())
```

attestation-service/attestation-service/src/bin/grpc/mod.rs
```diff
@@ -231,6 +240,7 @@ pub async fn start(socket: SocketAddr, config_path: Option<String>) -> Result<()
         .add_service(AttestationServiceServer::new(attestation_server.clone()))
         .add_service(ReferenceValueProviderServiceServer::new(attestation_server))
         .serve(socket)
-        .await?;
+        .await
+        .map_err(|e| ServiceError::StartService(e.to_string()))?;
```
mkulke, line 244: let's not wrap anyhow errors and convert them into strings.

attestation-service/attestation-service/src/bin/restful-as.rs
```diff
@@ -49,17 +50,30 @@ enum WebApi {
...
+pub enum RestfulError {
+    #[error("Creating service failed: {0}: {0}")]
+    Service(#[from] ServiceError),
+    #[error("Failed to read AS config file: {0}")]
+    Config(#[from] ConfigError),
+    #[error("Openssl errorstack: {0}")]
+    Openssl(#[from] openssl::error::ErrorStack),
+    #[error("io error")]
+    IO(#[from] std::io::Error),
+    #[error("Failed to start server: {0}")]
```
mkulke, line 63: see other comments about `{0}`

mkulke (changes requested):
a couple of smaller remarks, but we're getting there

attestation-service/attestation-service/src/config.rs
```diff
@@ -65,12 +73,9 @@ impl TryFrom<&Path> for Config {
...
-    type Error = anyhow::Error;
-    fn try_from(config_path: &Path) -> Result<Self, Self::Error> {
-        let file = File::open(config_path)
-            .map_err(|e| anyhow!("failed to open AS config file {}", e.to_string()))?;
-
-        serde_json::from_reader::<File, Config>(file)
-            .map_err(|e| anyhow!("failed to parse AS config file {}", e.to_string()))
+    type Error = ConfigError;
+    fn try_from(config_path: &Path) -> Result<Self, ConfigError> {
+        let file = File::open(config_path)?;
```
Xynnn007, line 78: The original `map_err(|e| anyhow!("failed to open AS config file {}", e.to_string()))?;`

will add extra context of the config file path and it is now opening AS config file. I am not sure the directly `?` here will result in a `IO(#[from] std::io::Error)` defined in `ConfigError`, which will only show an error information `io error`. This would make users confused.

My suggestion is that we can add some more context due to error type, s.t.
```rust
pub enum ConfigError {
    #[error("Failed to open CoCoAS' configuration file: {0}")]
    IO(#[from] std::io::Error),
    #[error("Illegal format of the content of the configuration file: {0}")]
    SerdeJson(#[from] serde_json::Error),
}
```

This would make the error information clearer. wdyt?

attestation-service/attestation-service/src/bin/restful-as.rs
```diff
@@ -49,17 +50,30 @@ enum WebApi {
...
+    #[error("Creating service failed")]
+    Service(#[from] ServiceError),
+    #[error("Failed to read AS config file")]
+    Config(#[from] ConfigError),
+    #[error("Openssl errorstack")]
+    Openssl(#[from] openssl::error::ErrorStack),
+    #[error("io error")]
+    IO(#[from] std::io::Error),
+    #[error("Error")]
+    Anyhow(#[from] anyhow::Error),
```
Xynnn007, line 64: If it is set `#[error("Error")]`, user will see a very simple error string `Error` which will be confusing.

For anyhow, we can use `#[error(transparent)]` due to https://docs.rs/thiserror/1.0.57/thiserror/

attestation-service/attestation-service/src/bin/restful-as.rs
```diff
@@ -49,17 +50,30 @@ enum WebApi {
...
 
+#[derive(Error, Debug)]
+pub enum RestfulError {
+    #[error("Creating service failed")]
+    Service(#[from] ServiceError),
+    #[error("Failed to read AS config file")]
+    Config(#[from] ConfigError),
+    #[error("Openssl errorstack")]
+    Openssl(#[from] openssl::error::ErrorStack),
+    #[error("io error")]
```
Xynnn007, line 61: we'd better add some more details about the error, or include the original error inside the error information, s.t.
```
#[error("io error: {0}")]
```

attestation-service/attestation-service/src/bin/restful-as.rs
```diff
@@ -49,17 +50,30 @@ enum WebApi {
     Policy,
 }
 
+#[derive(Error, Debug)]
+pub enum RestfulError {
+    #[error("Creating service failed")]
+    Service(#[from] ServiceError),
+    #[error("Failed to read AS config file")]
+    Config(#[from] ConfigError),
+    #[error("Openssl errorstack")]
```
Xynnn007, line 59: we'd better add some more details about the error, or include the original error inside the error information, s.t.
```rust
#[error("HTTPS set failed due to openssl error: {0}")]
```

attestation-service/attestation-service/src/lib.rs
```diff
@@ -77,6 +78,18 @@ pub enum Data {
...
+#[derive(Error, Debug)]
+pub enum ServiceError {
+    #[error("io error")]
+    IO(#[from] std::io::Error),
+    #[error("Parse error")]
+    ParseError(#[from] strum::ParseError),
+    #[error("Create rvps failed.")]
+    Rvps(#[from] RvpsError),
+    #[error("Error")]
+    Anyhow(#[from] anyhow::Error),
```
Xynnn007, line 90: ditto

attestation-service/attestation-service/src/rvps/mod.rs
```diff
@@ -55,10 +56,24 @@ impl Default for RvpsConfig {
...
+    #[error("feature `rvps-grpc` or `rvps-builtin` should be enabled")]
+    FeatureNotEnabled,
+    #[error("Serde Json Error")]
+    SerdeJson(#[from] serde_json::Error),
+    #[error("Returned status")]
+    Status(#[from] tonic::Status),
+    #[error("tonic transport error")]
+    TonicTransport(#[from] tonic::transport::Error),
+    #[error("Error")]
+    Anyhow(#[from] anyhow::Error),
```
Xynnn007, line 70: ditto

attestation-service/attestation-service/src/bin/restful-as.rs
```diff
@@ -49,17 +50,30 @@ enum WebApi {
     Policy,
 }
 
+#[derive(Error, Debug)]
+pub enum RestfulError {
+    #[error("Creating service failed")]
```
Xynnn007, line 55: we'd better add some more details about the error, or include the original error inside the error information, s.t.
```rust
#[error("Creating service failed: {0}")]
```

attestation-service/attestation-service/src/bin/restful-as.rs
```diff
@@ -49,17 +50,30 @@ enum WebApi {
...
+    Service(#[from] ServiceError),
+    #[error("Failed to read AS config file")]
```
Xynnn007, line 57: we'd better add some more details about the error, or include the original error inside the error information, s.t.
```
#[error("Failed to read AS config file: {0}")]
```

attestation-service/attestation-service/src/config.rs
mkulke: I agree. I don't think we need the `{0}` when wrapping errors, though.

mkulke (commented):
almost there, I think we need to keep the error messages specific when we refactor from `anyhow!` or `.context()`. pasted some examples in the recent comments to show how this can be achieved.

attestation-service/attestation-service/src/bin/restful-as.rs
Xynnn007: Oh. I use `#[from]` macro like the following
```rust
use thiserror::Error;

#[derive(Error, Debug)]
enum TestError {
    #[error("io error")]
    IO(#[from] std::io::Error)
}


fn t() -> Result<(), TestError> {
    let _a = std::fs::File::open("./nothisfile")?;
    Ok(())
}

fn main() -> Result<(), TestError> {
    let a = t();
    println!("{}", a.expect_err("ss"));
    Ok(())
}
```

The output is 
```
io error
```
without any context.

If convert `#[error("io error")]` to `#[error("io error: {0}")]`, the output will be
```
io error: No such file or directory (os error 2)
```

mkulke: oh, that's bad. so, it's up to the logger impl to print the cause or not. We could add  `{0}` to the message. afaiu this will just call the .fmt() of the wrapped struct, in this case the error cause.

I found it a bit unfortunate to see those deeply nested log lines e.g. in API responses ("Error: ... Error: ... Error ...") that look like flattened backtraces and maybe even reveal too much of the internals on an API endpoint. They are sometimes called "Big Ball of Mud" errors: (see [this chapter](https://www.lpalmieri.com/posts/error-handling-rust/#avoid-ball-of-mud-error-enums) from the the zero2prod rust book). They are bad because:

> We do not want to expose the implementation details of the fallible routines that get mapped to UnexpectedError by subscribe - it must be opaque.

E.g. you don't want to log the rate limit error that you get from your managed database backend to the consumer.

There seems to be consensus that duplicating the source on the error message is an antipattern, see this [SO answer](https://stackoverflow.com/questions/62869360/should-an-error-with-a-source-include-that-source-in-the-display-output).

That said, if the logger doesn't respect "source", I don't know how this information can still be surfaced in the log output, but limited to the consumer, especially with the autogenerated RPC code. I'll open a dedicated issue about this, and it should be fine in the PR.

Xynnn007: Overall, I do not like the flattened format. But compared to a non-detailed error information, I have to choose the flatten one.

You are right that the error information should care about the boundary, e.g. `AttestationService` lib should response a detailed error information, while gRPC/RESTful responsed error should not expose internal details.

To me, if way 2 in https://stackoverflow.com/a/63620301 would be better in crates (e.g. "failed to open file XXX" is better than "io error", which will do great help to debug), while we can cut the details of the errors in concrete binary implementations to avoid explosure of internal information.

mkulke: apparently, the log create doesn't support structured logging yet (looks like they plan to add it, though). fwiw in the the `tracing` crate it looks like this:

```rust
        error!(error = %e, source = e.source(), "could not do this");
``` 

which will produce:

```bash
2024-03-04T10:41:17.556335Z ERROR rust_error_logging: could not do this error=some thing with context source=some other thing failed
```

Xynnn007: Well. This could be a problem but up to now it is ok to me without structured information.

attestation-service/attestation-service/src/config.rs
```diff
@@ -63,12 +75,9 @@ impl TryFrom<&Path> for Config {
...
-    type Error = anyhow::Error;
-    fn try_from(config_path: &Path) -> Result<Self, Self::Error> {
-        let file = File::open(config_path)
-            .map_err(|e| anyhow!("failed to open AS config file {}", e.to_string()))?;
-
-        serde_json::from_reader::<File, Config>(file)
-            .map_err(|e| anyhow!("failed to parse AS config file {}", e.to_string()))
+    type Error = ConfigError;
+    fn try_from(config_path: &Path) -> Result<Self, ConfigError> {
+        let file = File::open(config_path).map_err(ConfigError::FileParse)?;
```
mkulke, line 80: nothing is being parsed here, that's just an IO error, I don't think we need `ConfigError::FileParse`

attestation-service/attestation-service/src/lib.rs
```diff
@@ -86,20 +101,20 @@ pub struct AttestationService {
...
         if !config.work_dir.as_path().exists() {
             fs::create_dir_all(&config.work_dir)
                 .await
-                .context("Create AS work dir failed: {:?}")?;
+                .map_err(ServiceError::CreateDir)?;
         }
 
         let policy_engine = PolicyEngineType::from_str(&config.policy_engine)
-            .map_err(|_| anyhow!("Policy Engine {} is not supported", &config.policy_engine))?
+            .map_err(ServiceError::ParseError)?
```
mkulke, line 112: `ServiceError::UnsupportedPolicy` would be better maybe?

attestation-service/attestation-service/src/lib.rs
```diff
@@ -115,10 +130,9 @@ impl AttestationService {
 
     /// Set Attestation Verification Policy.
     pub async fn set_policy(&mut self, input: SetPolicyInput) -> Result<()> {
```
mkulke, line 132: this is still anyhow, should/can we convert it?

attestation-service/attestation-service/src/bin/restful-as.rs
```diff
@@ -49,17 +50,38 @@ enum WebApi {
...
+    Config(#[from] ConfigError),
+    #[error("Openssl errorstack")]
+    Openssl(#[from] openssl::error::ErrorStack),
+    #[error("read HTTPS private key")]
+    ReadHttpsKey(#[source] std::io::Error),
+    #[error("read HTTPS private key from pem")]
+    ReadHttpsKeyFromPem(#[source] openssl::error::ErrorStack),
```
mkulke, line 64:
```suggestion
    #[error("failed to get HTTPS private key from pem")]
    ReadHttpsKeyFromPem(#[source] openssl::error::ErrorStack),
```

attestation-service/attestation-service/src/bin/restful-as.rs
```diff
@@ -49,17 +50,38 @@ enum WebApi {
...
 
+#[derive(Error, Debug)]
+pub enum RestfulError {
+    #[error("Creating service failed")]
+    Service(#[from] ServiceError),
+    #[error("Failed to read AS config file")]
+    Config(#[from] ConfigError),
+    #[error("Openssl errorstack")]
+    Openssl(#[from] openssl::error::ErrorStack),
+    #[error("read HTTPS private key")]
```
mkulke, line 61:
```suggestion
    #[error("failed to read HTTPS private key")]
```

mkulke (commented):
looks good, small nits from my side. @Xynnn007 comments should are left to be addressed, mostly adding `{0}` to extend the error message, where appropriate.

attestation-service/attestation-service/src/config.rs
kartikjoshi21: Thats true.

attestation-service/attestation-service/src/lib.rs
kartikjoshi21: This one has dependency from `kbs/src/api/src/attestation/coco/builtin.rs:25`.  we can ignore this one as part of initial refactoring. Later on things can be extended to broader directories one all the basic refactoring is done in these modules. wdyt ?

mkulke: yup

mkulke (approved):
lgtm, thanks for accommodating the comments

fitzthum (commented):
Looks pretty good.

My only question is about how you chose the names for error enums. It seems like you are creating an enum for each file, but this might not be optimal. For instance we end up with the same errors in multiple enums (like the config error). It also means that the names of the enums aren't super meaningful. The `RestfulError` enum contains a bunch of stuff that isn't realy related to Rest but is just part of that file. `ServiceError` is a really general term.

I wonder if there is a way to break up the errors into different logical groups, e.g. `ConfigError`, `ProtocolError`, `CryptoError`, and then share these enums between files.

I guess this isn't too big of a deal, though. If @Xynnn007 and @mkulke are happy with the PR then we can go ahead as is.

mkulke:
> Looks pretty good.
> 
> My only question is about how you chose the names for error enums. It seems like you are creating an enum for each file, but this might not be optimal. For instance we end up with the same errors in multiple enums (like the config error). It also means that the names of the enums aren't super meaningful. The `RestfulError` enum contains a bunch of stuff that isn't realy related to Rest but is just part of that file. `ServiceError` is a really general term.
> 
> I wonder if there is a way to break up the errors into different logical groups, e.g. `ConfigError`, `ProtocolError`, `CryptoError`, and then share these enums between files.

ah indeed, in some of those cases a per-file error doesn't make much sense, I agree. e.g. in the case of RestfulError & GrpcError, maybe they could be split into ProtocolError and ,idk, "ServiceError"?

Xynnn007 (approved):
Thanks @kartikjoshi21

PR #328, opened 2024-02-15
Verifier: Refactor errors in sgx module

kartikjoshi21:
Fixes: #231


attestation-service/verifier/src/sgx/mod.rs
```diff
@@ -41,6 +42,36 @@ struct SgxEvidence {
...
+    TeeGetQuoteSupplemtalData(String),
+    #[error("Verification completed with Terminal result: {0}")]
+    QuoteVerificationResult(String),
+    #[error("generate_parsed_claims failed: {0}")]
+    GenerateParsedClaims(String),
+    #[error("anyhow error: {0}")]
+    Anyhow(#[from] anyhow::Error),
+    #[error("ecdsa quote verification failed: {0}")]
+    ECDSAQuoteVerification(String),
+}
```
mkulke, line 73: that's a large enum of errors. It's a lot more than the errors that have been touched below, are they used somewhere?

mkulke (changes requested):
with thiserror we do not have to recreate every error explicitly (it will also lose information if the original error is reduced to a string). The following code illustrates how it works.



```rust
use anyhow;
use thiserror;
use thiserror::Error;
use std::fmt;

#[derive(Debug, Clone)]
struct OtherError;

impl std::error::Error for OtherError {}

impl fmt::Display for OtherError {
    fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
        write!(formatter, "some other thing failed")
    }
}

fn with_other_error() -> Result<(), OtherError> {
    Ok(())
}

#[derive(Error, Debug)]
pub enum MyError {
    #[error("some thing")]
    Something,
    #[error("some other thing")]
    Other(#[from] OtherError)
}

fn with_thiserror() -> Result<(), MyError> {
    with_other_error()?;
    Err(MyError::Something)
}

fn top_level_with_anyhow() -> anyhow::Result<()> {
    with_thiserror()?;
    Ok(())
}
```

`top_level_with_anyhow()`  returns an anyhow Result, it calls `with_thiserror()` which returns `MyError`, a thiserror enum as a Result. `MyError` has both its own type as well as a wrapper for `OtherError` which is a custom defined type, which is returned by `with_other_error()`. 

We don't need to recreate any errors. `MyError` converts into anyhow implicitly and for `OtherError` we create a simple ´#from` wrapper in the `MyError` struct.

mkulke:
@kartikjoshi21 

note, in some cases we want to provide context to the source error because it isn't specific enough. that's how `.context("...")` is being used here with anyhow.

Usually the source error should provide the context we need but sometimes it's not enough and we need to provide context from the caller site:

```rust
fn with_other_error() -> Result<(), OtherError> {
    Err(OtherError)
}

#[derive(Error, Debug)]
pub enum MyError {
    #[error("some thing")]
    Something,
    #[error("some other thing")]
    Other(#[from] OtherError),
    #[error("some other thing with context")]
    FailedDoingSomethingSpecific(#[source] OtherError),
}

fn with_thiserror() -> Result<(), MyError> {
    // we're doing something specific, that we want provide context about which
    // the source error doesn't have.
    with_other_error().map_err(|e| MyError::FailedDoingSomethingSpecific(e))?;
    // or shorter even, since the constructor to `MyError::FailedDoingSomethingSpecific`
    // can act as a closure
    with_other_error().map_err(MyError::FailedDoingSomethingSpecific)?;
    // we can just bubble up the source error, we don't need context
    with_other_error()?;
    Err(MyError::Something)
}
```

mythi:
@kartikjoshi21 are you still working on this? One higher-level comment is that many of the errors introduced here are generic to DCAP (covers both TDX and SGX). We have also discussed getting those two verifier modules to reuse code (see https://github.com/confidential-containers/trustee/pull/404#issuecomment-2148827562) so perhaps we need a set to DCAP errors instead?

dcmiddle:
I think this PR is stale and considering the proposed deprecation of enclave-cc I am going to close it.
Should @kartikjoshi21 or @mythi want to pick it back up again, though you can always re-open it.

PR #329, opened 2024-02-15
kbs/config: use Intel PCS to fetch SGX/TDX quote verification collateral

mythi:
DRAFT until we figure out the actions caching to stay compliant with the PCS Service terms: "Caching of the Public Certificate Key (PCK) and Trusted Computing Based (TCB) information is required to avoid unnecessary high-frequency calls to the Intel Provisioning Certification Services (PCS). "


mythi:
We got #323 merged and in that the CI builds its own config and takes care of the caching. I'm thinking we could leave the 'default' here as is and just close this. @mkulke thoughts?

mkulke:
fine w/ me.

PR #336, opened 2024-02-22
ci: fix DCAP package install

mythi:
ubuntu-latest is not focal anymore and will eventually move to 24.04. Therefore, use a pinned ubuntu-<version> runner and match the DCAP packages repository path with it.

The same fix was done in https://github.com/confidential-containers/guest-components/pull/350


Xynnn007 (approved):
lgtm

wainersm (approved):
Thanks @mythi !

PR #339, opened 2024-02-26
Add configuration file for RVPS and add support for JSON fs storage

### LINKED ISSUE #179: RVPS doesn't observe attestation-service's work-dir config
It will use a hardcoded value

https://github.com/confidential-containers/attestation-service/blob/871ae511286bc08bf0086c572f7ba96ed17662f3/attestation-service/src/rvps/store/local_fs/mod.rs#L18

IMO it should be `${work_dir}/reference_values` instead

### LINKED ISSUE #338: Q: Would a JSON (yaml/toml)-backed RVPS store make sense?
At the moment we rely on an embedded database (sled) to store reference values, that makes a kbs/rvps deployment stateful, which is operationally challenging (think replication). Maybe it makes sense to have the option to provide a flat file store (can be read-only maybe?)  for rvps, so it can be backed by k8s configmaps or similar mechanisms.

Xynnn007:
Close #179 
Close #338 

This PR adds a JSON file storage backend. To leverage this, we need to specify some parameters like the file path of the JSON. Thus also add an overall RVPS config to set different parameters of RVPS. To avoid backward compatibility problems, if no config is provided, a legacy configuration file will be used as default.

### DIFF
--- a/attestation-service/attestation-service/src/lib.rs
+++ b/attestation-service/attestation-service/src/lib.rs
@@ -174,6 +174,7 @@ impl AttestationService {
             .map_err(|e| anyhow!("Verifier evaluate failed: {e:?}"))?;
 
         let flattened_claims = flatten_claims(tee, &claims_from_tee_evidence)?;
+        debug!("flattened_claims: {:#?}", flattened_claims);
 
         let tcb_json = serde_json::to_string(&flattened_claims)?;
 
@@ -181,6 +182,7 @@ impl AttestationService {
             .get_reference_data(flattened_claims.keys())
             .await
             .map_err(|e| anyhow!("Generate reference data failed: {:?}", e))?;
+        debug!("reference_data_map: {:#?}", reference_data_map);
 
         let evaluation_report = self
             .policy_engine
--- a/attestation-service/attestation-service/src/lib.rs
+++ b/attestation-service/attestation-service/src/lib.rs
@@ -97,9 +97,7 @@ impl AttestationService {
             .map_err(|_| anyhow!("Policy Engine {} is not supported", &config.policy_engine))?
             .to_policy_engine(config.work_dir.as_path())?;
 
-        let rvps = config
-            .rvps_config
-            .to_rvps()
+        let rvps = rvps::initialize_rvps_client(&config.rvps_config)
             .await
             .context("create rvps failed.")?;
 
--- a/attestation-service/attestation-service/src/rvps/mod.rs
+++ b/attestation-service/attestation-service/src/rvps/mod.rs
@@ -32,6 +32,7 @@ pub struct RvpsConfig {
     /// Specify the underlying storage type of RVPS, e.g.
     ///
     /// - `localfs`: store inside local filesystem.
+    /// - `localjson`: store inside local json file.
     ///
     /// Only used when feature `rvps-builtin` is enabled
     #[serde(default = "String::default")]
--- a/attestation-service/attestation-service/src/rvps/mod.rs
+++ b/attestation-service/attestation-service/src/rvps/mod.rs
@@ -5,7 +5,9 @@
 
 use anyhow::*;
 use log::{info, warn};
+use reference_value_provider_service::config::{Config as RvpsCrateConfig, DEFAULT_STORAGE_TYPE};
 use serde::Deserialize;
+use serde_json::{json, Value};
 
 /// The interfaces of Reference Value Provider Service
 /// * `verify_and_extract` is responsible for verify a message and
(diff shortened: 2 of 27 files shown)


Xynnn007:
Oh. Let me fix the deployment parts and make it draft temporarily.

attestation-service/rvps/src/store/local_json/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+impl Store for LocalJson {
+    fn set(&mut self, name: String, rv: ReferenceValue) -> Result<Option<ReferenceValue>> {
+        let file = fs::read(&self.file_path)?;
+        let mut dic: HashMap<String, ReferenceValue> = serde_json::from_slice(&file)?;
+        let res = dic.insert(name, rv);
+        let contents = serde_json::to_vec(&dic)?;
+        fs::write(&self.file_path, contents)?;
```
mkulke, line 42: If we want to implement write (I would not recommend it), we probably want to wrap it in mutex.

attestation-service/rvps/src/store/local_json/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+        })
+    }
+}
+
+impl Store for LocalJson {
+    fn set(&mut self, name: String, rv: ReferenceValue) -> Result<Option<ReferenceValue>> {
+        let file = fs::read(&self.file_path)?;
+        let mut dic: HashMap<String, ReferenceValue> = serde_json::from_slice(&file)?;
+        let res = dic.insert(name, rv);
+        let contents = serde_json::to_vec(&dic)?;
```
mkulke, line 41: shouldn't the file be `ReferenceValue[]`, instead of a not a top level key/value  object?

attestation-service/attestation-service/src/lib.rs
```diff
@@ -174,13 +172,15 @@ impl AttestationService {
             .map_err(|e| anyhow!("Verifier evaluate failed: {e:?}"))?;
 
         let flattened_claims = flatten_claims(tee, &claims_from_tee_evidence)?;
+        debug!("flattened_claims: {:#?}", flattened_claims);
```
mkulke, line 175: debug leftover

attestation-service/attestation-service/src/lib.rs
```diff
@@ -174,13 +172,15 @@ impl AttestationService {
...
 
         let tcb_json = serde_json::to_string(&flattened_claims)?;
 
         let reference_data_map = self
             .get_reference_data(flattened_claims.keys())
             .await
             .map_err(|e| anyhow!("Generate reference data failed: {:?}", e))?;
+        debug!("reference_data_map: {:#?}", reference_data_map);
```
mkulke, line 183: debug leftover

attestation-service/attestation-service/src/rvps/mod.rs
```diff
@@ -27,59 +29,64 @@ pub mod grpc;
...
-                            warn!("No RVPS address provided and will launch a built-in rvps");
-                            Ok(Box::new(builtin::Rvps::new(&self.store_type)?) as Box<dyn RvpsApi + Send + Sync>)
-                        } else {
-                            Err(anyhow!("either feature `rvps-grpc` or `rvps-builtin` should be enabled."))
-                        }
+pub async fn initialize_rvps_client(config: &RvpsConfig) -> Result<Box<dyn RvpsApi + Send + Sync>> {
+    cfg_if::cfg_if! {
+        if #[cfg(feature = "rvps-grpc")] {
+            if config.remote_addr.as_ref().is_some_and(|addr| !addr.is_empty()) {
+                let remote_addr = config.remote_addr.as_ref().expect("must be Some");
```
mkulke, line 72: it's probably better to give a more descriptive panic message ("remote addr should be set" or something)

attestation-service/rvps/README.md
```diff
@@ -75,6 +80,20 @@ Run
...
+```json
+{
+    "store_type": "LocalFs",
+    "store_config": {
+        "file_path": "/opt/confidential-containers/attestation-service/reference_values"
+    }
+}
+```
+- `store_type`: backend storage type to store reference values. Currently `LocalFs` and `LocalJson` are supported.
+- `store_config`: extra parameters for different kinds of `store_type`. This is also a JSON map object. The concrete content is different due to different `store_type`.
```
mkulke, line 95:
```suggestion
- `store_config`: optional extra parameters for different kinds of `store_type`. This is also a JSON map object. The concrete content is different due to different `store_type`.
```

attestation-service/attestation-service/src/lib.rs
Xynnn007: I imported this rom your PR. Do you want to keep this? I think it's ussful for some debug cases so leave it as-is.

mkulke: sure, we can leave it

attestation-service/attestation-service/src/rvps/mod.rs
Xynnn007: Yes. I am refactoring this and try to avoid `unwrap`s

attestation-service/rvps/README.md
Xynnn007: Thanks. Fixed

attestation-service/rvps/src/bin/rvps.rs
```diff
@@ -41,9 +39,17 @@ async fn main() -> Result<()> {
     info!("CoCo RVPS: {version}");
 
     let cli = Cli::parse();
+    let config = Config::from_file(&cli.config).unwrap_or_else(|e| {
+        warn!(
+            "fail to read config from {}. Error: {e:?}. Use a default one.",
```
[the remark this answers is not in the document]
Xynnn007: Nice, fixed

attestation-service/rvps/src/store/local_json/mod.rs
Xynnn007: Ah. Yes. let me fix this.

attestation-service/rvps/src/store/local_json/mod.rs
Xynnn007: Yes. filesystem write would really bring synchronization and mutual exclusion problem. I implement this as I want it to be fully realized, for some cases users might want to use client tool to register reference values.

We surely cannot keep the file be accessed inclusively. And I do not want to make it complexer by importing Locks, do you think it would be a good option to leave it as-is without changing, but add extra note like the following?

> This is a very simple implementation. When you need to call the write interface, you need to pay attention. Its synchronization is not very good and can have unexpected behaviors. Be sure of what you are doing and then use write.

Xynnn007: But surely, out of my intention of implementing fs, if you still suggest to use lock here. I'd like to implement this : )

attestation-service/rvps/src/store/local_json/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl LocalJson {
+    pub fn new(config: Option<Value>) -> Result<Self> {
+        let config = match config {
+            Some(v) => serde_json::from_value(v)?,
+            None => Config::default(),
+        };
+        Ok(Self {
+            file_path: config.file_path.unwrap_or_else(|| {
+                warn!("json file storage backend use default path {FILE_PATH}");
```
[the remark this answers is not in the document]
Xynnn007: Fixed, thanks

attestation-service/rvps/src/store/local_json/mod.rs
mkulke: I wouldn't protect against manual external file access either, I think it's clear that doing that while the service is running might cause corruption.

but we wiould need a mutex for concurrent access in the service. those operations are not atomic.

mkulke (approved):
lgtm

attestation-service/rvps/src/config.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2022 Alibaba Cloud
```
fitzthum, line 1: nit: `2022` -> `2024`

fitzthum (approved):
LGTM

One tiny nit

Issue #342, opened 2024-03-01
Attestation-Service: Add IBM Secure Execution verifier driver


huoqifeng:
Corresponding to the s390x attester driver in guest-component tracked in issue in [guest-components](https://github.com/confidential-containers/guest-components/issues/485), this issue will be used to add the s390x verifier driver in Attestation-Service to verify the s390x evidence.

Issue #242, opened 2023-11-27
KBS (protocol) enhancements to reportdata generation


mythi:
The original case was reported in #162 triggered by my work on #159 and a bit of #151 too. This issue is an RFC proposal based on ideas I've prepared with @Xynnn007 to get CoCo attesters to generate quotes/reportdata in a generic way so that they can be made consumable by different AS backends.

The source of the problem is: `hash(nonce || runtime data)` and how to keep the attester vs verifier in sync. Current CoCo setup uses `sha384` hash for all TEEs and the nonce is from the KBS session. This fails with Intel trust authority that expects `sha256` and `sha512` for SGX and TDX, respectively. IOW: CoCo attester generated evidence cannot be verified by Intel Trustauthority.

For this to work with non-CoCo attestation-services, two problems need to be sorted out:

1. it must be possible to seed `nonce` from the Attestation service
2. it must be possible to specify hash algorithm to be used by the attester.

This focuses on 2.

# Current RCAR handshake protocol

1. __Request__ from Client to Server
```json
{
    "version": "0.1.0",
    "tee": "$tee",
    "extra-params": {}
}
```

2. __Challenge__ from Server to Client: 
```json
{
    "nonce": "$nonce",
    "extra-params": {}
}
```

3. __Attestation__ from Client to Server
```json
{
    "tee-pubkey": "$pubkey",
    "tee-evidence": {}
}
```

4. __Response__ from Server to Client: 
```json
{
    "protected": "$jose_header",
    "encrypted_key": "$encrypted_key",
    "iv": "$iv",
    "ciphertext": "$ciphertext",
    "tag": "$tag"
}
```

# Suggested protocol changes

1. add an extra param "supported-hash-algorithms" to __Request__. This
helps the client to tell the server what hash algorithms are supported.

```json
{
    "version": "0.1.0",
    "tee": "$tee",
    "extra-params": {
        "supported-hash-algorithms": ["sha256", "sha384", "sha512"]
    }
}
```

2. add an extra param "selected-hash-algorithm" to __Challenge__. This
tells the hash algorithm that the server selects from the ones that
client supports.

```json
{
    "nonce": "$nonce",
    "extra-params": {
        "selected-hash-algorithm": "sha512"
    }
}
```

In this way, the KBS could select the hash algorithm depending on the TEE type and attestation service used.

For forward compatibility, the server should NOT response a __Challenge__ with
[...]

PR #345, opened 2024-03-06
Verifier: Add IBM Secure Execution driver framework

huoqifeng:
Fixes: #342 
This is kbs side code and related with PR: https://github.com/confidential-containers/guest-components/pull/492/
Depends on: 
 - [x] https://github.com/virtee/kbs-types/issues/26

# The IBM SE Remote Attestation flow:
image

- The verifier generate the encrypted attestation-request based on hkd, CA, signing_key, a measurement key  and a nonce, the encrypted data is protected by a symmetric attestation request protection key,  which is encrypted using the Host-key document
- Verifier sends the request to attester
- Firmware on the Attester's system decrypts the request via private host-key and calculates the evidence based on the encrpted part of the request (Measurement key + nonce)
- Attester send the evidence to verifier
- Verifier recalculates the evidence based on the Configuration UID, Additional data, user-data, guest image hashes, and nonce. (its a HMAC-SHA512 with the measurement key as secret)
- if both HMACs, the one from the Firmware and the calculated one from the verifier match -> attestation success

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ regorus = { version = "0.1.5", default-features = false, features = ["regex", "b
 rstest = "0.18.1"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.89"
-serde_with = { version = "1.11.0", features = ["base64"] }
+serde_with = { version = "1.11.0", features = ["base64", "hex"] }
 serial_test = "0.9.0"
 sha2 = "0.10"
 shadow-rs = "0.19.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ clap = { version = "4", features = ["derive"] }
 config = "0.13.3"
 env_logger = "0.10.0"
 hex = "0.4.3"
-kbs-types = "0.5.3"
+kbs-types = "0.6.0"
 jsonwebtoken = { version = "9", default-features = false }
 log = "0.4.17"
 prost = "0.11.0"
@@ -38,6 +38,7 @@ regorus = { version = "0.1.5", default-features = false, features = ["regex", "b
 rstest = "0.18.1"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.89"
+serde_with = { version = "1.11.0", features = ["base64"] }
 serial_test = "0.9.0"
 sha2 = "0.10"
 shadow-rs = "0.19.0"
--- a/attestation-service/attestation-service/Cargo.toml
+++ b/attestation-service/attestation-service/Cargo.toml
@@ -13,6 +13,7 @@ az-tdx-vtpm-verifier = [ "verifier/az-tdx-vtpm-verifier" ]
 snp-verifier = [ "verifier/snp-verifier" ]
 csv-verifier = [ "verifier/csv-verifier" ]
 cca-verifier = [ "verifier/cca-verifier" ]
+se-verifier  = [ "verifier/se-verifier" ]
 
 # Only for testing and CI
 rvps-builtin = [ "reference-value-provider-service" ]
--- a/attestation-service/attestation-service/src/bin/grpc/mod.rs
+++ b/attestation-service/attestation-service/src/bin/grpc/mod.rs
@@ -15,7 +15,10 @@ use tonic::transport::Server;
 use tonic::{Request, Response, Status};
 
 use crate::as_api::attestation_service_server::{AttestationService, AttestationServiceServer};
-use crate::as_api::{AttestationRequest, AttestationResponse, SetPolicyRequest, SetPolicyResponse};
+use crate::as_api::{
+    AttestationRequest, AttestationResponse, ChallengeRequest, ChallengeResponse, SetPolicyRequest,
+    SetPolicyResponse,
+};
 
 use crate::rvps_api::reference_value_provider_service_server::{
     ReferenceValueProviderService, ReferenceValueProviderServiceServer,
@@ -37,6 +40,7 @@ fn to_kbs_tee(tee: &str) -> anyhow::Result<Tee> {
         "azsnpvtpm" => Tee::AzSnpVtpm,
         "cca" => Tee::Cca,
         "aztdxvtpm" => Tee::AzTdxVtpm,
+        "se" => Tee::Se,
         other => bail!("Unsupported TEE type: {other}"),
     };
 
@@ -193,6 +197,39 @@ impl AttestationService for Arc<RwLock<AttestationServer>> {
         debug!("Attestation Token: {}", &attestation_token);
 
         let res = AttestationResponse { attestation_token };
+        Ok(Response::new(res))
+    }
+
+    async fn get_attestation_challenge(
+        &self,
+        request: Request<ChallengeRequest>,
+    ) -> Result<Response<ChallengeResponse>, Status> {
+        let request: ChallengeRequest = request.into_inner();
+        info!("get_attestation_challenge API called.");
+        debug!("get_attestation_challenge: {request:#?}");
+
+        let inner_tee = request
+            .inner
+            .get("tee")
+            .ok_or(Status::aborted("Error parse inner_tee tee"))?;
+        let tee_params = request
+            .inner
+            .get("tee_params")
+            .map_or(Err(Status::aborted("Error parse inner_tee tee_params")), Ok)?;
+        let tee = to_kbs_tee(&inner_tee)
+            .map_err(|e| Status::aborted(format!("Error parse TEE type: {e}")))?;
+
+        let attestation_challenge = self
+            .read()
+            .await
+            .attestation_service
+            .generate_supplemental_challenge(tee, tee_params.clone())
+            .await
+            .map_err(|e| Status::aborted(format!("Challenge: {e:?}")))?;
+
+        let res = ChallengeResponse {
+            attestation_challenge,
+        };
         Ok(Response::new(res))
     }
 }
--- a/attestation-service/attestation-service/src/bin/restful/mod.rs
+++ b/attestation-service/attestation-service/src/bin/restful/mod.rs
@@ -1,7 +1,7 @@
-use std::sync::Arc;
+use std::{collections::HashMap, sync::Arc};
 
 use actix_web::{body::BoxBody, web, HttpRequest, HttpResponse, ResponseError};
-use anyhow::{bail, Context};
+use anyhow::{anyhow, bail, Context};
 use attestation_service::{AttestationService, HashAlgorithm};
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use kbs_types::Tee;
@@ -45,6 +45,14 @@ pub struct AttestationRequest {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
+pub struct ChallengeRequest {
+    // ChallengeRequest uses HashMap to pass variables like:
+    // tee, tee_params etc
+    #[serde(flatten)]
+    inner: HashMap<String, String>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case")]
 enum Data {
     Raw(String),
@@ -62,6 +70,7 @@ fn to_tee(tee: &str) -> anyhow::Result<Tee> {
         "csv" => Tee::Csv,
         "sample" => Tee::Sample,
         "aztdxvtpm" => Tee::AzTdxVtpm,
+        "se" => Tee::Se,
         other => bail!("tee `{other} not supported`"),
     };
 
@@ -177,6 +186,36 @@ pub async fn set_policy(
         .context("set policy")?;
 
     Ok(HttpResponse::Ok().body(""))
+}
+
+/// This handler uses json extractor
+pub async fn get_challenge(
+    request: web::Json<ChallengeRequest>,
+    cocoas: web::Data<Arc<RwLock<AttestationService>>>,
+) -> Result<HttpResponse> {
+    info!("get_challenge API called.");
+    let request: ChallengeRequest = request.into_inner();
+
+    debug!("get_challenge: {request:#?}");
+    let inner_tee = request
+        .inner
+        .get("tee")
+        .as_ref()
+        .map(|s| s.as_str())
+        .ok_or(anyhow!("Failed to get inner tee"))?;
+    let tee_params = request
+        .inner
+        .get("tee_params")
+        .ok_or(anyhow!("Failed to get inner tee_params"))?;
+
+    let tee = to_tee(inner_tee)?;
+    let challenge = cocoas
+        .read()
+        .await
+        .generate_supplemental_challenge(tee, tee_params.to_string())
+        .await
+        .context("generate challenge")?;
+    Ok(HttpResponse::Ok().body(challenge))
 }
 
 /// GET /policy
--- a/attestation-service/attestation-service/src/lib.rs
+++ b/attestation-service/attestation-service/src/lib.rs
@@ -274,6 +274,17 @@ impl AttestationService {
     pub async fn register_reference_value(&mut self, message: &str) -> Result<()> {
         self.rvps.verify_and_extract(message).await
     }
+
+    pub async fn generate_supplemental_challenge(
+        &self,
+        tee: Tee,
+        tee_parameters: String,
+    ) -> Result<String> {
+        let verifier = verifier::to_verifier(&tee)?;
+        verifier
+            .generate_supplemental_challenge(tee_parameters)
+            .await
+    }
 }
 
 /// Get the expected init/runtime data and potential claims due to the given input
--- a/attestation-service/docs/parsed_claims.md
+++ b/attestation-service/docs/parsed_claims.md
@@ -93,6 +93,14 @@ The claim inherit the fields from the SEV-SNP claim with and additional `tpm` hi
 
 Note: The TD Report and TD Quote are fetched during early boot in this TEE. Kernel, Initrd and rootfs are measured into the vTPM's registers.
 
+## IBM Secure Execution (SE)
+- `se.version`: The version this quote structure.
+- `se.cuid`: The config uid.
+- `se.hdr.tag`: SE header tag (seht)
+- `se.image.phkh`: SE image public host key hash
+- `se.attestation.phkh`: SE attestation public host key hash
+- `se.user_data`: Custom attestation key owner data.
+
 ## AMD SEV-SNP
 
 - `snp.measurement` Launch Digest covering initial guest memory
--- a/attestation-service/docs/parsed_claims.md
+++ b/attestation-service/docs/parsed_claims.md
@@ -95,11 +95,11 @@ Note: The TD Report and TD Quote are fetched during early boot in this TEE. Kern
 
 ## IBM Secure Execution (SE)
 - `se.version`: The version this quote structure.
-- `se.cuid`: The config uid.
-- `se.hdr.tag`: SE header tag (seht)
+- `se.cuid`: The unique ID of the attested guest (configuration uniqe ID).
+- `se.hdr.tag`: SE header tag.
 - `se.image.phkh`: SE image public host key hash
 - `se.attestation.phkh`: SE attestation public host key hash
-- `se.user_data`: Custom attestation key owner data.
+- `se.user_data`: Optional custom attestation owner data, could be key:value pairs collected on guest.
 
 ## AMD SEV-SNP
 
--- a/attestation-service/protos/attestation.proto
+++ b/attestation-service/protos/attestation.proto
@@ -77,8 +77,18 @@ message SetPolicyRequest {
 }
 message SetPolicyResponse {}
 
+message ChallengeRequest {
+    // ChallengeRequest uses HashMap to pass variables like:
+    // tee, tee_params etc
+    map<string, string> inner = 1;
+}
+message ChallengeResponse {
+    string attestation_challenge = 1;
+}
+
 service AttestationService {
     rpc AttestationEvaluate(AttestationRequest) returns (AttestationResponse) {};
     rpc SetAttestationPolicy(SetPolicyRequest) returns (SetPolicyResponse) {};
+    rpc GetAttestationChallenge(ChallengeRequest) returns (ChallengeResponse) {};
     // Get the GetPolicyRequest.user and GetPolicyRequest.tee specified Policy(.rego)
 }
--- a/attestation-service/verifier/src/lib.rs
+++ b/attestation-service/verifier/src/lib.rs
@@ -28,6 +28,9 @@ pub mod csv;
 #[cfg(feature = "cca-verifier")]
 pub mod cca;
 
+#[cfg(feature = "se-verifier")]
+pub mod se;
+
 pub fn to_verifier(tee: &Tee) -> Result<Box<dyn Verifier + Send + Sync>> {
     match tee {
         Tee::Sev => todo!(),
@@ -99,6 +102,16 @@ pub fn to_verifier(tee: &Tee) -> Result<Box<dyn Verifier + Send + Sync>> {
                 }
             }
         }
+
+        Tee::Se => {
+            cfg_if::cfg_if! {
+                if #[cfg(feature = "se-verifier")] {
+                    Ok(Box::<se::SeVerifier>::default() as Box<dyn Verifier + Send + Sync>)
+                } else {
+                    bail!("feature `se-verifier` is not enabled for `verifier` crate.")
+                }
+            }
+        }
     }
 }
 
@@ -152,6 +165,17 @@ pub trait Verifier {
         expected_report_data: &ReportData,
         expected_init_data_hash: &InitDataHash,
     ) -> Result<TeeEvidenceParsedClaim>;
+
+    /// Generate the supplemental challenge
+    ///
+    /// Some TEE like IBM SE need a `challenge` generated on verifier side
+    /// and pass it to attester side. This challenge is used by attester to
+    /// generate the evidence
+    ///
+    /// A optional `tee_parameters` comes from the attester side as the input.
+    async fn generate_supplemental_challenge(&self, _tee_parameters: String) -> Result<String> {
+        Ok(String::new())
+    }
 }
 
 /// Padding or truncate the given data slice to the given `len` bytes.
--- a/attestation-service/verifier/src/se/ibmse.rs
+++ b/attestation-service/verifier/src/se/ibmse.rs
@@ -7,9 +7,10 @@ use crate::TeeEvidenceParsedClaim;
 use anyhow::{anyhow, bail, Context, Result};
 use core::result::Result::Ok;
 use log::{debug, info, warn};
+use openssl::ec::EcKey;
 use openssl::encrypt::{Decrypter, Encrypter};
 use openssl::pkey::{PKey, Private, Public};
-use openssl::rsa::{Padding, Rsa};
+use openssl::rsa::Padding;
 use pv::attest::{
     AdditionalData, AttestationFlags, AttestationItems, AttestationMeasAlg, AttestationMeasurement,
     AttestationRequest, AttestationVersion,
@@ -33,10 +34,10 @@ const DEFAULT_SE_CERTIFICATE_REVOCATION_LISTS_ROOT: &str =
 const DEFAULT_SE_IMAGE_HEADER_FILE: &str = "/run/confidential-containers/ibmse/hdr/hdr.bin";
 
 const DEFAULT_SE_MEASUREMENT_ENCR_KEY_PRIVATE: &str =
-    "/run/confidential-containers/ibmse/rsa/encrypt_key.pem";
+    "/run/confidential-containers/ibmse/ec/encrypt_key.pem";
 
 const DEFAULT_SE_MEASUREMENT_ENCR_KEY_PUBLIC: &str =
-    "/run/confidential-containers/ibmse/rsa/encrypt_key.pub";
+    "/run/confidential-containers/ibmse/ec/encrypt_key.pub";
 
 macro_rules! env_or_default {
     ($env:literal, $default:ident) => {
@@ -51,9 +52,7 @@ fn list_files_in_folder(dir: &str) -> Result<Vec<String>> {
     let mut file_paths = Vec::new();
 
     for entry in fs::read_dir(dir)? {
-        let entry = entry?;
-        let path = entry.path();
-
+        let path = entry?.path();
         if path.is_file() {
             if let Some(path_str) = path.to_str() {
                 file_paths.push(path_str.to_string());
@@ -64,6 +63,7 @@ fn list_files_in_folder(dir: &str) -> Result<Vec<String>> {
     Ok(file_paths)
 }
 
+#[repr(C)]
 #[serde_as]
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SeAttestationResponse {
@@ -100,6 +100,7 @@ pub struct SeAttestationClaims {
     tag: [u8; 16],
 }
 
+#[repr(C)]
 #[serde_as]
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SeAttestationRequest {
@@ -116,57 +117,57 @@ pub struct SeAttestationRequest {
 }
 
 #[derive(Debug)]
-pub struct RealSeVerifier {
-    rsa_private_key: PKey<Private>,
-    rsa_public_key: PKey<Public>,
+pub struct SeVerifierImpl {
+    private_key: PKey<Private>,
+    public_key: PKey<Public>,
 }
 
-impl RealSeVerifier {
+impl SeVerifierImpl {
     pub fn new() -> Result<Self> {
         let pri_key_file = env_or_default!(
             "SE_MEASUREMENT_ENCR_KEY_PRIVATE",
             DEFAULT_SE_MEASUREMENT_ENCR_KEY_PRIVATE
         );
         let priv_contents = fs::read(pri_key_file)?;
-        let rsa_private_key = Rsa::private_key_from_pem(&priv_contents)?;
-        let rsa_private_key = PKey::from_rsa(rsa_private_key)?;
+        let private_key = EcKey::private_key_from_pem(&priv_contents)?;
+        let private_key = PKey::from_ec_key(private_key)?;
 
         let pub_key_file = env_or_default!(
             "SE_MEASUREMENT_ENCR_KEY_PUBLIC",
             DEFAULT_SE_MEASUREMENT_ENCR_KEY_PUBLIC
         );
         let pub_contents = fs::read(pub_key_file)?;
-        let rsa = Rsa::public_key_from_pem(&pub_contents)?;
-        let rsa_public_key = PKey::from_rsa(rsa)?;
+        let rsa = EcKey::public_key_from_pem(&pub_contents)?;
+        let public_key = PKey::from_ec_key(rsa)?;
 
         Ok(Self {
-            rsa_private_key,
-            rsa_public_key,
+            private_key,
+            public_key,
         })
     }
 
     fn decrypt(&self, ciphertext: &[u8]) -> Result<Vec<u8>> {
-        let mut decrypter = Decrypter::new(&self.rsa_private_key)?;
+        let mut decrypter = Decrypter::new(&self.private_key)?;
         decrypter.set_rsa_padding(Padding::PKCS1)?;
 
         let buffer_len = decrypter.decrypt_len(ciphertext)?;
-        let mut decrypted_hmac_key = vec![0; buffer_len];
-        let decrypted_len = decrypter.decrypt(ciphertext, &mut decrypted_hmac_key)?;
-        decrypted_hmac_key.truncate(decrypted_len);
+        let mut decrypted = vec![0; buffer_len];
+        let decrypted_len = decrypter.decrypt(ciphertext, &mut decrypted)?;
+        decrypted.truncate(decrypted_len);
 
-        Ok(decrypted_hmac_key)
+        Ok(decrypted)
     }
 
     fn encrypt(&self, text: &[u8]) -> Result<Vec<u8>> {
-        let mut encrypter = Encrypter::new(&self.rsa_public_key)?;
+        let mut encrypter = Encrypter::new(&self.public_key)?;
         encrypter.set_rsa_padding(Padding::PKCS1)?;
     
         let buffer_len = encrypter.encrypt_len(text)?;
-        let mut encrypted_hmac_key = vec![0; buffer_len];
-        let len = encrypter.encrypt(text, &mut encrypted_hmac_key)?;
-        encrypted_hmac_key.truncate(len);
+        let mut encrypted = vec![0; buffer_len];
+        let len = encrypter.encrypt(text, &mut encrypted)?;
+        encrypted.truncate(len);
     
-        Ok(encrypted_hmac_key)
+        Ok(encrypted)
     }
 
     pub fn evaluate(&self, evidence: &[u8]) -> Result<TeeEvidenceParsedClaim> {
@@ -182,13 +183,9 @@ impl RealSeVerifier {
             .decrypt(&se_response.encr_request_nonce)
             .context("decrypt Request Nonce")?;
 
-        if nonce.len() != 16 {
-            bail!("The nonce vector must have exactly 16 elements.");
-        }
-
         let nonce_array: [u8; 16] = nonce
             .try_into()
-            .map_err(|_| anyhow!("Failed to convert nonce from Vec<u8> to [u8; 16]."))?;
+            .map_err(|_| anyhow!("Failed to convert nonce from Vec<u8> to [u8; 16], It must have exactly 16 elements."))?;
 
         let meas_key = PKey::hmac(&meas_key)?;
         let items = AttestationItems::new(
@@ -205,11 +202,8 @@ impl RealSeVerifier {
         if !measurement.eq_secure(&se_response.measurement) {
             debug!("Recieved: {:?}", se_response.measurement);
             debug!("Calculated: {:?}", measurement.as_ref());
-            warn!("Attestation measurement verification failed. Calculated and received attestation measurement are not equal.");
             bail!("Failed to verify the measurement!");
         }
-
-        // TODO check self.user_data.image_btph with previous saved value
 
         let mut att_flags = AttestationFlags::default();
         att_flags.set_image_phkh();
@@ -222,8 +216,6 @@ impl RealSeVerifier {
         let attestation_phkh = add_data
             .attestation_public_host_key_hash()
             .ok_or(anyhow!("Failed to get attestation_public_host_key_hash."))?;
-
-        // TODO image_phkh and attestation_phkh with previous saved value
 
         let claims = SeAttestationClaims {
             cuid: se_response.cuid,
@@ -253,10 +245,13 @@ impl RealSeVerifier {
         let verifier =
             CertVerifier::new(certs.as_slice(), crls.as_slice(), Some(root_ca_path), false)?;
 
+        let mut attestation_flags = AttestationFlags::default();
+        attestation_flags.set_image_phkh();
+        attestation_flags.set_attest_phkh();
         let mut arcb = AttestationRequest::new(
             AttestationVersion::One,
             AttestationMeasAlg::HmacSha512,
-            AttestationFlags::default(),
+            attestation_flags,
         )?;
 
         let hkds_root = env_or_default!(
--- a/attestation-service/verifier/src/se/ibmse.rs
+++ b/attestation-service/verifier/src/se/ibmse.rs
@@ -18,7 +18,7 @@ use pv::misc::{open_file, read_certs};
 use pv::request::{BootHdrTags, CertVerifier, HkdVerifier, ReqEncrCtx, Request, SymKeyType};
 use pv::uv::ConfigUid;
 use serde::{Deserialize, Serialize};
-use serde_with::{base64::Base64, serde_as};
+use serde_with::{base64::Base64, hex::Hex, serde_as};
 use std::{env, fs};
 
 const DEFAULT_SE_HOST_KEY_DOCUMENTS_ROOT: &str = "/run/confidential-containers/ibmse/hkds";
@@ -87,16 +87,16 @@ pub struct SeAttestationResponse {
 #[serde_as]
 #[derive(Debug, Serialize, Deserialize)]
 pub struct SeAttestationClaims {
-    #[serde_as(as = "Base64")]
+    #[serde_as(as = "Hex")]
     cuid: ConfigUid,
-    #[serde_as(as = "Base64")]
+    #[serde_as(as = "Hex")]
     user_data: Vec<u8>,
     version: u32,
-    #[serde_as(as = "Base64")]
+    #[serde_as(as = "Hex")]
     image_phkh: Vec<u8>,
-    #[serde_as(as = "Base64")]
+    #[serde_as(as = "Hex")]
     attestation_phkh: Vec<u8>,
-    #[serde_as(as = "Base64")]
+    #[serde_as(as = "Hex")]
     tag: [u8; 16],
 }
 
--- a/attestation-service/verifier/src/se/ibmse.rs
+++ b/attestation-service/verifier/src/se/ibmse.rs
@@ -1,0 +1,311 @@
+// Copyright (C) Copyright IBM Corp. 2024
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use crate::TeeEvidenceParsedClaim;
+use anyhow::{anyhow, bail, Context, Result};
+use core::result::Result::Ok;
+use log::{debug, info, warn};
+use openssl::encrypt::{Decrypter, Encrypter};
+use openssl::pkey::{PKey, Private, Public};
+use openssl::rsa::{Padding, Rsa};
+use pv::attest::{
+    AdditionalData, AttestationFlags, AttestationItems, AttestationMeasAlg, AttestationMeasurement,
+    AttestationRequest, AttestationVersion,
+};
+use pv::misc::{open_file, read_certs};
+use pv::request::{BootHdrTags, CertVerifier, HkdVerifier, ReqEncrCtx, Request, SymKeyType};
+use pv::uv::ConfigUid;
+use serde::{Deserialize, Serialize};
+use serde_with::{base64::Base64, serde_as};
+use std::{env, fs};
+
+const DEFAULT_SE_HOST_KEY_DOCUMENTS_ROOT: &str = "/run/confidential-containers/ibmse/hkds";
+
+const DEFAULT_SE_CERTIFICATES_ROOT: &str = "/run/confidential-containers/ibmse/certs";
+
+const DEFAULT_SE_CERTIFICATE_ROOT_CA: &str = "/run/confidential-containers/ibmse/certs/ca";
+
+const DEFAULT_SE_CERTIFICATE_REVOCATION_LISTS_ROOT: &str =
+    "/run/confidential-containers/ibmse/crls";
+
+const DEFAULT_SE_IMAGE_HEADER_FILE: &str = "/run/confidential-containers/ibmse/hdr/hdr.bin";
+
+const DEFAULT_SE_MEASUREMENT_ENCR_KEY_PRIVATE: &str =
+    "/run/confidential-containers/ibmse/rsa/encrypt_key.pem";
+
+const DEFAULT_SE_MEASUREMENT_ENCR_KEY_PUBLIC: &str =
+    "/run/confidential-containers/ibmse/rsa/encrypt_key.pub";
+
+macro_rules! env_or_default {
+    ($env:literal, $default:ident) => {
+        match env::var($env) {
+            Ok(env_path) => env_path,
+            Err(_) => $default.into(),
+        }
+    };
+}
+
+fn list_files_in_folder(dir: &str) -> Result<Vec<String>> {
+    let mut file_paths = Vec::new();
+
+    for entry in fs::read_dir(dir)? {
+        let entry = entry?;
+        let path = entry.path();
+
+        if path.is_file() {
+            if let Some(path_str) = path.to_str() {
+                file_paths.push(path_str.to_string());
+            }
+        }
+    }
+
+    Ok(file_paths)
+}
+
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SeAttestationResponse {
+    #[serde_as(as = "Base64")]
+    measurement: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    additional_data: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    user_data: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    cuid: ConfigUid,
+    #[serde_as(as = "Base64")]
+    encr_measurement_key: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    encr_request_nonce: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    image_hdr_tags: BootHdrTags,
+}
+
+#[repr(C)]
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SeAttestationClaims {
+    #[serde_as(as = "Base64")]
+    cuid: ConfigUid,
+    #[serde_as(as = "Base64")]
+    user_data: Vec<u8>,
+    version: u32,
+    #[serde_as(as = "Base64")]
+    image_phkh: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    attestation_phkh: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    tag: [u8; 16],
+}
+
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SeAttestationRequest {
+    #[serde_as(as = "Base64")]
+    request_blob: Vec<u8>,
+    measurement_size: u32,
+    additional_size: u32,
+    #[serde_as(as = "Base64")]
+    encr_measurement_key: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    encr_request_nonce: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    image_hdr_tags: BootHdrTags,
+}
+
+#[derive(Debug)]
+pub struct RealSeVerifier {
+    rsa_private_key: PKey<Private>,
+    rsa_public_key: PKey<Public>,
+}
+
+impl RealSeVerifier {
+    pub fn new() -> Result<Self> {
+        let pri_key_file = env_or_default!(
+            "SE_MEASUREMENT_ENCR_KEY_PRIVATE",
+            DEFAULT_SE_MEASUREMENT_ENCR_KEY_PRIVATE
+        );
+        let priv_contents = fs::read(pri_key_file)?;
+        let rsa_private_key = Rsa::private_key_from_pem(&priv_contents)?;
+        let rsa_private_key = PKey::from_rsa(rsa_private_key)?;
+
+        let pub_key_file = env_or_default!(
+            "SE_MEASUREMENT_ENCR_KEY_PUBLIC",
+            DEFAULT_SE_MEASUREMENT_ENCR_KEY_PUBLIC
+        );
+        let pub_contents = fs::read(pub_key_file)?;
+        let rsa = Rsa::public_key_from_pem(&pub_contents)?;
+        let rsa_public_key = PKey::from_rsa(rsa)?;
+
+        Ok(Self {
+            rsa_private_key,
+            rsa_public_key,
+        })
+    }
+
+    fn decrypt(&self, ciphertext: &[u8]) -> Result<Vec<u8>> {
+        let mut decrypter = Decrypter::new(&self.rsa_private_key)?;
+        decrypter.set_rsa_padding(Padding::PKCS1)?;
+
+        let buffer_len = decrypter.decrypt_len(ciphertext)?;
+        let mut decrypted_hmac_key = vec![0; buffer_len];
+        let decrypted_len = decrypter.decrypt(ciphertext, &mut decrypted_hmac_key)?;
+        decrypted_hmac_key.truncate(decrypted_len);
+
+        Ok(decrypted_hmac_key)
+    }
+
+    fn encrypt(&self, text: &[u8]) -> Result<Vec<u8>> {
+        let mut encrypter = Encrypter::new(&self.rsa_public_key)?;
+        encrypter.set_rsa_padding(Padding::PKCS1)?;
+    
+        let buffer_len = encrypter.encrypt_len(text)?;
+        let mut encrypted_hmac_key = vec![0; buffer_len];
+        let len = encrypter.encrypt(text, &mut encrypted_hmac_key)?;
+        encrypted_hmac_key.truncate(len);
+    
+        Ok(encrypted_hmac_key)
+    }
+
+    pub fn evaluate(&self, evidence: &[u8]) -> Result<TeeEvidenceParsedClaim> {
+        info!("IBM SE verify API called.");
+
+        // evidence is serialized SeAttestationResponse String bytes
+        let se_response: SeAttestationResponse = serde_json::from_slice(evidence)?;
+
+        let meas_key = self
+            .decrypt(&se_response.encr_measurement_key)
+            .context("decrypt Measurement Key")?;
+        let nonce = self
+            .decrypt(&se_response.encr_request_nonce)
+            .context("decrypt Request Nonce")?;
+
+        if nonce.len() != 16 {
+            bail!("The nonce vector must have exactly 16 elements.");
+        }
+
+        let nonce_array: [u8; 16] = nonce
+            .try_into()
+            .map_err(|_| anyhow!("Failed to convert nonce from Vec<u8> to [u8; 16]."))?;
+
+        let meas_key = PKey::hmac(&meas_key)?;
+        let items = AttestationItems::new(
+            &se_response.image_hdr_tags,
+            &se_response.cuid,
+            Some(&se_response.user_data),
+            Some(&nonce_array),
+            Some(&se_response.additional_data),
+        );
+
+        let measurement =
+            AttestationMeasurement::calculate(items, AttestationMeasAlg::HmacSha512, &meas_key)?;
+
+        if !measurement.eq_secure(&se_response.measurement) {
+            debug!("Recieved: {:?}", se_response.measurement);
+            debug!("Calculated: {:?}", measurement.as_ref());
+            warn!("Attestation measurement verification failed. Calculated and received attestation measurement are not equal.");
+            bail!("Failed to verify the measurement!");
+        }
+
+        // TODO check self.user_data.image_btph with previous saved value
+
+        let mut att_flags = AttestationFlags::default();
+        att_flags.set_image_phkh();
+        att_flags.set_attest_phkh();
+        let add_data = AdditionalData::from_slice(&se_response.additional_data, &att_flags)?;
+        debug!("additional_data: {:?}", add_data);
+        let image_phkh = add_data
+            .image_public_host_key_hash()
+            .ok_or(anyhow!("Failed to get image_public_host_key_hash."))?;
+        let attestation_phkh = add_data
+            .attestation_public_host_key_hash()
+            .ok_or(anyhow!("Failed to get attestation_public_host_key_hash."))?;
+
+        // TODO image_phkh and attestation_phkh with previous saved value
+
+        let claims = SeAttestationClaims {
+            cuid: se_response.cuid,
+            user_data: se_response.user_data.clone(),
+            version: AttestationVersion::One as u32,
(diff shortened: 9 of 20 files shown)


kbs/src/api/src/attestation/mod.rs
```diff
@@ -32,6 +32,9 @@ pub trait Attest: Send + Sync {
     /// Verify Attestation Evidence
     /// Return Attestation Results Token
     async fn verify(&self, tee: Tee, nonce: &str, attestation: &str) -> Result<String>;
+
+    /// generate the challenge payload to pass to attester based on Tee and nonce
+    async fn generate_challenge(&self, tee: Tee, nonce: &str) -> Result<Challenge>;
```
Xynnn007, line 37: Hi @huoqifeng , currently we will get the challenge nonce from KBS side [here](https://github.com/confidential-containers/trustee/blob/main/kbs/src/api/src/session.rs#L78). 

And we have decoupled the code. Now the challenge will be check binding by calling Attestation Service's [reportdata/runtimedata api parameter](https://github.com/confidential-containers/trustee/blob/main/kbs/src/api/src/attestation/coco/grpc.rs#L104C9-L104C100)

So I think what is needed to be done on the verifier side is only (If SE is similiar with TDX/SNP/CCA ...)
1. implement a verifier plugin of `attester`
2. add se to https://github.com/confidential-containers/trustee/blob/main/attestation-service/attestation-service/src/bin/grpc/mod.rs#L29-L41, https://github.com/confidential-containers/trustee/blob/main/attestation-service/attestation-service/src/bin/restful/mod.rs#L54-L69, https://github.com/confidential-containers/trustee/blob/main/attestation-service/protos/attestation.proto#L5-L15
3. Document update
(I might ignore something else that could be figured out due to CI error)

huoqifeng: @Xynnn007 thanks for the good suggestions, there are some specific requirements for IBM SE, like:
- On kbs side, the `challenge` need include an `attestation request`.
like https://github.com/confidential-containers/trustee/blob/main/kbs/src/api/src/session.rs#L78, IBM SE has an `attestation request` need be generated by `verifier`, this request `must` be generated on verifier and pass to attester. We can use the `nonce` field but the meaning has been changed from a random str to something generated from verifier driver, alternative as I made in this draft PR, we can also use field `extra_params`, anyway, IBM SE need pass `attestation request` generated by verifier but not only random string.
- On guest side, this `attestation request` will be used to issue evidence
Like in https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/kbs_protocol/src/client/rcar_client.rs#L184, we need `runtime_data` can be deserialized in `attester` driver, because IBM SE attester requires the `attestation request` passed from verifier side to generate the evidence. 

Lets talk about it this afternoon and make it clear. Thanks!

huoqifeng: @Xynnn007 https://github.com/ibm-s390-linux/s390-tools/tree/master/pvattest, this is a reference for c approach, rust is in-developing.

huoqifeng:
@Xynnn007 @fitzthum I'd like start the review process of the API change to support IBM SE. May you please help on that?

Xynnn007:
Oh. I am still worried about the breakage for the PR. As it couple the KBS logic and AS logic to prepare an `attreq.bin`. After some reading of https://www.ibm.com/docs/en/linuxonibm/pdf/l130se03.pdf, I think of some different design ways for this and I must ignore something.

#### Way 1: 

Con: We need to embed the host cert key chain (`host key doc`, `CA cert`, `IBM_sign_key_cert`) into the guest image

The logic of `attester` side:
1. After receiving `challenge`, generate an `arp.key` by `HKDF(challenge || host_key_doc || CA cert || IBM_sign_key_cert)`
2. Generate a report using the `arp.key` and other materials using `pvattest`

The logic of `verifier` side:
1. Use the same `host key doc`, `CA cert`, `IBM_sign_key_cert` and `challenge` to generate a same `arp.key`, and verify with `pvattest verify -i attresp.bin --arpk arp.key --hdr hdr.bin`

Rationality:
Looking at the entire attestation protocol of pvattest, the most critical thing is this `arp.key`, which is used for anti-replay, similar to the Nonce of other architectures, so we combine platform information `host_key_doc || intermediate CA certificate || IBM sign key cert` together with `challenge`.  The strong binding information is used as the derivation factor of this key to ensure that the attester cannot use an instance on another platform to forge evidence that can pass verification - because the `arp.key` of the verifier side must be based on the certificate of the target environment.

BTW, if SE guest could support initdata (either though `hostdata` mechanism or vTPM), we could inject those documents when launching the guest without embed them inside the guest.

#### Way 2: Change the `verifier` API

This is the current implementation.

Con: Bring extra `generate_challenge_extra_params()` for all verifiers and might couple the current design of `KBS` and `AS`.

This means every time KBS receives a RCAR request, it should first call the backend AS to get an `extra_params`. This API is somehow specific for CoCo hinting that Intel-TA and other attestation services might not support such.

#### Way 3: Integrite IBM SE verifier logic inside KBS, and attester logic in KBS-client

Con: No `se` plugin will be support in `attester` and `verifier` crate, and thus not be support by `AS`. `KBS` should store `arp.key`s and `se_guest.hdr`s

This is based on a different attestation protocol of SE. We can introduce some conditional judgement for SE on the both KBS and KBS client side.

#### Way 4: Integrite IBM SE verifier logic inside KBS but attester logic in `attester` crate

Con: `KBS` should store `arp.key`s and `se_guest.hdr`s.

On the attester side, we can reuse `report_data/challenge` field for SE to present the `host key doc`, `CA cert`, `IBM_sign_key_cert` and `attreq.bin` bundle and generate an `attresp.bin`

On the verifier side, `KBS` will implement the generation of `attreq.bin`, and the verification of `attresp.bin`.

huoqifeng:
> Oh. I am still worried about the breakage for the PR. As it couple the KBS logic and AS logic to prepare an `attreq.bin`. After some reading of https://www.ibm.com/docs/en/linuxonibm/pdf/l130se03.pdf, I think of some different design ways for this and I must ignore something.
> #### Way 1:
> 
> Con: We need to embed the host cert key chain (`host key doc`, `CA cert`, `IBM_sign_key_cert`) into the guest image
> 
> The logic of `attester` side:
> 
>     1. After receiving `challenge`, generate an `arp.key` by `HKDF(challenge || host_key_doc || CA cert || IBM_sign_key_cert)`
> 
>     2. Generate a report using the `arp.key` and other materials using `pvattest`
> 
> 
> The logic of `verifier` side:
> 
>     1. Use the same `host key doc`, `CA cert`, `IBM_sign_key_cert` and `challenge` to generate a same `arp.key`, and verify with `pvattest verify -i attresp.bin --arpk arp.key --hdr hdr.bin`
> 
> 
> Rationality: Looking at the entire attestation protocol of pvattest, the most critical thing is this `arp.key`, which is used for anti-replay, similar to the Nonce of other architectures, so we combine platform information `host_key_doc || intermediate CA certificate || IBM sign key cert` together with `challenge`. The strong binding information is used as the derivation factor of this key to ensure that the attester cannot use an instance on another platform to forge evidence that can pass verification - because the `arp.key` of the verifier side must be based on the certificate of the target environment.
> 
> BTW, if SE guest could support initdata (either though `hostdata` mechanism or vTPM), we could inject those documents when launching the guest without embed them inside the guest.
> #### Way 2: Add a `get_challenge` to the `verifier` API
> 
> This is the current implementation.
> 
> Con: `get_challenge()` might mean nothing to other platforms.
> 
> This means every time KBS receives a RCAR request, it should first call the backend AS to call `get_challenge()`.
> 
> This hints that CoCoAS should also support 4-pass protocol besides current 2-pass protocol, s.t. two ways to use CoCoAS
> 
>     1. `get_challenge()` and then [evaluate()](https://github.com/confidential-containers/trustee/blob/main/attestation-service/protos/attestation.proto#L92)
> 
>     2. directly [evaluate()](https://github.com/confidential-containers/trustee/blob/main/attestation-service/protos/attestation.proto#L92)
> 
> 
> cc @jialez0
> #### Way 3: Integrite IBM SE verifier logic inside KBS, and attester logic in KBS-client
> 
> Con: No `se` plugin will be support in `attester` and `verifier` crate, and thus not be support by `AS`. `KBS` should store `arp.key`s and `se_guest.hdr`s
> 
> This is based on a different attestation protocol of SE. We can introduce some conditional judgement for SE on the both KBS and KBS client side.
> #### Way 4: Integrite IBM SE verifier logic inside KBS but attester logic in `attester` crate
> 
> Con: `KBS` should store `arp.key`s and `se_guest.hdr`s.
> 
> On the attester side, we can reuse `report_data/challenge` field for SE to present the `host key doc`, `CA cert`, `IBM_sign_key_cert` and `attreq.bin` bundle and generate an `attresp.bin`
> 
> On the verifier side, `KBS` will implement the generation of `attreq.bin`, and the verification of `attresp.bin`.

Thanks @Xynnn007 for good suggestions, I think `way 2` approach is used in this PR, `generate_challenge_extra_params` is added in verifier with a default impplementation. I'm happy to chnage the function name. CC @jialez0

huoqifeng:
@Xynnn007 @jialez0 we also want to add the `extra-params` in `Attestation` like https://github.com/huoqifeng/kbs-types/blob/271ef4bac54035fb25fa44b24e5bbca844cd057c/src/lib.rs#L74-L75, how do you think?

fitzthum:
This is a tricky one. If we get this extra challenge information from the AS, does that mean that the AS will have to maintain some per-connection state?

huoqifeng:
> This is a tricky one. If we get this extra challenge information from the AS, does that mean that the AS will have to maintain some per-connection state?

@fitzthum No need keep the state. For IBM SE, the attester will embed the original encrypted attestation request in evidence also, the verifier will extract the request first and then perform the verifying.

Xynnn007:
Well, as we might not change the low level of s390x attestation workflow involving pvattest, there are now some best ways for me as declared in https://github.com/confidential-containers/trustee/pull/345#issuecomment-1999175504

Way 1 if IBM SE cert chain for the guest will not change during the whole lifetime of the guest
Way 2 if we agree on a new API of get_nonce() to CoCoAS
Way 3 will surely achieve that s390x get resources after remote attestation, even if no code is added to attester and CoCo-AS

attestation-service/verifier/src/lib.rs
```diff
@@ -152,6 +166,12 @@ pub trait Verifier {
         expected_report_data: &ReportData,
         expected_init_data_hash: &InitDataHash,
     ) -> Result<TeeEvidenceParsedClaim>;
+
+    async fn generate_challenge_extra_params(
```
jialez0, line 170: Do not use `extra_params` as function name here, because this is a conception of KBS protocol. I think `generate_challenge` is enough.

attestation-service/verifier/src/lib.rs
```diff
@@ -152,6 +166,12 @@ pub trait Verifier {
...
+        &self,
```
jialez0, line 171: I think there may be some extra input parameters when call this function in the future, so can you remain a optional input parameter here? Such as `tee_parameters: Option<Vec<u8>>`.

attestation-service/attestation-service/src/lib.rs
```diff
@@ -239,6 +239,11 @@ impl AttestationService {
     pub async fn register_reference_value(&mut self, message: &str) -> Result<()> {
         self.rvps.verify_and_extract(message).await
     }
+
+    pub async fn generate_challenge_extra_params(&self, tee: Tee) -> Result<String> {
```
jialez0, line 243: `generate_challenge_extra_params` -> `generate_challenge`

attestation-service/attestation-service/src/bin/restful/mod.rs
jialez0, line 1: Why don't you add generate challenge API in restful and grpc AS implementation? (What I can see is that new API is only in the AS-library)

jialez0:
> Well, as we might not change the low level of s390x attestation workflow involving pvattest, there are now some best ways for me as declared in [#345 (comment)](https://github.com/confidential-containers/trustee/pull/345#issuecomment-1999175504)
> 
> Way 1 if IBM SE cert chain for the guest will not change during the whole lifetime of the guest Way 2 if we agree on a new API of get_nonce() to CoCoAS cc @jialez0 Way 3 will surely achieve that s390x get resources after remote attestation, even if no code is added to attester and CoCo-AS

@Xynnn007 I prefer way 2, I think AS is reasonable to provide this `generate_challenge` (or `get_nonce`) API. What's more, I think AS maintain a simple `Nonce Map` is not a very unreasonable thing, it is not a true "state save" like KBS sessions, and the Nonce Map here will be a very lightweight structure. cc @fitzthum

Xynnn007:
> @Xynnn007 I prefer way 2, I think AS is reasonable to provide this `generate_challenge` (or `get_nonce`) API. What's more, I think AS maintain a simple `Nonce Map` is not a very unreasonable thing, it is not a true "state save" like KBS sessions, and the Nonce Map here will be a very lightweight structure. cc @fitzthum

Some more contexts are, as a separate attestation service, CoCoAS should have its own mechanism to prevent reply attack when the user does not use KBS.

So we add the new API `get_nonce()` and binds it to a same context of `attestation()` to let the AS check if the evidence is fresh enough. At the same time, only call `attestation()` to check an evidence is also supported.

attestation-service/attestation-service/src/bin/restful/mod.rs
huoqifeng: Checking, will add it also. Thanks!

attestation-service/verifier/src/lib.rs
huoqifeng: fixed

attestation-service/verifier/src/lib.rs
huoqifeng: fixed.

attestation-service/attestation-service/src/lib.rs
huoqifeng: fixed

attestation-service/attestation-service/src/bin/restful/mod.rs
huoqifeng: Fixed

fitzthum (commented):
A few comments. Sorry for the delay. I don't have a full picture without the actual attestation code, but I think this looks pretty good. I am open to adding support for the challenge generation.

I do wonder if there is a better name for `generate_challenge`. The current name makes it sound like somethign that every verifier would implement. Maybe `generate_supplemental_challenge` or something like that would be less confusing.

attestation-service/verifier/src/se/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+        _tee_parameters: Option<Vec<u8>>,
+    ) -> Result<String> {
+
+        // TODO replace FakeSeAttest with real crate
+        let attester = FakeSeAttest::default();
+
+        let hkds: Vec<String> = vec![String::new(); 2];
+        let certk = "cert_file_path";
+        let signk = "sign_file_path";
+        let arpk = "arpk_file_path";
```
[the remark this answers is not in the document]
huoqifeng: @fitzthum yes, it's placeholder for now, will replace it when https://crates.io/crates/s390_pv is ready.

attestation-service/verifier/src/se/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    let arpk = "arpk_file_path";
+    let hdr = "hdr_file_path";
+    let se = attester.verify(evidence, arpk, hdr)
+                .await
+                .context("Verify SE attestation evidence failed: {:?}")?;
+
+    let claims_map = json!({
+        "serial_number": format!("{}", "SE-ID"),
+        "measurement": format!("{}", BASE64_STANDARD.encode(se.clone())),
+        "report_data": format!("{}", BASE64_STANDARD.encode(se.clone())),
```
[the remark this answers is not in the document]
huoqifeng: Yes, I'll add this into doc once the crates is ready. Thanks! Tobin

huoqifeng:
> A few comments. Sorry for the delay. I don't have a full picture without the actual attestation code, but I think this looks pretty good. I am open to adding support for the challenge generation.
> 
> I do wonder if there is a better name for `generate_challenge`. The current name makes it sound like somethign that every verifier would implement. Maybe `generate_supplemental_challenge` or something like that would be less confusing.

Thanks @fitzthum , I'm open to the fn name. @Xynnn007 @jialez0 what's your suggestions on the fn name?

kbs/src/api/src/attestation/mod.rs
```diff
@@ -89,4 +93,25 @@ impl AttestationService {
...
+            AttestationService::CoCoASgRPC(inner) => {
+                inner.generate_challenge(tee, tee_parameters).await
+            }
+            #[cfg(any(feature = "coco-as-builtin", feature = "coco-as-builtin-no-verifier"))]
+            AttestationService::CoCoASBuiltIn(inner) => {
+                inner.generate_challenge(tee, tee_parameters).await
+            }
+            #[cfg(feature = "intel-trust-authority-as")]
+            AttestationService::IntelTA(inner) => {
+                inner.generate_challenge(tee, tee_parameters).await
```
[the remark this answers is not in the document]
huoqifeng: @fitzthum Shall we allow `kbs/src/api/src/attestation/intel_trust_authority/mod.rs` to decide error or a real value returned?

jialez0:
> > A few comments. Sorry for the delay. I don't have a full picture without the actual attestation code, but I think this looks pretty good. I am open to adding support for the challenge generation.
> > I do wonder if there is a better name for `generate_challenge`. The current name makes it sound like somethign that every verifier would implement. Maybe `generate_supplemental_challenge` or something like that would be less confusing.
> 
> Thanks @fitzthum , I'm open to the fn name. @Xynnn007 @jialez0 what's your suggestions on the fn name?

I'm OK with `generate_supplemental_challenge`.

attestation-service/verifier/src/se/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+        // TODO replace FakeSeAttest with real crate
+        let attester = FakeSeAttest::default();
+
+        // TODO replace the placeholder
+        let hkds: Vec<String> = vec![String::new(); 2];
+        let certk = "cert_file_path";
+        let signk = "sign_file_path";
+        let arpk = "arpk_file_path";
+
+        let extra_params = attester.create(hkds, certk, signk, arpk)
```
Xynnn007, line 44:
```suggestion
        let challenge = attester.create(hkds, certk, signk, arpk)
```

The rename matches the function name to avoid confusion.

attestation-service/verifier/src/se/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+    let arpk = "arpk_file_path";
+    let hdr = "hdr_file_path";
+
+    let se = attester.verify(evidence, arpk, hdr)
+                .await
+                .context("Verify SE attestation evidence failed: {:?}")?;
+
+    let claims_map = json!({
+        "serial_number": format!("{}", "SE-ID"),
+        "measurement": format!("{}", BASE64_STANDARD.encode(se.clone())),
```
Xynnn007, line 70: Btw I'd suggest use `hex` rather than base64 for all claims. Wdyt ? @fitzthum @mkulke

attestation-service/verifier/src/se/seattest.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright (C) Copyright IBM Corp. 2024
```
Xynnn007, line 1: Do we need this file? IMO things in `se/mod.rs` are enough.

kbs/src/api/src/attestation/coco/grpc.rs
```diff
@@ -125,6 +126,27 @@ impl Attest for GrpcClientPool {
...
     }
+
+    async fn generate_supplemental_challenge(
+        &self,
+        tee: Tee,
+        tee_parameters: Option<Vec<u8>>,
+    ) -> Result<String> {
+        let req = tonic::Request::new(ChallengeRequest {
+            tee: to_grpc_tee(tee).into(),
+            tee_params: String::from_utf8(tee_parameters.unwrap()).unwrap(),
```
Xynnn007, line 137: We'd try to avoid using `unwrap()`s. See https://github.com/confidential-containers/trustee/blob/main/attestation-service/attestation-service/src/bin/grpc/mod.rs#L194-L195 for example

kbs/src/api/src/attestation/mod.rs
```diff
@@ -32,6 +32,10 @@ pub trait Attest: Send + Sync {
     /// Verify Attestation Evidence
     /// Return Attestation Results Token
     async fn verify(&self, tee: Tee, nonce: &str, attestation: &str) -> Result<String>;
+
+    /// generate the challenge payload to pass to attester based on Tee and nonce
+    async fn generate_supplemental_challenge(&self, tee: Tee, tee_parameters: Option<Vec<u8>>)
```
Xynnn007, line 37: As `extra_params` are defined as `String`, we can directly use `tee_parameters: Option<String>` here. helps to avoid `as_bytes` and `from_utf8` on both side.

attestation-service/protos/attestation.proto
```diff
@@ -88,8 +89,17 @@ message SetPolicyRequest {
 }
 message SetPolicyResponse {}
 
+message ChallengeRequest {
+    Tee tee           = 1;
+    string tee_params = 2;
```
Xynnn007, line 94: We might need some refactor for the `tee_params`. It is marked `string` here, but the source could be a JSON map for better organization. So `map<string, string>` might be better somehow.

BTW I do not see any `tee_params` generated from the `attester` side. Can you give a link?

Xynnn007 (commented):
cool. Some comments..

attestation-service/verifier/src/se/seattest.rs
huoqifeng: Agreed, I'll drop this file later when the IBM SE crate ready.

attestation-service/verifier/src/se/mod.rs
huoqifeng: Good point, fixed.

attestation-service/protos/attestation.proto
huoqifeng: Right, grpc and rest approach is new added in https://github.com/confidential-containers/trustee/pull/345/commits/3a6b84e8c8f6403a8801f5ed61ffd30d7f1bb3b0 according to @jialez0 's suggestion. IBM SE generates challenges directly.

kbs/src/api/src/attestation/mod.rs
huoqifeng: Good point! fixed.

kbs/src/api/src/attestation/coco/grpc.rs
huoqifeng: Fixed

huoqifeng:
@Xynnn007 @fitzthum @jialez0 addressed most of the comments, may you please have a look again? And we'll still need some time to get https://crates.io/crates/s390_pv ready, do you think it's OK to add the implementation details, claims details in a new PR? We focus the API change and framework in this PR?

Xynnn007:
> @Xynnn007 @fitzthum @jialez0 addressed most of the comments, may you please have a look again? And we'll still need some time to get https://crates.io/crates/s390_pv ready, do you think it's OK to add the implementation details, claims details in a new PR? We focus the API change and framework in this PR?

Personally, it is a complete function, and the integration of API changes alone cannot make it work properly. Thus I prefer to wait until upstream is ready and then review and merge it together?

mkulke:
> Personally, it is a complete function, and the integration of API changes alone cannot make it work properly. Thus I prefer to wait until upstream is ready and then review and merge it together?

I agree. Also, from experience it's easier to merge attester + verifier together, since often you need to iterate on one of the components to make the full worklfow work (e.g. the evidence schema). So I'd recommend to start with a release on the external crate, then the attester module and finally the verifier (remember to bump the guest-components in this repo's kbs client). 

@huoqifeng You might also want to run an e2e test for the whole flow in `kbs/test` on an SE enclave. Not sure about the option to get an SE enclave as a runner, but even running it manually is probably.

I'll read the "Attestation Flow in IBM SE for Linux" document to understand better the need to extend the API for the SE challenge. before I can make an educated review.

Still, it would be good if you could describe in simple words what you are proposing and why the changes are required in the description of the PR.

huoqifeng:
> > Personally, it is a complete function, and the integration of API changes alone cannot make it work properly. Thus I prefer to wait until upstream is ready and then review and merge it together?
> 
> I agree. Also, from experience it's easier to merge attester + verifier together, since often you need to iterate on one of the components to make the full worklfow work (e.g. the evidence schema). So I'd recommend to start with a release on the external crate, then the attester module and finally the verifier (remember to bump the guest-components in this repo's kbs client).
> 
> @huoqifeng You might also want to run an e2e test for the whole flow in `kbs/test` on an SE enclave. Not sure about the option to get an SE enclave as a runner, but even running it manually is probably.
> 
> I'll read the "Attestation Flow in IBM SE for Linux" document to understand better the need to extend the API for the SE challenge. before I can make an educated review.
> 
> Still, it would be good if you could describe in simple words what you are proposing and why the changes are required in the description of the PR.

@mkulke @Xynnn007 @fitzthum @jialez0 I updated the description in the PR to explain the IBM SE Attestation flow and you can check it an see why a `generate_supplemental_challengs` is required in IBM SE case. I think which is also an option for other TEE in future.

huoqifeng:
> Does the Verifier (attestation-service in our case) have access to key material that would not be available of the Attester (attestation-agent in our case)?

The `a measurement key` is generated by verifier and will be passed to attester in encrypted format. This key is only visible in hardware/firmware but not in attester's user space.

> would it make sense that the attester creates the "attestion request" using the Challenge retrieved from the Verifier as nonce?
No, because of the above logic, the request can only be created and encrypted by verifier. I think this is also the philosophy described in RATS.

attestation-service/verifier/src/lib.rs
```diff
@@ -152,6 +166,13 @@ pub trait Verifier {
         expected_report_data: &ReportData,
         expected_init_data_hash: &InitDataHash,
     ) -> Result<TeeEvidenceParsedClaim>;
+
+    async fn generate_supplemental_challenge(
```
mkulke, line 170: we probably want to add api docs for this trait fn

kbs/src/api/src/attestation/coco/grpc.rs
```diff
@@ -125,6 +126,27 @@ impl Attest for GrpcClientPool {
...
     }
+
+    async fn generate_supplemental_challenge(
+        &self,
+        tee: Tee,
+        tee_parameters: Option<String>,
+    ) -> Result<String> {
+        let req = tonic::Request::new(ChallengeRequest {
+            tee: to_grpc_tee(tee).into(),
+            tee_params: tee_parameters.unwrap(),
```
mkulke, line 137: we shouldn't use unwrap/panic in library code. you can turn a None option into a Result using `Option::ok_or()`.

Do we have to use Option for `tee_parameters`? on kbs_types' Request type, the extra_params propery is [not an option](https://github.com/virtee/kbs-types/blob/3f70b78f18e39cfcd3c5225f56b49117c95ee79b/src/lib.rs#L48)

mkulke:
> @mkulke hope it makes sense.

@huoqifeng thanks for the details, I understand it better, probably not fully yet. It looks like the SE Attestation scheme requires a special attestation challenge to be set by the SE Verifier. The SE Attester will need it to create an evidence (while other attesters are able to consume a non-structured byte array).

huoqifeng:
> > @mkulke hope it makes sense.
> 
> @huoqifeng thanks for the details, I understand it better, probably not fully yet. It looks like the SE Attestation scheme requires a special attestation challenge to be set by the SE Verifier. The SE Attester will need it to create an evidence (while other attesters are able to consume a non-structured byte array).

Right, IBM SE verifier actually verify the evidence and returns true or false only, but we do inject information for further checking, like: Additional data, user-data, guest image hashes mentioned in the flow. I'll add the details when the crate is ready.

attestation-service/verifier/src/lib.rs
huoqifeng: Added, thanks!

kbs/src/api/src/attestation/coco/grpc.rs
huoqifeng: Sounds good to me, will fix it.

huoqifeng:
@fitzthum @Xynnn007 @jialez0 @mkulke , the IBM SE implementation has been added, would you please check it again?

Cargo.toml
```diff
@@ -30,13 +30,14 @@ clap = { version = "4", features = ["derive"] }
...
 hex = "0.4.3"
-kbs-types = "0.5.3"
+kbs-types = "0.6.0"
 jsonwebtoken = { version = "9", default-features = false }
 log = "0.4.17"
 prost = "0.11.0"
 regorus = { version = "0.1.5", default-features = false, features = ["regex", "base64", "time"] }
 rstest = "0.18.1"
 serde = { version = "1.0", features = ["derive"] }
+serde_with = { version = "1.11.0", features = ["base64"] }
```
Xynnn007, line 40: Let's put `serde_with` after `serde_json` to follow dictionary order

Cargo.toml
```diff
@@ -47,3 +48,4 @@ tokio = { version = "1.23.0", features = ["full"] }
 tempfile = "3.4.0"
 tonic = "0.8.1"
 tonic-build = "0.8.0"
+s390_pv = "0.10.0"
```
Xynnn007, line 51: Let's move `s390_pv` to `attestation-service/verifier` crate's `Cargo.toml`

attestation-service/attestation-service/src/bin/restful/mod.rs
```diff
@@ -44,6 +44,12 @@ pub struct AttestationRequest {
     policy_ids: Vec<String>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ChallengeRequest {
+    tee: String,
+    tee_params: String,
+}
```
Xynnn007, line 51: BTW, we might need to refactor this request struct. As we might have more abstract contexts apart from `tee` and `tee_params`, like
```rust
pub struct ChallengeRequest {
    #[flatten]
    inner: HashMap<String, String>,
}
```

attestation-service/docs/parsed_claims.md
```diff
@@ -93,6 +93,13 @@ The claim inherit the fields from the SEV-SNP claim with and additional `tpm` hi
 
 Note: The TD Report and TD Quote are fetched during early boot in this TEE. Kernel, Initrd and rootfs are measured into the vTPM's registers.
 
+## IBM Secure Execution (SE)
+- `se.version`: The version this quote structure.
+- `se.cuid`: The config uid.
+- `se.hdr.tag`: SE Header Tag (seht)
+- `se.image.phkh`: SE image Public host key hash
+- `se.attestation.phkh`: SE attestation Public host key hash
+- `se.user_data`: Custom attestation key owner data.
```
Xynnn007, line 102: Nice to see the parsed claims here. Need to add a new empty line here.

attestation-service/protos/attestation.proto
Xynnn007: Recall this. See the comment in `attestation-service/attestation-service/src/bin/restful/mod.rs`

attestation-service/verifier/src/lib.rs
```diff
@@ -152,6 +166,21 @@ pub trait Verifier {
...
     ) -> Result<TeeEvidenceParsedClaim>;
+
+    /// Generate the supplemental challenge
+    /// 
+    /// Some TEE like IBM SE need a `challenge` generated on verifier side
+    /// and pass it to attester side. This challenge is used by attester to
+    /// generate the evidence
+    ///
+    /// A optional `tee_parameters` comes from the attester side as the input.
+
```
Xynnn007, line 177: If we want to use inline comment, we could delete extra empty lines between `///` lines and function definitions.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,318 @@
...
+    request::{BootHdrTags, CertVerifier, HkdVerifier, ReqEncrCtx, Request, SymKeyType},
+    uv::ConfigUid,
+};
+use serde::{Deserialize, Serialize};
+use serde_json;
+use serde_with::{base64::Base64, serde_as};
+use std::{fs::File, io::Read};
+
+fn encrypt_measurement_key(key: &[u8], rsa_public_key: &PKey<Public>) -> Vec<u8> {
+    let mut encrypter = Encrypter::new(rsa_public_key).unwrap();
```
Xynnn007, line 28: Let's try to avoid all the `unwrap()`s and replace with `anyhow` error handling.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,318 @@
...
+    let buffer_len = encrypter.encrypt_len(key).unwrap();
+    let mut encrypted_hmac_key = vec![0; buffer_len];
+    let len = encrypter.encrypt(key, &mut encrypted_hmac_key).unwrap();
+    encrypted_hmac_key.truncate(len);
+
+    encrypted_hmac_key
+}
+
+fn decrypt_measurement_key(key: &[u8], rsa_private_key: &PKey<Private>) -> Vec<u8> {
+    let mut decrypter = Decrypter::new(rsa_private_key).unwrap();
```
Xynnn007, line 40: Let's try to avoid all the `unwrap()`s and replace with `anyhow` error handling.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,318 @@
...
+}
+
+impl SeAttestationRequest {
+    pub fn from_slice(request: &[u8]) -> Result<Self> {
+        Ok(serde_json::from_slice(request).unwrap())
+    }
+
+    pub fn from_string(request: &str) -> Result<Self> {
+        Ok(serde_json::from_str(request).unwrap())
+    }
```
Xynnn007, line 96: we have `serde_json::from_str` and `serde_json::from_slice` already ,thus no need to wrap the function again here.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,318 @@
...
+        Ok(Self {
+            measurement: measurement.to_vec(),
+            additional_data: additional_data.to_vec(),
+            user_data: user_data.to_vec(),
+            cuid: *cuid,
+            encr_measurement_key: encr_measurement_key.to_vec(),
+            encr_request_nonce: encr_request_nonce.to_vec(),
+            image_hdr_tags: *image_hdr_tags,
+        })
+    }
```
Xynnn007, line 201: If this function is only to specify every field of the `SeAttestationResponse`, we could directly use
```rust
let a = SeAttestationResponse {
...
}
```

rather than define it again.

attestation-service/verifier/src/se/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+pub mod ibmse;
+
+const SE_HOST_KEY_DOCUMENTS_ROOT: &str = "/run/confidential-containers/ibmse/hkds";
+const SE_CERTIFICATES_ROOT: &str = "/run/confidential-containers/ibmse/certs";
+const SE_CERTIFICATE_ROOT_CA: &str = "/run/confidential-containers/ibmse/certs/ca";
+const SE_CERTIFICATE_REVOCATION_LISTS_ROOT: &str = "/run/confidential-containers/ibmse/crls";
+const SE_IMAGE_HEADER_FILE: &str = "/run/confidential-containers/ibmse/hdr/hdr.bin";
+const SE_MEASUREMENT_ENCR_KEY_PRIVATE: &str =
+    "/run/confidential-containers/ibmse/rsa/encrypt_key.pem";
+const SE_MEASUREMENT_ENCR_KEY_PUBLIC: &str =
```
Xynnn007, line 21: Keep in mind that these files are set-up on the trustee side -- Trustee runs on a trusted place and does not need to be in TEE. So we'd better make the files configurable. `/run` might not be a good place.

kbs/src/api/src/attestation/coco/grpc.rs
```diff
@@ -118,6 +134,28 @@ impl Attest for GrpcClientPool {
 
         Ok(token)
     }
+
+    async fn generate_supplemental_challenge(
+        &self,
+        tee: Tee,
+        tee_parameters: String,
+    ) -> Result<String> {
+        let grpc_tee = to_grpc_tee(tee).unwrap();
```
Xynnn007, line 143: We can use
```
let tee = serde_json::to_string(&tee)
            .context("CoCo AS client: serialize tee type failed.")?
            .trim_end_matches('"')
            .trim_start_matches('"')
            .to_string();
```
similiarly to what we use in `verify()` without defining `to_grpc_tee()`.

Xynnn007 (commented):
Good. We are getting there.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,318 @@
...
+
+    pub fn verify(&self, priv_key_file: &str) -> Result<SeAttestationClaims> {
+        let mut file = File::open(priv_key_file)?;
+        let mut contents = vec![];
+        file.read_to_end(&mut contents)?;
```
mkulke, line 206: Does it make sense to re-read and re-process the file in every verify operation or could the private key be initialized once?

background: the top-level fns in the verifiers are async and we shouldn't perform costly sync code in those. There are async fs operations, but maybe we don't need it if we parse the key once.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        if nonce.len() != 16 {
+            return Err(anyhow!("The nonce vector must have exactly 16 elements."));
+        }
+        let boxed_slice: Box<[u8]> = nonce.into_boxed_slice();
+        let boxed_array: Box<[u8; 16]> = match boxed_slice.try_into() {
+            Ok(ba) => ba,
+            Err(_) => return Err(anyhow!("Failed to convert nonce from Vec<u8> to [u8; 16].")),
+        };
+        let nonce_array: [u8; 16] = *boxed_array;
```
mkulke, line 222: decrypt_measurement_key returns a Vec<u8>, so we can do just that, no?

```rust
let nonce_array: [u8; 16] = nonce
  .try_into()
  .ok_or(anyhow!("Failed to convert nonce from Vec<u8> to [u8; 16]."))?;
```

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,318 @@
...
+            &self.image_hdr_tags,
+            &self.cuid,
+            Some(&self.user_data),
+            Some(&nonce_array),
+            Some(&self.additional_data),
+        );
+
+        let measurement =
+            AttestationMeasurement::calculate(items, AttestationMeasAlg::HmacSha512, meas_key)
+                .unwrap();
```
mkulke, line 235: the unwrap should be handled

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        if !measurement.eq_secure(&self.measurement) {
+            debug!("Recieved: {:?}", self.measurement);
+            debug!("Calculated: {:?}", measurement.as_ref());
+            warn!("Attestation measurement verification failed. Calculated and received attestation measurement are not equal.");
+            return Err(anyhow!("Failed to verify the measurement!"));
```
mkulke, line 241: Err(anyhow!(...) can be shortened to bail!

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,318 @@
...
+
+    let se_request = SeAttestationRequest::create(
+        hkds,
+        certs,
+        crls,
+        root_ca,
+        &mut image_hdr_tags,
+        pub_key_file,
+    )
+    .unwrap();
```
mkulke, line 300: The unwrap should be handled

kbs/src/api/src/http/attest.rs
```diff
@@ -17,11 +18,17 @@ pub(crate) async fn auth(
...
+    attestation_service: web::Data<Arc<AttestationService>>,
 ) -> Result<HttpResponse> {
     info!("Auth API called.");
     debug!("Auth Request: {:?}", &request);
 
-    let session = SessionStatus::auth(request.0, **timeout)
+    let extra_params = attestation_service
+        .generate_supplemental_challenge(request.tee, request.extra_params.clone())
+        .await
+        .unwrap();
```
mkulke, line 29: the unwrap should be handled

mkulke (commented):
thx. a couple of comments, mostly related to panics/unwrap.

Cargo.toml
huoqifeng: Fixed, thanks!

Cargo.toml
huoqifeng: Fixed, thanks!

attestation-service/attestation-service/src/bin/restful/mod.rs
huoqifeng: Fixed, thanks!

attestation-service/docs/parsed_claims.md
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/lib.rs
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/se/ibmse.rs
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/se/ibmse.rs
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/se/ibmse.rs
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/se/mod.rs
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,318 @@
...
+            let hk = read_file(hkd, "host-key document")?;
+            let certs = read_certs(&hk)?;
+            if certs.is_empty() {
+                warn!("The host key document in '{hkd}' contains empty certificate!");
+            }
+            if certs.len() != 1 {
+                warn!("The host key document in '{hkd}' contains more than one certificate!")
+            }
+            // Panic: len is == 1 -> unwrap will succeed/not panic
+            let c = certs.first().unwrap();
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/se/ibmse.rs
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/se/ibmse.rs
huoqifeng: Fixed, thanks!

kbs/src/api/src/http/attest.rs
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/se/ibmse.rs
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/se/ibmse.rs
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,318 @@
...
+        let mut contents = vec![];
+        file.read_to_end(&mut contents)?;
+        let rsa = Rsa::public_key_from_pem(&contents)?;
+        let rsa_public_key = &PKey::from_rsa(rsa)?;
+
+        let conf_data = arcb.confidential_data();
+        let encr_measurement_key =
+            encrypt_measurement_key(conf_data.measurement_key(), rsa_public_key);
+        let encr_request_nonce =
+            encrypt_measurement_key(conf_data.nonce().clone().unwrap().value(), rsa_public_key);
```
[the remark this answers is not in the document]
huoqifeng: Fixed, thanks!

attestation-service/protos/attestation.proto
huoqifeng: Fixed, thanks!

attestation-service/verifier/src/se/ibmse.rs
huoqifeng: OK, I'll use lazy load and cache in memory to do that. Thanks!

huoqifeng:
@mkulke @Xynnn007 thanks for the review and good suggestions! I have updated it in new commit, would you please check it again?

attestation-service/attestation-service/src/bin/grpc/mod.rs
```diff
@@ -195,6 +199,43 @@ impl AttestationService for Arc<RwLock<AttestationServer>> {
...
+        let tee_params = request
+            .inner
+            .get("tee_params")
+            .map_or(Err(Status::aborted("Error parse inner_tee tee_params")), Ok)?;
+        let tee = to_kbs_tee(&inner_tee).map_or(
+            Err(Status::aborted(format!(
+                "Error parse TEE type: {inner_tee}"
+            ))),
+            Ok,
+        )?;
```
mkulke, line 224:
```suggestion
        let tee = to_kbs_tee(&inner_tee).ok_or(Status::aborted(format!("Error parse TEE type: {inner_tee}")))?;
```

attestation-service/attestation-service/src/bin/restful/mod.rs
```diff
@@ -44,6 +44,14 @@ pub struct AttestationRequest {
     policy_ids: Vec<String>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ChallengeRequest {
+    // tee: String,
+    // tee_params: String,
```
mkulke, line 50: leftovers?

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,302 @@
...
+use serde_with::{base64::Base64, serde_as};
+use std::{fs::File, io, io::Read, sync::Mutex};
+
+lazy_static::lazy_static! {
+    static ref PUB_KEY_FILE_CONTENTS: Mutex<Option<Vec<u8>>> = Mutex::new(None);
+    static ref PRI_KEY_FILE_CONTENTS: Mutex<Option<Vec<u8>>> = Mutex::new(None);
+}
+
+fn get_cached_file_or_read(filename: &str, content_ref: &Mutex<Option<Vec<u8>>>) -> io::Result<Vec<u8>> {
+    let mut guard = content_ref.lock().unwrap();
```
mkulke, line 34: let's avoid the unwrap

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,302 @@
...
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct UserData {
+    #[serde_as(as = "Base64")]
+    image_btph: Vec<u8>,
+}
+
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SeAttestationClaims {
```
mkulke, line 83: for those structs, do you need to set the memory layout to `#[repr(c)]` maybe? If they originate from c libraries rust/llvm might decide to reorder/pad the fields.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,302 @@
...
+
+        if !measurement.eq_secure(&self.measurement) {
+            debug!("Recieved: {:?}", self.measurement);
+            debug!("Calculated: {:?}", measurement.as_ref());
+            warn!("Attestation measurement verification failed. Calculated and received attestation measurement are not equal.");
+            bail!("Failed to verify the measurement!");
+        }
+        
+        // let userdata = serde_json::from_slice(&self.user_data)?;
+        // debug!("user_data: {:?}", userdata);
```
mkulke, line 228: leftover?

attestation-service/attestation-service/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-default = [ "restful-bin", "rvps-grpc", "rvps-builtin", "all-verifier" ]
```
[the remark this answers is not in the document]
huoqifeng: This was mis-picked from Choi other PR when build on s390x machine, I'll leave it to Choi and using original one.

mkulke (approved):
lgtm.

I'll suggest to wait for the attester module to be merged first and then bump the kbs/tools/client guest-component dependency revision for e2e tests

Xynnn007 (approved):
Thanks. lgtm. Let's get this merged after https://github.com/confidential-containers/guest-components/pull/492

huoqifeng:
@mkulke @Xynnn007 thank you very much for reviewing the PR and gave so many useful suggestions! thanks for your approval.

kbs/src/api/src/http/attest.rs
```diff
@@ -17,11 +18,19 @@ pub(crate) async fn auth(
...
     map: web::Data<SessionMap>,
     timeout: web::Data<i64>,
+    attestation_service: web::Data<Arc<AttestationService>>,
 ) -> Result<HttpResponse> {
     info!("Auth API called.");
     debug!("Auth Request: {:?}", &request);
 
-    let session = SessionStatus::auth(request.0, **timeout)
+    let extra_params = attestation_service
+        .generate_supplemental_challenge(request.tee, request.extra_params.clone())
```
mythi, line 27: we have #242 RFC to suggest how `extra_params` are used. perhaps we can review the new usages in it first and have them documented before any code gets added.

Xynnn007: Sounds good. The only flaws not covered well in this PR is how KBS interacts with AS. In the beginning we have refactored AS & KBS to [avoid coupling](https://github.com/confidential-containers/trustee/pull/240). So the core idea is that we should not use a structure of KBS as AS's API.

An alternative way is to put the things into `nonce` rather than `extra_params` for IBM SE here. `extra_params` could be decided & appended in the kbs side by the AS plugin type.


btw, I still prefer to integrate IBM SE inside only KBS part rather than also covering AS part. Let's see in future.

Xynnn007: In this way, the [nonce()](https://github.com/confidential-containers/trustee/pull/163/commits/1e3b172a5c16d357f68799e0b2ed4c0075fdc362#diff-b83a2a4e4135ca8f4193335a1a8d660b5f1dadf86f8cad301b44c4f2a6867a2eR37) API means the same as `generate_supplemental_challenge()` here.

huoqifeng: > `generate_supplemental_challenge`

It's similar for `nonce()` and `generate_supplemental_challenge()`, the difference is: for IBM SE which is generated by verifier as part of the challenge. Do you think reasonable to delegate nonce() generation from KBS to verifier?

huoqifeng: The key point is should AS or KBS generate the attestation challenge. If challenge is specific to verifier or holds verifier specific info, it should be generated by AS/Verifier (IBM SE is such as case).

Xynnn007: I know that. What I mean are
1. The challenges generated by IBM SE might better to be put to `nonce` rather than `extra_params`
2. `nonce` in @mythi 's PR also aims to be from AS (although the default impl is from KBS), thus the same as `generate_supplemental_challenge`
3. `extra_params` would have some other semantice as #242 proposes. All the contents now is determined by the AS type, but no need to interact with AS to get.
4. Also, as 1stR-A in R-C-A-R are both handled by backend AS, we could rename `generate_supplemental_challenge` in KBS side to `generate_challenge` and directly returns a `Challenge` structure.

Let me write a commit for this.

huoqifeng: The original names is `generate_challenge` :-)

Xynnn007: See this commit https://github.com/Xynnn007/kbs/commit/1655c266f67cbdb421659366af655e50662cb68f

huoqifeng: > See this commit [Xynnn007@1655c26](https://github.com/Xynnn007/kbs/commit/1655c266f67cbdb421659366af655e50662cb68f)

I agree with this approach, this is actually my original thoughts :-), @Xynnn007 do you want me picking it and revise or do you want get the commit merged firstly?

Xynnn007: If possible, pick it together with this pr

mythi: LGTM too, seems it also allows to set  per-AS `"supported-hash-algorithms": ["sha256", "sha384", "sha512"]` as suggested by #242.

But perhaps squash the commits a  bit?

huoqifeng: @Xynnn007 I changed the guest-components accordingly also here https://github.com/confidential-containers/guest-components/pull/492/commits/cdeabd46330defd3e401175a1d7b3c4bf7e8e8e3

huoqifeng:
Squashed commits.

Xynnn007 (approved):
lgtm

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,311 @@
...
+    encr_request_nonce: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    image_hdr_tags: BootHdrTags,
+}
+
+#[repr(C)]
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SeAttestationClaims {
+    #[serde_as(as = "Base64")]
```
Xynnn007, line 90: I suggest that we use hex/base16 to encode the parsed claims. This would help users have a common way to set policy. see https://github.com/confidential-containers/trustee/issues/372

huoqifeng: Yeah, I'm happy to keep aligned with other drivers.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,311 @@
...
+    encr_request_nonce: Vec<u8>,
+    #[serde_as(as = "Base64")]
+    image_hdr_tags: BootHdrTags,
+}
+
+#[derive(Debug)]
+pub struct RealSeVerifier {
+    rsa_private_key: PKey<Private>,
+    rsa_public_key: PKey<Public>,
```
steffen-eiden, line 121: I'd prefer EC keys, but that's up to the projects maintainers.

SE attestation uses EC-keys with spec521r1 curve

kbs/src/api/src/attestation/mod.rs
```diff
@@ -32,6 +34,21 @@ pub trait Attest: Send + Sync {
...
+
+        thread_rng()
+            .try_fill(&mut nonce[..])
+            .map_err(anyhow::Error::from)?;
+
+        let nonce = STANDARD.encode(&nonce);
+        Ok(Challenge {
+            nonce,
+            extra_params: String::new(),
+        })
```
fitzthum, line 50: Do we really need to implement this method here? It's just a fallback, right?

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,311 @@
...
+        }
+    };
+}
+
+fn list_files_in_folder(dir: &str) -> Result<Vec<String>> {
+    let mut file_paths = Vec::new();
+
+    for entry in fs::read_dir(dir)? {
+        let entry = entry?;
+        let path = entry.path();
```
fitzthum, line 55: I think you can probably collapse these two lines into the next one.

fitzthum (commented):
Looks pretty good. A few comments.

attestation-service/verifier/src/se/ibmse.rs
huoqifeng: Good point! thank you, fixed.

kbs/src/api/src/attestation/mod.rs
huoqifeng: Yes, this is a fallback.

kbs/src/api/src/attestation/coco/grpc.rs
```diff
@@ -118,6 +123,46 @@ impl Attest for GrpcClientPool {
...
     }
+
+    async fn generate_challenge(&self, tee: Tee, tee_parameters: String) -> Result<Challenge> {
+        let nonce = match tee {
+            Tee::Se => {
+                let tee = serde_json::to_string(&tee)
+                    .context("CoCo AS client: serialize tee type failed.")?
+                    .trim_end_matches('"')
+                    .trim_start_matches('"')
+                    .to_string();
```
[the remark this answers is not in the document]
huoqifeng: Good point, thank you Tobin, fixed.

attestation-service/verifier/src/se/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+    async fn evaluate(
+        &self,
+        evidence: &[u8],
+        _expected_report_data: &ReportData,
+        _expected_init_data_hash: &InitDataHash,
+    ) -> Result<TeeEvidenceParsedClaim> {
+        let se_verifier = ONCE
+            .get_or_try_init(|| async { RealSeVerifier::new() })
+            .await?;
+        warn!("IBM SE does not support initdata.");
```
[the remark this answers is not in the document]
huoqifeng: Yeah, I'll remove it and add a comment here.

attestation-service/docs/parsed_claims.md
```diff
@@ -93,6 +93,14 @@ The claim inherit the fields from the SEV-SNP claim with and additional `tpm` hi
 
 Note: The TD Report and TD Quote are fetched during early boot in this TEE. Kernel, Initrd and rootfs are measured into the vTPM's registers.
 
+## IBM Secure Execution (SE)
+- `se.version`: The version this quote structure.
+- `se.cuid`: The config uid.
+- `se.hdr.tag`: SE header tag (seht)
```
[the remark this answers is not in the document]
huoqifeng: Good point, thank you Steffen!

attestation-service/docs/parsed_claims.md
```diff
@@ -93,6 +93,14 @@ The claim inherit the fields from the SEV-SNP claim with and additional `tpm` hi
...
+- `se.image.phkh`: SE image public host key hash
+- `se.attestation.phkh`: SE attestation public host key hash
+- `se.user_data`: Custom attestation key owner data.
```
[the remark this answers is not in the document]
huoqifeng: Good point, thank you Steffen!

attestation-service/docs/parsed_claims.md
```diff
@@ -93,6 +93,14 @@ The claim inherit the fields from the SEV-SNP claim with and additional `tpm` hi
 
 Note: The TD Report and TD Quote are fetched during early boot in this TEE. Kernel, Initrd and rootfs are measured into the vTPM's registers.
 
+## IBM Secure Execution (SE)
+- `se.version`: The version this quote structure.
+- `se.cuid`: The config uid.
```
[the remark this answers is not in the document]
huoqifeng: Good point, thank you Steffen.

attestation-service/verifier/src/lib.rs
```diff
@@ -152,6 +165,17 @@ pub trait Verifier {
...
+    /// Generate the supplemental challenge
+    ///
+    /// Some TEE like IBM SE need a `challenge` generated on verifier side
+    /// and pass it to attester side. This challenge is used by attester to
+    /// generate the evidence
+    ///
+    /// A optional `tee_parameters` comes from the attester side as the input.
+    async fn generate_supplemental_challenge(&self, _tee_parameters: String) -> Result<String> {
+        Ok(String::new())
```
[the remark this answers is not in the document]
huoqifeng: It's the default implementation as a fallback.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,311 @@
...
+            Some(&se_response.additional_data),
+        );
+
+        let measurement =
+            AttestationMeasurement::calculate(items, AttestationMeasAlg::HmacSha512, &meas_key)?;
+
+        if !measurement.eq_secure(&se_response.measurement) {
+            debug!("Recieved: {:?}", se_response.measurement);
+            debug!("Calculated: {:?}", measurement.as_ref());
+            warn!("Attestation measurement verification failed. Calculated and received attestation measurement are not equal.");
```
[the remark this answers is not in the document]
huoqifeng: Yes, agreed, fixed.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,311 @@
...
+            bail!("Failed to verify the measurement!");
+        }
+
+        // TODO check self.user_data.image_btph with previous saved value
```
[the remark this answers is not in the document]
huoqifeng: Removed.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,311 @@
...
+        let add_data = AdditionalData::from_slice(&se_response.additional_data, &att_flags)?;
+        debug!("additional_data: {:?}", add_data);
+        let image_phkh = add_data
+            .image_public_host_key_hash()
+            .ok_or(anyhow!("Failed to get image_public_host_key_hash."))?;
+        let attestation_phkh = add_data
+            .attestation_public_host_key_hash()
+            .ok_or(anyhow!("Failed to get attestation_public_host_key_hash."))?;
+
+        // TODO image_phkh and attestation_phkh with previous saved value
```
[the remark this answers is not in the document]
huoqifeng: Removed

attestation-service/verifier/src/se/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+use async_trait::async_trait;
+use ibmse::RealSeVerifier;
+use log::warn;
+use tokio::sync::OnceCell;
+
+use crate::{InitDataHash, ReportData, TeeEvidenceParsedClaim, Verifier};
+
+pub mod ibmse;
+
+static ONCE: OnceCell<RealSeVerifier> = OnceCell::const_new();
```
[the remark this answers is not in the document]
huoqifeng: Good point, fixed.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,311 @@
...
+            .context("decrypt Measurement Key")?;
+        let nonce = self
+            .decrypt(&se_response.encr_request_nonce)
+            .context("decrypt Request Nonce")?;
+
+        if nonce.len() != 16 {
+            bail!("The nonce vector must have exactly 16 elements.");
+        }
+
+        let nonce_array: [u8; 16] = nonce
```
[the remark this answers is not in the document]
huoqifeng: Merged these 2.

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,311 @@
...
+            }
+        }
+    }
+
+    Ok(file_paths)
+}
+
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SeAttestationResponse {
```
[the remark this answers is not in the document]
Xynnn007: For other archs, there should be a 3rd crate define this. For SE, `pv_s390x` might be a good choice.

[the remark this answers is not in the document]
huoqifeng: Yeah, pv_s390x is an option, another option is to create a repo under IBM org. Can we do this refine later? Which should be se only change. While in this PR, there is some AS/KBS API change, I'd like to separate. @steffen-eiden @Xynnn007 WDYT?

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,311 @@
...
+        if !measurement.eq_secure(&se_response.measurement) {
+            debug!("Recieved: {:?}", se_response.measurement);
+            debug!("Calculated: {:?}", measurement.as_ref());
+            warn!("Attestation measurement verification failed. Calculated and received attestation measurement are not equal.");
+            bail!("Failed to verify the measurement!");
+        }
+
+        // TODO check self.user_data.image_btph with previous saved value
+
+        let mut att_flags = AttestationFlags::default();
```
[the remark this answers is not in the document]
huoqifeng: @steffen-eiden the flags bind to the claims, we'll have to change claims format if flags change. Most likely we'll use fixed flags even more options provided. We need change flags and claims at same time at least. Can we leave this change to later?

attestation-service/verifier/src/se/ibmse.rs
```diff
@@ -0,0 +1,311 @@
...
+        let items = AttestationItems::new(
+            &se_response.image_hdr_tags,
+            &se_response.cuid,
+            Some(&se_response.user_data),
+            Some(&nonce_array),
+            Some(&se_response.additional_data),
+        );
+
+        let measurement =
+            AttestationMeasurement::calculate(items, AttestationMeasAlg::HmacSha512, &meas_key)?;
```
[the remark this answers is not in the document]
huoqifeng: @steffen-eiden same as flags, I'd like to leave this as later change, wdyt?

attestation-service/verifier/src/se/ibmse.rs
huoqifeng: I'm happy to use ECIES approach, after some searching I don't find a handy openssl example, I think pure rust approach is not good for us. Let me add it in my todo. @steffen-eiden WDYT?

attestation-service/verifier/src/se/ibmse.rs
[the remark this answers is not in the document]
steffen-eiden: Yes, IMO we can to that later if more flags and algorithms are supported by the s390 attestation flow.
IDK anything about the compatibility claims of this project so just make sure you do not block further extensions to the claims, request, and response structs.

-- mark as resolved --

attestation-service/verifier/src/se/ibmse.rs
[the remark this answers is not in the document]
steffen-eiden: same as above

-- mark as resolved --

attestation-service/verifier/src/se/ibmse.rs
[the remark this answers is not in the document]
steffen-eiden: you can always refactor later IMO - if you want move fast with this PR i don't see an issue here.

if you meant with `pv_s390x` the [pv crate](https://docs.rs/s390_pv/0.10.0/s390_pv/) I have to say no.
This crate should not contain any implementation details about the attestation.

attestation-service/verifier/src/se/ibmse.rs
steffen-eiden: Instead of reading in a RSA key as pem-file just read in an ec key as pem-file (and rename the variable)
->
`SE_MEASUREMENT_ENCR_KEY_PRIVATE` should contain an EC key in PEm format and you are using EC keys instead of RSA keys.

the PKey abstracts over the exact key type (RSA, EC, HMAC, ....) see [OpenSSL documenatation (C)](https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_new.html)

huoqifeng: oh, you don't mean ECIES but just replace the RSA key to EC, OK.

attestation-service/verifier/src/se/mod.rs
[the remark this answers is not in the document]
fitzthum: I mean `ONCE` -> `VERIFIER`.

attestation-service/verifier/src/se/mod.rs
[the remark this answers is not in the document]
fitzthum: See how the AZ SNP verifier [handles it](https://github.com/confidential-containers/trustee/blob/main/attestation-service/verifier/src/az_snp_vtpm/mod.rs#L103-L105)

attestation-service/verifier/src/se/mod.rs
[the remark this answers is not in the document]
huoqifeng: oh, got it, fixed, thank you @fitzthum

attestation-service/verifier/src/se/mod.rs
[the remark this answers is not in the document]
huoqifeng: Got it, fixed, thank you @fitzthum

fitzthum (approved):
LGTM

I do want to emphasize that since this verifier does not check the report data, it is not secure. Hopefully that will be addressed in a follow-up. I think this PR is fine for getting all the support code in.

huoqifeng:
> LGTM
> 
> I do want to emphasize that since this verifier does not check the report data, it is not secure. Hopefully that will be addressed in a follow-up. I think this PR is fine for getting all the support code in.

Thank you! @fitzthum , yes, the checking algorithm checks the evidence from TEE guest and re-calculated value from verifier, which also based on same SE OS Image. We'll add additional check in a follow up PR.

Xynnn007:
https://github.com/confidential-containers/guest-components/pull/492 was merged. we should update the rev in https://github.com/confidential-containers/trustee/blob/main/kbs/tools/client/Cargo.toml#L21

Issue #333, opened 2024-02-20
Release KBS Client Binary


fitzthum:
It would be handy if we released a binary for the KBS client. Unlike the services that we already release, we probably don't want this to be a standalone container image. We could use something like [oras](https://github.com/oras-project/oras) to provide a binary via a registry. We should be able to push a staged and release version of the client tool just like we do with the other components.

We'll also want to consider multiple platforms.

wainersm:
It would be useful for the Kata CI where we are working to have attestation tests (https://github.com/kata-containers/kata-containers/pull/9114) and we'd like to avoid having to compile the client. Kata CI intents to consume released versions of KBS project, so technically a binary within a tarball each release would be sufficient, however, we might end up on a situation where we need to track a commit (to consume a fix) so that having a staged binaries much more flexible.

One concern, regardless whether an executable in tarball or oras, is that it will be required to be statically compiled. Due to some dependencies on attesters implementations, it might be not feasible. An alternative to try is to compile with `sample_only` feature, reducing the dependencies, but then the resulting kbs-client will be only used for configuration.

Cc @mkulke @fitzthum @portersrc

wainersm:
Hi @portersrc !

> Not sure if you'd confirmed the dependencies on `sample_only`, but it's indeed a trimmer list (no libcrypto dependency, etc.):
> 
> ```
> $ ldd kbs-client
>         linux-vdso.so.1 (0x00007ffdd55d6000)
>         libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f2ae2099000)
>         libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f2ae1fb2000)
>         libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2ae0a00000)
>         /lib64/ld-linux-x86-64.so.2 (0x00007f2ae20ca000)```
> ```

It seems easy to get the sample_only built statically on Ubuntu then. Thanks for checking it out!

portersrc:
[PR 349](https://github.com/confidential-containers/trustee/pull/349) is merged, which seems to have successfully triggered an `oras push`, and kbs-client is now available here:
https://github.com/confidential-containers/trustee/pkgs/container/staged-images%2Fkbs-client
```
[...]

PR #349, opened 2024-03-08
Build and push kbs-client binary

portersrc:
This PR attempts to partially address https://github.com/confidential-containers/trustee/issues/333

It adds a workflow to build a statically linked kbs-client and push it to ghcr.io.

Limitations: It only covers x86-64 linux for now. It only supports the `sample_only` kbs-client.

### DIFF
--- a/.github/workflows/kbs-client-build-and-push.yaml
+++ b/.github/workflows/kbs-client-build-and-push.yaml
@@ -1,0 +1,39 @@
+name: Build and push kbs-client
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build_and_push:
+    env:
+      RUSTC_VERSION: 1.76.0
+    runs-on: ubuntu-22.04
+
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v4
+    - name: Install rust toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: ${{ env.RUSTC_VERSION }}
+        override: true
+        profile: minimal
+    - name: Log in to ghcr.io
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - name: Build a statically linked kbs-client for x86_64 linux
+      working-directory: kbs
+      run: |
+        make cli-static-x86_64-linux
+    - name: Push to ghcr.io
+      working-directory: target/x86_64-unknown-linux-gnu/release
+      run: |
+        commit_sha=${{ github.sha }}
+        oras push \
+          ghcr.io/confidential-containers/staged-images/kbs-client:sample_only-x86_64-linux-gnu-${commit_sha},latest \
+          kbs-client
--- a/kbs/Makefile
+++ b/kbs/Makefile
@@ -34,6 +34,17 @@ passport-resource-kbs:
 cli:
 	cargo build -p kbs-client --locked --release --no-default-features --features $(CLI_FEATURES)
 
+.PHONY: cli-static-x86_64-linux
+cli-static-x86_64-linux:
+	cargo build \
+      -p kbs-client \
+      --target=x86_64-unknown-linux-gnu \
+      --config "target.x86_64-unknown-linux-gnu.rustflags = '-C target-feature=+crt-static'" \
+      --locked \
+      --release \
+      --no-default-features \
+      --features sample_only
+
 install-kbs:
 	install -D -m0755 ../target/release/kbs $(INSTALL_DESTDIR)
 


.github/workflows/kbs-client-build-and-push.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Build a statically linked kbs-client for x86_64 linux
+      working-directory: kbs
+      run: |
+        make cli-static-x86_64-linux
+    - name: Push to ghcr.io
+      working-directory: target/x86_64-unknown-linux-gnu/release
+      run: |
+        commit_sha=${{ github.sha }}
+        oras push \
+          ghcr.io/confidential-containers/staged-images/kbs-client:${commit_sha},latest \
```
Xynnn007, line 38: Do we need to change to `sample-only-x86-64-linux-gnu-${commit_sha}`?

Potentially, we might have
`tdx-x86-64-linux-gnu-${commit_sha}`
`yyy-x86-64-windows-${commit_sha}`
...

If I understand correctly?

Xynnn007 (approved):
Thanks for this! Only a little question.

.github/workflows/kbs-client-build-and-push.yaml
portersrc: Yes, thanks for this catch!

.github/workflows/kbs-client-build-and-push.yaml
```diff
@@ -0,0 +1,39 @@
...
+        password: ${{ secrets.GITHUB_TOKEN }}
+    - name: Build a statically linked kbs-client for x86_64 linux
+      working-directory: kbs
+      run: |
+        make cli-static-x86_64-linux
+    - name: Push to ghcr.io
+      working-directory: target/x86_64-unknown-linux-gnu/release
+      run: |
+        commit_sha=${{ github.sha }}
+        oras push \
```
fitzthum, line 37: Is oras installed already?

fitzthum (commented):
Nice. One question

.github/workflows/kbs-client-build-and-push.yaml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build_and_push:
+    env:
+      RUSTC_VERSION: 1.76.0
+    runs-on: ubuntu-22.04
```
wainersm, line 12: Hi @portersrc ! I don't remember the last time I built static executables on Ubuntu, so wondering if you need to install the static libc package. On my fedora 39 laptop I had to install glibc-static to test this static build.

.github/workflows/kbs-client-build-and-push.yaml
wainersm: good question.

kbs/Makefile
```diff
@@ -34,6 +34,17 @@ passport-resource-kbs:
 cli:
 	cargo build -p kbs-client --locked --release --no-default-features --features $(CLI_FEATURES)
 
+.PHONY: cli-static-x86_64-linux
+cli-static-x86_64-linux:
```
wainersm, line 38: I'm still unsure it should be hardcoded the arch as it's not cross compiling to x86_64, it is?

Another thing is whether it should specify the target `--target=x86_64-unknown-linux-gnu` as one might just static compile with musl, for example.

Would be possible to statically build with `cli` target but tweaking options via `RUSTFLAGS`?

.github/workflows/kbs-client-build-and-push.yaml
portersrc: It should be there. I tested it was there, but also a github action runner from this PR says its installed software is [here](https://github.com/confidential-containers/trustee/actions/runs/8266650115/job/22615127521?pr=349#step:1:10) (see [ORAS under cli tools](https://github.com/actions/runner-images/blob/ubuntu22/20240304.1/images/ubuntu/Ubuntu2204-Readme.md#cli-tools)).

.github/workflows/kbs-client-build-and-push.yaml
portersrc: I don't think I checked this; great question; will check it now.

kbs/Makefile
portersrc: When this target is executed on our github runner, it's not cross compiling, as you point out.

Statically building with musl could be cool/useful/better. I didn't try this. I'm not partial to gnu libc or musl, but bear in mind musl would also need to be present. Do you want me to add support for musl, as well, or do you want to support both? (I'm OK either way.)

I thought it better to make an explicit target than to leverage the `cli` target. If someone clones the repo and tries to build kbs-client, I want it to just work. If the dependencies aren't there, it should either fail gracefully or have some meaningful way to grab the dependencies. So, `RUSTFLAGS` would be a little cleaner (because it would just re-use the `cli` target), but there'd be all sorts of unsupported combinations there. For instance, `cli` supports `$(CLI_FEATURES)`, but we're only expecting `sample_only` to work. Similarly, we don't have any other checks for the host machine or the cross compilation libraries for building for other targets. Hence, I tried to make an explicit Makefile target with explicit flags.

.github/workflows/kbs-client-build-and-push.yaml
portersrc: I can confirm that `gcc -static` works on a [simple test](https://github.com/portersrc/infra/actions/runs/8327016995/job/22783806866#step:8:8). Also 22.04 package list includes [libc6-dev](https://packages.ubuntu.com/jammy/allpackages?format=txt.gz), and the (debian) [file list for libc6-dev](https://packages.debian.org/sid/amd64/libc6-dev/filelist) includes libc.a, etc. (So I think we're OK?)

The easier thing is to just verify that this workflow succeeds before merging, but I don't see this workflow getting kicked... Do you know how/why?

kbs/Makefile
portersrc: Follow-up: I thought your idea might be better, so I tried it. It runs into a weird issue:
```
$ RUSTFLAGS="--target=x86_64-unknown-linux-gnu -C target-feature=+crt-static" make cli CLI_FEATURES=sample_only
...
error: cannot produce proc-macro for `async-trait v0.1.77` as the target `x86_64-unknown-linux-gnu` does not support these crate types
make: *** [Makefile:35: cli] Error 101
```
An issue is raised for this, it seems: https://github.com/rust-lang/rust/issues/71651
In the meantime, this suggests that an extra Makefile target might be needed, though.

.github/workflows/kbs-client-build-and-push.yaml
fitzthum: Ah I guess they are adding stuff on top of Ubuntu

kbs/Makefile
wainersm: hmm.... I think it is okay to leave your first implementation

.github/workflows/kbs-client-build-and-push.yaml
portersrc: Thanks for the pointers! I looked at the yaml file, but to test at least part of this PR, I did something cruder: I changed to `on: pull_request` and commented out the ghcr login and push. This at least checks that `make cli-static-x86_64-linux` completes on the runner. It built fine (GHA result [here](https://github.com/confidential-containers/trustee/actions/runs/8473999540/job/23219496678?pr=349)). I changed it back to `on: push`, as intended for this PR. But I guess we need it to merge in order to verify the oras step works (though I tested with a personal repo, and it looks OK).

fitzthum (approved):
LGTM

portersrc:
I just noticed that the "latest" tag didn't get applied correctly:
https://github.com/confidential-containers/trustee/actions/runs/8502445853/job/23286620681#step:6:13
I will look at it.

PR #351, opened 2024-03-19
AS: Optimize policy management mechanism

jialez0:
Added API for listing policies


attestation-service/attestation-service/src/bin/restful/mod.rs
```diff
@@ -172,3 +172,48 @@ pub async fn set_policy(
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct RemovePolicyRequest {
+    pub policy_ids: Vec<String>,
+}
+
+pub async fn remove_policy(
+    input: web::Json<RemovePolicyRequest>,
+    cocoas: web::Data<Arc<RwLock<AttestationService>>>,
+) -> Result<HttpResponse> {
+    info!("set policy.");
```
mkulke, line 207: wrong log statement

mkulke (commented):
can you elaborate more about the security issues you want to avoid by using an in-memory store?

Will this cause problems for HA deployments?

Would an abstract store like we use for reference values and kbs secrets make sense?

attestation-service/attestation-service/src/bin/restful/mod.rs
```diff
@@ -172,3 +172,48 @@ pub async fn set_policy(
...
+
+    let policy_list = cocoas
+        .read()
+        .await
+        .list_policy()
+        .await
+        .context("get policys")?;
+
+    let body = serde_json::to_string(&serde_json::json!({
+        "policys": policy_list,
```
fitzthum, line 189: `policys` -> `policies` (and above)

attestation-service/attestation-service/src/bin/restful/mod.rs
```diff
@@ -172,3 +172,48 @@ pub async fn set_policy(
...
+        .content_type("application/json")
+        .body(body))
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct RemovePolicyRequest {
+    pub policy_ids: Vec<String>,
+}
+
+pub async fn remove_policy(
```
fitzthum, line 203: maybe this should be `remove_policies` if the input is a list of policies?

attestation-service/attestation-service/src/lib.rs
```diff
@@ -121,6 +121,22 @@ impl AttestationService {
...
+    /// Remove Attestation Verification Policy.
+    pub async fn remove_policy(&mut self, policy_id: String) -> Result<()> {
+        self.policy_engine
+            .remove_policy(policy_id)
+            .await
+            .map_err(|e| anyhow!("Cannot Remove Policy: {:?}", e))
+    }
+
+    /// Get all Attestation Verification Policy.
+    pub async fn list_policy(&self) -> Result<Vec<crate::policy_engine::PolicyListEntry>> {
```
fitzthum, line 133: Maybe this should be `list_policies`

attestation-service/attestation-service/src/policy_engine/mod.rs
```diff
@@ -53,4 +61,8 @@ pub trait PolicyEngine {
     ) -> Result<HashMap<String, (PolicyDigest, EvaluationResult)>>;
 
     async fn set_policy(&mut self, input: SetPolicyInput) -> Result<()>;
```
fitzthum, line 63: i know this isn't part of your PR, but I wonder if we should rename policy to `add_policy` since we support multiple policies.

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -29,31 +30,16 @@ pub struct GoString {
...
-                .ok_or_else(|| anyhow!("Policy DirPath to string failed"))?,
-        );
-        default_policy_path.push("default.rego");
-        if !default_policy_path.as_path().exists() {
-            let policy = std::include_str!("default_policy.rego").to_string();
-            fs::write(&default_policy_path, policy)?;
-        }
+    pub fn new() -> Result<Self> {
+        let default_policy = std::include_str!("default_policy.rego");
+        let mut policy_map = HashMap::new();
```
fitzthum, line 39: The downside to this is that the policies will disappear once the AS is stopped. Is that going to be an issue? In theory we could have multiple storage backends for the policies, but that might be overkill.

fitzthum (commented):
Nice. I made a few comments. 

Btw are you thinking about making similar changes for the resource policies in the KBS?

attestation-service/verifier/test_data/cca_evidence_claim_output.txt
```diff
@@ -0,0 +1 @@
+Object {"platform": Object {"cca-platform-challenge": String("AQICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgIC"), "cca-platform-sw-components": Array [Object {"measurement-type": String("BL"), "measurement-value": String("AwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwM="), "signer-id": String("BAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQ="), "version": String("3.4.2")}]}, "realm": Object {"cca-realm-extensible-measurements": Array [String("Q0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQw=="), String("Q0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQw=="), String("Q0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQw=="), String("Q0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQw==")], "cca-realm-hash-algo-id": String("sha-256"), "cca-realm-initial-measurement": String("Q0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQw=="), "cca-realm-personalization-value": String("QURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBRA=="), "cca-realm-public-key-hash-algo-id": String("sha-512")}}
```
Xynnn007, line 1: This file is generated automatically by unit tests in `verifier`. I'd suggest to  add them (together with other `parse_*.txt`) to `.gitignore`

attestation-service/attestation-service/src/policy_engine/mod.rs
jialez0: I think using set policy here is more accurate because we not only use this API to add new policies, but also to update some already set policies.

jialez0:
> can you elaborate more about the security issues you want to avoid by using an in-memory store?
> 
> Will this cause problems for HA deployments?

@mkulke The reason why we store real-time policies in memory instead of on disk is because in cloud deployment scenarios, the AS provided by CSP (or other trusted third parties) to tenants usually needs to run in a TEE, and the disk of this TEE is not always encrypted. If we place the verification policy on the disk, it will not make users believe in the security of that.

BTW, What is HA deployments？

> The downside to this is that the policies will disappear once the AS is stopped. Is that going to be an issue? In theory we could have multiple storage backends for the policies, but that might be overkill.

@fitzthum I don't think the disappearance of in-memory policies due to the stop of AS will become a problem, at least in the current productization scenario, the lifecycle of policies will not exceed that of AS. Even when we use AS as an independent service instance, the long-term retention of policy content will not be the responsibility of AS itself. Policy owners need to have clear records and awareness of the policies they set to AS (so that they can use the correct policy ID when performing validation).

> Would an abstract store like we use for reference values and kbs secrets make sense?

@mkulke @fitzthum This is the ultimate solution, but it will bring many extra code and work, and I don't think we really need it now, at least not yet.

mkulke:
> @mkulke The reason why we store real-time policies in memory instead of on disk is because in cloud deployment scenarios, the AS provided by CSP (or other trusted third parties) to tenants usually needs to run in a TEE, and the disk of this TEE is not always encrypted. If we place the verification policy on the disk, it will not make users believe in the security of that.

thanks for clarifying. But that reasoning would apply to all state that we currently have in a kbs/as deployment, no?

e.g. reference values, secrets and the other policies.

If that's the case, I'm not sure it's a good idea to handle this particular kind of policies in a special way. There could be a simple storage trait abstraction like we have for the other state and then we can have an in-memory k/v implementation.

> BTW, What is HA deployments？

High-Availability deployments, i.e. >1 replicas for a service. I think for most production deployments this is a realistic scenario. A memory-backed service will have state per instance. e.g. a policy will be updated on replica 1, but not on replica 2. If the state is backed by hard disk, you can still have a stateless deployment by using shared volume technologies, configmaps, etc. In-memory state is really local to a process. 

There are also other problems with local state: the service will not survive a restart, hence it cannot be moved to another node, (rolling) update don't work, etc..

see [this discussion](https://github.com/confidential-containers/trustee/issues/290) about kbs deployment scenarios.

> the lifecycle of policies will not exceed that of AS

can you explain why that is case and what the assumed lifecycle of AS is? a deployment or a process? 

> > Would an abstract store like we use for reference values and kbs secrets make sense?
> 
> @mkulke @fitzthum This is the ultimate solution, but it will bring many extra code and work, and I don't think we really need it now, at least not yet.

a store trait abstraction should not add a lot of code or complexity, IMO. I agree it's a bit unfortunate that we deal with this individually for every bit of state, but I think we need to have a discussion about state in general. kbs/as wants to be a CRUD service for certain resources, so we should consider a proper data-store that could potentially be used for all the state.

bpradipt:
> > @mkulke The reason why we store real-time policies in memory instead of on disk is because in cloud deployment scenarios, the AS provided by CSP (or other trusted third parties) to tenants usually needs to run in a TEE, and the disk of this TEE is not always encrypted. If we place the verification policy on the disk, it will not make users believe in the security of that.
> 
> thanks for clarifying. But that reasoning would apply to all state that we currently have in a kbs/as deployment, no?
> 
> e.g. reference values, secrets and the other policies.
> 
> If that's the case, I'm not sure it's a good idea to handle this particular kind of policies in a special way. There could be a simple storage trait abstraction like we have for the other state and then we can have an in-memory k/v implementation.
> 
> > BTW, What is HA deployments？
> 
> High-Availability deployments, i.e. >1 replicas for a service. I think for most production deployments this is a realistic scenario. A memory-backed service will have state per instance. e.g. a policy will be updated on replica 1, but not on replica 2. If the state is backed by hard disk, you can still have a stateless deployment by using shared volume technologies, configmaps, etc. In-memory state is really local to a process.
> 
> There are also other problems with local state: the service will not survive a restart, hence it cannot be moved to another node, (rolling) update don't work, etc..
> 
> see [this discussion](https://github.com/confidential-containers/trustee/issues/290) about kbs deployment scenarios. cc @bpradipt
> 

There has to be a persistent store for policies and reference values otherwise we hit the issues that @mkulke mentioned.

jialez0:
@mkulke @fitzthum @bpradipt @Xynnn007 Now this PR is only focus on adding `list` and `delete` APIs for policy management. I agree that AS need a persistent store for policies, so we can still use filesystem for storing.

Please take a look and this PR, and let's try to get it merged :-)

attestation-service/attestation-service/src/bin/restful-as.rs
```diff
@@ -95,7 +95,12 @@ async fn main() -> Result<(), RestfulError> {
     let server = HttpServer::new(move || {
         App::new()
             .service(web::resource(WebApi::Attestation.as_ref()).route(web::post().to(attestation)))
-            .service(web::resource(WebApi::Policy.as_ref()).route(web::post().to(set_policy)))
+            .service(
+                web::resource(WebApi::Policy.as_ref())
+                    .route(web::post().to(set_policy))
+                    .route(web::get().to(list_policies))
```
Xynnn007, line 101: Do we need an API to list concrete policy content? like
GET `/policy` would return policy list
GET  `/policy/<id>` would return content of policy <id>

attestation-service/attestation-service/src/lib.rs
```diff
@@ -134,6 +134,22 @@ impl AttestationService {
         Ok(())
     }
 
+    /// Remove Attestation Verification Policy.
+    pub async fn remove_policies(&mut self, policy_id: String) -> Result<()> {
+        self.policy_engine
+            .remove_policies(policy_id)
+            .await
+            .map_err(|e| anyhow!("Cannot Remove Policy: {:?}", e))
```
Xynnn007, line 142:
```suggestion
            .context("Cannot Remove Policy")
```

attestation-service/attestation-service/src/lib.rs
```diff
@@ -134,6 +134,22 @@ impl AttestationService {
...
+    }
+
+    /// Get all Attestation Verification Policy.
+    pub async fn list_policies(&self) -> Result<Vec<crate::policy_engine::PolicyListEntry>> {
+        self.policy_engine
+            .list_policies()
+            .await
+            .map_err(|e| anyhow!("Cannot List Policy: {:?}", e))
```
Xynnn007, line 150:
```suggestion
            .context("Cannot List Policy")
```

attestation-service/attestation-service/src/bin/restful/mod.rs
```diff
@@ -172,3 +172,48 @@ pub async fn set_policy(
...
+    let policy_list = cocoas
+        .read()
+        .await
+        .list_policies()
+        .await
+        .context("get policies")?;
+
+    let body = serde_json::to_string(&serde_json::json!({
+        "policies": policy_list,
+    }))
```
Xynnn007, line 190: We could directly use the `serde_json::to_string(&policy_list)`?

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -146,7 +147,7 @@ impl PolicyEngine for OPA {
             bail!("OPA Policy Engine only support .rego policy");
         }
 
-        let policy_bytes = base64::engine::general_purpose::URL_SAFE_NO_PAD
+        let policy_bytes = base64::engine::general_purpose::STANDARD
```
Xynnn007, line 150: Note that we are changing base64 encoding here, which would cause compability problem. please ensure this is intended.

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -161,6 +162,60 @@ impl PolicyEngine for OPA {
...
+
+    async fn remove_policies(&mut self, policy_id: String) -> Result<()> {
+        let policy_dir_path = self
+            .policy_dir_path
+            .to_str()
+            .ok_or_else(|| anyhow!("Miss Policy DirPath"))?;
+        let policy_file_path = format!("{policy_dir_path}/{policy_id}.rego");
+        tokio::fs::remove_file(policy_file_path)
+            .await
+            .map_err(|e| anyhow!("Remove OPA policy file failed: {:?}", e))?;
```
Xynnn007, line 174:
```suggestion
            .context("Remove OPA policy file failed")?;
```

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -161,6 +162,60 @@ impl PolicyEngine for OPA {
...
+            }
+        }
+
+        let mut policy_list = Vec::new();
+
+        for id in policy_ids.iter() {
+            let policy_file_path = format!("{policy_dir_path}/{id}.rego");
+            let policy = tokio::fs::read_to_string(policy_file_path)
+                .await
+                .map_err(|e| anyhow!("Read OPA policy file failed: {:?}", e))?;
```
Xynnn007, line 205:
```suggestion
                .context("Read OPA policy file failed")?;
```

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -161,6 +162,60 @@ impl PolicyEngine for OPA {
...
+
+            let mut hasher = Sha384::new();
+            hasher.update(policy.as_bytes());
+            let digest = hasher.finalize().to_vec();
+            policy_list.push(PolicyListEntry {
+                id: id.to_string(),
+                digest: base64::engine::general_purpose::STANDARD.encode(digest),
```
Xynnn007, line 212: Encoding issue. Mentioned before.

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -216,18 +271,23 @@ mod tests {
...
-            policy: base64::engine::general_purpose::URL_SAFE_NO_PAD.encode(policy),
+            policy: base64::engine::general_purpose::STANDARD.encode(policy),
         };
 
         assert!(opa.set_policy(input).await.is_ok());
+        let policy_list = opa.list_policies().await.unwrap();
+        let list_output = format!("{:?}", policy_list);
+        std::fs::write("../tests/tmp/list_policy_output.txt", list_output).unwrap();
+        assert!(opa.list_policies().await.is_ok());
+        assert!(opa.remove_policies("test".to_string()).await.is_ok());
```
Xynnn007, line 291: We can also add another sentence to check `list_policies()` after the removing.

Xynnn007 (commented):
Thanks! Some nits

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
jialez0: According to feedback from some AS users, it is more reasonable to change to the standard encoding format, as URL secure padding will prohibit `\n` at the end of policy content, which is unfriendly for users to manually copy policy content and perform base64 encoding on the command line.

But this is not what this PR needs to pay attention to, so I have temporarily removed the modifications here. In the future, we can implement this modification in the new PR.

attestation-service/attestation-service/src/bin/restful-as.rs
jialez0: At present, one GET endpoint to get all polices can already meet the requirements. I think the new endpoint you mentioned can be implemented when further requirements arise in the future.

jialez0: Good idea, I'll implement this.

mkulke:
> @mkulke @fitzthum @bpradipt @Xynnn007 Now this PR is only focus on adding `list` and `delete` APIs for policy management. I agree that AS need a persistent store for policies, so we can still use filesystem for storing.
> 
> Please take a look and this PR, and let's try to get it merged :-)

Would a `delete()` endpoint create the same state problems for deployment with multiple replicas? (delete on one, but still present on others)

Xynnn007:
> Would a `delete()` endpoint create the same state problems for deployment with multiple replicas? (delete on one, but still present on others)

I think this problem should be covered by potential refactoring storage trait of OPA policy, hinting that the backend storage should handle this, not the AS.

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -161,6 +162,60 @@ impl PolicyEngine for OPA {
...
+
+        let mut policy_list = Vec::new();
+
+        for id in policy_ids.iter() {
+            let policy_file_path = format!("{policy_dir_path}/{id}.rego");
+            let policy = tokio::fs::read_to_string(policy_file_path)
+                .await
+                .context("Read OPA policy file failed")?;
+
+            let mut hasher = Sha384::new();
```
mkulke, line 207: is the use of this particular hash algo documented somewhere? It might be sufficient to add it as a comment to the API fn or PolicyListEntry struct

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -229,5 +284,17 @@ default allow = true"
...
+        std::fs::write("../tests/tmp/list_policy_output.txt", list_output).unwrap();
+        assert!(opa.list_policies().await.is_ok());
+        assert!(opa.remove_policies("test".to_string()).await.is_ok());
+        let policy_list = opa.list_policies().await.unwrap();
+        let list_output = format!("{:?}", policy_list);
+        std::fs::write(
+            "../tests/tmp/list_policy_after_remove_output.txt",
+            list_output,
+        )
+        .unwrap();
```
mkulke, line 298: are we missing an assert about the content of policy_list?

mkulke:
> I think this problem should be covered by potential refactoring storage trait of OPA policy, hinting that the backend storage should handle this, not the AS. Another deploy model is to use k8s like tools. we'd also use fs type and mount to a given volume that is driven by underlying specific storage plugin (CSI).

Ah sorry not sure I understand:

Is your point that stateful operations are ok in general, because users are supposed to mount r/w CSI volumes as a backend? (If that's the case then we should point that out in the docs, and probably add a volume mount in the kubernetes manifests that we ship)

Or are you suggesting that we shouldn't implement stateful ops at the moment? (and manage policy objects with e.g. object store APIs, if an object store volume is mounted).

Xynnn007:
> Is your point that stateful operations are ok in general, because users are supposed to mount r/w CSI volumes as a backend? (If that's the case then we should point that out in the docs, and probably add a volume mount in the kubernetes manifests that we ship)
> 
> Or are you suggesting that we shouldn't implement stateful ops at the moment? (and manage policy objects with e.g. object store APIs, if an object store volume is mounted).

Almost. Two ways in my mind
1. If we deploy CoCoAS within k8s, we can do you mentioned. We need doc updates.
2. If we deploy CoCoAS without k8s, we can define a "new object store API" implementation as you mentioned, like a postgres/MySQL cluster client.

Certainly, the two ways are not distinguished and can be mixed in some way.

mkulke:
> Almost. Two ways in my mind
> 
> 1. If we deploy CoCoAS within k8s, we can do you mentioned. We need doc updates.
> 2. If we deploy CoCoAS without k8s, we can define a "new object store API" implementation as you mentioned, like a postgres/MySQL cluster client.
> 
> Certainly, the two ways are not distinguished and can be mixed in some way.

thanks for clarifying, makes sense.

(note: even in a k8s deployment having the policies stored in a datastore makes sense, and there's also non-k8s deployments with load balancer + replicas, where state on the local hd would be a problem)

So should we defer stateful operations (like delete() in this PR) until we introduce an abstract storage interface for policies? currently in a k8s deployment Policies could be managed via ConfigMaps that are mounted, any changes should be propagated to the pod (by default in a eventually-consistent fashion, but I think this can be tweaked)

jialez0:
@mkulke Now this PR is only adding GET API, you can take a look at it again.

mkulke (approved):
lgtm, thanks for accommodating!

PR #354, opened 2024-03-22
Verifeir: Add support for TDX quote v5

### LINKED ISSUE #352: AS/Verifier: TDX Quote format updation
There are two TDX quote formats in DCAP:
-  [_sgx_quote4_t](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/quote_wrapper/common/inc/sgx_quote_4.h#L141-L144)
-  [_sgx_quote5_t](https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/quote_wrapper/common/inc/sgx_quote_5.h#L106-L118).

The two quote formats are different so TDX quote verification will fail when it comes to compare `report_data` field as the offset is different.

We should support both formats of quote.

Xynnn007:
Intel DCAP defines two different formats of quotes, s.t. v4 and v5.

Current implementation only supports v4 and if we get a v5 quote it will parse the wrong fields like report_data. This PR mainly add support for v5 and they both can be supported by runtime parsing. Also adds a unit test and doc update.

Fixes #352

cc @jialez0 @arronwy


Xynnn007:
Fixed lint error

arronwy (approved):
Thanks @Xynnn007 LGTM!

PR #355, opened 2024-03-22
Use the Trustee name in a few more places

fitzthum:
Fixup a few places where we refer to the "KBS Cluster." Now we call that Trustee.


Xynnn007 (approved):
LGTM

PR #357, opened 2024-03-29
kbs: switch to Regorus for resource policy

fitzthum:
1) Switch from Go-based policy engine to rust-based Regorus 
2) Switch from Anyhow to thiserror for resource policy 
3) Add several unit tests and test policies

This does not touch the AS policy. I will do that later although the plan may change based on whether/how we use EAR tokens.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ kbs-types = "0.5.3"
 jsonwebtoken = { version = "9", default-features = false }
 log = "0.4.17"
 prost = "0.11.0"
+regorus = { version = "0.1.2", default-features = false, features = ["regex", "base64", "time"] }
 rstest = "0.18.1"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.89"
--- a/kbs/sample_policies/README.md
+++ b/kbs/sample_policies/README.md
@@ -1,6 +1,8 @@
 This directory contains sample policy files to configure the policy engine
 of the KBS. You can use these files to write your own policies.
 
+There are also several example policies used [for testing](../test/data). 
+
 | File | Description |
 | ---  | ---         |
 |[allow_all.rego](./allow_all.rego)|Equivalent to turning off the policy engine. Release resources unconditionally|
--- a/kbs/src/api/src/policy_engine/mod.rs
+++ b/kbs/src/api/src/policy_engine/mod.rs
@@ -7,12 +7,37 @@ use async_trait::async_trait;
 use serde::Deserialize;
 use std::path::PathBuf;
 use std::sync::Arc;
+use thiserror::Error;
 use tokio::sync::Mutex;
 
 #[cfg(feature = "opa")]
 mod opa;
 
 const DEFAULT_POLICY_PATH: &str = "/opa/confidential-containers/kbs/policy.rego";
+
+#[derive(Error, Debug)]
+pub enum ResourcePolicyError {
+    #[error("Failed to evaluate resource policy {0}")]
+    EvaluationError(#[from] anyhow::Error),
+
+    #[error("Failed to load data for resource policy")]
+    DataLoadError,
+
+    #[error("Invalid resource path format")]
+    ResourcePathError,
+
+    #[error("Resource Policy IO Error: {0}")]
+    IOError(#[from] std::io::Error),
+
+    #[error("Decoding (base64) resource policy failed: {0}")]
+    DecodeError(#[from] base64::DecodeError),
+
+    #[error("Failed to load input for resource policy")]
+    InputError,
+
+    #[error("Failed to load resource policy")]
+    PolicyLoadError,
+}
 
 /// Resource policy engine interface
 #[async_trait]
@@ -26,11 +51,14 @@ pub(crate) trait PolicyEngineInterface: Send + Sync {
     /// ([decide_result, extra_output])
     /// decide_result: Boolean value to present whether the evaluate is passed or not.
     /// extra_output: original ouput from policy engine.
-    async fn evaluate(&self, resource_path: String, input_claims: String)
-        -> Result<(bool, String)>;
+    async fn evaluate(
+        &self,
+        resource_path: String,
+        input_claims: String,
+    ) -> Result<bool, ResourcePolicyError>;
 
     /// Set policy (Base64 encode)
-    async fn set_policy(&mut self, policy: String) -> Result<()>;
+    async fn set_policy(&mut self, policy: String) -> Result<(), ResourcePolicyError>;
 }
 
 /// Policy engine configuration.
@@ -55,7 +83,7 @@ pub(crate) struct PolicyEngine(pub Arc<Mutex<dyn PolicyEngineInterface>>);
 
 impl PolicyEngine {
     /// Create and initialize PolicyEngine
-    pub async fn new(config: &PolicyEngineConfig) -> Result<Self> {
+    pub async fn new(config: &PolicyEngineConfig) -> Result<Self, ResourcePolicyError> {
         let policy_engine: Arc<Mutex<dyn PolicyEngineInterface>> = {
             cfg_if::cfg_if! {
                 if #[cfg(feature = "opa")] {
--- a/kbs/src/api/src/policy_engine/opa/mod.rs
+++ b/kbs/src/api/src/policy_engine/opa/mod.rs
@@ -2,29 +2,11 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::policy_engine::PolicyEngineInterface;
-use anyhow::{anyhow, Result};
+use crate::policy_engine::{PolicyEngineInterface, ResourcePolicyError};
 use async_trait::async_trait;
 use base64::Engine;
-use serde_json::Value;
-use std::ffi::CStr;
 use std::fs;
-use std::os::raw::c_char;
 use std::path::PathBuf;
-
-// Link import cgo function
-#[link(name = "cgo")]
-extern "C" {
-    pub fn evaluateGo(policy: GoString, data: GoString, input: GoString) -> *mut c_char;
-}
-
-/// String structure passed into cgo
-#[derive(Debug)]
-#[repr(C)]
-pub struct GoString {
-    pub p: *const c_char,
-    pub n: isize,
-}
 
 #[derive(Debug, Clone)]
 pub struct Opa {
@@ -32,11 +14,10 @@ pub struct Opa {
 }
 
 impl Opa {
-    pub fn new(policy_path: PathBuf) -> Result<Self> {
+    pub fn new(policy_path: PathBuf) -> Result<Self, ResourcePolicyError> {
+        std::fs::create_dir_all(policy_path.parent().unwrap())?;
+
         if !policy_path.as_path().exists() {
-            if !policy_path.as_path().parent().unwrap().exists() {
-                std::fs::create_dir_all(policy_path.parent().unwrap())?;
-            }
             let policy = std::include_str!("default_policy.rego").to_string();
             fs::write(&policy_path, policy)?;
         }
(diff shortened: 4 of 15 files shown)


fitzthum:
Unfortunately when I rebase (and pick up az-snp-vtpm 0.5.2) I run into this dependency conflict. Not sure we can fix this ourselves? @mkulke @anakrish

```
error: failed to select a version for `once_cell`.
    ... required by package `jsonschema v0.17.1`
    ... which satisfies dependency `jsonschema = "^0.17.1"` of package `regorus v0.1.2`
    ... which satisfies dependency `regorus = "^0.1.2"` of package `api-server v0.1.0 (/home/tobin/kbs/kbs/src/api)`
    ... which satisfies path dependency `api-server` of package `kbs v0.1.0 (/home/tobin/kbs/kbs/src/kbs)`
versions that meet the requirements `^1.17` are: 1.19.0, 1.18.0, 1.17.2, 1.17.1, 1.17.0

all possible versions conflict with previously selected packages.

  previously selected package `once_cell v1.0.1`
    ... which satisfies dependency `once_cell = ">=1, <=1.14.0"` of package `mbox v0.6.1`
    ... which satisfies dependency `mbox = "^0.6.0"` of package `tss-esapi v7.4.0`
    ... which satisfies dependency `tss-esapi = "^7.4"` of package `az-cvm-vtpm v0.5.2`
    ... which satisfies dependency `az-cvm-vtpm = "^0.5.2"` of package `az-snp-vtpm v0.5.2`
    ... which satisfies dependency `az-snp-vtpm = "^0.5.2"` of package `attester v0.1.0 (https://github.com/confidential-containers/guest-components.git?rev=21b2c536b4d6c5c1442b53916c908b54dde136e8#21b2c536)`
    ... which satisfies git dependency `attester` of package `kbs_protocol v0.1.0 (https://github.com/confidential-containers/guest-components.git?rev=21b2c536b4d6c5c1442b53916c908b54dde136e8#21b2c536)`
    ... which satisfies git dependency `kbs_protocol` of package `kbs-client v0.1.0 (/home/tobin/kbs/kbs/tools/client)`

failed to select a version for `once_cell` which could resolve this conflict
make: *** [Makefile:21: background-check-kbs] Error 101
```

Cargo.toml
```diff
@@ -31,6 +31,8 @@ hex = "0.4.3"
 kbs-types = "0.5.3"
 log = "0.4.17"
 prost = "0.11.0"
+#regorus = "0.1.2"
+regorus = { git = "https://github.com/microsoft/regorus" }
```
anakrish, line 35: We could configure regorus to support only those builtins that will be used by the policies. The way it is used currently, we might not even need the `arc` feature since neither the Engine instance nor the `Value` instances are shared between threads.

```suggestion
regorus = { git = "https://github.com/microsoft/regorus",  default-features=false, features=["arc"] }
```

anakrish:
> Unfortunately when I rebase (and pick up az-snp-vtpm 0.5.2) I run into this dependency conflict. Not sure we can fix this ourselves? @mkulke @anakrish
> 
> ```
> error: failed to select a version for `once_cell`.
>     ... required by package `jsonschema v0.17.1`
>     ... which satisfies dependency `jsonschema = "^0.17.1"` of package `regorus v0.1.2`
>     ... which satisfies dependency `regorus = "^0.1.2"` of package `api-server v0.1.0 (/home/tobin/kbs/kbs/src/api)`
>     ... which satisfies path dependency `api-server` of package `kbs v0.1.0 (/home/tobin/kbs/kbs/src/kbs)`
> versions that meet the requirements `^1.17` are: 1.19.0, 1.18.0, 1.17.2, 1.17.1, 1.17.0
> 
> all possible versions conflict with previously selected packages.
> 
>   previously selected package `once_cell v1.0.1`
>     ... which satisfies dependency `once_cell = ">=1, <=1.14.0"` of package `mbox v0.6.1`
>     ... which satisfies dependency `mbox = "^0.6.0"` of package `tss-esapi v7.4.0`
>     ... which satisfies dependency `tss-esapi = "^7.4"` of package `az-cvm-vtpm v0.5.2`
>     ... which satisfies dependency `az-cvm-vtpm = "^0.5.2"` of package `az-snp-vtpm v0.5.2`
>     ... which satisfies dependency `az-snp-vtpm = "^0.5.2"` of package `attester v0.1.0 (https://github.com/confidential-containers/guest-components.git?rev=21b2c536b4d6c5c1442b53916c908b54dde136e8#21b2c536)`
>     ... which satisfies git dependency `attester` of package `kbs_protocol v0.1.0 (https://github.com/confidential-containers/guest-components.git?rev=21b2c536b4d6c5c1442b53916c908b54dde136e8#21b2c536)`
>     ... which satisfies git dependency `kbs_protocol` of package `kbs-client v0.1.0 (/home/tobin/kbs/kbs/tools/client)`
> 
> failed to select a version for `once_cell` which could resolve this conflict
> make: *** [Makefile:21: background-check-kbs] Error 101
> ```

You could just not use the `jsonschema` feature in Regorus that bring in the dependency.
https://github.com/microsoft/regorus/blob/3d98c3b12ecbdfdfabdcfd6947c2019413883a9f/Cargo.toml#L38

The following table lists various Rego builtins and which feature each one depends on.
https://github.com/microsoft/regorus/blob/main/docs/builtins.md

For OPA compatibility, all features are enabled by default. But most applications won't need or use all the OPA builtins.

kbs/src/api/src/policy_engine/mod.rs
```diff
@@ -7,13 +7,38 @@ use async_trait::async_trait;
...
+    #[error("Resource Policy IO Error")]
+    IOError,
+
+    #[error("Decoding (base64) resource policy failed.")]
+    DecodeError,
+
+    #[error("Failed to laod input for resource policy")]
+    InputError,
+
+    #[error("Failed to load policy for resource policy")]
```
Xynnn007, line 38:
```suggestion
    #[error("Failed to load resource policy")]
```

kbs/test/data/policy_4.rego
```diff
@@ -0,0 +1,31 @@
+
```
Xynnn007, line 1: maybe we should delete the empty lines

Xynnn007 (commented):
Nice PR and good unit tests.

Cargo.toml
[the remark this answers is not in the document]
fitzthum: Whoops. That's leftover from development when I was thinking about using the new `eval_rule` thing.

fitzthum: This seems to work. Thanks for the advice.

kbs/sample_policies/README.md
```diff
@@ -1,7 +1,9 @@
...
 of the KBS. You can use these files to write your own policies.
 
+There are also several example policies used [for testing](../test/data). 
+
 | File | Description |
 | ---  | ---         |
 |[allow_all.rego](./allow_all.rego)|Equivalent to turning off the policy engine. Release resources unconditionally|
-|[deny_all.rego](./deny_all.rego)|Deny all resources release|
\ No newline at end of file
+|[deny_all.rego](./deny_all.rego)|Deny all resources release|
```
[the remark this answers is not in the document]
fitzthum: Yeah I think we should refactor a few things in this repo including the docs.

kbs/test/data/policy_4.rego
fitzthum: The GO OPA engine is very picky about the format of the policy and would often fail if there were any extra lines. This test is supposed to make sure the Regorus is more resilient. I added a comment to the policy file because it does look odd.

fitzthum:
Hmm. Looks like there is some issue with building Regorus inside of our docker container

```
60.32 error: failed to run custom build command for `regorus v0.1.2`
60.32 
60.32 Caused by:
60.32   process didn't exit successfully: `/usr/src/kbs/target/release/build/regorus-d111907e1736880c/build-script-build` (exit status: 101)
60.32   --- stderr
60.32   thread 'main' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/regorus-0.1.2/build.rs:20:10:
60.32   called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
60.32   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
60.32 warning: build failed, waiting for other jobs to finish...
81.04 error: failed to compile `kbs v0.1.0 (/usr/src/kbs/kbs/src/kbs)`, intermediate artifacts can be found at `/usr/src/kbs/target`.
81.04 To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

I don't see a similar issue building locally. I have also tried pinning to v0.1.0 and v0.1.1 and the upstream head. Any ideas @anakrish

Cargo.toml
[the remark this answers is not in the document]
anakrish: `eva;_rule` would be the way to go in future. It avoids having to parse a query, and also returns the value of a rule directly.

anakrish:
> Hmm. Looks like there is some issue with building Regorus inside of our docker container
> 
> ```
> 60.32 error: failed to run custom build command for `regorus v0.1.2`
> 60.32 
> 60.32 Caused by:
> 60.32   process didn't exit successfully: `/usr/src/kbs/target/release/build/regorus-d111907e1736880c/build-script-build` (exit status: 101)
> 60.32   --- stderr
> 60.32   thread 'main' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/regorus-0.1.2/build.rs:20:10:
> 60.32   called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
> 60.32   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> 60.32 warning: build failed, waiting for other jobs to finish...
> 81.04 error: failed to compile `kbs v0.1.0 (/usr/src/kbs/kbs/src/kbs)`, intermediate artifacts can be found at `/usr/src/kbs/target`.
> 81.04 To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
> ```
> 
> I don't see a similar issue building locally. I have also tried pinning to v0.1.0 and v0.1.1 and the upstream head. Any ideas @anakrish

Hmm. I haven't seen such an issue before. The failing build command is trying to do this:
https://github.com/microsoft/regorus/blob/0ebcb568cc577602c9ac64ae491ca1e45683deaa/build.rs#L17-L21
```rust
    let output = Command::new("git")
        .args(["rev-parse", "HEAD"])
        .output()
        .unwrap();
    let git_hash = String::from_utf8(output.stdout).unwrap();
```

It is likely that `git` is not present in the container.

Can you point me to the container image so that I can try it out locally?

Xynnn007 (approved):
lgtm. cc @jialez0

fitzthum:
Ok, thanks @anakrish. I added `git` to our container image so we don't have to pick up a new version.

@jialez0 I would be curious about your feedback since you've worked on the policy stuff previously.

jialez0 (approved):
LGTM. Thanks @fitzthum

kbs/src/api/src/policy_engine/opa/mod.rs
```diff
@@ -2,43 +2,25 @@
...
-
 #[derive(Debug, Clone)]
 pub struct Opa {
     policy_path: PathBuf,
 }
 
 impl Opa {
-    pub fn new(policy_path: PathBuf) -> Result<Self> {
+    pub fn new(policy_path: PathBuf) -> Result<Self, ResourcePolicyError> {
+        std::fs::create_dir_all(policy_path.parent().unwrap())
```
mkulke, line 18: is there a reason for assuming policy_path will never be / or should we return an Error?

kbs/src/api/src/policy_engine/opa/mod.rs
```diff
@@ -2,43 +2,25 @@
...
+            .map_err(|_| ResourcePolicyError::IOError)?;
+
         if !policy_path.as_path().exists() {
-            if !policy_path.as_path().parent().unwrap().exists() {
-                std::fs::create_dir_all(policy_path.parent().unwrap())?;
-            }
             let policy = std::include_str!("default_policy.rego").to_string();
-            fs::write(&policy_path, policy)?;
+            fs::write(&policy_path, policy).map_err(|_| ResourcePolicyError::IOError)?;
```
mkulke, line 23: If we add a `from` directive for the source errors, we don't need the explicit mapping (can just use `fs::write(&policy_path, policy)?;`) and keep the context of the original error.

```rust
#[derive(Error, Debug, PartialEq)]
pub enum ResourcePolicyError {
    #[error("io error")]
    IoError(#[from] io::Error),
    ...
}
```

kbs/src/api/src/policy_engine/opa/mod.rs
```diff
@@ -51,105 +33,211 @@ impl PolicyEngineInterface for Opa {
...
+
+        // Add policy as data
+        engine
+            .add_policy_from_file(self.policy_path.clone())
+            .map_err(|_| ResourcePolicyError::PolicyLoadError)?;
+
+        // Add resource path as data
+        let resource_path_object =
+            regorus::Value::from_json_str(&format!("{{\"resource-path\":\"{}\"}}", resource_path))
+                .map_err(|_| ResourcePolicyError::ResourcePathError)?;
```
mkulke, line 47: same concern es with the other errors, we discard the information carried in the original error, which might be useful for debugging.

kbs/src/api/src/policy_engine/opa/mod.rs
```diff
@@ -51,105 +33,211 @@ impl PolicyEngineInterface for Opa {
...
-                "svn": svn.to_string(),
+                "productId": product_id,
+                "svn": svn
             }
         })
         .to_string()
     }
 
+    async fn set_policy_from_file(opa: &mut Opa, path: &str) -> Result<(), ResourcePolicyError> {
+        let policy = std::fs::read(PathBuf::from(path.to_string())).unwrap();
```
mkulke, line 98: I wouldn't resort to unwrap if the fn already returns a Result

mkulke (commented):
I have some concerns about dropping error sources, but otherwise lg!

kbs/src/api/src/policy_engine/opa/mod.rs
fitzthum: Ok, `io::Error` does not implement `PartialEq`, which breaks the asserts below. This is generally the reason why I don't have context in any of the errors, but I agree it would be better to carry more info so let me see if there is a good workaround.

kbs/src/api/src/policy_engine/opa/mod.rs
fitzthum: `create_dir_all` creates the path recursively if it does not already exist. It returns an error only if it tries to create the path and is unable to.

kbs/src/api/src/policy_engine/opa/mod.rs
fitzthum: This is a helper for a test. I pass up the error from `set_policy` so that it can be checked in the tests but if we fail to read the policy file specified in the test itself I figure we should just explode immediately.

kbs/src/api/src/policy_engine/opa/mod.rs
mkulke: ah, yes that's a bit unergonomic, but I wouldn't discard the source error because of this. there is a couple of ways, but the simplest would be to drop the `PartialEq` derive and do assertions via pattern matching or a helper macro (assert_matches):

```rust
use std::io;
use thiserror::Error;

#[derive(Error, Debug)]
pub enum MyError {
    #[error("fail")]
    IoFail(#[from] io::Error),
    #[error("sth else")]
    SthElse,
}

fn fails() -> Result<(), MyError> {
    let err = Err(io::Error::new(io::ErrorKind::NotFound, "fail"));
    err?;
    Ok(())
}

#[test]
fn test_fail() {
    let MyError::IoFail(_) = fails().unwrap_err() else {
        panic!("Expected an IoFail error");
    };
}
```

fitzthum: Ok, turns out the `matches!` macros have to match against a specific pattern so I made a `compare_errors` function instead with a big match statement in it. Still slightly weird but this is the best way I can think of. Also, it turns out that thiserror doesn't support multiple variants coming from the same source. Since regorus uses anyhow for all its errors this is slightly limiting. I opted to include the error just for the evaluate case (since i can only do one). I guess I could make one variant for all the regorus errors, but this seems fine.

mkulke (approved):

Issue #153, opened 2023-09-01
kbs crash


mythi:
With #152 included, I set up the "cluster" using `docker-compose up -d` (only kbs image modified). The following client request makes KBS to crash(?):

```
  let evidence_provider = Box::new(NativeEvidenceProvider::new().unwrap());

    let mut client = KbsClientBuilder::with_evidence_provider(evidence_provider, server.as_str())
        .add_kbs_cert(cert.as_str())
        .build()
        .unwrap();

    rt.block_on(async {
        secret = client
            .get_resource(
                format!("kbs:///foo/bar/{set_key}")
                    .as_str()
                    .try_into()
                    .unwrap(),
            )
            .await
            .unwrap();
    });
```

Client error:
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: RcarHandshake("error sending
request for url (https://localhost:8080/kbs/v0/attest): connection error: unexpected end of file")'
```

not much logs on the KBS side after

```
[2023-09-01T12:10:13Z INFO  api_server::http::resource] Get pkey from auth header
[2023-09-01T12:10:13Z INFO  actix_web::middleware::logger] 172.27.0.1 "GET /kbs/v0/resource/foo/bar/wrap-key HTTP/2.0" 401 173 "-" "attestation-agent-kbs-client/0.1.0" 0.000243
[2023-09-01T12:10:13Z INFO  api_server::http::attest] request: Json(Request { version: "0.1.0", tee: Sgx, extra_params: "" })
[2023-09-01T12:10:13Z INFO  actix_web::middleware::logger] 172.27.0.1 "POST /kbs/v0/auth HTTP/2.0" 200 74 "-" "attestation-agent-kbs-client/0.1.0" 0.000122
[2023-09-01T12:10:13Z INFO  api_server::http::attest] Cookie 64610b58b1c747e8bd42cd4e1f452576 attestation Json(Attestation { tee_pubkey: TeePubKey { kty: "RSA", alg: "RSA1_5", k_mod: " ...
...
[2023-09-01T12:10:16Z INFO  api_server] Starting HTTPS server at [0.0.0.0:8080]
```

the container PID changes which suggests it got restarted.

mythi:
NB: the same client code works with in `grpc-as` mode

Xynnn007:
@mythi Could you provide some guide to reproduce this?

mythi:
> @mythi Could you provide some guide to reproduce this?

[...]

PR #368, opened 2024-04-17
docker: Use Ubuntu 22.04 as kbs base image

mkulke:
related to: #153

When using debian:stable-slim as a base image the sgx verification libraries will crash when attempting to fetch an sgx collateral during td quote validation.

Using Ubuntu 22.04 as a base image fixes the crash, and is consistent with AS which uses the same base image.

(discussion [here](https://cloud-native.slack.com/archives/C039JSH0807/p1713270650840409?thread_ts=1713264500.545019&cid=C039JSH0807))

Also removed the tdx attest lib which is not required for verification and pointed the sgx repo to the correct ubuntu version.


mythi (approved):
LGTM, thanks! I wonder if it was good to have L14 aligned?

dcmiddle (approved):

fitzthum (approved):
LGTM

PR #375, opened 2024-04-24
k8s: docs: DCAP kustomization + non-release images

mythi:
The KBS crash problem (ee1a8cebb6) was difficult to debug with the "built-in-as" image since all the logs from AS were lost.

Make the k8s sample deployment to use AS separately. This also makes it aligned with what docker-compose offers by default.

### DIFF
--- a/kbs/config/kubernetes/README.md
+++ b/kbs/config/kubernetes/README.md
@@ -58,6 +58,16 @@ Once you deploy the KBS, you can use the services' nodeport and the Kubernetes n
 echo $(kubectl get nodes -o jsonpath='{.items[0].status.addresses[0].address}'):$(kubectl get svc kbs -n coco-tenant -o jsonpath='{.spec.ports[0].nodePort}')
 ```
 
+## Optional: Use custom Intel DCAP configuration
+
+If you would like to override the default `sgx_default_qcnl.conf` in the KBS/AS images, copy/configure one into `custom_pccs/` directory and deploy using:
+
+```bash
+export DEPLOYMENT_DIR=custom_pccs
+```
+
+NB: this currently builds on `nodeport` kustomization.
+
 ## Deploy KBS
 
 Deploy KBS by running the following command:
--- a/kbs/config/kubernetes/README.md
+++ b/kbs/config/kubernetes/README.md
@@ -44,6 +44,19 @@ kustomize edit add resource ingress.yaml
 popd
 ```
 
+## Optional: Use non-release images
+
+Sometimes it may be desirable to deploy KBS with an image that is not what is set in the repo (typically
+the latest release image). To change the deployment to use a staging build, set the image using `kustomize`:
+
+```bash
+pushd base
+kustomize edit set image kbs-container-image=ghcr.io/confidential-containers/staged-images/kbs:65ee7e1acccd13dcb515058e71c5f8bfb4281e35
+popd
+```
+
+The available image tags can be found in the [CoCo packages listing](https://github.com/orgs/confidential-containers/packages?repo_name=trustee).
+
 ## Optional: Expose KBS using Nodeport
 
 If you would like to expose KBS service using Nodeport then export the following environment variable:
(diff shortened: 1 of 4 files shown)


fitzthum:
Unfortunately I strongly prefer using the builtin AS. When the KBS and AS are separate the attack surface increases because they are communicating over the network. We have already found one issue with the API between the two and I suspect there could be more.

bpradipt:
Using an all-in-one KBS image is useful in real scenarios where one want to run it as a single binary or container image using docker/podman without using compose. The same can also be run via systemd.  So I would also prefer to have the inbuilt AS support to be tested in CI.

fitzthum:
> Do you mean this is not meant to be just a test setup for CI but something people would use for real? Previously the only argument for a built-in was just that it's easier for developers to use...

We support both but I would prefer having our sample deployment use the builtin AS because it will make users more likely to adopt that model.

mythi:
> > Do you mean this is not meant to be just a test setup for CI but something people would use for real? Previously the only argument for a built-in was just that it's easier for developers to use...
> 
> We support both but I would prefer having our sample deployment use the builtin AS because it will make users more likely to adopt that model.

As noted in the commit message, this aligns with the docker-compose sample we have. I'm fine moving this to an overlay (i.e., away from `base`). Would that work?

fitzthum:
> As noted in the commit message, this aligns with the docker-compose sample we have. I'm fine moving this to an overlay (i.e., away from base). Would that work?

I would be fine with adding an overlay that allows people to use a separate AS if they want to.

kbs/config/kubernetes/base/kustomization.yaml
```diff
@@ -4,16 +4,22 @@ namespace: coco-tenant
 
 images:
 - name: kbs-container-image
-  newName: ghcr.io/confidential-containers/key-broker-service
-  newTag: built-in-as-v0.8.2
+  newName: ghcr.io/confidential-containers/staged-images/kbs-grpc-as
+  newTag: dc01f454264fb4350e5f69eba05683a9a1882c41
+- name: as-container-image
+  newName: ghcr.io/confidential-containers/staged-images/coco-as-grpc
+  newTag: f76cf9272a0bc0be45864f4b49ba857793fd52f0
```
fitzthum, line 11: What we have discussed previously regarding these images is that the kubernetes deployment is intended for end users and such will always be pinned to the most recent release image.

The docker compose file on the other hand will always pull the latest staging images. This is for developers.

This helps simplify the release a bit. Note if you add a new overlay that depends on the release version, that you should also update the release doc and the new release script to make sure it is updated.

mythi: I could not find the release images for AS but looking at it again I can see they are available after all. I'll get this fixed tomorrow and see if I can get moved to an overlay nicely.

fitzthum: For some reason the naming convention for the release images and staged images is different.

fitzthum: https://github.com/confidential-containers/trustee/issues/377

mythi: I'm just noticing what the issue is: I first checked "packages from trustee repo" only. If I look at all the packages/images published by confidential-containers org, I can see attestation-service with 0.8.2 tags.

I believe we are not publishing attestation-services images from trustee repo, which probably needs fixing?

mythi: Ok nevermind, just saw the issue. Thanks!

mythi:
> > As noted in the commit message, this aligns with the docker-compose sample we have. I'm fine moving this to an overlay (i.e., away from base). Would that work?
> 
> I would be fine with adding an overlay that allows people to use a separate AS if they want to.

I decided to drop this idea for now since it's not super critical. I proposed the changes only because I had the work done while I was debugging the crash. Instead, I'm now proposing to add an optional doc entry on how users can use custom images if needed.

I've now verified the crash is gone with the latest staged images following the steps documented in this PR.

fidencio (approved):
lgtm, thanks @mythi!

fitzthum (approved):
NIce

PR #376, opened 2024-04-24
KBS: fix session status

### LINKED ISSUE #374: KBS: session panic
If multiple users send lots of `attest` http requests to KBS, it may result in session panic errors.

kbs logs: 
session-panic

user logs:  
`[WARN] "/kbs/v0/attest": error sending request for url (http://localhost:8080/kbs/v0/attest): connection closed before message completed`

kbs config: 

```toml
sockets = ["0.0.0.0:8080"]
auth_public_key = "/opt/confidential-containers/kbs/user-keys/public.pub"
insecure_http = true

[attestation_token_config]
attestation_token_type = "CoCo"

[grpc_config]
as_addr = "http://as:50004"
```

cc @Xynnn007

Xynnn007:
When an attested session sends a new attest request, the session should not do anything else but only return the old token if it is still valid. 

Close #374

### DIFF
--- a/kbs/src/api/src/http/attest.rs
+++ b/kbs/src/api/src/http/attest.rs
@@ -52,13 +52,31 @@ pub(crate) async fn attest(
         let session = session.get();
 
         debug!("Session ID {}", session.id());
-        let attestation_str = serde_json::to_string_pretty(&attestation.0)
-            .map_err(|_| Error::AttestationFailed("Failed to serialize Attestation".into()))?;
-        debug!("Attestation: {attestation_str}");
 
         if session.is_expired() {
             raise_error!(Error::ExpiredCookie);
         }
+
+        if let SessionStatus::Attested { token, .. } = session {
+            debug!(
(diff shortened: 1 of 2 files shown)


fitzthum (approved):
Changes look fine, but why are we calling the `/attest` endpoint multiple times in the first place? Shouldn't this only happen once per guest/token?

Xynnn007:
> Changes look fine, but why are we calling the `/attest` endpoint multiple times in the first place? Shouldn't this only happen once per guest/token?

Yes. For normal cases, `attest` endpoint would be called only once. But due to network error, retry from client (async) or some unexpected reason, one or more duplicated requests could reach. The error reported by @ChengyuZhu6 is caught when doing performance test, but I think it should be treated more gerenicly. We should better promote the robust of the code.

ChengyuZhu6 (approved):
LGTM ! Thanks for the fix @Xynnn007 ! I have tested the patch on my local machine and it worked as expected.

Issue #364, opened 2024-04-11
KBS: build s390x key-broker-service image in CI


huoqifeng:
I'd like to have s390x kbs OCI image built in CI.
A cross-platform build or by a s390x runner in GHA both works for me.

huoqifeng:
@BbolroC @magowan @stevenhorsman @bpradipt

PR #383, opened 2024-05-07
Enable artifacts for s390x

BbolroC:
This PR enables all artifacts (container images and binaries) for `s390x` alongside `x86_64`. The self-hosted runner to build them is already registered as `s390x-runner-01` to the organisation.

The change was verified with a private runner as follows:

AS: https://github.com/BbolroC/trustee/actions/runs/8982469105
KBS: https://github.com/BbolroC/trustee/actions/runs/8982469108
KBS-client: https://github.com/BbolroC/trustee/actions/runs/8982469107

The artifacts are available at:

- https://github.com/BbolroC/trustee/pkgs/container/staged-images%2Fcoco-as-grpc
- https://github.com/BbolroC/trustee/pkgs/container/staged-images%2Fcoco-as-restful
- https://github.com/BbolroC/trustee/pkgs/container/staged-images%2Fkbs
- https://github.com/BbolroC/trustee/pkgs/container/staged-images%2Fkbs-grpc-as
- https://github.com/BbolroC/trustee/pkgs/container/staged-images%2Fkbs-client

Fixes: #364

### DIFF
--- a/.github/workflows/as-build-and-push.yaml
+++ b/.github/workflows/as-build-and-push.yaml
@@ -7,10 +7,18 @@ on:
 
 jobs:
   build_and_push:
-    runs-on: ubuntu-latest
+    permissions:
+      packages: write
     strategy:
       fail-fast: false
       matrix:
+        instance:
+          - ubuntu-latest
+          - s390x
+        tag:
+          - coco-as-grpc
+          - coco-as-restful
+          - rvps
         include:
           - docker_file: attestation-service/Dockerfile.as-grpc
             tag: coco-as-grpc
@@ -21,8 +29,17 @@ jobs:
           - docker_file: attestation-service/rvps/Dockerfile
             tag: rvps
             name: RVPS
+    runs-on: ${{ matrix.instance }}
 
     steps:
+    - name: Take a pre-action for self-hosted runner
+      run: |
+        # NOTE: Use file checking instead triggering a step based on a runner type
+        # to avoid updating the step for each new self-hosted runner.
+        if [ -f "${HOME}/script/pre_action.sh" ]; then
+          "${HOME}/script/pre_action.sh" cc-trustee
+        fi
+
     - name: Checkout code
       uses: actions/checkout@v4
 
@@ -39,4 +56,57 @@ jobs:
     - name: Build ${{ matrix.name }} Container Image
       run: |
         commit_sha=${{ github.sha }}
-        DOCKER_BUILDKIT=1 docker build -t ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:${commit_sha} -t ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:latest . -f ${{ matrix.docker_file }} --push
+        arch=$(uname -m)
+        DOCKER_BUILDKIT=1 docker build -f ${{ matrix.docker_file }} --push --build-arg ARCH=${arch} \
+          -t ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:${commit_sha}-${arch} \
+          -t ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:latest-${arch} .
+
+    - name: Take a post-action for self-hosted runner
+      if: always()
+      run: |
+        # Please check out the note in the pre-action step for the reason of using file checking
+        if [ -f "${HOME}/script/post_action.sh" ]; then
+          "${HOME}/script/post_action.sh" cc-trustee
+        fi
+
+  publish_multi_arch_image:
+    needs: build_and_push
+    permissions:
+      packages: write
+    strategy:
+      fail-fast: false
+      matrix:
+        tag:
+          - coco-as-grpc
+          - coco-as-restful
+          - rvps
+        include:
+          - tag: coco-as-grpc
+            name: gRPC CoCo-AS
+          - tag: coco-as-restful
+            name: RESTful CoCo-AS
+          - tag: rvps
+            name: RVPS
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: Login to GHCR Container Registry
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Publish Multi-arch Image for ${{ matrix.name }}
+      run: |
+        commit_sha=${{ github.sha }}
+        docker manifest create ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:${commit_sha} \
+          --amend ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:${commit_sha}-s390x \
+          --amend ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:${commit_sha}-x86_64
+        docker manifest push ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:${commit_sha}
+        docker manifest create ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:latest \
+          --amend ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:latest-s390x \
+          --amend ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:latest-x86_64
+        docker manifest push ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:latest
--- a/.github/workflows/kbs-client-build-and-push.yaml
+++ b/.github/workflows/kbs-client-build-and-push.yaml
@@ -7,14 +7,28 @@ on:
 
 jobs:
   build_and_push:
+    strategy:
+      fail-fast: false
+      matrix:
+        arch:
+          - x86_64
+          - s390x
     env:
       RUSTC_VERSION: 1.76.0
-    runs-on: ubuntu-22.04
+    runs-on: ${{ matrix.arch == 'x86_64' && 'ubuntu-22.04' || 's390x' }}
     permissions:
       contents: read
       packages: write
 
     steps:
+    - name: Take a pre-action for self-hosted runner
+      run: |
+        # NOTE: Use file checking instead triggering a step based on a runner type
+        # to avoid updating the step for each new self-hosted runner.
+        if [ -f "${HOME}/script/pre_action.sh" ]; then
+          "${HOME}/script/pre_action.sh" cc-trustee
+        fi
+
     - name: Check out code
       uses: actions/checkout@v4
     - name: Install rust toolchain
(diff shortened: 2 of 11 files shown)


BbolroC:
For reviewers, this PR does not cover the full enablement for s390x. Makefiles for `AS` and `KBS` are not yet enabled because `cargo install` is used in the workflows for them. I will raise a follow-up PR to deal with this by making the affected workflows use `make`. Thanks. 😉

kbs/docker/Dockerfile
```diff
@@ -25,22 +27,24 @@ RUN curl -fsSL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.k
     wget \
     clang \
     cmake \
-    libtss2-dev \
+    libtss2-dev && \
+    if [ "${ARCH}" = "x86_64" ]; then apt-get install -y --no-install-recommends \
     libsgx-dcap-quote-verify-dev \
-    libtdx-attest-dev
+    libtdx-attest-dev; fi
```
fitzthum, line 33: I think this is fine, but really it might make more sense to connect all this TDX-specific stuff to a TDX-related flag rather than the arch. TDX could always be set to false for s390x

attestation-service/attestation-service/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-default = [ "restful-bin", "rvps-grpc", "rvps-builtin", "all-verifier" ]
+default = [ "restful-bin", "rvps-grpc", "rvps-builtin" ]
```
fitzthum, line 7: I'm on the fence about this. Maybe it would be better for us to add a dummy s390x feature or merge @huoqifeng's PR. wdyt @Xynnn007?

attestation-service/attestation-service/Cargo.toml
```diff
@@ -62,7 +62,12 @@ thiserror = { workspace = true, optional = true }
 tokio.workspace = true
 tonic = { workspace = true, optional = true }
 uuid = { version = "1.1.2", features = ["v4"] }
-verifier = { path = "../verifier", default-features = false }
+
+[target.'cfg(not(target_arch = "s390x"))'.dependencies]
+verifier = { path = "../verifier", default-features = false, features = ["all-verifier"] }
+
+[target.'cfg(target_arch = "s390x")'.dependencies]
+verifier = { path = "../verifier", default-features = false, features = ["csv-verifier"] }
```
fitzthum, line 70: So once the s390x verifier is in, will that only be able to be built (and run) on an s390x machine? Will we be able to support validating s390x evidence on x86?

fitzthum (commented):
Looks pretty good.

You might want to check the r[elease helper script](https://github.com/confidential-containers/trustee/blob/main/hack/release-helper.sh) and make sure that it will account for the multiarch images you are addding.

attestation-service/attestation-service/Cargo.toml
fitzthum: nvm i think this is fine. hadn't read the bottom part yet.

.github/workflows/kbs-client-build-and-push.yaml
```diff
@@ -7,14 +7,30 @@ on:
...
+    strategy:
+      matrix:
+        instance:
+          - ubuntu-22.04
+          - s390x
+        include:
+          - instance: ubuntu-22.04
+            arch: x86_64
+          - instance: s390x
+            arch: s390x
```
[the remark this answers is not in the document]
BbolroC: Yep, what you are suggesting seems like:

```
jobs:
  build_and_push:
    strategy:
      matrix:
        arch:
          - x86_64
          - s390x
    env:
      RUSTC_VERSION: 1.76.0
    runs-on: ${{ matrix.arch == 'x86_64' ? 'ubuntu-22.04' : 's390x' }}
```

It looks more concise and cleaner. Thanks for the suggestion.

kbs/Makefile
```diff
@@ -2,6 +2,8 @@ AS_TYPE ?= coco-as
 HTTPS_CRYPTO ?= rustls
 POLICY_ENGINE ?=
 
+ARCH := $(shell uname -m)
```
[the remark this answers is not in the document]
BbolroC: It sounds good. I will bring the check. Thanks!

attestation-service/attestation-service/Cargo.toml
BbolroC: Yes, once the s390x verifier comes in, if it is included in `all-verifier`, we could see x86 support validating s390x evidence, but not vice versa.

kbs/docker/Dockerfile
BbolroC: Yes, we could introduce a TDX variable in the build script and then set it to true/false based on the arch, even we could have more fine-grained control over different boolean variables for TEE types. But as discussed in Slack, this could be handled in a follow-up PR. wdyt, @Xynnn007 ?

Xynnn007: reasonable. In future, we need to add all plugin flags and by default set them all to true.

kbs/docker/Dockerfile
```diff
@@ -49,14 +53,14 @@ RUN apt-get update && \
...
+RUN if [ "${ARCH}" = "x86_64" ]; then curl -fsSL https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | \
+    gpg --dearmor --output /usr/share/keyrings/intel-sgx.gpg; fi
+RUN if [ "${ARCH}" = "x86_64" ]; then echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/intel-sgx.gpg] https://download.01.org/intel-sgx/sgx_repo/ubuntu jammy main' | tee /etc/apt/sources.list.d/intel-sgx.list; fi
 RUN apt-get update
-RUN apt-get install -y --no-install-recommends \
+RUN if [ "${ARCH}" = "x86_64" ]; then apt-get install -y --no-install-recommends \
     libsgx-dcap-default-qpl \
     libsgx-dcap-quote-verify \
-    tpm2-tools
+    tpm2-tools; fi
```
Xynnn007, line 63: Not related to this PR, but we need to do some optimization upon this dockerfile.

Xynnn007 (commented):
Looks great. we might need to wait #345 to be merged and then merge this. because the feature flag in Cargo.toml is not prepared I think.

huoqifeng:
> Looks great. we might need to wait #345 to be merged and then merge this. because the feature flag in Cargo.toml is not prepared I think.

@Xynnn007 might a circle dependency here, the remain work in #345 requires we can build on s390x because of the IBM SE crate. What about we merge #345 and then this PR and then add the IBM SE crate?

.github/workflows/as-build-and-push.yaml
```diff
@@ -21,8 +29,15 @@ jobs:
           - docker_file: attestation-service/rvps/Dockerfile
             tag: rvps
             name: RVPS
+    runs-on: ${{ matrix.instance }}
 
     steps:
+    - name: Take a pre-action for self-hosted runner
+      run: |
+        if [ -f "${HOME}/script/pre_action.sh" ]; then
+          "${HOME}/script/pre_action.sh" cc-trustee
```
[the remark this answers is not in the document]
BbolroC: Good point! Thanks. Yes, we can follow your suggestion. However, when a new architecture is introduced, we'll need to revisit that condition. But if we're only checking for file existence, an ARM maintainer, for example, can simply add a script to the specified location for the self-hosted runner, eliminating the need to update their integration workflow.

As discussed in Slack, I will put a short note for why the file checking is introduced above the step name. 😉

attestation-service/Dockerfile.as-grpc
```diff
@@ -3,44 +3,47 @@
...
 FROM rust:latest as builder
+ARG ARCH=x86_64
 
 WORKDIR /usr/src/attestation-service
 COPY . .
 
 # Install golang
-RUN wget https://go.dev/dl/go1.20.1.linux-amd64.tar.gz && \
-    tar -C /usr/local -xzf go1.20.1.linux-amd64.tar.gz
+RUN [ "${ARCH}" = "x86_64" ] && GOARCH="amd64" || GOARCH="${ARCH}" && \
```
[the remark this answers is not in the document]
BbolroC: Yeah, this could be confusing to those who are not familiar with bash. I will make it clearer with `if else fi`. Thanks!

kbs/docker/Dockerfile
BbolroC: Yep, we should do that as other plugins come in to build their stuff.

BbolroC:
> > Looks great. we might need to wait #345 to be merged and then merge this. because the feature flag in Cargo.toml is not prepared I think.
> 
> @Xynnn007 might a circle dependency here, the remain work in #345 requires we can build on s390x because of the IBM SE crate. What about we merge #345 and then this PR and then add the IBM SE crate?

Thanks for the comment. I also discussed it with @Xynnn007.  We ended up with waiting until #345 is merged and updating `Cargo.toml` accordingly. 😉 Let's merge yours!!

BbolroC:
> Looks pretty good.
> 
> You might want to check the r[elease helper script](https://github.com/confidential-containers/trustee/blob/main/hack/release-helper.sh) and make sure that it will account for the multiarch images you are addding.

It seems that container images will not be affected by this PR, but an oras artifact `kbc-client`, which is not handled by the script. So this PR does not hurt the release process. 😉

BbolroC:
Here are the verification result for the last update:

- AS: https://github.com/BbolroC/trustee/actions/runs/8999917487
- KBS: https://github.com/BbolroC/trustee/actions/runs/8999917482
- KBS-client: https://github.com/BbolroC/trustee/actions/runs/8999917484

portersrc (approved):

fitzthum (approved):
I'm still a bit suspicious of the release script but we can fix that up afterwards if needed.

BbolroC:
> I'm still a bit suspicious of the release script but we can fix that up afterwards if needed.

Now, I got your point. I was only considering whether the multi-arch images hurt the release helper script or not. But you wanted me to enable the script to publish the multi-arch release images. I have updated the PR. Please take a look. 😉

hack/release-helper.sh
```diff
@@ -81,21 +81,30 @@ function tag_and_push_packages() {
...
+            docker push ${ghcr_repo}/${release_pkg_name}:${release_tag_full}-${arch}
+        done
 
-        # push it (i.e. release it)
-        docker push ${ghcr_repo}/${release_pkg_name}:${release_tag_full}
+        # Publish the multi-arch manifest
+        docker manifest create ${ghcr_repo}/${release_pkg_name}:${release_tag_full} \
+            --amend ${ghcr_repo}/${release_pkg_name}:${release_tag_full}-x86_64 \
+            --amend ${ghcr_repo}/${release_pkg_name}:${release_tag_full}-s390x
+        docker manifest push ${ghcr_repo}/${release_pkg_name}:${release_tag_full}
```
portersrc, line 107: Thanks for handling this release-helper.sh change with this PR! I wonder if there was a minor bug in the earlier script, though: It never added a `latest` tag to the new release. If you want, maybe add something like that here:
```
        docker manifest create ${ghcr_repo}/${release_pkg_name}:${latest} \
            --amend ${ghcr_repo}/${release_pkg_name}:${release_tag_full}-x86_64 \
            --amend ${ghcr_repo}/${release_pkg_name}:${release_tag_full}-s390x
        docker manifest push ${ghcr_repo}/${release_pkg_name}:${latest}
```

BbolroC:
> Are we waiting on #345 for this?

Yep!

genjuro214:
@BbolroC , I'm trying to build s390x kbs images based on this PR in our internal pipeline. The problem is we don't have s390x node in the pipeline. So I have to build s390x image on x86 platform.

First, I tried docker buildx with `--platform` like:
```
docker buildx build -t ${registry}/kbs:latest --platform linux/s390x --load --build-arg ARCH=s390x --build-arg HTTPS_CRYPTO=openssl . -f kbs/docker/Dockerfile
```
It works and we don't need more changes on your PR. But the performance is very low. It takes hours to build the rust in this way.

So I also tried to use rust cross-compile support, like:
```
RUSTFLAGS=" -C linker=s390x-linux-gnu-gcc" CGO_ENABLED=1 GOOS=linux GOARCH=s390x cargo install --path kbs/src/kbs --no-default-features --features coco-as-builtin,resource,openssl,opa --target "s390x-unknown-linux-gnu"
```
It also works, but requires some changes based on your PR. I think we did similar works in [guest-components](https://github.com/confidential-containers/guest-components): 
https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/Makefile#L72

I will try those changes in my fork based on your fork, and raise PR after this PR is merged.

BbolroC:
@genjuro214 thanks for your work. I think the best resolution would be not using the self-hosted runner in terms of capability and maintenance. Why don't you create a new PR based on this for the best resolution and merge it instead of this one?

BbolroC:
@Xynnn007 Could you resume reviewing this? I have rebased this onto main as #345 is merged.

Xynnn007 (approved):
lgtm. Thanks!

Issue #386, opened 2024-05-15
AS/Verifier: Support quote verification in multi-thread in tdx under high concurrent load


ChengyuZhu6:
We have observed a bottleneck in the quote verification feature in tdx. When AS processes a high volume of quote verification requests concurrently, the The time required for quote verification is increased.

ChengyuZhu6:
cc  @arronwy @Xynnn007

PR #387, opened 2024-05-15
AS/verifier: Enhance quote verification with multi-thread support in tdx

ChengyuZhu6:
When the qvl is linked to a process, it needs to know the proper enclave loading policy. But the default policy of qvl is SGX_QL_PERSISTENT, which means that all threads will share single QvE instance.

So we set QvE loading policy to multi-thread to enhance quote verification with multi-thread support in tdx.

Fixes: #386

### DIFF
--- a/attestation-service/verifier/src/tdx/quote.rs
+++ b/attestation-service/verifier/src/tdx/quote.rs
@@ -441,7 +441,7 @@ pub async fn ecdsa_quote_verification(quote: &[u8]) -> Result<()> {
     }
 
     // get collateral
-    let _collateral = match tee_qv_get_collateral(quote) {
+    let collateral = match tee_qv_get_collateral(quote) {
         Ok(c) => {
             debug!("tee_qv_get_collateral successfully returned.");
             Some(c)
@@ -451,8 +451,6 @@ pub async fn ecdsa_quote_verification(quote: &[u8]) -> Result<()> {
             None
         }
     };
-
-    let p_collateral: Option<&[u8]> = None;
 
     // set current time. This is only for sample purposes, in production mode a trusted time should be used.
     //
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
Thanks for the optimization!

mythi:
This change is odd and I would not merge it just yet. We are explicitly doing untrusted quote verification (i.e., without QVE) because `if qve_report_info is None, the API will use QVL library to perform quote verification` so the enclave load policy should not have an impact here.

ChengyuZhu6:
> This change is odd and I would not merge it just yet. We are explicitly doing untrusted quote verification (i.e., without QVE) because `if qve_report_info is None, the API will use QVL library to perform quote verification` so the enclave load policy should not have an impact here.

Hey @mythi. Initially, I held this belief. 
However, in DCAP code, both QVE and QVL utilize the function `tee_verify_evidence`. It’s important to note that the configuration of the policy will influence untrusted quote verification as well. https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/45554a754ba8c03342cc394831fa7f04db08805c/QuoteVerification/dcap_quoteverify/sgx_dcap_quoteverify.cpp#L513

mythi:
So this PR is just hiding a bug that is on DCAP side?

ChengyuZhu6:
> So this PR is just hiding a bug that is on DCAP side?

I think it's not to hide the bug but to mitigate the impact on our end while the DCAP team addresses the issue. We’ve documented the underlying bug and intend to communicate it to the DCAP team.

mythi:
> > So this PR is just hiding a bug that is on DCAP side?
> 
> I think it's not to hide the bug but to mitigate the impact on our end while the DCAP team addresses the issue. We’ve documented the underlying bug and intend to communicate it to the DCAP team.

Thanks! That's what I was going to ask next. Please submit an issue in https://github.com/intel/SGXDataCenterAttestationPrimitives/issues and our code here should refer to that because current implementation is not logical at all.

Issue #388, opened 2024-05-15
Rename Container Image Build Process from gRPC to RESTful


ChengyuZhu6:
The current name `Build gRPC AS Container Image` on CI suggests that we are building a gRPC-based AS docker image, while in fact, it is a RESTful AS image. This could lead to confusion and should be rectified.

PR #389, opened 2024-05-15
workflows: Rename Docker build step from gRPC to RESTful

ChengyuZhu6:
Updates the naming in the CI build step to accurately reflect the RESTful AS image, replacing the previous gRPC name.

Fixes: #388


Xynnn007 (approved):
Nice catch!

PR #390, opened 2024-05-17
AS | Refactor the policy module

Xynnn007:
This PR mainly includes the following work
- Replace golang OPA with `regorus` in CoCo AS.
- Add some documents for the policy of CoCo AS.
- Some refactoring for policy management, mainly for user-friendly
- Make SNP's runtime data check optional for CoCoAS, which is enforced when the requester is KBS.

I also did some performance test with SNP evidence check. After replacing golang OPA with `regorus`, the throughput was increased by approximately 200%.

cc @jialez0


ChengyuZhu6:
Nice patch!

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -56,6 +40,12 @@ impl OPA {
 
         Ok(Self { policy_dir_path })
     }
+
+    fn is_validated_policy_id(policy_id: &str) -> bool {
```
fitzthum, line 44: nit: `is_validated_policy_id` -> `is_valid_policy_id`

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -88,82 +81,56 @@ impl PolicyEngine for OPA {
...
-            // treated as succees.
-            let allow = res_kv
-                .get("allow")
-                .map(|a| a.as_bool())
-                .map(|a| a.unwrap_or(true))
-                .unwrap_or(true);
+            // Add TCB claims as input
+            engine.set_input_json(&input).context("set input")?;
 
+            let allow = engine.eval_bool_query("data.policy.allow".to_string(), false)?;
```
fitzthum, line 99: I would prefer supporting policies with arbitrary return value not just true/false. I guess we might be overhauling this to support EAR anyway tho

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -88,82 +81,56 @@ impl PolicyEngine for OPA {
...
             if !allow {
-                bail!("TEE evidence does not pass policy {policy_id}, reason: {policy_res}");
-            } else {
-                let evaluation_result =
-                    serde_json::from_str(&policy_res).context("serialize OPA result")?;
-                res.insert(policy_id.to_owned(), (policy_hash, evaluation_result));
+                bail!("TEE evidence does not pass policy {policy_id}");
```
fitzthum, line 101: Should we bail here or return an evaluation report that shows that the policy was false?

attestation-service/tests/coco-as/policy/example-1.rego
fitzthum, line 1: We might want to do `import rego.v1` for the examples.

attestation-service/tests/coco-as/policy/restful-check.json
fitzthum, line 1: is this the same as the grpc one above? maybe we could share

fitzthum (commented):
Looks good. A few comments.

attestation-service/verifier/src/snp/mod.rs
```diff
@@ -92,16 +92,15 @@ impl Verifier for Snp {
...
+            let expected_report_data =
+                regularize_data(expected_report_data, 64, "REPORT_DATA", "SNP");
 
-        if expected_report_data != report.report_data {
-            bail!("Report Data Mismatch");
-        }
+            if expected_report_data != report.report_data {
+                bail!("Report Data Mismatch");
+            }
+        };
```
fitzthum, line 103: Maybe we should add a warning if no expected report data is provided?

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -88,82 +81,56 @@ impl PolicyEngine for OPA {
...
-            let decision_buf: *mut c_char =
-                unsafe { evaluateGo(policy_go, reference_go, input_go) };
-            let decision_str: &CStr = unsafe { CStr::from_ptr(decision_buf) };
-            let policy_res = decision_str.to_str()?.to_string();
-            debug!("OPA Evaluated: {}", policy_res);
-            if policy_res.starts_with("Error::") {
-                bail!("OPA verification failed: {policy_res}");
-            }
+            // Add policy as data
+            engine
```
anakrish, line 85: Since `policy` has already been read from file, you could use 
[Engine::add_policy](https://docs.rs/regorus/0.1.5/regorus/struct.Engine.html#method.add_policy) to avoid a reread from disk.

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
Xynnn007: Yes. Maybe we can change this in the PR that supports EAR later. For now the main work is to replace golang with regorus without any logic change, e.g. return value with true/false.

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
Xynnn007: Does the first parameter `path` result in a fs reading?

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
Xynnn007: The error handling here will finally be raised to a [RPC call error](https://github.com/confidential-containers/trustee/blob/main/attestation-service/attestation-service/src/bin/grpc/mod.rs#L195) for gRPC AS or [an internal error](https://github.com/confidential-containers/trustee/blob/main/attestation-service/attestation-service/src/bin/restful/mod.rs#L23-L30) for restful AS. If the policy all pass, an attestation token will be returned. So if we want a "failure report" to be returned, it would be practical to embed the reasons inside the error body returned to the client.

I did not find an API to gather the reason (like concrete field unmatched info) when a policy fails to check, thus only return the error log id. @anakrish Could you help with this?

attestation-service/tests/coco-as/policy/example-1.rego
Xynnn007: Updated and also default policies.

attestation-service/tests/coco-as/policy/restful-check.json
Xynnn007: Ther are different. The only place is the `tee` field. Restful uses a string while grpc uses a number enum

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
anakrish: No. It is intended to be the path/name of the file from which the policy string came, but you can pass in any string.
It is mainly used in error messages to indicate which policy file the error originated from. In the example below, `mypolicy.rego` was passed in as the path.
```
Error: 
--> mypolicy.rego:3:11
  |
3 | x = count1(input.data)
  |           ^
error: could not find function count1
```

Xynnn007: Thanks!

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
anakrish: If more information is needed, then the policy must be written to provide additional information. For example, the policy in [rego tutorial](https://www.openpolicyagent.org/docs/latest/#2-try-opa-eval) not only has a rule for `allow`, but also has rules to capture the `violations`. 

To avoid executing two rules (`allow`, `violations`) separately, you could instead add a `violations` rule and rewrite the `allow` rule as below

```rego

violations := [ [k,v] | # Gather k, v and any other information.
   some k, v in input
   not judge_field(k, v)
]

allow := {
  "result": count(violations) == 0,
   "violations": violations
}

```
And use [eval_rule](https://docs.rs/regorus/0.1.5/regorus/struct.Engine.html#method.eval_rule) to evaluate `"data.policy.allow"`.

Another example is https://github.com/open-telemetry/weaver/blob/4401e3d2603121e6c21d0b4c6b47d8ef45151505/crates/weaver_checker/data/policies/otel_policies.rego#L13 where violations are gathered as json objects.

jialez0 (approved):
LGTM.

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
mkulke: I'm not sure we should use an Error for modelling control-flow of what seems to be valid code path: A policy doesn't pass validation. it's something that is expected and not really an internal error.

As far as I understand the semantics of this evaluate() fn seems to be:

- verify an input body against a set of ref values and a list of policies
- return a list of policy ids with their sha384 at eval time
- abort early on first failed validation

(maybe add a fn doc statement to make the behaviour clear obvious)

So, I think a signature like this could make sense?

```rust
    pub enum EvalOutcome {
       Passed(HashMap<String, PolicyDigest>),
       FailedWith { policy_id: String },
    }

    async fn evaluate(...) -> Result<EvalOutcome> {
    ...
    }
```

attestation-service/attestation-service/src/policy_engine/mod.rs
```diff
@@ -35,38 +22,25 @@ impl PolicyEngineType {
...
-    ) -> Result<HashMap<String, (PolicyDigest, EvaluationResult)>>;
+    ) -> Result<HashMap<String, PolicyDigest>>;
 
-    async fn set_policy(&mut self, input: SetPolicyInput) -> Result<()>;
+    async fn set_policy(&mut self, policy_id: String, policy: String) -> Result<()>;
 
-    async fn list_policies(&self) -> Result<Vec<PolicyListEntry>>;
+    /// The result is tuples. The first member is the policy id, and the
+    /// second member is the digest of the policy (using **Sha384**).
+    async fn list_policies(&self) -> Result<Vec<(String, PolicyDigest)>>;
```
mkulke, line 43: any reason why we replace the structs with tuples?

attestation-service/docs/policy.md
```diff
@@ -0,0 +1,84 @@
+# Attestation Policy (Developing)
+
+CoCo AS provides a flexible policy support based on OPA to facilitate the customized verification rules.
```
mkulke, line 3: Is it still accurate to call refer to OPA or should it be "based on Rego"?

fitzthum (approved):
LGTM.

There are still some possible refinements, but maybe we should handle that separately. We might be making some major changes here anyway if we are going to support EAR,

attestation-service/docs/policy.md
```diff
@@ -0,0 +1,84 @@
+# Attestation Policy (Developing)
+
+CoCo AS provides a flexible policy support based on OPA to facilitate the custormized verification rules.
```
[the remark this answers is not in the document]
Xynnn007: Nice idea

attestation-service/attestation-service/src/policy_engine/mod.rs
Xynnn007: I once think the struct definition is deep inside the lib code, which would make trouble for client developers. Maybe we can move the definition into only `restful-as.rs`, just above the api handler function. Hopefully which will make the code more readable?

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
Xynnn007: > I'm not sure we should use an Error for modelling control-flow of what seems to be valid code path: A policy doesn't pass validation. it's something that is expected and not really an internal error.
> 
> As far as I understand the semantics of this evaluate() fn seems to be:
> 
> * verify an input body against a set of ref values and a list of policies
> * return a list of policy ids with their sha384 at eval time
> * abort early on first failed validation
> 
> (maybe add a fn doc statement to make the behaviour clear obvious)
> 
> So, I think a signature like this could make sense?
> 
> ```rust
>     pub enum EvalOutcome {
>        Passed(HashMap<String, PolicyDigest>),
>        FailedWith { policy_id: String },
>     }
> 
>     async fn evaluate(...) -> Result<EvalOutcome> {
>     ...
>     }
> ```

I vote for some docs upon the function for now.

Overall return enum makes sense to both tell the programmer that the function of this and the result. However a tricky thing is that if any policy does not pass, we should raise an error to the caller due to current logic.

The internal error thing is a workaround because all the error types inside attestation-service lib is `anyhow`, and we can easily map it to a HTTP error type [here](https://github.com/confidential-containers/trustee/blob/main/attestation-service/attestation-service/src/bin/restful/mod.rs#L18). Up to now it looks fine to me, as all errors we can just treat as `internal error`. Or, we can extend the successful response body format to support either a successful token, or a failure information in later PR.

attestation-service/docs/policy.md
Xynnn007: Based on rego sounds good!

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
mkulke: I agree it's a good idea to return an error to the caller, it's just the wrong error, IMO. "internal error" is reseved to signal the client that something is broken on the server, i.e. the client cannot do anything about it ([grpc](https://grpc.github.io/grpc/core/md_doc_statuscodes.html), [rest](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) semantics)

> Internal errors. This means that some invariants expected by the underlying system have been broken. This error code is reserved for serious errors.

> 500 Internal Server Error server error response code indicates that the server encountered an unexpected condition that prevented it from fulfilling the request.

if a policy doesn't pass or the client calls the with `policy_ids=vec!(["does-not-exist"])`, it's not an internal error, it's a user error.

Xynnn007: Right. we need to define a more generalized 200 response, with either token or error info.

mkulke: we can still return an error, but we should maybe discern:
- user errors (e.g. policy doesn't exist => 400)
- internal errors (e.g. cannot perform ls in /opt/... => 500)
- result of the validation (the upper-layer logic, probably in get_token() that is calling the fn should decide whether fail-to-pass is good or bad, so it could be either 400 or 200 for a client)

attestation-service/attestation-service/src/policy_engine/mod.rs
mkulke: > I once think the struct definition is deep inside the lib code, which would make trouble for client developers.

if the fn is public it's possible to [re-export](https://doc.rust-lang.org/reference/items/use-declarations.html#use-visibility) such a struct to make it appear on the module of the fn:

```rust
pub use crate::some::where::deep::PolicyListEntry;
```

fitzthum: I wasn't a huge fan of those structs tbh. They seemed a bit unnecessary to me.

Xynnn007: @mkulke I hope that the current PR looks good to you. we can do some change upon error handling in another PR, as this PR does not focus on/change that behavior. Feel free to give any more suggestions or merge it.

mkulke: 👍 I added #394 to track it

PR #398, opened 2024-05-28
tdx: sgx: Bump DCAP dependency

fidencio:
Background motivation: Building Trustee on CentOS 9 stream will break, due to an incompatibility of rust-bindgen and clang, as shown by the error below:
```
  clang diag: warning: .: 'linker' input unused [-Wunused-command-line-argument]
  thread 'main' panicked at /home/cloud-user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bindgen-0.60.1/src/ir/context.rs:861:9:
  "_sgx_ql_qv_supplemental_t_union_(anonymous_at_/usr/x86_64-intel-sgx/include/sgx_qve_header_h_95_5)" is not a valid Ident
warning: build failed, waiting for other jobs to finish...
make: *** [Makefile:22: grpc-as] Error 101
```

The rust-bindgen version causing this issue was coming from the SGXDataCenterAttestationPrimitives repo.  With that in mind, and considering that the DCAP version used so far is 1+ years old, let's bump it and solve those two issues at the same time.

Unfortunately, we're also adapting the code a little bit due to the API changes between 1.16 and 1.21.  Due to the bump, we're also updating the the container used to build the project, which brings in a newer version of clang.


mkulke (approved):
lgtm

PR #400, opened 2024-05-29
kbs-client: encode policies with nopad-url-b64

mkulke:
The underlying kbs code expects a string in a specific b64 encoding. The client will fail to set certain policies if standard b64 encoding is used.

Not sure what the reasoning behind the b64 encoding of a policy body is, but it is very deep into the call graph in the the policy engines' [implementation](https://github.com/confidential-containers/trustee/blob/4b0b47f382e1fca216462dd7ae0764f8da056b36/kbs/src/api/src/policy_engine/opa/mod.rs#L62), so it's reasonable to change it at the client, I think.

### DIFF
--- a/kbs/tools/client/src/lib.rs
+++ b/kbs/tools/client/src/lib.rs
@@ -5,7 +5,7 @@
 //! KBS client SDK.
 
 use anyhow::{anyhow, bail, Result};
-use base64::engine::general_purpose::STANDARD;
+use base64::engine::general_purpose::URL_SAFE_NO_PAD;
 use base64::Engine;
 use jwt_simple::prelude::{Claims, Duration, Ed25519KeyPair, EdDSAKeyPairLike};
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
LGTM and thanks for this. This is great we are facing the bombs deployed before and try to dig them out although the process might be tough. In history, we are mixing STANDARD and URL_SAFE_NO_PAD alphabets and we should pay more attention for this in future.

mkulke:
> we are facing the bombs deployed before

😅, it's been confusing to debug why some policies would fail, but it should soon be covered by some tests hopefully, as I stumbled over this while refining some tests. It's probably a good idea to verify such input expectations earlier on the api level.

fitzthum (approved):
Ah I think this explains some of the weird behavior that I had seen in the past.

PR #401, opened 2024-05-29
CI: set expected tee in policy within the kbs e2e test

mkulke:
The e2e tests used to remove the tee != sample restriction from the policy prior to testing. Since we want to avoid using the sample attester accidentally, we can set the expected TEE in the policy.

The TEE e2e tests would fail here, because the interface of the workflow has been changed (introduced mandatory `tee` input) and it will be executed with the workflow from the main branch.

I ran the workflow is a discrete branch to test [here ](https://github.com/confidential-containers/trustee/actions/runs/9284165792/job/25545957778)(using `tee: aztdxvtpm`)


fitzthum (approved):
LGTM. You can rebase to fix the padding issue.

fidencio (approved):
lgtm, thanks @mkulke!

PR #404, opened 2024-06-05
AS | Fix SGX verifier & Optimization

Xynnn007:
This PR includes work
1. Fix a bug that the retrieved collateral was not used by DCAP verification
2. Optimize the performance of SGX quote by set DCAP policy
3. some lint fixes

cc @arronwy

### DIFF
--- a/attestation-service/verifier/src/sgx/mod.rs
+++ b/attestation-service/verifier/src/sgx/mod.rs
@@ -11,13 +11,14 @@ use std::{
 use anyhow::*;
 use async_trait::async_trait;
 use base64::Engine;
+use intel_tee_quote_verification_rs::{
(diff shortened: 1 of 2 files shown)


arronwy (approved):
Thanks @Xynnn007 , verified with this patch, benchmark results at AS API level can achieve our performance target.

mythi:
[0bf1b35](https://github.com/confidential-containers/trustee/pull/404/commits/0bf1b35713c103fe247fc5ff96ff6e4821018be0) seems to be the same change we did for TDX two weeks back. It made me wonder would be make sense to unify the DCAP quote steps between the two TEEs to deduplicate code. We have `pub async fn ecdsa_quote_verification()` which looks similar in two places.

mythi (approved):
lgtm, we can work on harmonizing dcap code later

jialez0 (approved):

Issue #151, opened 2023-08-29
`Nonce` is not used in Amber AS


Xynnn007:
https://github.com/confidential-containers/kbs/blob/main/src/api/src/attestation/amber/mod.rs#L51

The `nonce` parameter is a crytical parameter in KBS RCAR protocol. See https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#challenge

It helps to avoid reply attack from the attestation-agent side.

Currently Amber attestation service does not check the nonce field, thus having some secure problems.

cc @Lu-Biao

mythi:
I'm looking into this.

PR #406, opened 2024-06-07
intel-trust-authority-as: add runtime data to attestation request

mythi:
Closes: #151 

By adding runtime data to the appraisal request and having the reportdata
correctly hashed in the quote, ITA returns it back in the token claims
under attester_runtime_data.

For this to work, the Kata rootfs must be built with a modified
guest-components with sha512 hashing:

```    
--- a/attestation-agent/kbs_protocol/src/client/rcar_client.rs
+++ b/attestation-agent/kbs_protocol/src/client/rcar_client.rs
@@ -13,7 +13,7 @@ use log::{debug, warn};
 use resource_uri::ResourceUri;
 use serde::Deserialize;
 use serde_json::json;
-use sha2::{Digest, Sha384};
+use sha2::{Digest, Sha512};
    
 use crate::{
     api::KbsClientCapabilities,
@@ -189,7 +189,7 @@ impl KbsClient<Box<dyn EvidenceProvider>> {
         nonce: String,
     ) -> Result<String> {
         debug!("Challenge nonce: {nonce}");
-        let mut hasher = Sha384::new();
+        let mut hasher = Sha512::new();
         hasher.update(runtime_data);

         let ehd = match tee {
```    

Otherwise, ITA responds 400 / bad request.
    
This change is still safe because ITA AS with KBS get-resource isn't working without this either.

### DIFF
--- a/kbs/src/api/src/attestation/intel_trust_authority/mod.rs
+++ b/kbs/src/api/src/attestation/intel_trust_authority/mod.rs
@@ -5,10 +5,12 @@
 use super::Attest;
 use anyhow::*;
 use async_trait::async_trait;
+use base64::{engine::general_purpose::STANDARD, Engine};
 use jsonwebtoken::{decode, decode_header, jwk, Algorithm, DecodingKey, Validation};
 use kbs_types::{Attestation, Tee};
 use reqwest::header::{ACCEPT, CONTENT_TYPE};
 use serde::{Deserialize, Serialize};
+use serde_json::json;
 use std::fs::File;
 use std::io::BufReader;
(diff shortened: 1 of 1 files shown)


fitzthum (approved):
LGTM

I assume the changes you mention for guest components are just an example. We'll need to come up with a better way to negotiate the hashing method.

mythi:
> We'll need to come up with a better way to negotiate the hashing method.

Yes, it's just an example/test setup that makes, e.g., `kbs-client` to work. Proper fixing will be implemented as part of #242

PR #408, opened 2024-06-12
AS/verifier: support AA eventlog in TDX

Xynnn007:
This is an alignment with guest-components side AA eventlog.

See https://github.com/confidential-containers/guest-components/pull/548

cc @arronwy @binxing @mythi


attestation-service/verifier/src/eventlog/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    pub fn to_parsed_claims(&self) -> Map<String, Value> {
+        let mut aael = Map::new();
+        for eventlog in &self.events {
+            let key = format!("{}/{}", eventlog.domain, eventlog.operation);
+            let item = Value::String(eventlog.content.clone());
+            match aael.get_mut(&key) {
+                Some(value) => value
+                    .as_array_mut()
+                    .expect("Only array can be inserted")
```
mkulke, line 154: should we panic here?

attestation-service/verifier/src/eventlog/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        let hash_algorithm = hash_algorithm
+            .try_into()
+            .context("parse Hash Algorithm in INIT entry")?;
+        let init_state = hex::decode(init_state).context("parse init state in INIT entry")?;
+
+        let events = eventline
+            .iter()
+            .map(|line| AAEvent::from_str(line))
+            .collect::<Result<Vec<AAEvent>>>()?;
```
mkulke, line 79: can be a little less intimidating when written as `let events: Vec<AAEvent> = eventline.iter.map(AAEvent::from_str).collect()?;`

attestation-service/verifier/src/eventlog/mod.rs
Xynnn007: This is some logic that will never panic. The initial `aael` is empty. If a key is `None`, a `Value::Array` will be inserted. Thus whenever a `Some` is found, it must be a `Value::Array`.

Here, the same `operation` could have multiple different events, thus we make all different events an `Array` of a same `domain/operation`.

attestation-service/verifier/src/eventlog/mod.rs
Xynnn007: My compiler will warn for this.
`line` is `&&str` but `AAEvent::from_str` only accepts `&str`. Thus we need to explicitly write the calling routine rather than simply the name of function to leverage auto dereference.

Also, https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect gives exampl. Once a `map` fails, the returned will be the error. It is very useful but looks usual.

mkulke: ah 2x yes:

it needs to be `into_iter()` to not get a double ref, I guess.
I didn't spot the result short circuit, I think turbofish is required  here, if you want to squeeze it it one line.

I would consider renaming the variable though, it should not be singular (eventline), it looks like we are iterating over characters of a str otherwise. maybe you can call the initial variable different:

```rust
        let all_lines: Vec<&str> = input.lines().collect();

        let (initline, eventlines) = event_lines
            .split_first()
            .ok_or(anyhow!("at least one line should be included in AAEL"))?;
```

attestation-service/verifier/src/eventlog/mod.rs
mkulke: mmh I understand that it would be an internal (programmer) error that should never happen. future refactorings might modify the logic and introduce breakage. and then it's often nicer to look at an error log line rather than a crashed AS (at least that's my experience with the tdx attester). But yeah, that's arguable whether we want to panic on internal errors, I don't think we follow a common practice here, so i'm fine with either.

Xynnn007: You are right that risks will hide here. Let me add a comment to tell the future refactoring writers.

attestation-service/verifier/src/eventlog/mod.rs
Xynnn007: Yes, let me rename the variables.  `into_iter()` still not work ; |

Xynnn007:
btw, @mkulke do you think we should make a spec for CoCo about typical events, e.g. `CreateContainer`?

mkulke:
> btw, @mkulke do you think we should make a spec for CoCo about typical events, e.g. `CreateContainer`?

I think we probably want to have a naming convention for `Domain` and `Operation`, but I'm not sure what typical events would be, in practice we probably follow the kata-agent API somehow, but the content of the logline depends on the payload:

e.g. `github.com/confidential-containers CreateContainer docker.io/library/alpine` implies that we extract the Image from a CreateContainer RPC payload, for `SetPolicy` the policy digest would be a meaningful content, etc...

So, I would suggest to create an RFC issue to discuss whether there should be a schema for CoCo events.

mkulke (approved):

Issue #394, opened 2024-05-23
AS: PolicyEngine should differentiate user and internal errors


mkulke:
At the moment we conflate user errors (policy fails to pass, requested policy does not exist) with internal errors (io errors). Signatures and code of those functions should be adapted, so that grpc/http api layers can produce different error codes instead of returning an "internal error" in all cases.

[prior discussion](https://github.com/confidential-containers/trustee/pull/390/files#r1605278778)

PR #409, opened 2024-06-12
opa: Refactor opa module errors

kartikjoshi21:
Fixes: #394

### DIFF
--- a/attestation-service/attestation-service/src/policy_engine/mod.rs
+++ b/attestation-service/attestation-service/src/policy_engine/mod.rs
@@ -1,3 +1,4 @@
+use crate::policy_engine::opa::RegoError;
 use anyhow::Result;
 use async_trait::async_trait;
 use serde::Deserialize;
@@ -37,13 +38,13 @@ pub trait PolicyEngine {
         reference_data_map: HashMap<String, Vec<String>>,
         input: String,
         policy_ids: Vec<String>,
-    ) -> Result<HashMap<String, PolicyDigest>>;
+    ) -> Result<HashMap<String, PolicyDigest>, RegoError>;
 
-    async fn set_policy(&mut self, policy_id: String, policy: String) -> Result<()>;
+    async fn set_policy(&mut self, policy_id: String, policy: String) -> Result<(), RegoError>;
 
     /// The result is a map. The key is the policy id, and the
     /// value is the digest of the policy (using **Sha384**).
-    async fn list_policies(&self) -> Result<HashMap<String, PolicyDigest>>;
+    async fn list_policies(&self) -> Result<HashMap<String, PolicyDigest>, RegoError>;
 
-    async fn get_policy(&self, policy_id: String) -> Result<String>;
+    async fn get_policy(&self, policy_id: String) -> Result<String, RegoError>;
 }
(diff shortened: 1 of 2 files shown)


attestation-service/attestation-service/src/policy_engine/mod.rs
```diff
@@ -13,6 +15,40 @@ pub enum PolicyEngineType {
     OPA,
 }
 
+#[derive(Error, Debug)]
+pub enum OpaError {
+    #[error("Failed to create policy directory: {0}")]
```
Xynnn007, line 20: I prefer to add an empty line inbetween error enums

Xynnn007 (commented):
Nice PR. I think we need `cargo fmt` to make lint checker happy.

fitzthum (approved):
LGTM, but gotta fixup the linter

attestation-service/attestation-service/src/policy_engine/mod.rs
```diff
@@ -13,6 +15,40 @@ pub enum PolicyEngineType {
     OPA,
 }
 
+#[derive(Error, Debug)]
+pub enum OpaError {
```
mkulke, line 19: I think this shold be in the opa module, no?

attestation-service/attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -2,40 +2,39 @@
...
+use anyhow::{Context, Result};
 use async_trait::async_trait;
 use base64::Engine;
 use sha2::{Digest, Sha384};
 use std::collections::HashMap;
 use std::fs;
 use std::path::PathBuf;
 
-use super::{PolicyDigest, PolicyEngine};
+use super::{OpaError, PolicyDigest, PolicyEngine};
```
mkulke, line 13: it's inconsistent, but should we maybe calling it "RegoError", since we are using regorus now and directly map its errors?

attestation-service/attestation-service/src/policy_engine/mod.rs
kartikjoshi21: Renamed it to RegoError and keeping it here, so that we dont have to include opa module things here.

attestation-service/attestation-service/src/policy_engine/mod.rs
kartikjoshi21: Fixed the linter , This error enum will look pretty big if i add empty line between each error ? I see other files are not using empty lines as well, should we follow the same here ?

Xynnn007: Yeah, we can keep it as-is in this PR and re-organize that later altogether.

attestation-service/attestation-service/src/policy_engine/mod.rs
mkulke: I would still move it into the `opa` module, rego == opa, the module name is not 100% correct anymore since we replaced OPA with regorus. the errors are specific to the `opa` module. e.g. if we move opa/rego code behind a feature toggle at a later point, we don't have to untangle the rego errors from the generic `policy_engine` mod, also.

mkulke (approved):

PR #412, opened 2024-06-13
chore: bump guest-components and reqwest

mythi:
updating kbs-client alone is not enough but it also depends on reqwest 0.12.


fitzthum (approved):
LGTM

PR #414, opened 2024-06-13
kbs: simplify tee-pubkey reading from the attestation token

mythi:
get-resource API searches `tee-pubkey` claim from the attestation token. Instead of doing the search by ourselves, simply pass the desired path and leverage `pointer()` method to do the search. This simplifies the code when the search path needs to be made configurable.

One such case as of today is when attestation service provided tokens carry `tee-pubkey` under a different claims structure. We piggyback the existing cargo features to let each AS implementation define their own search paths compile time.

### DIFF
--- a/kbs/src/api/src/attestation/mod.rs
+++ b/kbs/src/api/src/attestation/mod.rs
@@ -14,6 +14,11 @@ use intel_trust_authority::*;
 use kbs_types::{Challenge, Tee};
 use rand::{thread_rng, Rng};
 
+#[cfg(not(feature = "intel-trust-authority-as"))]
+pub const AS_TOKEN_TEE_PUBKEY_PATH: &str = "/customized_claims/runtime_data/tee-pubkey";
+#[cfg(feature = "intel-trust-authority-as")]
+pub const AS_TOKEN_TEE_PUBKEY_PATH: &str = "/attester_runtime_data/tee-pubkey";
+
 #[cfg(feature = "coco-as")]
 #[allow(missing_docs)]
 pub mod coco;
--- a/kbs/src/api/src/http/resource.rs
+++ b/kbs/src/api/src/http/resource.rs
@@ -19,6 +19,9 @@ use crate::raise_error;
 
 use super::*;
 
+#[cfg(feature = "as")]
+const TOKEN_TEE_PUBKEY_PATH: &str = AS_TOKEN_TEE_PUBKEY_PATH;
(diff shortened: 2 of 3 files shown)


fitzthum:
Can you add PR description and some commit messages?

mythi:
> Can you add PR description and some commit messages?

Done

kbs/src/api/src/attestation/mod.rs
```diff
@@ -14,6 +14,11 @@ use intel_trust_authority::*;
 use kbs_types::{Challenge, Tee};
 use rand::{thread_rng, Rng};
 
+#[cfg(any(feature = "coco-as-builtin", feature = "coco-as-grpc"))]
```
fitzthum, line 17: Should this just be `#[cfg(not(feature = "intel-trust-authority-as"))]`?

fitzthum (commented):
Just one comment.

I wonder if we should incorporate the variable pubkey path into the Attest trait at some point. This seems fine for now.

kbs/src/api/src/http/resource.rs
```diff
@@ -19,6 +19,11 @@ use crate::raise_error;
 
 use super::*;
 
+#[cfg(feature = "as")]
+const TOKEN_TEE_PUBKEY_PATH: &str = AS_TOKEN_TEE_PUBKEY_PATH;
+#[cfg(not(feature = "as"))]
+const TOKEN_TEE_PUBKEY_PATH: &str = "/customized_claims/runtime_data/tee-pubkey";
```
Xynnn007, line 25: There are two places defining TOKEN_TEE_PUBKEY_PATH, like [before](https://github.com/confidential-containers/trustee/pull/414/files#diff-b83a2a4e4135ca8f4193335a1a8d660b5f1dadf86f8cad301b44c4f2a6867a2eR20). Is it possible to define only on one place and use a reference on the other side?

mythi: I wanted it to be like that but I found the problem that `TOKEN_TEE_PUBKEY_PATH` needs to be defined even completely without Attest  (i.e., `as` not enabled) so some "default" must be kept here.

kbs/src/api/src/attestation/mod.rs
mythi: that works also. the proposed `cfg` was a copy-paste from the other coco-as* configs in this file.

fitzthum (approved):
LGTM

PR #418, opened 2024-06-14
ibmse: add debug_assertions for debug and release branch

huoqifeng:
Add `#[cfg(debug_assertions)]` to make sure non-release code never built in release build


fitzthum (commented):
I think I might prefer if you kept the env var but only had it be an option in the debug case rather than always skipping verification in the debug builds.

PR #420, opened 2024-06-17
ibmse: add debug_assertions for debug and release branch

huoqifeng:
Add #[cfg(debug_assertions)] to make sure non-release code never built in release build

### DIFF
--- a/attestation-service/verifier/src/se/README.md
+++ b/attestation-service/verifier/src/se/README.md
@@ -44,7 +44,7 @@ openssl pkey -in kbs.key -pubout -out kbs.pem
 
 ## Build KBS
 ```
-cargo install --locked --path kbs/src/kbs --no-default-features --features coco-as-builtin,openssl,resource,opa 
+cargo install --locked --debug --path kbs/src/kbs --no-default-features --features coco-as-builtin,openssl,resource,opa 
 ```
 
 ## (Option 1) Launch KBS as a program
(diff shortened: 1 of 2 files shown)


huoqifeng:
https://github.com/confidential-containers/trustee/pull/418 was closed, I guess it's because of my force push? I created this new PR to track it.

huoqifeng:
@Xynnn007 @fitzthum

attestation-service/verifier/src/se/README.md
```diff
@@ -107,6 +107,7 @@ export SE_SKIP_CERTS_VERIFICATION=true
 ```
 DOCKER_BUILDKIT=1 docker build -t ghcr.io/confidential-containers/staged-images/kbs:latest --build-arg KBS_FEATURES=coco-as-builtin,openssl,resource,opa . -f kbs/docker/Dockerfile
 ```
+>Note: Please add `--debug` in statement like `cargo install` in file `kbs/docker/Dockerfile` if you're using a development host key document to skip HKD's signature verification.
```
fitzthum, line 110: And the env var must be set

fitzthum (approved):
One comment regarding the docs but otherwise LGTM

Xynnn007 (approved):
Seems that `debug_assertions` would only affact building time code https://doc.rust-lang.org/reference/conditional-compilation.html#debug_assertions, hinting that we do not explicit set debug env when running? cc @fitzthum

attestation-service/verifier/src/se/README.md
huoqifeng: @fitzthum for developer, I think we can just revise commands like `cargo install ...` in file `kbs/docker/Dockerfile` directly?

fitzthum:
> Seems that debug_assertions would only affact building time code

Yes, but I don't think we should be skipping certs whenever debug is enabled. Instead having debug enabled should allow the option to skip the certs.

Issue #419, opened 2024-06-14
kbs build warning when no `as` features are enabled


mythi:
This is seen in "KBS e2e with sample TEE" under build bins which runs `cargo build -p kbs --locked --release --no-default-features --features rustls,resource,opa`

```
warning: unused import: `sync::Arc`
  --> kbs/src/api/src/lib.rs:29:28
   |
29 | use std::{net::SocketAddr, sync::Arc};
   |                            ^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

PR #421, opened 2024-06-17
Fix KBS AS build warning

larrydewey:
This introduces a compiler feature check for `as` when importing `sync::Arc`. This removes the `unused_imports` warning when building as specified in #419.

Resolves: #419


fitzthum (approved):
LGTM

PR #424, opened 2024-06-20
intel-trust-authority-as: add error message log

pawelpros:
- Added logging error message for failed appraisal request.
- Fixed typos

Sample output:

 `ERROR api_server::http::error] Attestation failed: Attestation request failed: response status=400 Bad Request, message=Invalid nonce and/or run time data provided in the request`

### DIFF
--- a/kbs/src/api/src/attestation/intel_trust_authority/mod.rs
+++ b/kbs/src/api/src/attestation/intel_trust_authority/mod.rs
@@ -35,6 +35,11 @@ struct Claims {
     policy_ids_unmatched: Option<Vec<serde_json::Value>>,
 }
 
+#[derive(Deserialize, Debug)]
+struct ErrorResponse {
+    error: String,
+}
+
 #[derive(Clone, Debug, Deserialize)]
 pub struct IntelTrustAuthorityConfig {
     pub base_url: String,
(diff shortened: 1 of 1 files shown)


kbs/src/api/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -82,18 +88,23 @@ impl Attest for IntelTrustAuthority {
...
+        if status != reqwest::StatusCode::OK {
+            let body = resp
+                .json::<ErrorResponse>()
+                .await
+                .map_err(|e| anyhow!("Deserialize error response failed: {:?}", e))?;
             bail!(
-                "Attestation request failed: respone status={}",
-                resp.status()
+                "Attestation request failed: response status={}, message={}",
+                status, body.error.unwrap_or_else(|| body.message.unwrap_or_else(|| "-".to_string()))
```
Xynnn007, line 99: I think you mean
```suggestion
                body.error.unwrap_or("-".to_string()), body.message.unwrap_or("-".to_string())
```
?

pawelpros: For testing I'm using this instance: https://api-poc-user1.project-amber-smas.com
It should always respond with error property but initial call after a while responds with "message" rather than expected "error" -> they both are never accessible in the same time

fidencio: @Xynnn007, the main thing is that we'll have have **body.error** or **body.message**, but never **both**.
With that in mind, I think @pawelpros' code does the correct thing, trying to unwrap the first, and if that fails, goes for the second.

pawelpros: Let's go with expected error format -> that "message" is probably specific for env and for now simpliest solution should do the trick.

fidencio (approved):
lgtm, thanks @pawelpros!

mkulke (approved):

PR #426, opened 2024-06-21
AS/verifier: fix tdx quote verification unit test

### LINKED ISSUE #399: AS: tdx-quote verification unit test fails
there is a tdx quote verification unit test that fails atm. the test is disabled by default, most likely because you need a PCCS configured on the test system.

```bash
cargo t -p verifier --no-default-features --features tdx-verifier -- verify_tdx_quote::case_2
    Finished test [unoptimized + debuginfo] target(s) in 0.21s
     Running unittests src/lib.rs (target/debug/deps/verifier-62425084b3596eb5)

running 1 test
test tdx::quote::tests::test_verify_tdx_quote::case_2 ... FAILED

failures:

---- tdx::quote::tests::test_verify_tdx_quote::case_2 stdout ----
thread 'tdx::quote::tests::test_verify_tdx_quote::case_2' panicked at attestation-service/verifier/src/tdx/quote.rs:540:9:
Err(tee_verify_quote failed: 0xe066)


failures:
    tdx::quote::tests::test_verify_tdx_quote::case_2

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 76 filtered out; finished in 0.02s

error: test failed, to rerun pass `-p verifier --lib`
```

We should either remove that test or maybe consider rewriting the logic slightly so that we can store all external dependencies as fixtures for the unit test and cover real verification in an e2e test

Xynnn007:
The example quotes need to configure a proper PCCS provided by Aliyun.

Close #399


Xynnn007:
cc @mkulke

attestation-service/verifier/src/tdx/quote.rs
```diff
@@ -528,16 +528,40 @@ mod tests {
...
     async fn test_verify_tdx_quote(#[case] quote: &str) {
+        use std::path::Path;
+
         let quote_bin = fs::read(quote).unwrap();
+        let mut original_content = None;
+
+        if Path::new(SGX_QCNL_CONFIG_PATH).exists() {
+            original_content = Some(fs::read(SGX_QCNL_CONFIG_PATH).unwrap());
+        }
+        let _ = fs::write(SGX_QCNL_CONFIG_PATH, SGX_QCNL_CONFIG).unwrap();
```
mkulke, line 558: I'm not sure it's a good idea to write a global config file to a machine which is running the unix test, I wouldn't like that a as a developer who is checking out the repo, even if we attempt to write an existing file back. I would rather add instructions to the documentation block and skip this test, until we have dynamic config file paths

Xynnn007: Agreed. Let's make it still ignored.

mkulke (approved):

fitzthum (approved):
Ok for now

PR #429, opened 2024-06-25
CLI: specify ATTESTER to build kbs-client

genjuro214:
We can't build kbs-client on s390x host with default feature including all attesters. 

I tried to add a new variable ATTESTER in Makefile, so that we can specify separate attester when building kbs-client

* by default, `make cli` will build kbs-client with all attesters
* keep feature sample_only: `CLI_FEATURES=sample_only make cli`
* specify ATTESTER: `ATTESTER=se-attester make cli`

### DIFF
--- a/kbs/Makefile
+++ b/kbs/Makefile
@@ -8,7 +8,8 @@ ifeq ($(filter $(ARCH),x86_64 s390x),)
 	$(error "Unsupported architecture: $(ARCH)")
 endif
 
-CLI_FEATURES ?= default
+CLI_FEATURES ?=
+ATTESTER ?=
 
 COCO_AS_INTEGRATION_TYPE ?= builtin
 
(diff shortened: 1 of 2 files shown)


huoqifeng:
`se-only` feature tag looks weird to me, maybe, considering to add ATTESTER like guest-components? like https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/Makefile#L28

huoqifeng:
@Xynnn007 @fitzthum @arronwy may you help have a look?

huoqifeng (approved):
LGTM

Xynnn007 (approved):

kbs/Makefile
```diff
@@ -20,6 +21,14 @@ else
   AS_FEATURE = $(AS_TYPE)
 endif
 
+ifndef CLI_FEATURES
+  ifdef ATTESTER
+    CLI_FEATURES = "sample_only,$(ATTESTER)"
```
fitzthum, line 26: It's a bit misleading to use both `sample_only` and another attester, but I guess we can live with that.

fitzthum (approved):
See one nit

PR #431, opened 2024-06-25
kbs: Revert support for configurable policy

mkulke:
There is existing test code that leverages the set-policy endpoint. In the current kubernetes deployment configuration that functionality is disabled, as it's backed by read-only configMap.

context [here](https://github.com/kata-containers/kata-containers/pull/9896)

This reverts commit d25fb27923081acad8232453dcf71322c79b8119.

Added remark about state into the Readme.


fitzthum (approved):
LGTM

PR #433, opened 2024-06-26
doc: add attestation policy guide for ibmse verifier

huoqifeng:
- Added guide on how to set attestation policy for ibmse verifier


huoqifeng:
@Xynnn007 @fitzthum may you please have a look at this PR? We added rvps guide for IBM SE. I did not find a good place for the file `parse_hdr.py`, so put it under `verifer\hack`, any suggestions?

attestation-service/verifier/src/se/README.md
```diff
@@ -167,10 +179,207 @@ services:
...
+default allow = false
+
+converted_version := sprintf("%v", [input["se.version"]])
+
+allow if {
+    input["se.attestation_phkh"] == data.reference["se.attestation_phkh"][_]
+    input["se.image_phkh"] == data.reference["se.image_phkh"][_]
+    input["se.tag"] == data.reference["se.tag"][_]
+    input["se.user_data"] == data.reference["se.user_data"][_]
+    converted_version == data.reference["se.version"][_]
```
Xynnn007, line 377: Actually we could directly set the hexs calculated by `parse_hdr.py` on the right side of the `==` here without interact with RVPS right? Current RVPS is not stable and would change frequently, thus I suggest use AS policy only. Or, it might cause a lot of document changes when RVPS changes, which would easily be ignored and thus leaving a bunch of out-of-date documents.

huoqifeng: Yeah, I know, we actually tried set the value directly, I'm fine to set the value here instead reference value as 1st option. Thanks!

huoqifeng: Refined, thanks! @Xynnn007

Xynnn007 (approved):
lgtm

attestation-service/verifier/hack/parse_hdr.py
fitzthum, line 1: Can you name this something to do with s390x or SE?

fitzthum (commented):
one comment but this seems good

attestation-service/verifier/hack/parse_hdr.py
huoqifeng: Yes, @fitzthum , renamed it `se_parse_hdr.py`. thanks!

fitzthum (approved):

PR #434, opened 2024-06-27
ci: test use https in kbs e2e test

mkulke:
This test will enable https for the token-kbs. The resource-kbs will remain http.

The test folder structure has been reworked a bit, so we don't end up littering the filesystem with temporary files, all temp resources should now end up in `./work` and can be safely removed without overwriting unrelated files.

### DIFF
--- a/kbs/.gitignore
+++ b/kbs/.gitignore
@@ -6,11 +6,8 @@
 data
 
 # test
-test/*
-!test/Makefile
-!test/data/
-test/data/attestation-service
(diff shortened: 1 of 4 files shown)


fitzthum (approved):
LGTM

Xynnn007:
Nice to see the outcome https://github.com/confidential-containers/trustee/actions/runs/9695875325/job/26756574226?pr=434#step:10:45

PR #436, opened 2024-07-01
KBS: Enable deployment for s390x

BbolroC:
The following changes have been made to enable the KBS deployment for s390x:

- Made the mount point `/opa/confidential-containers/kbs` writable (for both arches, reviving https://github.com/confidential-containers/trustee/pull/431)
- Differentiated the configuration for x86_64 and s390x in overlays and nodeport
- Updated the documentation

Detailed explanations can be found in each commit message. Verification was performed on a test machine for each platform.


Xynnn007 (approved):
Thanks @BbolroC . The init-container way seems work-able for both configMap and writable. I think also needs reviews from @mkulke 

A question from my side is why we marked policy as volume rather than configMap. What if a stateful volume? In this way we can probably store the policy in front of time (even for DBs).

mkulke (commented):
why do we need to add a persistent volume? could we just mount the configmap into the init-container (read-only) and copy the file to the (writable) container storage?

BbolroC:
> why do we need to add a persistent volume? could we just mount the configmap into the init-container (read-only) and copy the file to the (writable) container storage?

Do you see any cons for using the PV? Why I chose the PV is that I just don't want to open any credentials used for the SE verifier upstream.

mkulke:
> > why do we need to add a persistent volume? could we just mount the configmap into the init-container (read-only) and copy the file to the (writable) container storage?
> 
> Do you see any cons for using the PV? Why I chose the PV is that I just don't want to open any credentials used for a SE verifier to upstream. For me, a PV is the easiest and simplest way of encapsulating those data for kustomize.

why not use the kustomize [secret generator](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/secretGeneratorPlugin.md#secret-values-from-local-files)?

kbs/config/kubernetes/README.md
```diff
@@ -9,7 +9,7 @@ We will see how to deploy KBS (with builtin Attestation Service) on a Kubernetes
 Create a secret that you want to be served using this instance of KBS:
 
 ```bash
-echo "This is my super secert" > overlays/key.bin
+echo "This is my super secert" > overlays/$(uname -m)/key.bin
```
fitzthum, line 12: Should probably fix `secert` -> `secret`

kbs/config/kubernetes/base/deployment.yaml
```diff
@@ -12,6 +12,19 @@ spec:
       labels:
         app: kbs
     spec:
+      initContainers:
+      - command:
+        - sh
+        - -c
+        - cp -r /config/$(dirname $(readlink /config/policy.rego))/* /opa/confidential-containers/kbs/
```
fitzthum, line 19: This doesn't support updates at runtime or provisioning resources, right? Maybe this is fine in the short term.

fitzthum (commented):
Seems good. Two comments.

kbs/config/kubernetes/README.md
BbolroC: Yep, I will change it. Thanks!

BbolroC:
> A question from my side is why not we marked policy as volume rather than configMap. What if a stateful volume? In this way we can probably store the policy in front of time (even for DBs).

I'm not sure what the original idea behind the policy configuration was. We might be able to improve the configuration later after some investigation. Thanks.

BbolroC:
> why not use the kustomize [secret generator](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/secretGeneratorPlugin.md#secret-values-from-local-files)?

The configuration is specific to s390x, and a way of configuring the credentials via PV aligns with the documentation at https://github.com/confidential-containers/trustee/tree/main/attestation-service/verifier/src/se. This means that once a user has prepared the credentials following the documentation, they can simply export `IBM_SE_CREDS_DIR` to point to the root directory of the credentials.

fitzthum (approved):
LGTM. I think we may change back to the simple config map approach soon, but I guess it's fine to use the init container

PR #437, opened 2024-07-03
Fix broken SE link

fitzthum:
The refactor of the repository broke this link.

We also have some issues with rustup not being available on the az runners. I wasn't sure if we should add rustup installation to the tests or do it locally. wdyt @mkulke


mkulke (approved):

mkulke:
yup I made a fix in #438

PR #439, opened 2024-07-09
config: fix custom pccs deployment for TDX

fitzthum:
We recently split the nodeport yaml into an s390x
and an x86_64 directory, but we forgot to update
the custom_pccs yaml to point to the correct one.

For now let's assume that the custom_pccs will
always run on x86_64 since it's for TDX.
We might revisit that assumption in the future.

### DIFF
--- a/kbs/config/kubernetes/custom_pccs/kustomization.yaml
+++ b/kbs/config/kubernetes/custom_pccs/kustomization.yaml
@@ -3,7 +3,7 @@ kind: Kustomization
 namespace: coco-tenant
 
 resources:
-- ../nodeport
+- ../nodeport/x86_64
 
 patches:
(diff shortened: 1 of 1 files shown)


wainersm:
> We recently split the nodeport yaml into an s390x and an x86_64 directory, but we forgot to update the custom_pccs yaml to point to the correct one.
> 
> For now let's assume that the custom_pccs will always run on x86_64 since it's for TDX. We might revisit that assumption in the future.

Maybe we don't even need the nodeport manifests in per-arch overlays as it should be arch-agnostic. Anyway, this is a topic for another PR maybe :)

wainersm (approved):
Thanks @fitzthum !

PR #443, opened 2024-07-17
dep/verifier: fix eventlog with extra blank chars

Xynnn007:
Due to definition of AAEL, a ' ' is allowed as part of content field. Thus we now parse the original eventlog string using ' ' as separator to distinguish domain, operation and content only twice, and the things left are treated as the content of the eventlog, including extra ' 's.

### DIFF
--- a/deps/verifier/src/eventlog/mod.rs
+++ b/deps/verifier/src/eventlog/mod.rs
@@ -12,7 +12,7 @@ use hash::HashAlgorithm;
 use serde_json::{Map, Value};
 use sha2::{digest::FixedOutput, Digest, Sha256, Sha384, Sha512};
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Debug)]
 pub struct AAEvent {
     pub domain: String,
     pub operation: String,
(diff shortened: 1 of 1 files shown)


deps/verifier/src/eventlog/mod.rs
```diff
@@ -24,10 +24,7 @@ impl FromStr for AAEvent {
 
     fn from_str(input: &str) -> Result<Self> {
         let input_trimed = input.trim_end();
-        let sections: Vec<&str> = input_trimed.split(' ').collect();
-        if sections.len() != 3 {
-            bail!("Illegal AA event entry format. Should be `<domain> <operation> <content>`");
-        }
+        let sections: Vec<&str> = input_trimed.splitn(3, ' ').collect();
```
mkulke, line 27: I think this will panic later if the input string is "one" or "one two", so I would keep the length check

mkulke (commented):
Maybe add a negative test? or even more tests to spec out the log format a bit more (like can we use unicode/emojis in domain and operation,  is `"my domain"` w/ quotes accepted, etc..

Xynnn007:
> Maybe add a negative test? or even more tests to spec out the log format a bit more (like can we use unicode/emojis in domain and operation, is `"my domain"` w/ quotes accepted, etc..

yea, I think you raise really good points that we do not have a spec for the eventlog format. Initially I think it should be letters and numbers, the domain/operation should not have any blanks. Maybe we should make an issue to sync a spec for this, and then publish it on CoCo website due to https://github.com/confidential-containers/confidential-containers/issues/225#issuecomment-2263089133 ?

mythi:
@Xynnn007 this one can be closed?

PR #445, opened 2024-07-26
Bump kbs-types and kbs_protocol with a KBS protocol version change

mythi:
Ref: https://github.com/confidential-containers/guest-components/pull/628

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,9 +29,9 @@ config = "0.13.3"
 env_logger = "0.10.0"
 hex = "0.4.3"
 jwt-simple = "0.11"
-kbs_protocol = { git = "https://github.com/confidential-containers/guest-components.git", rev="9bd6f06a9704e01808e91abde130dffb20e632a5", default-features = false }
-kbs-types = "0.6.0"
-kms = { git = "https://github.com/confidential-containers/guest-components.git", rev="9bd6f06a9704e01808e91abde130dffb20e632a5", default-features = false }
+kbs_protocol = { git = "https://github.com/confidential-containers/guest-components.git", rev="cd16b445291ad401b4b53664266983f4927a370e", default-features = false }
+kbs-types = "0.7.0"
+kms = { git = "https://github.com/confidential-containers/guest-components.git", rev="cd16b445291ad401b4b53664266983f4927a370e", default-features = false }
 jsonwebtoken = { version = "9", default-features = false }
 log = "0.4.17"
 prost = "0.12"
--- a/kbs/docs/kbs_attestation_protocol.md
+++ b/kbs/docs/kbs_attestation_protocol.md
@@ -70,8 +70,8 @@ The payload format of the request is as follows:
 
 ```json
 {
-    /* Attestation protocol version number used by KBC */
-    "version": "0.1.0",
+    /* KBS protocol version number used by KBC */
+    "version": "0.1.1",
     /*
      * Type of HW-TEE platforms where KBC is located,
      * e.g. "intel-tdx", "amd-sev-snp", etc.
@@ -286,6 +286,10 @@ The authentication service is provided by the KBS through two endpoints:
    the attester and its attestation results with that cookie.
    2. An attestation challenge for the attester to take. This is the content of
    the response, set to a [KBS Challenge](#challenge) JSON payload.
+   3. In case of an error (such as if the KBS rejects the [KBS Request](#request) based
+   on `version` compatibility), an HTTP response with a 401 (`Unauthorized`) status code
+   together with ErrorInformation JSON payload.
+
 2. `/kbs/v0/attest` only accepts `POST` requests whose body is a [KBS Attestation](#attestation)
    JSON payload and the header contains a `Cookie` set to the value received in
    step 1.i. This is how the attester replies to attestation challenge received
--- a/kbs/src/http/attest.rs
+++ b/kbs/src/http/attest.rs
@@ -28,7 +28,7 @@ lazy_static! {
             build: BuildMetadata::EMPTY,
         };
 
-        VersionReq::parse(&format!("<={kbs_version}")).unwrap()
+        VersionReq::parse(&format!("={kbs_version}")).unwrap()
     };
 }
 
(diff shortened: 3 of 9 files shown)


mythi:
@Xynnn007 @mkulke @fitzthum this is ready for review. It's needed to get the RCAR `test_client` tests back in action

kbs/src/http/attest.rs
```diff
@@ -16,7 +16,7 @@ use serde_json::json;
 
 static KBS_MAJOR_VERSION: u64 = 0;
 static KBS_MINOR_VERSION: u64 = 1;
-static KBS_PATCH_VERSION: u64 = 0;
+static KBS_PATCH_VERSION: u64 = 1;
```
Xynnn007, line 19: Do you think we should make KBS forward compatible? Another view is still kbs is not stable release like `1.x`, thus we could live that?

cc @fitzthum @mkulke

Xynnn007 (commented):
Thanks for this. I only have some opens due to compatibility.

The `evidence` has been tranformed from `string` into `value`, which will give opportunity for combined attestation evidences like stated in https://github.com/confidential-containers/confidential-containers/issues/176. Next phase we might need to think about to leverage this, e.g. once GPU attestation joins Trustee Bucket Meal.

cc @imlk0 @zvonkok

kbs/src/http/attest.rs
mythi: we have not talked about this too much and I think there's two dimensions: version compatibility with `trustee` and `guest-components` releases and version compatibility with the KBS protocol versions. Currently, the `version` check ( I did not change that in https://github.com/confidential-containers/trustee/commit/c42dea0367b2767d8d2a220e4d4a42e901626846) is so that RCAR clients cannot use a newer protocol version but in real life only an exact version match works today (RCAR client 1.0.0 wont work with KBS 1.0.1).

mkulke:
> Thanks for this. I only have some opens due to compatibility.
> 
> The `evidence` has been tranformed from `string` into `value`, which will give opportunity for combined attestation evidences like stated in [confidential-containers/confidential-containers#176](https://github.com/confidential-containers/confidential-containers/issues/176). Next phase we might need to think about to leverage this, e.g. once GPU attestation joins Trustee Bucket Meal.
> 
> cc @imlk0 @zvonkok

@Xynnn007 I have opened an issue on kbs-types about this, but maybe we can also discuss it here. I suspect using serde_json on that type can lead to problems, because the internal representation of `Value` might change between versions of serde_json and then having a different version of serde_json in your tree will break compilation.

Is there any upside of using `Value` vs just a `Vec<u8>` for flexible `extra_params`? For both we need to have a manual parsing step.

kbs/src/http/attest.rs
mkulke: IMO it would be an improvement to have inflexible, but predictable protocol versioning for the time being. Supporting more options than a 1:1 version match might be impeding the quick iterations on the protocol before a 1.0 release of kbs

Cargo.toml
```diff
@@ -29,8 +29,8 @@ config = "0.13.3"
 env_logger = "0.10.0"
 hex = "0.4.3"
 jwt-simple = "0.11"
-kbs_protocol = { git = "https://github.com/confidential-containers/guest-components.git", rev="9bd6f06a9704e01808e91abde130dffb20e632a5", default-features = false }
-kbs-types = "0.6.0"
+kbs_protocol = { git = "https://github.com/confidential-containers/guest-components.git", rev="cd16b445291ad401b4b53664266983f4927a370e", default-features = false }
+kbs-types = "0.7.0"
 kms = { git = "https://github.com/confidential-containers/guest-components.git", rev="9bd6f06a9704e01808e91abde130dffb20e632a5", default-features = false }
```
huoqifeng, line 34: Do we want keep aligned for the commit in `kms` and `kbs_protocol` which are all come from  `guest-components`?

mythi: good catch, I had missed this second dependency. will update

mythi: updated.

kbs/docs/kbs_attestation_protocol.md
```diff
@@ -70,8 +70,8 @@ The payload format of the request is as follows:
 
 ```json
 {
-    /* Attestation protocol version number used by KBC */
-    "version": "0.1.0",
+    /* KBS protocol version number used by KBC */
+    "version": "0.1.1",
```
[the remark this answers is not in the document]
mythi: I'd just add a note in the release notes every time there's a bump

kbs/src/http/attest.rs
mythi: Makes sense. However, since we now have `trustee` and `guest-components` releases decoupled we need to document the version dependencies in the release notes when the protocol version changes.

mythi: > Supporting more options than a 1:1 version match might be impeding the quick iterations on the protocol before a 1.0 release of kbs

I think we are currently evidencing this :-) I could add change the version check to be an exact match and we go ahead with the current `tee-evidence` as `Value`?

mythi:
rebased and force-pushed due to some dependabot PRs being merged to avoid any merge conflicts.

mythi:
Just a reminder that the guest-components side of the same change is already merged and thus keeps some rcar client tests skipped. This also blocks the work with #242. Therefore, it would be good to get this merged. We can get back to https://github.com/virtee/kbs-types/issues/43 once the discussion gets to a conclusion and `kbs-types` maintainers agree.

fitzthum (approved):
LGTM. Sorry for the delay. I was trying to find the tab.

kbs/src/http/attest.rs
mythi: I pushed a separate commit for requiring an exact version match.

mythi:
@Xynnn007 this is ready to be merged. It was approved by @fitzthum and also @mkulke earlier

Xynnn007 (approved):

Xynnn007 (approved):

PR #446, opened 2024-07-26
verifier: Use the VirTEE CA Chain

larrydewey:
This introduces changes into the code to utilize the cert chain provided by the VirTEE sev crate, pushing resposibility for upkeep and maintainership back to the library itself. This also reduces duplication of code and expands functionality for future changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ jsonwebtoken = { version = "9", default-features = false }
 log = "0.4.17"
 prost = "0.12"
 regorus = { version = "0.1.5", default-features = false, features = ["regex", "base64", "time"] }
-reqwest = "0.12"
+reqwest = { version = "0.12", default-features = false }
 rstest = "0.18.1"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.89"
@@ -46,7 +46,7 @@ sha2 = "0.10"
 shadow-rs = "0.19.0"
 strum = { version = "0.25", features = ["derive"] }
 thiserror = "1.0"
-tokio = { version = "1", features = ["full"] }
+tokio = { version = "1", features = ["full"], default-features = false }
 tempfile = "3.4.0"
 tonic = "0.11"
 tonic-build = "0.11"
--- a/deps/verifier/Cargo.toml
+++ b/deps/verifier/Cargo.toml
@@ -28,7 +28,7 @@ byteorder = "1"
 cfg-if = "1.0.0"
 codicon = { version = "3.0", optional = true }
 # TODO: change it to "0.1", once released.
-csv-rs = { git = "https://github.com/openanolis/csv-rs", rev = "b74aa8c", optional = true }
+csv-rs = { version = "=0.1.0", git = "https://github.com/openanolis/csv-rs", rev = "b74aa8c", optional = true }
 eventlog-rs = { version = "0.1.3", optional = true }
 hex.workspace = true
 jsonwebkey = "0.3.5"
@@ -41,9 +41,9 @@ scroll = { version = "0.11.0", default-features = false, features = ["derive"], 
 serde.workspace = true
 serde_json.workspace = true
 serde_with = { workspace = true, optional = true }
-sev = { version = "3.1.1", features = ["openssl", "snp"], optional = true }
+sev = { version = "4.0.0", features = ["openssl", "snp"], optional = true }
 sha2.workspace = true 
-tokio = { workspace = true, optional = true, default-features = false }
+tokio = { workspace = true, optional = true }
 intel-tee-quote-verification-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", tag = "DCAP_1.21", optional = true }
 strum.workspace = true
 veraison-apiclient = { git = "https://github.com/chendave/rust-apiclient", branch = "token", optional = true }
--- a/deps/verifier/src/snp/mod.rs
+++ b/deps/verifier/src/snp/mod.rs
@@ -15,6 +15,7 @@ use openssl::{
     x509::{self, X509},
 };
 use serde_json::json;
+use sev::certs::snp::{ca::Chain as CaChain, Certificate, Chain};
 use sev::firmware::guest::AttestationReport;
 use sev::firmware::host::{CertTableEntry, CertType};
 use std::sync::OnceLock;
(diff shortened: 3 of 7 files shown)


larrydewey:
Waiting on https://github.com/virtee/sev/pull/206

ryansavino (approved):
LGTM

deps/verifier/src/snp/mod.rs
```diff
@@ -34,43 +35,43 @@ const LOADER_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .1);
...
+    vendor_certs: Chain,
 }
 
-pub(crate) fn load_milan_cert_chain() -> &'static Result<VendorCertificates> {
-    static MILAN_CERT_CHAIN: OnceLock<Result<VendorCertificates>> = OnceLock::new();
-    MILAN_CERT_CHAIN.get_or_init(|| {
-        let certs = X509::stack_from_pem(include_bytes!("milan_ask_ark_asvk.pem"))?;
-        if certs.len() != 3 {
-            bail!("Malformed Milan ASK/ARK/ASVK");
+fn read_certs_from_file() -> Vec<X509> {
```
fitzthum, line 41: Not sure of the value of splitting this into two functions. This new one should still refer to `milan` since it is only loading the milan certs. Also this isn't exactly reading them from a file.

deps/verifier/src/snp/mod.rs
```diff
@@ -34,43 +35,43 @@ const LOADER_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .1);
...
+pub(crate) fn load_milan_cert_chain() -> &'static Chain {
+    static MILAN_CERT_CHAIN: OnceLock<Chain> = OnceLock::new();
+    MILAN_CERT_CHAIN.get_or_init(|| {
+        let certs = read_certs_from_file();
+        Chain {
+            ca: CaChain {
+                ark: Certificate::from(certs[1].clone()),
+                ask: Certificate::from(certs[0].clone()),
+            },
+            vek: Certificate::from(certs[2].clone()),
```
fitzthum, line 62: I thought this was the `asvk`. Has the term changed? If so, please remove references to `asvk`. I thought `vek` already referred to the guest encryption key.

fitzthum (commented):
CI not looking great.

There are some formatting changes here. I'm not sure they are necessary, but if they are consider moving them to another commit.

Cargo.toml
```diff
@@ -34,9 +34,13 @@ kbs-types = "0.6.0"
 kms = { git = "https://github.com/confidential-containers/guest-components.git", rev="9bd6f06a9704e01808e91abde130dffb20e632a5", default-features = false }
 jsonwebtoken = { version = "9", default-features = false }
 log = "0.4.17"
-prost = "0.12"
-regorus = { version = "0.1.5", default-features = false, features = ["regex", "base64", "time"] }
-reqwest = "0.12"
+prost = "0.12.0"
```
mkulke, line 37: we probably want to avoid unrelated toml reformatting for a clean diff

deps/verifier/Cargo.toml
```diff
@@ -4,31 +4,44 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-default = [ "all-verifier" ]
```
mkulke, line 7: same here, let's not reformat the full toml.

deps/verifier/Cargo.toml
```diff
@@ -4,31 +4,44 @@ version = "0.1.0"
...
+    "verifier",
+], optional = true }
 base64 = "0.21"
 bincode = "1.3.3"
 byteorder = "1"
 cfg-if = "1.0.0"
 codicon = { version = "3.0", optional = true }
 # TODO: change it to "0.1", once released.
-csv-rs = { git = "https://github.com/openanolis/csv-rs", rev = "b74aa8c", optional = true }
+csv-rs = { version = "=0.1.0", git = "https://github.com/openanolis/csv-rs", rev = "b74aa8c", optional = true }
```
mkulke, line 44: is this intentionally both version and ref?

Cargo.toml
```diff
@@ -46,7 +50,7 @@ sha2 = "0.10"
 shadow-rs = "0.19.0"
 strum = { version = "0.25", features = ["derive"] }
 thiserror = "1.0"
-tokio = { version = "1", features = ["full"] }
+tokio = { version = "1", features = ["full"], default-features = false }
```
mkulke, line 53: why do we want to turn on the tokio default-features?

deps/verifier/Cargo.toml
```diff
@@ -37,13 +50,15 @@ kbs-types.workspace = true
...
+    "derive",
+], optional = true }
 serde.workspace = true
 serde_json.workspace = true
 serde_with = { workspace = true, optional = true }
-sev = { version = "3.1.1", features = ["openssl", "snp"], optional = true }
+sev = { version = "4.0.0", features = ["openssl", "snp"], optional = true }
 sha2.workspace = true 
-tokio = { workspace = true, optional = true, default-features = false }
+tokio = { workspace = true, optional = true }
```
mkulke, line 61: why do we want to turn on the tokio default-features?

deps/verifier/src/snp/mod.rs
```diff
@@ -34,43 +35,43 @@ const LOADER_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .1);
...
-            bail!("Malformed Milan ASK/ARK/ASVK");
+fn read_certs_from_file() -> Vec<X509> {
+    match X509::stack_from_pem(include_bytes!("milan_ask_ark_asvk.pem")) {
+        Result::Ok(certs) => {
+            if certs.len() != 3 {
+                panic!("Malformed Milan ASK/ARK/ASVK");
+            }
+            return certs;
         }
+        Result::Err(err) => panic!("Error reading certificates file: {err}"),
```
mkulke, line 49: I think initially there was an ask to avoid panicking if the parsing of the bundled X509 failed. I'm fine with this, since this is very const-y and unlikely to fail.

However, we're not reading a file here, the bytes are embedded in the binary.

deps/verifier/src/tdx/eventlog.rs
```diff
@@ -124,6 +124,7 @@ impl CcEventLog {
 
 /// Defined in TCG PC Client Platform Firmware Profile Specification section
 /// 'UEFI_PLATFORM_FIRMWARE_BLOB Structure Definition'
+#[allow(dead_code)]
```
mkulke, line 127: I assume this is unrelated?

Cargo.toml
larrydewey: I don't have any clear insight into why the project chose to have full tokio features enabled. The PR moves the `default-features = false` into the workspace toml. If we have it inside of the verifier `Cargo.toml`, it is completely ignored, and it throws a warning.

deps/verifier/Cargo.toml
larrydewey: I am not sure if it is a bug in Cargo, or something else, but if I didn't enforce the specific version of the crate, it was being mangled on re-build.

deps/verifier/Cargo.toml
larrydewey: Again, I am not sure. in this case, I am just removing the ignored `default-features = false` flag from the Cargo.toml and moving into the workspace level `Cargo.toml`

deps/verifier/src/snp/mod.rs
larrydewey: In this case, the `ASK` and the `VEK` are generic terms. The **A**MD **S**igning **K**ey (ASK) covers both the traditional `ASK` and the `ASVK`. The **V**ersioned **E**ndorsement **K**ey (VEK) covers `VCEK` or `VLEK`.

deps/verifier/src/tdx/eventlog.rs
larrydewey: It was to remove a warning when compiling, as the TDX library is unused.

deps/verifier/src/snp/mod.rs
larrydewey: I am splitting this into two functions because the former behavior of utilizing a `OnceLock` around a `Result` leaves room for unexpected behavior, as it returns a reference to a `Result`. I suppose I could re-implement this using a `lazy_static!` to ensure it is only allocated once, and then use a Mutex to make sure access is serial. 

Ultimately, I was just trying to get rid of the UB.

larrydewey: Renamed. Let me know what you think about the latest commit.

deps/verifier/src/snp/mod.rs
fitzthum: Isn't `certs[2]` going to be the `ASVK`? That shouldn't go into the `VEK` field. The current `VendorCertificates` struct holds both the `ASK` and `ASVK`. It doesn't like the `CaChain` struct does that so you might need to tweak the logic a bit.

deps/verifier/src/snp/mod.rs
larrydewey: Renamed the file to more accurately describe what is happening.

deps/verifier/src/snp/mod.rs
larrydewey: You are absolutely correct. I misunderstood your comment. I will have to think about this, as a certificate chain usually should contain one or the other, not both.

mkulke (commented):
@larrydewey maybe the compiler/cargo errors you are seeing are due to some specifics in your local build environment? (e.g. cargo, rust toolchain, etc.) In this case you can maybe take a look at https://github.com/confidential-containers/trustee/blob/81f904f0f5ec10db471922d86f03d90fa8e6385e/.github/workflows/kbs-rust.yml#L33-L38 to find a working configuration. The CI should define the canonical build configuration that we want to support.

larrydewey:
> @larrydewey maybe the compiler/cargo errors you are seeing are due to some specifics in your local build environment? (e.g. cargo, rust toolchain, etc.) In this case you can maybe take a look at
> 
> https://github.com/confidential-containers/trustee/blob/81f904f0f5ec10db471922d86f03d90fa8e6385e/.github/workflows/kbs-rust.yml#L33-L38
> 
> to find a working configuration. The CI should define the canonical build configuration that we want to support.

This is my default configuration, so I don't believe it is my workflow or configuration.

PR #449, opened 2024-07-30
kbs: add ProtocolVersion error

mythi:
kbs already supports checking the Request version but any version mismatch is not correctly returned to the client (nor checked by the current RCAR client handshake).

Add an explicit kbs ProtocolVersion error that is returned when the Request version is higher than what the KBS claims to support.

### DIFF
--- a/kbs/src/http/attest.rs
+++ b/kbs/src/http/attest.rs
@@ -11,7 +11,26 @@ use base64::engine::general_purpose::{STANDARD, URL_SAFE_NO_PAD};
 use base64::Engine;
 use kbs_types::Challenge;
 use log::{debug, error, info};
+use semver::{BuildMetadata, Prerelease, Version, VersionReq};
 use serde_json::json;
+
+static KBS_MAJOR_VERSION: u64 = 0;
+static KBS_MINOR_VERSION: u64 = 1;
+static KBS_PATCH_VERSION: u64 = 0;
+
+lazy_static! {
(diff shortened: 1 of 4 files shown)


larrydewey (approved):
LGTM

Xynnn007 (approved):
lgtm. Thanks @mythi

kbs/src/http/attest.rs
```diff
@@ -22,6 +41,14 @@ pub(crate) async fn auth(
 ) -> Result<HttpResponse> {
     info!("Auth API called.");
     debug!("Auth Request: {:?}", &request);
+    let version = Version::parse(&request.version).unwrap();
+    if !VERSION_REQ.matches(&version) {
+        raise_error!(Error::ProtocolVersion(format!(
+            "client {}{} server",
```
mkulke, line 47: nit: is this how the string is supposed to look?

mythi: It becomes like this:

```
ErrorInformation { error_type: \"https://github.com/confidential-containers/kbs/errors/ProtocolVersion\", detail: \"KBS Client Protocol Version Mismatch: client 0.1.1<=0.1.0 server\" }")
```

I iterated over a few alternatives. I'm open to any suggestions :-)

mkulke (approved):

kbs/src/http/attest.rs
mythi: pushed an update.

PR #451, opened 2024-07-31
Extend KBS to provide the resources required to create an encrypted overlay network

### LINKED ISSUE #396: Extend KBS to provide the resources required to create an encrypted overlay network
This is a follow-up on the encrypted overlay network support we proposed in the May/09/2024 meeting to secure data in-transit between CoCo PODs. We'd like to use this issue to discuss the plugin interface design and which implementation option seems more appropriate.

## Background

The overlay network would be created using [Nebula](https://nebula.defined.net/docs/guides/quick-start/).

In Nebula, a node can join the overlay network only if it provides a valid credential generated by the Nebula Certificate Authority (CA); the node can request a credential to the Nebula CA by providing the IP address and the name it wants to have in the Nebula overlay network. Later, the node can use the returned certificate and key to join the Nebula overlay network.

For example, in a deployment with 10 CoCo PODs, the Nebula overlay network would have 10 CoCo PODs and one Nebula Lighthouse (assumed to already exist). Each POD would have two network interfaces one created by kubernetes and another one created by Nebula. The traffic sent through the Nebula interface is automatically encrypted by Nebula.

## Proposed design related to KBS

- Run the Nebula CA co-located with the KBS so that it can leverage the KBS RCAR protocol and the KBS encrypted session. Communication between the KBS and the Nebula CA wouldn't need to be encrypted.
- Create a generic plugin interface for specialized resources. The Nebula CA could be implemented as a KBS plugin to generate credentials on demand for attested PODs that want to join the Nebula network.

## Plugin interface requirements

- Support for multiple parameters. Plugins may take multiple parameters to generate a specialized resource.
- Support for plugin initialization.  Plugins may need to be initialized, e.g. the Nebula CA needs to create/load the CA.pem and CA.key that will be used to sign certificates.

## Implementation options for the plugin interface

1. Re-use the existing KBS resource URI `/kbs/v0/resource/<repository>/<type>/<tag>`

	*Pros:* May have less duplicated code. Re-use of the KBS resource URI
	*Cons:* It may look like a hack. The plugin repository may need to be reserved.

	The `get-resource()` function would do the RCAR dance as necessary and then if repository=plugin the other two fields would actually be interpreted as `/kbs/v0/resource/plugin/<plugin_name>/<plugin_action>`. The <plugin_action> would be forwarded to the appropriate <plugin_name>.
	The plugin interface could be implemented as RepositoryConfig::Plugin, but for that we would need to properly select between Plugin or LocalFs when a KBS get-resource call is received. As a RepositoryConfig::Plugin, it could be initialized in the same function as LocalFs.
	
	If the <plugin_action> requires multiple parameters:

	- 1a. we could use [HTTP query string](https://en.wikipedia.org/wiki/Query_string)(question mark symbol), to expand the <plugin_action>, e.g. `/kbs/v0/resource/plugin/<plugin_name>/<plugin_action>?ip=10.10.10.10&name=pod1`. If that is not supported by the KBS protocol v0, we could try to replace the `?` by something not standard such as `;`.
	- 1b. If the option "1a" does not work, we could extend the KBS protocol v0 to support HTTP query string and then increase the protocol version to something like v0.1.

2. Add a new URI for plugins e.g. `/kbs/v0.1/plugin/<plugin_name>/<plugin_action>`

	*Pros:* Have a proper URI for plugins. Plugin workflow is better isolated from get-resource.
	*Cons:* The KBS protocol version may need to be increased to something like v0.1. May have some duplicated code compared to get-resource

	The plugin URI is not supposed to change the KBS state, so it would be accessible through the HTTP GET method
	The new `plugin-handler()` function would do the RCAR as needed (similar to the get-resource) and then it would forward the <plugin-action> to the proper <plugin-name>.
	The plugin interface would have its own PluginConfig enum for plugin initialization rather than a new RepositoryConfig enum value.
	Additional parameters for <plugin-action> could be provided by using HTTP Query String.

4. Other. Any thoughts?

cclaudio:
This PR requires https://github.com/confidential-containers/guest-components/pull/634 to be merged first.

This PR resolves #396 by adding the following changes:
- Add plugin interface
- Add nebula-plugin

### Steps to test it

- enable nebula in the kbs-config.toml:   `enabled_plugins = ["nebula"]`
- Generate User authentication key pair as in [quickstart guide](https://github.com/confidential-containers/trustee/blob/main/kbs/quickstart.md)
- `docker compose build`
- `docker compose up`
-  build kbs client: `cd kbs && ATTESTER=snp-attester make cli && sudo make install-cli`
-  `kbs-client --url http://127.0.0.1:8080 config --auth-private-key kbs/config/private.key set-resource-policy --policy-file  kbs/sample_policies/allow_all.rego`
-  `kbs-client --url http://127.0.0.1:8080 get-resource --path 'plugin/nebula/credential?ip[ip]=10.11.12.13&ip[netbits]=21&name=pod1' | base64 -d`

The last command should return the credential requested, e.g.:

```json
{
"node_crt":[45,45,45,45,45,66,69,71,73,78,32,78,69,66,85,76,65,32,67,69,82,84,73,70,73,67,65,84,69,45,45,45,45,45,10,67,109,69,75,66,72,66,118,90,68,69,83,67,89,50,89,114,70,67,65,56,80,47,47,68,121,106,83,117,113,83,49,66,106,67,97,49,113,88,69,66,106,111,103,83,85,66,117,100,74,104,101,84,75,105,49,122,50,86,66,77,102,112,66,10,68,103,105,120,77,56,87,81,90,55,97,81,109,99,114,68,114,53,52,53,47,50,74,75,73,68,116,87,73,57,109,43,70,98,120,80,119,107,52,54,122,71,90,88,116,56,70,103,112,74,52,71,52,120,88,57,76,97,108,56,78,55,98,116,10,120,65,43,68,69,107,65,76,110,114,80,88,87,88,116,51,113,102,77,109,87,48,102,110,68,88,97,111,48,90,104,112,108,54,65,104,49,115,82,47,115,115,120,120,80,56,109,99,108,99,78,87,101,110,47,76,82,84,68,48,112,101,122,68,10,105,101,76,82,75,79,70,109,121,104,111,74,107,118,66,73,86,109,53,82,104,54,43,56,68,89,115,68,10,45,45,45,45,45,69,78,68,32,78,69,66,85,76,65,32,67,69,82,84,73,70,73,67,65,84,69,45,45,45,45,45,10],
"node_key":[45,45,45,45,45,66,69,71,73,78,32,78,69,66,85,76,65,32,88,50,53,53,49,57,32,80,82,73,86,65,84,69,32,75,69,89,45,45,45,45,45,10,79,69,103,121,70,88,106,101,107,97,78,115,50,98,111,122,56,80,101,68,69,112,51,113,82,51,47,114,89,120,67,82,79,79,57,89,70,49,103,109,118,102,69,61,10,45,45,45,45,45,69,78,68,32,78,69,66,85,76,65,32,88,50,53,53,49,57,32,80,82,73,86,65,84,69,32,75,69,89,45,45,45,45,45,10],
"ca_crt":[45,45,45,45,45,66,69,71,73,78,32,78,69,66,85,76,65,32,67,69,82,84,73,70,73,67,65,84,69,45,45,45,45,45,10,67,107,115,75,71,85,53,108,89,110,86,115,89,83,66,68,81,83,66,109,98,51,73,103,86,72,74,49,99,51,82,108,90,83,66,76,81,108,77,111,109,43,43,103,116,81,89,119,109,57,97,108,120,65,89,54,73,76,54,55,105,73,115,79,10,51,83,68,71,76,67,49,84,121,54,78,82,77,121,48,57,56,56,84,110,53,51,77,47,71,88,112,80,77,84,99,57,66,43,85,73,81,65,69,83,81,80,75,67,107,97,99,72,79,89,86,121,66,104,57,69,102,106,74,107,65,110,48,72,10,81,73,72,102,114,106,51,83,76,48,67,118,49,77,81,109,81,98,107,55,99,89,116,81,70,82,114,111,104,57,51,104,121,55,99,81,55,112,51,99,56,105,86,110,67,56,109,80,97,107,113,70,47,101,66,84,48,82,111,67,69,119,111,61,10,45,45,45,45,45,69,78,68,32,78,69,66,85,76,65,32,67,69,82,84,73,70,73,67,65,84,69,45,45,45,45,45,10]
}
```

### DIFF
--- a/kbs/config/docker-compose/kbs-config.toml
+++ b/kbs/config/docker-compose/kbs-config.toml
@@ -2,6 +2,10 @@ sockets = ["0.0.0.0:8080"]
 auth_public_key = "/opt/confidential-containers/kbs/user-keys/public.pub"
 insecure_http = true
 
+[repository_config]
+type = "LocalFs"
+dir_path = "/opt/confidential-containers/kbs/storage"
+
 [plugin_manager_config]
 work_dir = "/opt/confidential-containers/kbs/plugin"
 enabled_plugins = []
--- a/kbs/config/docker-compose/kbs-config.toml
+++ b/kbs/config/docker-compose/kbs-config.toml
@@ -2,6 +2,10 @@ sockets = ["0.0.0.0:8080"]
 auth_public_key = "/opt/confidential-containers/kbs/user-keys/public.pub"
 insecure_http = true
 
+[plugin_manager_config]
+work_dir = "/opt/confidential-containers/kbs/plugin"
+enabled_plugins = []
+
 [attestation_token_config]
 attestation_token_type = "CoCo"
 
--- a/kbs/docker/coco-as-grpc/Dockerfile
+++ b/kbs/docker/coco-as-grpc/Dockerfile
@@ -2,6 +2,7 @@ FROM rust:latest as builder
 ARG ARCH=x86_64
 ARG HTTPS_CRYPTO=rustls
 ARG ALIYUN=false
+ARG PLUGINS=""
 
 WORKDIR /usr/src/kbs
 COPY . .
@@ -9,7 +10,7 @@ COPY . .
 RUN apt-get update && apt install -y protobuf-compiler git
 
 # Build and Install KBS
-RUN cd kbs && make AS_FEATURE=coco-as-grpc HTTPS_CRYPTO=${HTTPS_CRYPTO} POLICY_ENGINE=opa ALIYUN=${ALIYUN} && \
+RUN cd kbs && make AS_FEATURE=coco-as-grpc HTTPS_CRYPTO=${HTTPS_CRYPTO} POLICY_ENGINE=opa ALIYUN=${ALIYUN} PLUGINS=${PLUGINS} && \
     make install-kbs
 
 FROM ubuntu:22.04
--- a/kbs/docker/coco-as-grpc/Dockerfile
+++ b/kbs/docker/coco-as-grpc/Dockerfile
@@ -2,7 +2,7 @@ FROM rust:latest as builder
 ARG ARCH=x86_64
 ARG HTTPS_CRYPTO=rustls
 ARG ALIYUN=false
-ARG PLUGINS=""
+ARG PLUGINS="nebula-plugin"
 
 WORKDIR /usr/src/kbs
 COPY . .
@@ -13,8 +13,13 @@ RUN apt-get update && apt install -y protobuf-compiler git
 RUN cd kbs && make AS_FEATURE=coco-as-grpc HTTPS_CRYPTO=${HTTPS_CRYPTO} POLICY_ENGINE=opa ALIYUN=${ALIYUN} PLUGINS=${PLUGINS} && \
     make install-kbs
 
+# Install Nebula
+RUN wget https://github.com/slackhq/nebula/releases/download/v1.8.2/nebula-linux-amd64.tar.gz
+RUN tar -C /usr/local/bin -xzf nebula-linux-amd64.tar.gz
+
 FROM ubuntu:22.04
 
 LABEL org.opencontainers.image.source="https://github.com/confidential-containers/trustee/kbs"
 
 COPY --from=builder /usr/local/bin/kbs /usr/local/bin/kbs
+COPY --from=builder /usr/local/bin/nebula-cert /usr/local/bin/nebula-cert
--- a/kbs/src/http/resource.rs
+++ b/kbs/src/http/resource.rs
@@ -15,7 +15,7 @@ use rsa::{BigUint, Pkcs1v15Encrypt, RsaPublicKey};
 use serde::Deserialize;
 use serde_json::{json, Deserializer, Value};
 
-use crate::raise_error;
+use crate::{raise_error, resource::plugin::PluginManager};
 
 use super::*;
 
@@ -30,6 +30,7 @@ const TOKEN_TEE_PUBKEY_PATH: &str = "/customized_claims/runtime_data/tee-pubkey"
 pub(crate) async fn get_resource(
     request: HttpRequest,
     repository: web::Data<Arc<RwLock<dyn Repository + Send + Sync>>>,
+    repository_plugin: web::Data<Arc<RwLock<PluginManager>>>,
     #[cfg(feature = "as")] map: web::Data<SessionMap>,
     token_verifier: web::Data<Arc<RwLock<dyn AttestationTokenVerifier + Send + Sync>>>,
     #[cfg(feature = "policy")] policy_engine: web::Data<PolicyEngine>,
@@ -113,12 +114,25 @@ pub(crate) async fn get_resource(
         info!("Resource access request passes policy check.");
     }
 
-    let resource_byte = repository
-        .read()
-        .await
-        .read_secret_resource(resource_description)
-        .await
-        .map_err(|e| Error::ReadSecretFailed(format!("{e:?}")))?;
+    let resource_byte = if resource_description.repository_name == "plugin" {
+        repository_plugin
+            .read()
+            .await
+            .get_resource(
+                resource_description.resource_type.as_str(),
+                resource_description.resource_tag.as_str(),
+                request.query_string(),
+            )
+            .await
+            .map_err(|e| Error::ReadSecretFailed(e.to_string()))?
+    } else {
+        repository
+            .read()
+            .await
+            .read_secret_resource(resource_description)
+            .await
+            .map_err(|e| Error::ReadSecretFailed(e.to_string()))?
+    };
 
     let jwe = jwe(pubkey, resource_byte)?;
 
--- a/kbs/src/resource/plugin/mod.rs
+++ b/kbs/src/resource/plugin/mod.rs
@@ -1,0 +1,102 @@
+// Copyright (c) 2024 by IBM Inc.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#[cfg(feature = "nebula-plugin")]
+mod nebula;
+
+use anyhow::{anyhow, bail, Context, Result};
+use serde::Deserialize;
+use std::fs;
+use std::path::Path;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+#[cfg(feature = "nebula-plugin")]
+use crate::resource::plugin::nebula::NebulaPluginConfig;
+
+trait PluginBuild {
+    fn get_plugin_name(&self) -> &str;
+    fn create_plugin(&self, work_dir: &str) -> Result<Arc<RwLock<dyn Plugin + Send + Sync>>>;
+}
+
+#[derive(Clone, Debug, Default, Deserialize)]
+pub struct PluginManagerConfig {
+    work_dir: String,
+    enabled_plugins: Vec<String>,
+}
+
+impl PluginManagerConfig {
+    fn get_plugin_builders(&self) -> Vec<Box<dyn PluginBuild>> {
+        #[allow(unused_mut)]
+        let mut p: Vec<Box<dyn PluginBuild>> = Vec::new();
+
+        // List of all plugins supported
+        #[cfg(feature = "nebula-plugin")]
+        p.push(Box::new(NebulaPluginConfig::default()));
+
+        p
+    }
+
+    pub fn create_plugin_manager(&self) -> Result<Arc<RwLock<PluginManager>>> {
+        if !Path::new(&self.work_dir).exists() {
+            fs::create_dir_all(&self.work_dir)
+                .with_context(|| "Create resource plugin dir".to_string())?;
+        }
+
+        #[allow(unused_mut)]
+        let mut manager = PluginManager {
+            plugins: Vec::new(),
+        };
+
+        let plugin_builders = self.get_plugin_builders();
+
+        for plugin_name in self.enabled_plugins.iter() {
+            let builder = plugin_builders
+                .iter()
+                .find(|x| x.get_plugin_name() == plugin_name)
+                .ok_or(anyhow!(
+                    "Cargo {}-plugin feature is either not set or not supported",
+                    plugin_name,
+                ))?;
+
+            let plugin_dir = format!("{}/{}", self.work_dir, builder.get_plugin_name());
+            let plugin = builder.create_plugin(plugin_dir.as_str())?;
+            manager.plugins.push(plugin);
+
+            log::info!("{} plugin loaded", builder.get_plugin_name());
+        }
+
+        log::info!("{} plugin(s) loaded", manager.plugins.len());
+
+        Ok(Arc::new(RwLock::new(manager)))
+    }
+}
+
+#[async_trait::async_trait]
+trait Plugin {
+    async fn get_name(&self) -> &str;
+    async fn get_resource(&self, resource: &str, query_string: &str) -> Result<Vec<u8>>;
+}
+
+pub struct PluginManager {
+    plugins: Vec<Arc<RwLock<dyn Plugin + Send + Sync>>>,
+}
+
+impl PluginManager {
+    pub async fn get_resource(
+        &self,
+        plugin_name: &str,
+        resource: &str,
+        query_string: &str,
+    ) -> Result<Vec<u8>> {
+        for plugin in self.plugins.iter() {
+            let p = plugin.read().await;
+
+            if *plugin_name == *p.get_name().await {
+                return p.get_resource(resource, query_string).await;
+            }
+        }
+        bail!("Plugin {} not found", plugin_name)
+    }
+}
--- a/kbs/src/resource/plugin/nebula.rs
+++ b/kbs/src/resource/plugin/nebula.rs
@@ -1,0 +1,462 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// Copyright (c) 2024 by IBM Inc.
+
+//! The nebula plugin allows the KBS to deliver resources required to create
+//! an encrypted overlay network between nodes using [Nebula](https://github.com/slackhq/nebula),
+//!
+//! Within the Nebula overlay network, all communications between nodes are
+//! automatically encrypted.
+
+use anyhow::{anyhow, bail, Context, Result};
+use serde_qs;
+use std::ffi::OsString;
+use std::fmt;
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::sync::Arc;
+use tempfile::{tempdir_in, TempDir};
+use tokio::sync::RwLock;
+
+use super::{Plugin, PluginBuild};
+
+pub const PLUGIN_NAME: &str = "nebula";
+
+const NEBULA_CONFIG_PATH: &str = "/etc/kbs/plugin/nebula-config.toml";
+const CRT_FILENAME: &str = "node.crt";
+const KEY_FILENAME: &str = "node.key";
+
+// Required binaries must be in PATH
+const NEBULA_CERT_BIN: &str = "nebula-cert";
+
+/// Policies that define when a Nebula CA must be generated.
+/// They are documented in the nebula plugin config toml file
+#[repr(u32)]
+pub enum CaGenerationPolicy {
+    GenerateIfNotFound = 1,
+    NeverGenerate = 2,
+}
+
+/// Plugin configuration
+/// It is documented in the nebula plugin config toml file
+#[derive(Debug, Default, serde::Deserialize)]
+pub struct NebulaPluginConfig {
+    crt_path: String,
+    key_path: String,
+    ca_generation_policy: u32,
+    self_signed_ca_config: Option<SelfSignedCaConfig>,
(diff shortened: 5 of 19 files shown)


cclaudio:
Once https://github.com/confidential-containers/guest-components/pull/634 is merged, I can update this PR and remove its draft label.

kbs/src/resource/plugin/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        log::info!("{} plugin(s) loaded", manager.plugins.len());
+
+        Ok(Arc::new(RwLock::new(manager)))
+    }
+}
+
+#[async_trait::async_trait]
+trait RepositoryPlugin {
```
fitzthum, line 79: Better to just call this `Plugin` imo. The repository abstraction is slightly different and I we should actually try to avoid using the term repository anyway because it's part of the resource URI as well.

kbs/src/resource/plugin/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+    async fn get_name(&self) -> &str;
+    async fn get_plugin_resource(&self, resource: &str, query_string: &str) -> Result<Vec<u8>>;
```
fitzthum, line 81: This can just be `get_resource` I think

kbs/src/resource/plugin/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+pub struct RepositoryPluginManager {
+    plugins: Vec<Arc<RwLock<dyn RepositoryPlugin + Send + Sync>>>,
+}
+
+impl RepositoryPluginManager {
+    pub async fn dispatch_get_request(
```
fitzthum, line 89: can probably just call this `get_resource`

kbs/src/resource/plugin/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        &self,
+        plugin_name: &str,
+        resource: &str,
+        query_string: &str,
+    ) -> Result<Vec<u8>> {
+        for plugin in self.plugins.iter() {
+            let p = plugin.write().await;
```
fitzthum, line 96: Are we allowing plugins that are stateful?

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,469 @@
...
+const CRT_FILENAME: &str = "node.crt";
+const KEY_FILENAME: &str = "node.key";
+
+// Required binaries must be in PATH
+const NEBULA_CERT_BIN: &str = "nebula-cert";
+
+fn create_dirname(path: &Path) -> Result<()> {
+    if let Some(dir) = path.parent() {
+        if !dir.exists() {
+            fs::create_dir_all(dir).with_context(|| format!("Create {} dir", dir.display()))?;
```
fitzthum, line 37: Doesn't `create_dir_all` handle some of these cases itself?

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,469 @@
...
+        }
+    }
+    Ok(())
+}
+
+fn remove_file(path: &Path) -> Result<()> {
```
fitzthum, line 43: This also seems slightly unnecessary but I guess it doesn't hurt.

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,469 @@
...
+    if path.exists() {
+        fs::remove_file(path).with_context(|| format!("Remove {} file", path.display()))?;
+    }
+    Ok(())
+}
+
+fn read_file_to_vec(path: &Path) -> Result<Vec<u8>> {
```
fitzthum, line 50: I think you might be able just use `std::fs::read` for this. Your helper function is very c-like.

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,469 @@
...
+    duration: Option<String>,
+    groups: Option<String>,
+    ips: Option<String>,
+    out_qr: Option<String>,
+    subnets: Option<String>,
+}
+
+impl From<&SelfSignedCaConfig> for Vec<OsString> {
+    fn from(config: &SelfSignedCaConfig) -> Self {
+        let mut v: Vec<OsString> = Vec::new();
```
fitzthum, line 190: this variable could have a better name

fitzthum (commented):
Looks good. A few preliminary comments. I haven't gotten into the meat of the `nebula.rs` yet

kbs/src/resource/plugin/nebula.rs
cclaudio: Right, it handle some. I removed `create_dirname` and moved code around.

kbs/src/resource/plugin/nebula.rs
cclaudio: I removed it as well and moved code around

kbs/src/resource/plugin/nebula.rs
cclaudio: Indeed, thanks

kbs/src/resource/plugin/mod.rs
cclaudio: We're not, replaced by `plugin.read().await`. Thanks

cclaudio:
Feedback applied, thanks @fitzthum 

I also updated the guest-components revision in the `Cargo.toml` to point to latest, which includes the query string support required by this PR.

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    pub fn get_version(&self) -> Result<String> {
+        let output = Command::new(NEBULA_CERT_BIN).arg("--version").output()?;
+        Ok(String::from_utf8(output.stdout)?)
+    }
+
+    pub fn test_all(&self) -> Result<()> {
+        self.test_nebula_cert_sign()
+    }
```
fitzthum, line 411: Do you call this test anywhere?

It's pretty common to put test code in a separate module rather than implementing it alongside the real methods. Not sure it's a big deal.

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,462 @@
...
+
+        Ok(Self {
+            _temp_dir: temp_dir,
+            crt,
+            key,
+        })
+    }
+
+    /// Run "nebula-cert sign" to generate a credential
+    pub fn generate(
```
fitzthum, line 327: maybe call this `generate_credential`

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,462 @@
...
+                        NEBULA_CONFIG_PATH
+                    ))?;
+
+                    let mut params: Vec<OsString> = Vec::from(&ca_config);
+                    params.push("-out-crt".into());
+                    params.push(ca.crt.as_path().into());
+                    params.push("-out-key".into());
+                    params.push(ca.key.as_path().into());
+
+                    let status = Command::new(NEBULA_CERT_BIN)
```
fitzthum, line 103: Should we check if this binary actually exists before trying to use it?

fitzthum (commented):
A few more comments

kbs/src/http/resource.rs
```diff
@@ -113,12 +114,25 @@ pub(crate) async fn get_resource(
...
+            .await
+            .map_err(|e| Error::ReadSecretFailed(e.to_string()))?
+    } else {
+        repository
+            .read()
+            .await
+            .read_secret_resource(resource_description)
+            .await
+            .map_err(|e| Error::ReadSecretFailed(e.to_string()))?
+    };
```
Xynnn007, line 135: We also need to update the KBS PROTOCOL document under [docs](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/kbs_attestation_protocol.md) to show that `plugin` is reserved.

kbs/config/docker-compose/kbs-config.toml
```diff
@@ -2,6 +2,14 @@ sockets = ["0.0.0.0:8080"]
...
 insecure_http = true
 
+[repository_config]
+type = "LocalFs"
+dir_path = "/opt/confidential-containers/kbs/storage"
+
+[plugin_manager_config]
+work_dir = "/opt/confidential-containers/kbs/plugin"
+enabled_plugins = []
+
```
Xynnn007, line 12: This would have an overall workdir for all plugins. Maybe we should let every single plugin has its own configs, like the workdir of them own?

Some config format in my mind
```toml
insecure_http = true
insecure_api = true

[attestation_token_config]
attestation_token_type = "CoCo"

[grpc_config]
as_addr = "http://127.0.0.1:50004"
pool_size = 200

[repository_config]
type = "LocalFs"
dir_path = "/opt/confidential-containers/kbs/storage"

[plugins.nebula]
config_path = "/etc/kbs/plugin/nebula-config.toml"

[plugins.another]
config_a=xxx
config_b=yyy
```

If only `[plugins.nebula]` is explicitly specified in the toml, `nebula` be initialized. In this way every single plugin could have its own config, or specify the config path inside that.

This could be implemented via
```rust
#[derive(Clone, Debug, Deserialize)]
pub struct KbsConfig {
...
    pub plugins: Option<PluginManagerConfig>,
}

#[derive(Clone, Debug, Deserialize)]
struct PluginManagerConfig {
    #[cfg(feature = "nebula-plugin")]
    nebula: Option<NebulaConfig>,

    #[cfg(feature = "another-plugin")]
    another: Option<AnotherPluginConfig>,
}

#[cfg(feature = "nebula-plugin")]
#[derive(Clone, Debug, Deserialize)]
struct NebulaConfig {
    config_path: String,
}

#[cfg(feature = "nebula-plugin")]
#[derive(Clone, Debug, Deserialize)]
struct AnotherPluginConfig {
    config_a:  String,
    config_b: String,
}
```

kbs/docker/coco-as-grpc/Dockerfile
```diff
@@ -2,18 +2,24 @@ FROM rust:latest as builder
 ARG ARCH=x86_64
 ARG HTTPS_CRYPTO=rustls
 ARG ALIYUN=false
+ARG PLUGINS="nebula-plugin"
```
Xynnn007, line 5: I suggest that we leave default plugins empty. Also, we can add some guides in docs to show how to enable nebula when building KBS image.

For this we might need to add a new doc about plugins [here](https://github.com/confidential-containers/trustee/tree/main/kbs/docs).

kbs/src/http/resource.rs
```diff
@@ -113,12 +114,25 @@ pub(crate) async fn get_resource(
...
+            )
+            .await
+            .map_err(|e| Error::ReadSecretFailed(e.to_string()))?
+    } else {
+        repository
+            .read()
+            .await
+            .read_secret_resource(resource_description)
+            .await
+            .map_err(|e| Error::ReadSecretFailed(e.to_string()))?
```
Xynnn007, line 134:
```suggestion
            .map_err(|e| Error::ReadSecretFailed(format!("{e:?}")))?;
```

kbs/src/http/resource.rs
```diff
@@ -113,12 +114,25 @@ pub(crate) async fn get_resource(
...
+        repository_plugin
+            .read()
+            .await
+            .get_resource(
+                resource_description.resource_type.as_str(),
+                resource_description.resource_tag.as_str(),
+                request.query_string(),
+            )
+            .await
+            .map_err(|e| Error::ReadSecretFailed(e.to_string()))?
```
Xynnn007, line 127:
```suggestion
            .map_err(|e| Error::ReadSecretFailed(format!("{e:?}")))?;
```

Xynnn007 (changes requested):
Thanks @cclaudio for this. I made some suggestions from a user perspective.

kbs/src/resource/plugin/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+    fn create_plugin(&self, work_dir: &str) -> Result<Arc<RwLock<dyn Plugin + Send + Sync>>>;
+}
+
+#[derive(Clone, Debug, Default, Deserialize)]
+pub struct PluginManagerConfig {
+    work_dir: String,
+    enabled_plugins: Vec<String>,
+}
+
+impl PluginManagerConfig {
```
mkulke, line 29: Are we sure the `PluginManager/PluginManagerConfig/PluginBuilder` abstractions are required? I am concerned they introduce accidental complexity without a use case. Afaict we don't really need to "manage" plugins, they are activated and configured at launch, and do not change at runtime.

IMO a plugin could be an implementation of this trait:

```rust
trait Plugin {
    async fn get_resource(&self, resource: &str, query_string: &str) -> Result<Vec<u8>>;
}
```

when parsing `KbsConfig` a Plugin can be initialized and registered in a lookup map.

```rust
// we don't want to have dynamic strings for plugin lookup, they are known at compile-time
const NEBULA: &str = "nebula";
...
// this needs to be in the app's state for lookup in the route handlers
let mut enabled_plugins: BTreeMap<&'static str, Box<dyn Plugin + Send + Sync>>;
...
let nebula_plugin = nebula::new(&kbs_config)?;
enabled_plugins.insert(NEBULA, Box::new(nebula_plugin));
```

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,462 @@
...
+                    log::info!("Nebula CA found, loading it")
+                }
+            }
+            x => {
+                bail!("CaGenerationPolicy {x} not supported");
+            }
+        };
+
+        log::info!("nebula-cert binary: {}", ca.get_version()?.trim());
+        ca.test_all()?;
```
portersrc, line 129: This line looks like you want to move it to a test section now

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,462 @@
...
+        match config.ca_generation_policy {
+            x if x == CaGenerationPolicy::GenerateIfNotFound as u32 => {
+                if !ca.crt.exists() || !ca.key.exists() {
+                    // Clean-up in case the CA failed to generate last time
+                    if ca.crt.exists() {
+                        fs::remove_file(ca.crt.as_path())
+                            .with_context(|| format!("Remove {} file", ca.crt.display()))?;
+                    }
+                    if ca.key.exists() {
+                        fs::remove_file(ca.crt.as_path())
```
portersrc, line 79: maybe paste mistake: `ca.key`

cclaudio:
Thanks for all feedback! We are improving the plugin interface in the #502, thanks @Xynnn007 . We will need to revisit this PR later.

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,462 @@
...
+            work_dir: PathBuf::from(work_dir),
+        };
+
+        if let Some(parent) = ca.work_dir.as_path().parent() {
+            fs::create_dir_all(parent)
+                .with_context(|| format!("Create {} dir", parent.display()))?;
+        }
+
+        match config.ca_generation_policy {
+            x if x == CaGenerationPolicy::GenerateIfNotFound as u32 => {
```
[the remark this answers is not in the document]
cclaudio:
```
#11 132.6 70 |         match config.ca_generation_policy {
#11 132.6    |               --------------------------- this expression has type `u32`
#11 132.6 71 |             CaGenerationPolicy::GenerateIfNotFound => {
#11 132.6    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `CaGenerationPolicy`
```
Rust does not allow casting in the left side of a match case. So we have to either convert `config.ca_generation_policy` to `CaGenerationPolicy` or use the expression trick `x if x ==` in the match case.

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,462 @@
...
+                if !ca.crt.exists() || !ca.key.exists() {
```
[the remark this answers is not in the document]
cclaudio: Yeah, makes sense. I will create a helper function to generate the cert. Thanks.

kbs/src/resource/plugin/nebula.rs
```diff
@@ -0,0 +1,462 @@
...
+            }
+            x if x == CaGenerationPolicy::NeverGenerate as u32 => {
+                if !ca.crt.exists() || !ca.key.exists() {
+                    bail!("Nebula CA not found");
+                } else {
+                    log::info!("Nebula CA found, loading it")
+                }
+            }
+            x => {
+                bail!("CaGenerationPolicy {x} not supported");
```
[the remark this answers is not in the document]
cclaudio: I can convert `config.ca_generation_policy` to `CaGenerationPolicy` outside of the match to avoid this fallback.

kbs/src/resource/plugin/nebula.rs
cclaudio: It is called when the plugin is initialized. It runs simple tests to check if the provided `nebula-cert` is working.

cclaudio:
New PR submitted #539 for nebula. Closing this one.

PR #458, opened 2024-08-08
kbs: token: add verifier with JSON Web Keys

mythi:
Add a new token verifier that uses JSON Web Keys (JWK) from
the configured JWK Set sources.

JWK Sets can be provided locally using file:// URL schema
or they can be downloaded automatically via OpenID Connect configuration
https:// URLs providing a pointer via "jwks_uri".

### DIFF
--- a/kbs/src/lib.rs
+++ b/kbs/src/lib.rs
@@ -240,7 +240,7 @@ impl ApiServer {
 
         #[cfg(feature = "resource")]
         let token_verifier =
-            crate::token::create_token_verifier(self.attestation_token_config.clone())?;
+            crate::token::create_token_verifier(self.attestation_token_config.clone()).await?;
 
         #[cfg(feature = "policy")]
         let policy_engine = PolicyEngine::new(&self.policy_engine_config).await?;
--- a/kbs/src/token/jwk.rs
+++ b/kbs/src/token/jwk.rs
@@ -1,0 +1,161 @@
+// Copyright (c) 2024 by Intel Corporation
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::token::{AttestationTokenVerifier, AttestationTokenVerifierConfig};
+use anyhow::*;
+use async_trait::async_trait;
+use jsonwebtoken::{decode, decode_header, jwk, Algorithm, DecodingKey, Validation};
+use reqwest::{get, Url};
+use serde::Deserialize;
+use serde_json::Value;
+use std::fs::File;
+use std::io::BufReader;
+use std::result::Result::Ok;
+use std::str::FromStr;
+use thiserror::Error;
+
+const OPENID_CONFIG_URL_SUFFIX: &str = ".well-known/openid-configuration";
+
+#[derive(Error, Debug)]
+pub enum JwksGetError {
+    #[error("Invalid source path: {0}")]
+    SourcePath(String),
+    #[error("Failed to access source: {0}")]
+    SourceAccess(String),
+    #[error("Failed to deserialize source data: {0}")]
+    SourceDeserializeJson(String),
+}
+
+#[derive(Deserialize)]
+struct OpenIDConfig {
+    jwks_uri: String,
+}
+
+pub struct JwkAttestationTokenVerifier {
+    trusted_certs: jwk::JwkSet,
+}
+
+pub async fn get_jwks_from_file_or_url(p: &str) -> Result<jwk::JwkSet, JwksGetError> {
+    let mut url = Url::parse(p).map_err(|e| JwksGetError::SourcePath(e.to_string()))?;
+    match url.scheme() {
+        "https" => {
+            url.set_path(OPENID_CONFIG_URL_SUFFIX);
+
+            let oidc = get(url.as_str())
+                .await
+                .map_err(|e| JwksGetError::SourceAccess(e.to_string()))?
+                .json::<OpenIDConfig>()
+                .await
+                .map_err(|e| JwksGetError::SourceDeserializeJson(e.to_string()))?;
+
+            let jwkset = get(oidc.jwks_uri)
+                .await
+                .map_err(|e| JwksGetError::SourceAccess(e.to_string()))?
+                .json::<jwk::JwkSet>()
+                .await
+                .map_err(|e| JwksGetError::SourceDeserializeJson(e.to_string()))?;
+
+            Ok(jwkset)
+        }
+        "file" => {
+            let file = File::open(url.path())
+                .map_err(|e| JwksGetError::SourceAccess(format!("open {}: {}", url.path(), e)))?;
+
+            serde_json::from_reader(BufReader::new(file))
+                .map_err(|e| JwksGetError::SourceDeserializeJson(e.to_string()))
+        }
+        _ => Err(JwksGetError::SourcePath(format!(
+            "unsupported scheme {} (must be either file or https)",
+            url.scheme()
+        ))),
+    }
+}
+
+impl JwkAttestationTokenVerifier {
+    pub async fn new(config: &AttestationTokenVerifierConfig) -> Result<Self> {
+        let mut trusted_certs = jwk::JwkSet { keys: Vec::new() };
+
+        for path in config.trusted_certs_paths.iter() {
+            match get_jwks_from_file_or_url(path).await {
+                Ok(mut jwkset) => trusted_certs.keys.append(&mut jwkset.keys),
+                Err(e) => log::warn!("error getting JWKS: {:?}", e),
+            }
+        }
+
+        Ok(Self { trusted_certs })
+    }
+}
+
+#[async_trait]
+impl AttestationTokenVerifier for JwkAttestationTokenVerifier {
+    async fn verify(&self, token: String) -> Result<String> {
+        if self.trusted_certs.keys.is_empty() {
+            bail!("Cannot verify token since trusted JWK Set is empty");
+        };
+
+        let kid = decode_header(&token)
+            .context("Failed to decode attestation token header")?
+            .kid
+            .ok_or(anyhow!("Failed to decode kid in the token header"))?;
+
+        let key = &self
+            .trusted_certs
+            .find(&kid)
+            .ok_or(anyhow!("Failed to find Jwk with kid {kid} in JwkSet"))?;
+
+        let key_alg = key
+            .common
+            .key_algorithm
+            .ok_or(anyhow!("Failed to find key_algorithm in Jwk"))?
+            .to_string();
+
+        let alg = Algorithm::from_str(key_alg.as_str())?;
+
+        let dkey = DecodingKey::from_jwk(key)?;
+        let token_data = decode::<Value>(&token, &dkey, &Validation::new(alg))
+            .context("Failed to decode attestation token")?;
+
+        Ok(serde_json::to_string(&token_data.claims)?)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::token::jwk::get_jwks_from_file_or_url;
+    use rstest::rstest;
+
+    #[rstest]
+    #[case("https://", true)]
+    #[case("http://example.com", true)]
+    #[case("file:///does/not/exist/keys.jwks", true)]
+    #[case("/does/not/exist/keys.jwks", true)]
+    #[tokio::test]
+    async fn test_source_path_validation(#[case] source_path: &str, #[case] expect_error: bool) {
+        assert_eq!(
+            expect_error,
+            get_jwks_from_file_or_url(source_path).await.is_err()
+        )
+    }
+
+    #[rstest]
+    #[case(
+        "{\"keys\":[{\"kty\":\"oct\",\"alg\":\"HS256\",\"kid\":\"coco123\",\"k\":\"foobar\"}]}",
+        false
+    )]
+    #[case(
+        "{\"keys\":[{\"kty\":\"oct\",\"alg\":\"COCO42\",\"kid\":\"coco123\",\"k\":\"foobar\"}]}",
+        true
+    )]
+    #[tokio::test]
+    async fn test_source_reads(#[case] json: &str, #[case] expect_error: bool) {
+        let tmp_dir = tempfile::tempdir().expect("to get tmpdir");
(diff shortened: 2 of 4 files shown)


zvonkok:
Can you add a meaningfull description? Thanks.

kbs/src/lib.rs
```diff
@@ -11,6 +11,7 @@ extern crate anyhow;
 extern crate base64;
 extern crate env_logger;
 extern crate kbs_types;
+#[allow(unused_imports)]
```
Xynnn007, line 14: why this?

kbs/src/token/coco.rs
```diff
@@ -21,7 +21,7 @@ pub struct CoCoAttestationTokenVerifier {
 }
 
 impl CoCoAttestationTokenVerifier {
-    pub fn new(config: &AttestationTokenVerifierConfig) -> Result<Self> {
+    pub async fn new(config: &AttestationTokenVerifierConfig) -> Result<Self> {
```
Xynnn007, line 24: We do not need this if we want outer called to be async.

kbs/src/lib.rs
mythi: the compiler triggers an unused_imports warning when buiding without `as`. See the commit message https://github.com/confidential-containers/trustee/pull/458/commits/83fa39013576a86e98faadef3e5b0e5b4af81d28

kbs/src/token/coco.rs
mythi: ah ok, let me drop

mythi:
@Xynnn007 a bit offtopic but is there a reason why `CoCo` tokenverifier could not use `jsonwebtoken` also? AFAUI, we could have a single verifier that just understands local pem/jwks and OpenID Connect provided jwks:

`trusted_cert_paths = ["/path/to/foo.pem", "https://provider.openid.conf", "/path/to/bar.jwks"]`

kbs/src/token/coco.rs
mythi: done

mythi:
> Can you add a meaningfull description? Thanks.

done

Xynnn007:
> @Xynnn007 a bit offtopic but is there a reason why `CoCo` tokenverifier could not use `jsonwebtoken` also? AFAUI, we could have a single verifier that just understands local pem/jwks and OpenID Connect provided jwks:

I did not remember the concrete reason for it is a long time ago -- but I think it makes sense to use a common crate for the same functions.

mythi:
> > @Xynnn007 a bit offtopic but is there a reason why `CoCo` tokenverifier could not use `jsonwebtoken` also? AFAUI, we could have a single verifier that just understands local pem/jwks and OpenID Connect provided jwks:
> 
> I did not remember the concrete reason for it is a long time ago -- but I think it makes sense to use a common crate for the same functions.

I was also wondering if it let us to drop `dep:openssl`.

huoqifeng:
> I was also wondering if it let us to drop `dep:openssl`.
For some verifier like IBMSE, `openssl` is the only certified cryptography library.

mythi:
> > I was also wondering if it let us to drop `dep:openssl`.
> 
> For some verifier like IBMSE, `openssl` is the only certified cryptography library.

OK. I was mainly referring to the CoCo tokenverifier code.

mythi:
@jialez0 PTAL

mythi:
rebased + force pushed. however, this change is relatively independent and does not add any new dependencies. are there any concerns or reasons to wait with this?

kbs/src/lib.rs
Xynnn007: I suggest not to use macro mod-level, but the concrete place it warns. Or, this would let future unused-imports error not be raised.

Xynnn007 (approved):
Only two left comments.

kbs/src/token/jwk.rs
```diff
@@ -0,0 +1,166 @@
...
+struct OpenIDConfig {
+    jwks_uri: String,
+}
+
+pub struct JwkAttestationTokenVerifier {
+    trusted_certs: Option<jwk::JwkSet>,
+}
+
+pub async fn get_jwks_from_file_or_url(p: &str) -> Result<jwk::JwkSet, JwksGetError> {
+    match Url::parse(p) {
```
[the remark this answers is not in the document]
mythi: the current `match` combines all checks under it. It is an error also when some insecure scheme is used.

mythi:
> Only two left comments.

what's the other one?

Xynnn007:
@mythi it is https://github.com/confidential-containers/trustee/pull/458#discussion_r1730697409

kbs/src/token/jwk.rs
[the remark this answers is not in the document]
Xynnn007: Ok. A first impression is
```rust
// failed to parse
let url = Url::parse(p).map_err(...)?;

if url.scheme() == "https" {
    ...
}

if url.scheme() == "file" && Path::new(url.path()).exists() {
    ...
}

...
```

kbs/src/lib.rs
mythi: I can drop it. It has nothing to do with this PR.

mythi: @Xynnn007 btw, this is exactly where the place is:

```
warning: unused `#[macro_use]` import
  --> kbs/src/lib.rs:14:1
   |
14 | #[macro_use]
   | ^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: `kbs` (lib) generated 1 warning
```

kbs/src/token/jwk.rs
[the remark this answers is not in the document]
mythi: I can move the error check arm to be the first one?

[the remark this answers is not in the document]
mythi: moved

kbs/src/lib.rs
mythi: I dropped this and submitted and issue #480

kbs/src/token/jwk.rs
```diff
@@ -0,0 +1,166 @@
...
+                        Err(e) => log::warn!("error getting JWKS: {:?}", e),
+                    }
+                }
+
+                match keyset.keys.len() {
+                    0 => None,
+                    _ => Some(keyset),
+                }
+            }
+            None => None,
```
fitzthum, line 95: I wouldn't use a match here since there is really no logic in this case. You can just use `map` instead.

kbs/src/token/jwk.rs
```diff
@@ -0,0 +1,166 @@
...
+        let kid = decode_header(&token)
+            .context("Failed to decode attestation token header")?
+            .kid
+            .ok_or(anyhow!("Failed to decode kid in the token header"))?;
+
+        let key = keyset
+            .find(&kid)
+            .ok_or(anyhow!("Failed to find kid in trusted certificates"))?;
+
+        let alg = Algorithm::from_str(key.common.key_algorithm.unwrap().to_string().as_str())?;
```
fitzthum, line 118: Do we know that this key algorithm will always be there? You might split this line into two. It's a bit of a mouthful.

fitzthum (commented):
A few comments on style but looks good overall.

kbs/src/token/jwk.rs
Xynnn007: Agreed. With the type change of `trusted_certs_paths`, we can simply write as
```rust
    pub async fn new(config: AttestationTokenVerifierConfig) -> Result<Self> {
        let keys = stream::iter(config.trusted_certs_paths)
            .filter_map(|p| async move {
                match get_jwks_from_file_or_url(&p).await {
                    Ok(jwkset) => Some(jwkset.keys),
                    Err(e) => {
                        warn!("error getting JWKS: {e:?}");
                        None
                    }
                }
            })
            .concat()
            .await;

        Ok(Self { trusted_certs: jwk::JwkSet { keys } })
    }
```

where `futures_util` crate is needed.

kbs/src/token/jwk.rs
```diff
@@ -0,0 +1,166 @@
...
+    SourceDeserializeJson(String),
+}
+
+#[derive(Deserialize)]
+struct OpenIDConfig {
+    jwks_uri: String,
+}
+
+pub struct JwkAttestationTokenVerifier {
+    trusted_certs: Option<jwk::JwkSet>,
```
Xynnn007, line 37: Can we change this to `jwk::JwkSet`? Because inside it the `keys` is a `Vec`, hinting that `keys` could be an empty `Vec`.

kbs/src/token/mod.rs
```diff
@@ -22,13 +22,17 @@ pub trait AttestationTokenVerifier {
...
 #[derive(Deserialize, Debug, Clone)]
 pub struct AttestationTokenVerifierConfig {
     pub attestation_token_type: AttestationTokenVerifierType,
 
-    // Trusted Certificates file (PEM format) path to verify Attestation Token Signature.
+    // Trusted Certificates file (PEM format) path (for "CoCo") or a valid Url
+    // (file:// and https:// schemes accepted) pointing to a local JWKSet file
+    // or to an OpenID configuration url giving a pointer to JWKSet certificates
+    // (for "Jwk") to verify Attestation Token Signature.
     pub trusted_certs_paths: Option<Vec<String>>,
```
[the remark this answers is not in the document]
mythi: `CoCo` expects `Option`

kbs/src/token/jwk.rs
mythi: Right, I guess I blindly followed the `CoCo` approach so it ended up looking the same

kbs/src/token/mod.rs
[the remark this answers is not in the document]
Xynnn007: This is somehow a code style leftover. Vec is enough as it could be empty vec, we do not need to redefine option around it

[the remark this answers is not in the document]
mythi: OK, I'll get that changed.

kbs/src/token/jwk.rs
mythi: > where `futures_util` crate is needed.

the proposal looks nice but I wonder if this is justified for the functionality here?

kbs/src/token/jwk.rs
```diff
@@ -0,0 +1,166 @@
...
+            let jwkset = get(oidc.jwks_uri)
+                .await
+                .map_err(|e| JwksGetError::SourceAccess(e.to_string()))?
+                .json::<jwk::JwkSet>()
+                .await
+                .map_err(|e| JwksGetError::SourceDeserializeJson(e.to_string()))?;
+
+            Ok(jwkset)
+        }
+        Ok(url) if url.scheme() == "file" && Path::new(url.path()).exists() => {
```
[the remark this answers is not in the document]
mythi: right, letting `File::open()` to error is sufficient.

kbs/src/token/jwk.rs
mythi: yeah better check for `Option<KeyAlgorithm>` is justified

kbs/src/token/jwk.rs
Xynnn007: What the `stream::iter` does, is try to do `get_jwks_from_file_or_url` concurrently if multiple paths in `trusted_certs_paths` are given. This makes sense to me. wdyt?

kbs/src/token/mod.rs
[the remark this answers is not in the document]
mythi: I moved some of these to #483. Will rebase this PR on top of it.

kbs/src/token/jwk.rs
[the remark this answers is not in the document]
mythi: updated.

kbs/src/token/jwk.rs
mythi: updated.

kbs/src/token/jwk.rs
mythi: updated.

kbs/src/token/jwk.rs
mythi: My use-case is with one URL only but I understand we may have several `trusted_certs_path` in the future where parallelism *might* help to improve the KBS startup time.

fitzthum (approved):

fitzthum (approved):
See my comment on Ding's thread, but generally this seems fine.

kbs/src/token/jwk.rs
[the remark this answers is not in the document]
mythi: I just updated this part too, thanks for the feedback!

fitzthum:
Let's merge this if you are happy with it @Xynnn007

kbs/src/token/jwk.rs
```diff
@@ -0,0 +1,161 @@
...
+
+        let kid = decode_header(&token)
+            .context("Failed to decode attestation token header")?
+            .kid
+            .ok_or(anyhow!("Failed to decode kid in the token header"))?;
+
+        let key = &self
+            .trusted_certs
+            .find(&kid)
+            .ok_or(anyhow!("Failed to find Jwk with kid in JwkSet"))?;
```
Xynnn007, line 105:
```suggestion
            .ok_or(anyhow!("Failed to find Jwk with kid {kid} in JwkSet"))?;
```

kbs/src/token/mod.rs
```diff
@@ -22,25 +23,33 @@ pub trait AttestationTokenVerifier {
...
 #[derive(Deserialize, Debug, Clone)]
 pub struct AttestationTokenVerifierConfig {
     #[serde(default)]
     pub attestation_token_type: AttestationTokenVerifierType,
 
-    // Trusted Certificates file (PEM format) path to verify Attestation Token Signature.
+    // Trusted Certificates file (PEM format) path (for "CoCo") or a valid Url
+    // (file:// and https:// schemes accepted) pointing to a local JWKSet file
+    // or to an OpenID configuration url giving a pointer to JWKSet certificates
+    // (for "Jwk") to verify Attestation Token Signature.
```
Xynnn007, line 37: Let's use `///` instead of `//` to work as inline documents.

kbs/src/token/jwk.rs
mythi: added

kbs/src/token/mod.rs
mythi: changed to `///`

PR #462, opened 2024-08-12
ibmse: use hash rather than hex for initdata digest in claims

huoqifeng:
Use the initdata hash value directly rather than it's hex in claims.
It's more reasonable to use the fixed length hash value `initdata.digest` rather than its HEX in claims because it's more straight forward for user to set the hash value in attestation policy.

Related PR:
https://github.com/confidential-containers/guest-components/pull/616
https://github.com/confidential-containers/cloud-api-adaptor/pull/1991


huoqifeng:
@liudalibj @mkulke @Xynnn007 @fitzthum

mkulke (approved):

liudalibj (approved):
LGTM

PR #469, opened 2024-08-19
kbs: msic fix in self-signed-https.md

huoqifeng:
- Fix the miscs in self-signed-https.md

When I trying the https kbs, I found some misc problems and use this PR to fix them.


Xynnn007 (approved):
Thanks for the fix

PR #483, opened 2024-08-28
kbs: token: configuration cleanup

mythi:
Moving some changes from #458 to this separate PR.

While working on this, I noticed KBS can be run without `trusted_certs_paths` and have resource access approved. Are we OK with this or would it be better to have an explicit user opt-in?

### DIFF
--- a/kbs/src/token/coco.rs
+++ b/kbs/src/token/coco.rs
@@ -7,6 +7,7 @@ use anyhow::*;
 use async_trait::async_trait;
 use base64::engine::general_purpose::URL_SAFE_NO_PAD;
 use base64::Engine;
+use log::warn;
 use openssl::hash::MessageDigest;
 use openssl::pkey::PKey;
 use openssl::rsa::Rsa;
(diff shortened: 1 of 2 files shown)


kbs/src/token/coco.rs
```diff
@@ -90,8 +94,8 @@ impl AttestationTokenVerifier for CoCoAttestationTokenVerifier {
             }
         }
 
-        let Some(trusted_store) = &self.trusted_certs else {
-            log::warn!("No Trusted Certificate in Config, skip verification of JWK cert of Attestation Token");
+        if self.trusted_certs.all_certificates().is_empty() {
+            warn!("No Trusted Certificate in Config, skip verification of JWK cert of Attestation Token");
```
Xynnn007, line 98: No verification on the token against trust anchor is dangerous. I could get the point that for debug case. We'd better add a configuration in config file to control this and by default the trust anchor verification is enforced.

mythi: I'm not changing the logic. That's the same comment I was making in the cover letter.

fitzthum (approved):
LGTM

We might want to think about the case where there are no certs, but that can be a separate PR. Let's make an issue tho.

mythi:
> LGTM
> 
> We might want to think about the case where there are no certs, but that can be a separate PR. Let's make an issue tho.

Created #486

PR #491, opened 2024-09-06
kbs: ita: Set hash algorithm based on TEE type

jodh-intel:
If the TEE specifies the hash algorithms it can use [1], add the appropriate hash algorithm to the returned `Challenge` [2].

For backwards compatibility, do not return the selected hash algorithm if the TEE does not provide the list of hash algorithms it can use.

Partially-fixes: #242.

[1] - In the optional `extra-params.supported-hash-algorithms` list.
[2] - In `extra-params.selected-hash-algorithm`.


jodh-intel:
Once I've taken it out of draft mode, this PR should land with https://github.com/confidential-containers/guest-components/pull/712.

kbs/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -131,6 +158,84 @@ impl Attest for IntelTrustAuthority {
...
+            } else {
+                return Err(anyhow!(
+                    "Intel Trust Authority: expected array, found {hash_algorithms_found:?}"
+                ));
+            }
+        } else {
+            log::info!("ITA: generate_challenge: no TEE hash parameters, so falling back to legacy behaviour");
+
+            return generic_generate_challenge(tee, tee_parameters).await;
+        }
```
Xynnn007, line 198: We could use 
```rust
let Some(xxx) = yyy() else {
    bail!(".....");
}

// do next steps with xxx
```

to avoid nested `{ ... }`s .

kbs/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -7,14 +7,41 @@ use crate::token::{
...
+    static ref TDX_HASH_ALGORITHM: String = HashAlgorithm::Sha512.to_string().to_lowercase();
+
+    static ref ERR_NO_TDX_ALGORITHM: String = format!(
+        "Intel Trust Authority: {:?} not supported by TDX TEE",
+        *TDX_HASH_ALGORITHM
+    );
+    static ref ERR_NO_SGX_ALGORITHM: String = format!(
+        "Intel Trust Authority: {:?} not supported by SGX TEE",
+        *TDX_HASH_ALGORITHM
+    );
```
Xynnn007, line 36: We might not need these consts. For `SGX_HASH_ALGORITHM`, we can 
```rust
enum HashAlgorithm {
    #[strum(serialize = "sha256")]
    Sha256,
    ...
}

...
if supported_hash_algorithms.contains(HashAlgorithm::Sha256.as_ref()) {
...
}
```

For errors, I suggest that directly use `.context("...")` for extra error context.

jodh-intel:
@Xynnn007 - Thanks for reviewing! Branch updated.

/cc @mythi.

kbs/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -3,18 +3,27 @@
...
 
 use super::Attest;
+use crate::attestation::{generic_generate_challenge, make_nonce};
 use crate::token::{
     jwk::JwkAttestationTokenVerifier, AttestationTokenVerifier, AttestationTokenVerifierConfig,
     AttestationTokenVerifierType,
 };
 use anyhow::*;
 use async_trait::async_trait;
+use attestation_service::HashAlgorithm;
```
mythi, line 13: I had somehow missed this dependency earlier. This is slightly problematic as it is right now. This makes one AS to depend on another AS code just for the sake of `HashAlgorithm` enum. Not only that but `attestation-service/default` makes the build to include all of the CoCo-AS verifier dependencies, such as the vtpm stuff that now makes se CI docker build to fail because the ITA `Dockerfile` does not have `libtss2-dev` installed.

jodh-intel: Branch updated. I've made a local definition of `HashAlgorithm` to resolve that issue.

Once this and https://github.com/confidential-containers/guest-components/pull/712 land, I plan to tal at consolidating all the definitions though :smile:

kbs/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -131,6 +152,88 @@ impl Attest for IntelTrustAuthority {
...
+    async fn generate_challenge(
+        &self,
+        tee: Tee,
+        tee_parameters: serde_json::Value,
+    ) -> Result<Challenge> {
+        log::debug!("ITA: generate_challenge: tee: {tee:?}, tee_parameters: {tee_parameters:?}");
+
+        if tee_parameters.is_null() {
+            log::debug!(
+                "ITA: generate_challenge: no TEE parameters so falling back to legacy behavour"
```
mkulke, line 165:
```suggestion
                "ITA: generate_challenge: no TEE parameters so falling back to legacy behaviour"
```

kbs/src/attestation/mod.rs
```diff
@@ -40,6 +40,18 @@ pub async fn make_nonce() -> Result<String> {
     Ok(STANDARD.encode(&nonce))
 }
 
+pub(crate) async fn generic_generate_challenge(
+    _tee: Tee,
+    _tee_parameters: serde_json::Value,
+) -> Result<Challenge> {
```
mkulke, line 46: why does this have params if we don't use them? Should we maybe do `impl Default for Challenge`?

mkulke (commented):
I wonder if it would be better to move the hash_alg support to the protocol level instead of putting it into the schemaless extra_params field. it can be optional, defaulting to sha348. atm, this seems local to ITA and otherwise unspecified strings `supported/selected-hash-algorithms`.

Sadly, this would need a (breaking) change on kbs-types, but it would make things a bit more explicit:

```bash
Challenge {
   nonce String,
   hash_alg: HashAlgorithm
   extra_params: Value
}
```

Maybe, if we don't want to touch kbs-types we can also introduce trustee `ExtraParams` that are more than just Value, but specify the hash_alg field.

kbs/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -131,6 +152,88 @@ impl Attest for IntelTrustAuthority {
...
+        let Some(algorithms) = hash_algorithms_found.as_array() else {
+            return Err(anyhow!(
+                "ITA: expected array, found {hash_algorithms_found:?}"
+            ));
+        };
+
+        let hash_algorithms: Vec<String> = algorithms
+            .iter()
+            .filter_map(|s| s.as_str())
+            .map(|s| s.to_lowercase())
```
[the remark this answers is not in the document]
jodh-intel: It's a bit ugly now, but updated.

kbs/src/attestation/intel_trust_authority/mod.rs
jodh-intel: Fixed.

kbs/src/attestation/mod.rs
jodh-intel: I was trying to make this a minimal change and if you look at the existing code, you'll see the same unused params.

>  Should we maybe do impl Default for Challenge?

I thought of that, but `Challenge` is defined in the `kbs-types` crate, so that can be done at a later time imho.

mkulke:
> @mkulke - Thanks for the review. I agree that this can all be improved, but since we'd really like this PR and [confidential-containers/guest-components#712](https://github.com/confidential-containers/guest-components/pull/712) to be in the next release, and since testing these takes some time, can we defer some of these comments until after the upcoming release(s)?

I see `SUPPORTED_HASH_ALGORITHMS` is generalized for all TEE on the guest-components PR, but only ITA is using it at the moment. I don't think there's much harm in merging this as-is, but maybe we can have a follow up ticket to make the hash_alg a more explicit part of the AS protocol.

mkulke (approved):

fitzthum (approved):
LGTM. This should be part of the protocol imo but we can do that later.

mythi:
> > @mkulke - Thanks for the review. I agree that this can all be improved, but since we'd really like this PR and [confidential-containers/guest-components#712](https://github.com/confidential-containers/guest-components/pull/712) to be in the next release, and since testing these takes some time, can we defer some of these comments until after the upcoming release(s)?
> 
> I see `SUPPORTED_HASH_ALGORITHMS` is generalized for all TEE on the guest-components PR, but only ITA is using it at the moment. I don't think there's much harm in merging this as-is, but maybe we can have a follow up ticket to make the hash_alg a more explicit part of the AS protocol.

I can create the follow-up ticket with additional thoughts. `hash_alg` is just one part of the puzzle. What/how things are hashed requires thinking too.

kbs/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -3,17 +3,38 @@
...
 use reqwest::header::{ACCEPT, CONTENT_TYPE};
 use serde::{Deserialize, Serialize};
 use serde_json::json;
+use strum::{AsRefStr, Display, EnumString};
+
+const SUPPORTED_HASH_ALGORITHMS_JSON_KEY: &str = "supported-hash-algorithms";
+const SELECTED_HASH_ALGORITHM_JSON_KEY: &str = "selected-hash-algorithm";
+
+const ERR_NO_TEE_ALGOS: &str = "ITA: TEE does not support any hash algorithms";
+const ERR_INVALID_TEE: &str = "ITA: Unknown TEE specified";
```
Xynnn007, line 25: Let me do a refactoring about error information assertion later. Now it's good to me.

kbs/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -131,6 +152,87 @@ impl Attest for IntelTrustAuthority {
...
+        else {
+            log::info!("ITA: generate_challenge: no TEE hash parameters, so falling back to legacy behaviour");
+
+            return generic_generate_challenge(tee, tee_parameters).await;
+        };
+
+        let Some(algorithms) = hash_algorithms_found.as_array() else {
+            return Err(anyhow!(
+                "ITA: expected array, found {hash_algorithms_found:?}"
+            ));
```
Xynnn007, line 183:
```suggestion
            bail!(
                "ITA: expected array, found {hash_algorithms_found:?}"
            );
```

Xynnn007 (approved):
Overall lgtm. Only one nit. Thanks! @jodh-intel

Xynnn007:
Btw, shall we cut a release after ITA is supported on both gc and trustee side? 

cc @fitzthum @mkulke @mythi

mythi:
> Btw, shall we cut a release after ITA is supported on both gc and trustee side?

I'm about to submit a PR for kustomization (`kbs/config/kubernetes`) still once we get the final validation with these PRs completed.

mythi:
@Xynnn007 we have this one also fully tested so it's fine to merge. We will rebase #494 on top of `HEAD` after that.

fidencio:
I've triggered the e2e tests as well, so let's wait for them to finish.

fidencio:
Merged, thanks @jodh-intel for the work and all the reviewers for the reviews! :-)

PR #497, opened 2024-09-11
bump guest-components + ITA kustomization

mythi:


kbs/config/kubernetes/ita/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: coco-tenant
+
+resources:
+- ../overlays/x86_64
+
```
fidencio, line 7: Just a note fur the future, but this doesn't block this PR.

I've asked @mythi about having the ITA image set here by default, but it seems that even if we have that the kustomization is not picking up the correct image.

We'll solve this and add the image here as a follow-up.

fidencio (approved):
lgtm, thanks @mythi!

jodh-intel (approved):
Thanks, @mythi.

lgtm

Xynnn007:
Thanks. Do we need to firstly bump gc and then change the rev of trustee to use that one?

fidencio:
> Thanks. Do we need to firstly bump gc and then change the rev of trustee to use that one?

We're using the guest components hash of the latest commit, which includes the changes that we need, so I think we're good to go here @Xynnn007.

My next step will do bump everything on the Kata Containers side as soon as this one is merged.

Xynnn007 (approved):

PR #498, opened 2024-09-11
ita: Build the kustomization based on nodeport

fidencio:
While testing on an environment as close as possible to the Kata Containers CI, I've noticed that we need to build ITA based on nodeport in order to get it working as expected there.

The tests I ran earlier, in a different environment, already had this based on nodeport, but I didn't realise it while reviewing bdaa4b2185d.


huoqifeng (approved):
LGTM, thanks! @fidencio

mythi (approved):
+1

mkulke (approved):

PR #504, opened 2024-09-13
kbs: update kustomization yaml to v0.10.0 & fix release script

Xynnn007:


mythi:
The newly added `ita` overlay uses this as the base so the versions need updating there too to get the match.

mkulke (approved):

Xynnn007:
Added a commit to fix ITA image pushing. Once this get merged I will manually run image tagging script and then do 0.10.1 tagging

mythi (approved):

Xynnn007:
> @Xynnn007 I glanced at this after it went in. I'm wondering if there was a copy paste error [here](https://github.com/confidential-containers/trustee/pull/504/files#diff-966b7c0c30e4092f47aaabdb67842e7961303a249df947f48078680ed974d8c2R110).
> That is, should this...
> ```
>         docker manifest create ${ghcr_repo}/${release_pkg_name}:${release_tag_full} \
>             --amend ${ghcr_repo}/${release_pkg_name}:${release_tag_full}-x86_64 \
>             --amend ${ghcr_repo}/${release_pkg_name}:${release_tag_full}-s390x
>         docker manifest push ${ghcr_repo}/${release_pkg_name}:latest
> ```
> be changed to this...
> ```
>         docker manifest create ${ghcr_repo}/${release_pkg_name}:latest \
>             --amend ${ghcr_repo}/${release_pkg_name}:${release_tag_full}-x86_64 \
>             --amend ${ghcr_repo}/${release_pkg_name}:${release_tag_full}-s390x
>         docker manifest push ${ghcr_repo}/${release_pkg_name}:latest
> ```
> (The first line has "latest" as the tag.) If you ran it and it tagged correctly then it should be fine. I'm not too familiar with `docker manifest create`, but I wanted to check.

Oh, yes. You are right. I manually executed the scripts. We need another fix for this.

portersrc:
Hm ok. I'm also noticing that this would execute inside that loop a couple times, which is probably not wanted either. When I look at release stuff next week, I'll double-check and put a PR to try to tweak it.

Xynnn007:
> Hm ok. I'm also noticing that this would execute inside that loop a couple times, which is probably not wanted either. When I look at release stuff next week, I'll double-check and put a PR to try to tweak it.

Thanks for this! BTW the PR making script seems not workable on my side. Also, I think we could integrate the image tagging task to a GHA triggered by a release action. Thus, we only need to
- update kustomization and guest-components rev in a PR
- get it merged and wait for the image to be built
- tag the repo

PR #506, opened 2024-09-13
Added missing usage import and updated certs_file property in documentation

pawelpros:


fitzthum (approved):
LGTM

Seems like this will miss the release.

mythi (approved):

rvps/build.rs
```diff
@@ -1,4 +1,4 @@
-use std::process::exit;
+use std::process::{exit, Command};
```
mythi, line 1: `make lint` complains about the opposite: `error: unused import: `Command``

pawelpros: Yeah - it's related with checking in-toto feature with is by default not set -> changed approach, should be ok now

mythi (approved):

Issue #162, opened 2023-09-20
KBS implementation observations


mythi:
While working with #159 and a bit of #151 too I made some observations I wanted to capture here for discussion.

1. `Nonce` from the attestation-service
With the current flow KBS creates the nonce and it has a lifetime of the session (which is configurable in kbs parameters). AS does not verify the nonce is unique and fresh. Thus, it may be desirable to have the AS to generate the nonce with some lifetime so that it becomes easier to track if a nonce is valid for a one-time usage.

2. TEE runtime data/nonce hashing
AFAICS, this is a "compatibility issue" as there seems to be no specification for how this should be done. CoCo Attesters SHA384 hash the nonce and parts of `TeePubKey` (Related: https://github.com/confidential-containers/guest-components/issues/366) and add it to the quote. Amber uses SHA256 and the lower 32 bytes of "report data" to verify so there's a gap. MAA seems to be doing something similar: "Azure Attestation will validate the provided “Quote”, and will then  ensure that the SHA256 hash of the provided Enclave Held Data is expressed in the first 32 bytes of the reportData field in the quote." With CoCo Attesters/Verifiers being in our control there's no problem but what if the CoCo attester generated quote needs to be verified by some other AS? 

3. Session timeout
there seems to be a conflict with how long the attestation results are valid: KBS could set some long session timeout but the AS token can theoretically have much shorter lifetime which is ignored by the resource API (`get_attest_claims_from_session()`) and is left for the policy to check? Similarly, if 1. makes sense and the nonce has an expiration time from AS, `auth` could set the session timeout to follow that initially?

4. Token verification in several places
JWT token verification happens in several places. Amber checks the public key based on the user input provided in the config where as the CoCo sample token verification uses the `jwt` claim in the token.

Xynnn007:
Thanks for bringing this up @mythi 

Let me try to share some ideas upon the topics.

> 1. Nonce from the attestation-service
With the current flow KBS creates the nonce and it has a lifetime of the session (which is configurable in kbs parameters). AS does not verify the nonce is used only once so it seems possible to cheat AS to continuously assign tokens when the TEE quote and KBS nonce are known. It may also be desirable to have the AS to generate the nonce with some lifetime so that it becomes easier to track if a nonce is valid for a one-time usage.

[...]

Issue #99, opened 2023-05-25
Document protocol errors


katexochen:
When an error occurs during communication with the KBC, error types are returned as described in the [protocol documentation](https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#error-information). Should we document these errors somewhere? Currently, the type is just a broken link, like

https://github.com/confidential-containers/kbs/errors/UnAuthenticatedCookie

fitzthum:
Anyone know if there is at least a way we can configure those links to point to the correct place? I'm not sure if they're actually intended as functioning hyperlinks or just a general reference.

PR #514, opened 2024-09-27
KBS | Refactoring the codebase / update config file format / bring in plugin mechanism

### LINKED ISSUE #486: Add "insecure" opt-in for Trustee AS/KBS token signing/verification
### Describe the bug

Triggered by #483.

It seems it's rather easy to create an insecure environment with the default configs where AS token signer material and/or KBS tokenverifier `‎trusted_certs_paths` are not set.

The proposal is to add an "insecure" opt-in similar to how the existing `insecure_*` switches work to at least in the tokenverifier to fail tokens if the opt-in is not enabled.

### How to reproduce

Run Trustee KBS with one of the default `kbs/config/*.toml` without `trusted_certs_paths`. With that, the following takes place
https://github.com/confidential-containers/trustee/blob/5c973bd4855b116a7e4dc5195334360c79a75ee6/kbs/src/token/coco.rs#L93-L96

### CoCo version information

trustee v0.9.0

None

```shell
N/A
```

Xynnn007:
This PR includes the following work
1. Refactoring the KBS codebase to have plugin mechanism in, including attestation/resource/policy engine
2. Update the config file format to make it more tidy and scalable.
3. Combine CoCo/ITA Token verification logic.
4. Add client plugin mechanism in. All client plugin calls will be filtered by attestation token.

I am sorry to have such a huge PR and did some disruptive changes. I originally want to add a simple plugin mechanism but soon found that some architectural problems and hidden bugs, thus fixed them by the way. At last resulting such many changed lines.

This PR still works fine with kbs-client on the guest-components side and did not change the protocol.

Needs some attestion from ITA folks @mythi @jodh-intel to see if the changes upon ITA config need any change.
Nebula folks @cclaudio about #451

cc @fitzthum @jialez0 @mkulke 

Fixes #486


kbs/src/policy_engine/opa/mod.rs
```diff
@@ -32,33 +32,33 @@ impl PolicyEngineInterface for Opa {
...
             .set_input_json(&input_claims)
-            .map_err(|_| ResourcePolicyError::InputError)?;
+            .map_err(|_| PolicyEngineError::InputError)?;
 
         let res = engine.eval_bool_query("data.policy.allow".to_string(), false)?;
         Ok(res)
     }
 
-    async fn set_policy(&mut self, policy: String) -> Result<(), ResourcePolicyError> {
+    async fn set_policy(&mut self, policy: String) -> Result<(), PolicyEngineError> {
```
mkulke, line 61: can we change `policy: &str` in this signature, so we don't have potentially clone on the caller site?

kbs/src/policy_engine/mod.rs
```diff
@@ -83,16 +58,30 @@ pub(crate) struct PolicyEngine(pub Arc<Mutex<dyn PolicyEngineInterface>>);
...
+
+    pub async fn set_policy(&self, request: &[u8]) -> Result<()> {
+        let request: Value = serde_json::from_slice(request)
+            .map_err(|_| PolicyEngineError::IllegalSetPolicyRequest)?;
+        let policy = request
+            .pointer("/policy")
+            .ok_or(PolicyEngineError::IllegalSetPolicyRequest)?
+            .as_str()
+            .ok_or(PolicyEngineError::IllegalSetPolicyRequest)?
+            .to_string();
```
mkulke, line 83: see comment on `PolicyEngine::set_policy()` signature, ideally we can just pass the str reference

kbs/src/attestation/backend.rs
```diff
@@ -0,0 +1,362 @@
...
+        Ok(response)
+    }
+
+    pub async fn attest(&self, attestation: &[u8], request: HttpRequest) -> Result<HttpResponse> {
+        self.__attest(attestation, request)
+            .await
+            .map_err(|e| Error::RcarAttestFailed { source: e })
+    }
+
+    async fn __attest(
```
mkulke, line 213: this convention of creating "private functions w/ underscore prefixes doesn't look like idiomatic rust. can we inline `__myfunc()` in `myfunc()` or if they need to be separated can we put it into a module?

kbs/src/admin/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+        let bearer = Authorization::<Bearer>::parse(request)?.into_scheme();
+
+        let token = bearer.token();
+
+        let _claims = self
+            .public_key
+            .as_ref()
+            .unwrap()
+            .verify_token::<NoCustomClaims>(token, Some(VerificationOptions::default()))
+            .map_err(|e| Error::JwtVerificaitonFailed { source: e })?;
```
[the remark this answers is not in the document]
Xynnn007: Nice idea. Let me fix this

kbs/src/api_server.rs
```diff
@@ -0,0 +1,318 @@
...
+                .await
+                .map_err(|e| Error::ServerLaunchFailed { source: e.into() })
+        } else {
+            http_server
+                .bind(&http_config.sockets[..])
+                .map_err(|e| Error::HTTPFailed { source: e.into() })?
+                .run()
+                .await
+                .map_err(|e| Error::ServerLaunchFailed { source: e.into() })
+        }
```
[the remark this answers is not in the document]
Xynnn007: Good idea. Let me delete the else block

kbs/src/policy_engine/opa/mod.rs
Xynnn007: Nice idea

kbs/src/attestation/backend.rs
Xynnn007: Do you mean to use `#[inline]` upon `pub async fn attest(...)`?

mkulke: no, just move the __attest code into the attest function (unless there is a reason to keep them apart, then we can move it it into a module maybe)

Xynnn007: Oh. The wrapping is for error conversion. If move all logic from `__attest` to `attest`, we need a lot of `map_err` from `anyhow::Error` to `Error`

Xynnn007: Yes. The `#[from]` could automatically convert *ALL* `anyhow:Error` to `Error` with `?`. In this case we want to do some classfication upon the different `anyhow::Error` due to `Auth` and `Attest` of RCAR

mkulke: ok, if we want to mix anyhow and concrete error types we have to resort to gymnastics like this function wrapping. If we bother about the concrete errors I'd suggest declaring them explicitly in an error enum, at least in new code. so instead of 

```rust
request.cookie(KBS_SESSION_ID).context("cookie not found")?;
```

```rust
#[error("cookie not found")]
CookieNotFound,
...
request.cookie(KBS_SESSION_ID).ok_or(Error::CookieNotFound)?;
```

Xynnn007: Right. My original aim is to avoid too many error types. Overall, they are only `RCAR attest failed`, because of an `anyhow::Error` and `RCAR auth failed`, because of an `anyhow::Error`. The concrete error inside could be all coverted by `anyhow`. I like the concise error type definition style myself

kbs/src/api_server.rs
```diff
@@ -0,0 +1,317 @@
...
+
+                let public_key = core.token_verifier.extract_tee_public_key(claims)?;
+                let jwe = jwe(public_key, resource).map_err(|e| Error::JweError { source: e })?;
+
+                let res = serde_json::to_string(&jwe)?;
+
+                Ok(HttpResponse::Ok()
+                    .content_type("application/json")
+                    .body(res))
+            } else if request.method() == Method::POST {
```
fitzthum, line 239: I wonder if we should introduce a different endpoint for an admin to set resources. In the future we might want to allow an attested client (i.e. the CDH) to set a resource. That would collide with this.

kbs/src/api_server.rs
```diff
@@ -0,0 +1,317 @@
...
+            let response = plugin
+                .handle(body, query.into(), sub_path.into(), request.method())
+                .await?;
+            Ok(response)
+        }
+    }
+}
+
+/// Admin APIs.
+pub(crate) async fn admin(
```
fitzthum, line 291: This is a bit confusing since we have admin APIs in the client section above as well

kbs/src/plugins/sample.rs
```diff
@@ -0,0 +1,41 @@
...
+#[async_trait::async_trait]
+impl ClientPlugin for Sample {
+    async fn handle(
+        &self,
+        _body: Vec<u8>,
+        _query: String,
+        _path: String,
+        _method: &Method,
+    ) -> Result<HttpResponse> {
+        let response = HttpResponse::Ok().body("sample plugin response");
```
fitzthum, line 38: Can you expand this example slightly to have different logic for `GET` and `POST` and to verify the admin credentials for the `POST` case.

kbs/src/api_server.rs
```diff
@@ -0,0 +1,317 @@
...
+            }
+        }
+        plugin_name => {
+            // Plugin calls needs to be authorized by the Token and policy
+            let token = core
+                .get_attestation_token(&request)
+                .await
+                .map_err(|_| Error::TokenNotFound)?;
+
+            let claims = core.token_verifier.verify(token).await?;
```
fitzthum, line 266: I like that we're decoupling the plugin from the `get_resource` API. It will allow us to implement more complex plugins including ones that take `POST` requests.

I think there is a problem with the logic here. We are passing the request to the handler function in the plugin, allowing it to specify different behavior for `GET` and `POST` requests, but in this function we are verifying the token and the policy for the request regardless of whether it is `GET` or `POST`. Basically this means that this plugin endpoint can't be used to implement requests from an admin.

Maybe we should distinguish between `GET` and `POST` here and do separate checks? Note that there is an open question about what `POST` requests should represent. See my main comment about this.

kbs/src/api_server.rs
```diff
@@ -0,0 +1,317 @@
...
+        "attest" if request.method() == Method::POST => core
+            .attestation_service
+            .attest(&body, request)
+            .await
+            .map_err(From::from),
+        #[cfg(feature = "as")]
+        "attestation-policy" if request.method() == Method::POST => {
+            core.admin_auth.validate_auth(&request)?;
+
+            core.attestation_service.set_policy(&body).await?;
```
fitzthum, line 187: This might be out of scope of this PR, but there is an open question regarding the security of this endpoint when we are not using a built-in AS. Basically the AS doesn't really evaluate where these requests are coming from.

kbs/src/api_server.rs
```diff
@@ -0,0 +1,317 @@
...
+
+            Ok(HttpResponse::Ok().finish())
+        }
+        "resource-policy" if request.method() == Method::POST => {
```
fitzthum, line 191: I think get can also implement `GET` logic here at some point to return the policy that is currently enabled.

fitzthum (commented):
Great work. Lots of positive changes here.

I mention one problem in a couple of my comments. Let me expand on it here.

For some of our endpoints, we have `GET` and `POST` functionality. The `GET` request is usually called by the client in the guest, but the `POST` request has too different cases. Sometimes the `POST` request is a privileged admin function that is called by the `kbs-client` running in a trusted environment. Sometimes it is called by the client in the guest.

In some cases we could actually support both. For resources and plugins, for instance, we might want to have an admin be able to configure the resource/plugin, but also have the guest be able to make a `POST` request. Unfortunately we only have one endpoint for these two things. Maybe we should decide on a convention for how these two different things can be implemented.

I think this will also help resolve some confusion around the so-called `client` and `admin` APIs, which are currently intermixed.

fitzthum:
btw I wonder if we should implement the resource backend as a plugin now

kbs/src/resource/backend.rs
```diff
@@ -0,0 +1,128 @@
...
+use serde::Deserialize;
+
+use super::local_fs;
+use super::{Error, Result};
+
+type RepositoryInstance = Arc<dyn Repository>;
+
+/// Interface of a `Repository`.
+#[async_trait::async_trait]
+pub trait Repository: Send + Sync {
```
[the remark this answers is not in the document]
Xynnn007: Nice idea. How about `StorageBackend`

kbs/src/resource/backend.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub resource_type: String,
+    pub resource_tag: String,
+}
+
+impl ResourceDesc {
+    pub fn is_valid(&self) -> bool {
+        if &self.repository_name == "."
+            || &self.repository_name == ".."
+            || &self.resource_type == "."
+            || &self.resource_type == ".."
```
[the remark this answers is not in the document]
Xynnn007: This is for security reason. For example a resource description like `../../secret-file`. Due to the logic of LocalFs it will perform a reading out of the localfs workdir, thus causing unexpected problem

kbs/src/api_server.rs
```diff
@@ -0,0 +1,317 @@
...
+            .bind(&http_config.sockets[..])
+            .map_err(|e| Error::HTTPFailed { source: e.into() })?
+            .run()
+            .await
+            .map_err(|e| Error::ServerLaunchFailed { source: e.into() })
+    }
+}
+
+/// Client APIs. /kbs/v0/XXX
+pub(crate) async fn client(
```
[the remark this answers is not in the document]
Xynnn007: Yes. You are right. I want to move the set_policy, set_resource APIs to `kbs/v0/admin/...`. Thus we could have a distinguished barrier from client api and admin api. Maybe we can relatively change the logic of kbs-client. This change will not influence anything on the guest-components side. wdyt?

kbs/src/api_server.rs
Xynnn007: Yes. I think we can follow https://github.com/confidential-containers/trustee/pull/514#discussion_r1779304844. The shortage is that old kbs-client will not work fine with new kbs.
interesting idea about an attested client to set resource. Maybe we could talk about it when we meet such case

kbs/src/api_server.rs
Xynnn007: I want to move all admin APIs into this section. Let converge onto https://github.com/confidential-containers/trustee/pull/514#discussion_r1779304844

kbs/src/api_server.rs
Xynnn007: The api makes less sense when we use non-built-AS and we usually directly interact with AS to set the policy. The existing of this API is much like a workaround for all-in-one-kbs, where the admin should care about the policy. I personally turn to use separate AS and KBS rather than all-in-one-kbs.

Some more thoughts. There is a nice architecture from classic [Kerberos](https://en.wikipedia.org/wiki/Kerberos_(protocol)). It decouples AS (like CoCoAS in our case, to verify the valiadity of evidence), TGS (like the authorization part of KBS in our case) and SS (like the resource storage + policy engine of KBS in our case). This would have a clear barrier of each components and helps maintain each of them without mixing.

kbs/src/api_server.rs
Xynnn007: Yes. I will add this to `admin` api

Xynnn007:
@fitzthum Thanks for the review and nice suggestion. About admin and client api, my idea is to move old admin APIs like POST `kbs/v0/<plugin>` to POST `kbs/v0/admin/<plugin>`. All admin APIs should be under `kbs/v0/admin`, no matter it is resource setting or policy setting. This would help us to have a clear border.

btw, about the policy filter for APIs. Now it still works fine with legacy ways, s.t. GET resources. I want to also do a re-design for the policy engine module when refactoring the code. But it seems like an independent work thus not included inside this PR. Mainly for
1. change the policy format to allow filter both client plugins (including resource) and admin plugins
2. new point from yours, to add support for GET/POST requests
3. To have relative static policy (defines roles and privileges) and data (defines how to be a role) and dynamic input (claims to be checked to be a role) to leverage policy engine like [the example](https://www.openpolicyagent.org/docs/latest/ssh-and-sudo-authorization/#5-elevate-a-users-rights-through-policy).

kbs/src/token/mod.rs
```diff
@@ -2,54 +2,86 @@
...
 
-#[derive(Deserialize, Default, Debug, Clone, EnumString)]
+#[derive(Deserialize, Default, Debug, Clone, EnumString, PartialEq)]
 pub enum AttestationTokenVerifierType {
     #[default]
     CoCo,
-    Jwk,
+
+    #[serde(rename = "ITA")]
+    Ita,
```
mythi, line 24: this shouldn't be a user facing name/option.

kbs/src/token/mod.rs
```diff
@@ -2,54 +2,86 @@
...
-    pub trusted_certs_paths: Vec<String>,
+    pub trusted_jwk_sets: Vec<String>,
+
+    /// Whether a JWK that directly comes from the JWT token is allowed to verify
+    /// the signature. This is insecure as it will not check the endorsement of
+    /// the JWK. If this option is set to false, the JWK will be looked up from
+    /// the key store configured during launching the KBS with kid field in the JWT,
+    /// or be checked against the configured trusted CA certs.
+    #[serde(default = "bool::default")]
+    pub insecure_key: bool,
```
mythi, line 49: I tried hard not to break users then `Jwk` type was added. I don't see these changes absolutely necessary either. how would users know when to use trusted_certs_paths` or `trusted_jwk_sets`? IMO, for a proper token verifier unification, we don't have a "type" at all and users would only configure their trust sources.

mythi (changes requested):
I checked 542f64c921a0ef3bb18fbf3bb1ac09f824c2259a only. Can that be split to a PR of its own? it's a bit too much content in one PR.

kbs/src/resource/backend.rs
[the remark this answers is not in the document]
fitzthum: oh yeah. that doesn't seem totally rigorous. again maybe we should move this to `try_from` rather than always trying to remember to call the validate code.

kbs/src/api_server.rs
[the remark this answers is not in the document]
fitzthum: That's fine with me although I'm not sure it's entirely RESTful

fitzthum:
@Xynnn007 

> All admin APIs should be under kbs/v0/admin, no matter it is resource setting or policy setting. This would help us to have a clear border.

I think this is good. Maybe we should also have a generic `admin` endpoint for each plugin. Like I said, I think it would be good to move our existing resource implementation to a plugin. This would prove that the plugin interface is robust. That might be too much work for this PR, though.

One other thing, do you think we should `plugin` as part of the route for plugins. As it is people can't implement a plugin named `attest` because it will collide with the `attest` endpoint. Maybe that's not too big of a deal. Having a specific plugin route would make the function serving it a bit smaller.

kbs/src/plugins/sample.rs
```diff
@@ -0,0 +1,41 @@
...
+use super::{plugin_manager::ClientPlugin, Result};
+
+#[derive(Deserialize, Clone, Debug, PartialEq)]
+pub struct SampleConfig {
+    pub item: String,
+}
+
+pub struct Sample {
+    _item: String,
+}
```
cclaudio, line 19: Can you update the `kbs-client` to take requests for the Sample plugin?

kbs/src/plugins/sample.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use actix_web::{http::Method, HttpResponse};
+use serde::Deserialize;
+
+use super::{plugin_manager::ClientPlugin, Result};
+
+#[derive(Deserialize, Clone, Debug, PartialEq)]
+pub struct SampleConfig {
+    pub item: String,
+}
```
cclaudio, line 15: https://github.com/confidential-containers/trustee/pull/451#discussion_r1710574306 , it would be nice if the SampleConfig path is read from the `kbs-config.toml`?

kbs/src/api_server.rs
```diff
@@ -0,0 +1,317 @@
...
+
+            let plugin = core
+                .plugin_manager
+                .get(plugin_name)
+                .ok_or(Error::PluginNotFound {
+                    plugin_name: plugin_name.to_string(),
+                })?;
+            let body = body.to_vec();
+            let response = plugin
+                .handle(body, query.into(), sub_path.into(), request.method())
```
cclaudio, line 248: I think the plugin may/should want to encrypt the response payload. If so, the `handle()` should be able to access the public_key. You could pass it via parameter.

kbs/src/resource/backend.rs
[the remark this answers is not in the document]
Xynnn007: Fixed.

kbs/src/api_server.rs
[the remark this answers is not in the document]
Xynnn007: Nice catch for the RESTful design and client/admin APIs. Maybe I went to the wrong design way at first. We might only need one plugin system, and adds `admin_auth` to the `handle()`. This means it is the API that decides whether admin auth needs to be checked.

In this way, we can avoid extra `kbs/v0/admin/<plugin-name>`, but only have one API `kbs/v0/<plugin-name>`.

Wdyt?

kbs/src/token/mod.rs
Xynnn007: What name do you prefer to use?

kbs/src/token/mod.rs
Xynnn007: This is hard to do. Like I said in https://github.com/confidential-containers/trustee/issues/519#issuecomment-2398902746. Currently KBS will receive JWTs from either CoCoAS or ITA. I have no good idea to detect which concrete JWT is runtimely. So the design here is to configure the token type in the config file.

In future, I wonder if we can change the authentication way here. That means
1. Attestation module would be responsible to interactive with backend attestation services and get a `yes/no` result.
2. KBS will provision a certificate that certs the tee public key and an id inside.

In this way, the storage and other plugins will only need to check the endorsement of the client certificate and the  id. Then the policy engine (OPA) would do much easier job -- control API access due to the client id, rather than complex different attestation claims.

kbs/src/plugins/sample.rs
Xynnn007: Yes we can! This would give a good implementation example

kbs/src/api_server.rs
Xynnn007: Nice idea.

Xynnn007:
> btw I wonder if we should implement the resource backend as a plugin now

Resource backend is somehow different. If we want to implement it as a backend plugin, we would use a config item like
```toml
...
[[plugins]]
name = "ResourceStorage"
type = "LocalFs"
dir_path = "/tmp/kbs-resource"
```

While what we are using now is
```toml
...
[repository]
type = "LocalFs"
dir_path = "/tmp/kbs-resource"
```

The difference
1. If we use plugin mechanism, if and only if the `ResourceStorage` is explicitly claimed in the configuration file, the API will be open. This hints that delete feature `resource` and include that may be a good choice.
2. If we use legacy way, whether the API is open is decided during compilation time.

Xynnn007:
@fitzthum I added one commit that tries to make `resource` a plugin. Also, combine `admin` and `client` plugins into one plugins system. The concrete plugin would decide whether it could be accessed upon client auth (attestation) or admin auth. Please take a look if it is good, then I can do some commit tidy work

kbs/src/api_server.rs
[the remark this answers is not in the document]
fitzthum: Yeah the problem with that is that you can't implement a plugin that has both `GET` / `POST` for the guest and `GET` / `POST` for an admin, which some people might want to do (although note that posting from the guest is a bit risky since the requests aren't encrypted). 

It looks like having an `/admin` route is somewhat common so maybe that's ok. Honestly I don't have a strong preference here.

fitzthum:
> I added one commit that tries to make resource a plugin. Also, combine admin and client plugins into one plugins system. The concrete plugin would decide whether it could be accessed upon client auth (attestation) or admin auth. Please take a look if it is good, then I can do some commit tidy work

I really like moving the resource stuff to a plugin. I think it makes sense to get rid of the resource feature. In the future there might be some complex plugins that are behind feature flags (in addition to being enabled at runtime), but imo we should always build the resource support.

Like I said, I could go either way on the admin routes. Maybe another reviewer has an opinion there. 

Otherwise I think this is good.

Xynnn007:
> Like I said, I could go either way on the admin routes. Maybe another reviewer has an opinion there.
> 
> Otherwise I think this is good.

Nice. I wonder if @cclaudio could give some suggestions upon the design.

Also, let me put another PR for token affairs as @mythi mentioned at first.

kbs/src/api_server.rs
cclaudio: `encrypted()` gives some flexibility to the plugin in terms of encrypting the payload. However, I can't think of a plugin that would want its payload to be sent back unencrypted. If we don't have a use case for that we could just assume that all plugins want its payload to be encrypted.

Xynnn007: There is one in my mind is to return the public cert whose confidentiality does not need to be protected. The request from guest is the evidence and the CSR of tee public key.

kbs/src/api_server.rs
```diff
@@ -0,0 +1,242 @@
...
+                    plugin_name: plugin_name.to_string(),
+                })?;
+
+            let body = body.to_vec();
+            if plugin
+                .validate_auth(&body, query, sub_path, request.method())
+                .await?
+            {
+                // Plugin calls needs to be authorized by the admin auth
+                core.admin_auth.validate_auth(&request)?;
```
cclaudio, line 200: I am wondering what would be the scope of the admin auth (or any other auth) for plugins. The way it is implemented now, the plugin could use the admin auth for anything.

I can think of some ways to have a better separation of duties in order to reduce damage if an auth gets compromised. E.g. we could have one auth only for plugins; or picking a crazy idea from the jar, we could have a wallet maintained by admin and its keys/auth could be selected for plugins via rego policies.

Anyways, the design proposed in the PR looks good to me. We can make adjustments later as needed.

Xynnn007: I am not quite sure about the design for now. Before this PR, we have `set resource`, `set policy` like APIs who should be accessed via so-called "user public key auth". Also, we have APIs like `get resource` for clients with Attestation tokens. This PR, I just bring them together and rename to "admin auth" who needs old "user public key auth" and those who need attestation token as auth -- which I think better to decribe the two different roles up to now -- admin and client.

Also, the new design makes it easier for developers to integrate any new APIs that needs to be filtered by "admin auth" without changing too many codes.

Your opintion considers this issue more comprehensively and inspiringly. I think in future we will meet some new demands that require more roles for KBS and we can do some more extension upon the design then.

cclaudio: > I am not quite sure about the design for now.

Do you want to expand on that? We could try to improve the parts you are not sure or find a way to move on only with the parts we are good. Honestly, like other reviewers, I also see many good changes in the PR. The auth thing I mentioned before is not introduced here, it is just something that may hit us in the future as we evolve trustee.

I am finishing to implement the nebula plugin on top of this PR, next week I should be able to post it. If the PR is not merged until then :smile:, the nebula plugin could be used to help the discussion of lingering questions.

kbs/src/plugins/error.rs
```diff
@@ -0,0 +1,32 @@
...
+    PluginManagerInitialization {
+        #[source]
+        source: anyhow::Error,
+    },
+
+    #[error("Error happens to plugin {plugin_name}")]
+    PluginHandlerError {
+        #[source]
+        source: anyhow::Error,
+        plugin_name: String,
```
cclaudio, line 32: Should this be &'static str the same way as InitializePluginFailed?

kbs/src/api_server.rs
Xynnn007: So if I understand correctly, the idea is that it is the admin's policy that defines which API needs admin auth, not the code of plugin, right?

kbs/src/plugins/error.rs
Xynnn007: It is hard to, as here the plugin_name is decided at runtime, while `InitializePluginFailed` could be decided during compilation time. `&'static` requires it to be a compilation time determined.

Xynnn007:
I just rebased on top of #524 , ptal

kbs/config/docker-compose/kbs-config.toml
```diff
@@ -1,9 +1,18 @@
+[http_server]
 sockets = ["0.0.0.0:8080"]
-auth_public_key = "/opt/confidential-containers/kbs/user-keys/public.pub"
 insecure_http = true
 
-[attestation_token_config]
+[attestation_token]
```
lmilleri, line 5: I think the changes to the configuration files should be reported in the various readme files as well (it seems they're not there yet).

kbs/src/admin/error.rs
```diff
@@ -0,0 +1,30 @@
...
+use log::error;
+use strum::AsRefStr;
+use thiserror::Error;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Error, AsRefStr, Debug)]
+pub enum Error {
+    #[error("Admin Token verification failed")]
+    JwtVerificaitonFailed {
```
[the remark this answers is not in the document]
Xynnn007: Thanks! Fixed

kbs/config/docker-compose/kbs-config.toml
Xynnn007: Yes, let me add the documents

lmilleri: Thank you @Xynnn007. Can you please update doc/examples for the following as well?
- [http-server]
- [attestation_service]
- [admin]
- [plugins]
- [repository]
- [policy-engine]
I might have not included them all.

Or are you thinking to do it later? (e.g. later on in the review/other PR)

kbs/src/error.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use std::fmt::Write;
+
+use actix_web::{body::BoxBody, HttpResponse, ResponseError};
+use kbs_types::ErrorInformation;
+use log::error;
+use strum::AsRefStr;
+use thiserror::Error;
+
+const ERROR_TYPE_PREFIX: &str = "https://github.com/confidential-containers/kbs/errors";
```
fitzthum, line 15: This isn't changed in this PR, so we can come back to it, but I think it's somewhat confusing to include this URL in the errors because it doesn't actually point to anything.

kbs/src/plugins/resource/backend.rs
```diff
@@ -0,0 +1,120 @@
...
+            },
+            _ => return Err(Error::ParseResourceDescription),
+        };
+
+        if path_parts[0] == "."
+            || path_parts[0] == ".."
+            || path_parts[1] == "."
+            || path_parts[1] == ".."
+        {
+            return Err(Error::MalwaredResourceDescription);
```
fitzthum, line 55: I would just call this `IllegalResourceDesciprtion`. This could be the result of user error. It isn't necessarily malware.

fitzthum (commented):
A few more comments. I think this is pretty close.

kbs/config/docker-compose/kbs-config.toml
Xynnn007: Yes. I should do this. Thanks for the great suggestion.

kbs/src/policy_engine/error.rs
```diff
@@ -0,0 +1,36 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use log::error;
+use strum::AsRefStr;
+use thiserror::Error;
+
+pub type Result<T> = std::result::Result<T, PolicyEngineError>;
+
+#[derive(Error, AsRefStr, Debug)]
+pub enum PolicyEngineError {
```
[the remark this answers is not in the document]
Xynnn007: Yes, when we meet an error information from a built-in-as KBS would make sense.

kbs/src/policy_engine/mod.rs
```diff
@@ -2,63 +2,41 @@
...
 /// Resource policy engine interface
+///
+/// TODO: Use a better authentication and authorization policy
 #[async_trait]
 pub(crate) trait PolicyEngineInterface: Send + Sync {
-    /// Determine whether there is access to a specific path resource based on the input claims.
+    /// Determine whether there is access to a specific path based on the input claims.
     /// Input parameters:
-    /// resource_path: Required to be a string in three segment path format:<TOP>/<MIDDLE>/<TAIL>, for example: "my'repo/License/key".
+    /// accessed_path: Required to be a string in segments path format:<FIRST>/.../<END>, for example: "my'repo/License/key".
```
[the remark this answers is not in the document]
Xynnn007: Yes let me fix this. The `'` has lived for more than a year here XD

kbs/src/error.rs
Xynnn007: Yes. We have an issue for this for a long time https://github.com/confidential-containers/trustee/issues/99

kbs/src/plugins/resource/backend.rs
Xynnn007: Sounds reasonable. Let me fix this.

kbs/src/error.rs
fitzthum: yeah as you can see by my comment I hadn't realized how this was implemented. I guess it's pretty easy to adjust I'm just not sure where we should actually point the link or if we want one at all.

kbs/src/api_server.rs
```diff
@@ -0,0 +1,242 @@
...
+
+            Ok(HttpResponse::Ok().finish())
+        }
+        "resource-policy" if request.method() == Method::POST => {
+            core.admin_auth.validate_auth(&request)?;
+            core.policy_engine.set_policy(&body).await?;
+
+            Ok(HttpResponse::Ok().finish())
+        }
+        "resource-policy" if request.method() == Method::GET => {
```
fitzthum, line 180: I guess we might want to change the name of this endpoint if we're no longer calling it the resource policy. Gotta make sure that isn't going to break things tho. Probably if we change the kbs-client accordingly it will be fine. We could save this for another PR tho.

kbs/src/plugins/resource/error.rs
```diff
@@ -0,0 +1,46 @@
...
+    #[error("Illegal HTTP method. Only supports `GET` and `POST`")]
+    IllegalHttpMethod,
+
+    #[error("Local FS backend error")]
+    LocalFs {
+        #[source]
+        source: anyhow::Error,
+    },
+
+    #[error("Malwared Resource Description")]
```
fitzthum, line 35: Let's change the message here as well.

fitzthum (approved):
Two small comments but I think this is ready.

kbs/src/policy_engine/mod.rs
```diff
@@ -83,16 +61,32 @@ pub(crate) struct PolicyEngine(pub Arc<Mutex<dyn PolicyEngineInterface>>);
...
+    }
+
+    pub async fn set_policy(&self, request: &[u8]) -> Result<()> {
+        let request: Value = serde_json::from_slice(request)
+            .map_err(|_| KbsPolicyEngineError::IllegalSetPolicyRequest)?;
+        let policy = request
+            .pointer("/policy")
+            .ok_or(KbsPolicyEngineError::IllegalSetPolicyRequest)?
+            .as_str()
+            .ok_or(KbsPolicyEngineError::IllegalSetPolicyRequest)?;
```
ChengyuZhu6, line 85: Could we provide more detailed error messages? For example, "failed to identify 'policy' in policy request" for the first error, and ‘failed to return the associated str’ for the second error?

kbs/src/plugins/sample.rs
ChengyuZhu6, line 1: Are we planning to add more customized plugins (using the sample plugin as a template) in this directory? If so, perhaps we could create a subdirectory to store all plugin implementations, making it clearer.

kbs/test/config/kbs.toml
```diff
@@ -1,31 +1,37 @@
...
+[policy_engine]
 policy_path = "./work/kbs-policy.rego"
+
+[admin]
+auth_public_key = "./work/kbs.pem"
+
+[[plugins]]
+name = "resource"
+type = "LocalFs"
+dir_path = "./work/repository"
```
ChengyuZhu6, line 37: I recommend adding a blank line to the end of every toml file that lacks one in the repo .

ChengyuZhu6 (commented):
Nice work @Xynnn007 ! Just have some minor comments.

kbs/src/error.rs
Xynnn007: Probably we could add a link to somewhere in confidentialcontainers.org

kbs/src/api_server.rs
Xynnn007: Let me leave a TODO inside code. There is also an `attestation-policy` API, so it would be good to have an issue for the rename things? cc @jialez0

kbs/src/policy_engine/mod.rs
Xynnn007: Yes, let me add this

kbs/src/plugins/sample.rs
Xynnn007: nice idea. Let me put them into `implementations` subdir

kbs/test/config/kbs.toml
Xynnn007: Let me add them

Xynnn007:
Hi @fitzthum @ChengyuZhu6 I have done some changes due to your suggestions but not pushed as there might be still some changes after @cclaudio 's https://github.com/confidential-containers/trustee/pull/514#discussion_r1797181068 is resolved. Let me do a push after that.

kbs/src/api_server.rs
cclaudio: It seems a good idea to make use of policies to control which APIs (or plugin services) need admin auth. The policy could be defined by the plugin or the admin, there are tradeoffs in both cases, but I like the current design.

We don't know yet, but in the future we may have a plugin that wants to provide multiple services. IIUC, the current design allows the plugin to determine which of the services require admin_auth. Additionally, it is up to the plugin to make that decision based a policy e.g. loaded from a config file. Each plugin has its own requirements.

`            if plugin
                .validate_auth(&body, query, sub_path, request.method())
                .await?
            {
`

The admin does not need to know much about the business of the plugin (which plugin service should require admin_auth). That seems good to me. Later, if we think that's not enough, we can extend the admin to override some of the plugin's decision (not with the same granularity).

Xynnn007: Ok. So my understanding is to keep the logic as-is. Thanks for the explaination and if you have any new suggestion please raise them up

Xynnn007:
Pushed the changes.

kbs/src/plugins/implementations/resource/backend.rs
```diff
@@ -0,0 +1,120 @@
...
+pub struct ResourceDesc {
+    pub repository_name: String,
+    pub resource_type: String,
+    pub resource_tag: String,
+}
+
+impl TryFrom<&str> for ResourceDesc {
+    type Error = Error;
+
+    fn try_from(value: &str) -> Result<Self> {
```
mkulke, line 34: this is a bit confusing. maybe use a regex to specify a legal path instead?

e.g.certain path_parts can have `.` and `..` in their name (the tag), but if we have 2 slashes and a "default" repository then resource tag isn't allowed to have those characters, I think

I would add a couple of unit test cases, so we have specified the desired behavior

mkulke (commented):
some nits

kbs/src/api_server.rs
```diff
@@ -0,0 +1,251 @@
...
+                .await
+                .map_err(|e| Error::ServerLaunchFailed { source: e.into() });
+        }
+
+        http_server
+            .bind(&http_config.sockets[..])
+            .map_err(|e| Error::HTTPFailed { source: e.into() })?
+            .run()
+            .await
+            .map_err(|e| Error::ServerLaunchFailed { source: e.into() })
```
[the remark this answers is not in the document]
Xynnn007: Nice idea. Merged them

kbs/src/plugins/implementations/resource/backend.rs
Xynnn007: Cool. Regex is a nice idea. 

> e.g.certain path_parts can have . and .. in their name (the tag), but if we have 2 slashes and a "default" repository then resource tag isn't allowed to have those characters, I think

I use a regex to force the `repo`s, `type`s and `tag`s can only be `[A-Za-z0-9_-]+`. I think the restrict would be fine to avoid unknown potential risks?

ChengyuZhu6 (approved):
LGTM Thanks @Xynnn007

kbs/config/docker-compose/kbs-config.toml
Xynnn007: please check https://github.com/Xynnn007/kbs/blob/refactor-kbs/kbs/docs/config.md which is the view of current PR

fitzthum (approved):
LGTM

kbs/src/api_server.rs
cclaudio: Yep, looks good to me. Thanks

fitzthum:
ok are we ready to merge this?

Xynnn007:
> ok are we ready to merge this?

hi are you ok with this? @mkulke  @mythi

mkulke (approved):
lgtm

kbs/src/attestation/backend.rs
```diff
@@ -0,0 +1,362 @@
...
+#[derive(Clone)]
+pub struct AttestationService {
+    /// Attestation Module
+    inner: Arc<dyn Attest>,
+
+    /// A concurrent safe map to keep status of RCAR status
+    session_map: Arc<SessionMap>,
+
+    /// Max timeout between `auth` and `attest` request of a client
+    timeout: i64,
```
mythi, line 105: in #162 I raised the question about this `timeout` and IIRC @fitzthum mentioned it somewhere too. I'm asking the question again since we are making big config changes: do we need to have this configurable for KBS or should it just follow the token `exp`.

With the current CDH setup, is it possible, for example to set 1 minute here but CDH can continue to pull secrets as long as the token is valid (5min by default)?

mythi (commented):
Looks great work to me and I could not spot anything that would block merging. I made some suggestions about "Intel TA" wording usage.

My biggest open is about keeping the KBS session `timeout` configurable but that logic isn't changing in this PR so it's not a blocker either, I guess.

kbs/config/kbs-config-intel-trust-authority.toml
```diff
@@ -1,10 +1,15 @@
...
 insecure_http = true
+
+[admin]
 insecure_api = true
 
-[attestation_token_config]
-trusted_certs_paths = ["https://portal.trustauthority.intel.com"]
+[attestation_token]
+trusted_jwk_sets = ["https://portal.trustauthority.intel.com"]
+insecure_key = true
```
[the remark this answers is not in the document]
Xynnn007: Yes, deleted this line

kbs/docs/config.md
```diff
@@ -60,190 +58,259 @@ For Attestation Services like Intel TA, there will only be a `kid` field inside
...
+| Property                   | Type                        | Description                                          | Default |
+|----------------------------|-----------------------------|-----------------------------------------------------|----------|
+| `timeout`            | Integer                      | The maximum time (in minutes) between RCAR handshake's `auth` and `attest` requests             |  5       |
+| `as_addr`                 | String                      | The URL of the remote CoCoAS |  `http://127.0.0.1:50004`       |
+| `pool_size`   | Integer         | The connections between KBS and CoCoAS are maintained in a conenction pool. This property determines the max size of the pool                      | `100`             |
 
-| Property  | Type   | Description                  | Required | Default                  |
-|-----------|--------|------------------------------|----------|--------------------------|
-| `as_addr` | String | Attestation service address. | No       | `http://127.0.0.1:50004` |
+#### Intel TA
```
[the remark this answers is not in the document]
Xynnn007: Fixed. It is `&reg;`

kbs/test_data/configs/intel-ta-1.toml
```diff
@@ -0,0 +1,31 @@
+[attestation_token]
+trusted_jwk_sets = ["/etc/ca", "/etc/ca2"]
+
+[attestation_service]
+type = "intel_ta"
+base_url = "example.io"
+api_key = "this-is-a-key"
+certs_file = "/etc/ita-cert.pem"
```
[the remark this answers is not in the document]
Xynnn007: fixed

kbs/docs/config.md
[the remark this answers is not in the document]
mythi: oh, yes sorry!

kbs/src/attestation/backend.rs
Xynnn007: This timeout as noted here is not the expiration of the token. It is just the maximum delay between an `auth` request and an `attest` request. That means if the `attest` comes too late after `auth`, the RCAR handshake will fail. You raised a good idea about the inner `timeout` for session. I think we should distinguish the two different `timeout`s. Btw, do you think we need to add another option to specify the timeout of the session time, or use a hardcode one? cc @fitzthum 

> With the current CDH setup, is it possible, for example to set 1 minute here but CDH can continue to pull secrets as long as the token is valid (5min by default)?

Actaully it can for now : |  The CDH uses the token (via HTTP Auth header of request) to pull secrets rather than reuse the session.

Xynnn007: This is a little tricky. Actually the both timeout are the session timeout. What it will influence is when a secret access comes, the token will firstly be looked up from the session. If the session is expired, the http auth header will be checked to get the token.

As for whether the resource retrievement is allowed, it is the checking against the token that matters. Some different cases
1. timeout > token expiration time: the token can be got. When checking against the token, kbs will realize the token is expired then failed to pull secret.
2. timeout < token expiration time: when trying to get token from session, the kbs will realize the session is expired. Then try to get token from http auth header. If a token is found, the secret pulling will succeed. If no tokens are found, the secret pulling will fail.

Xynnn007: Yup. I added a commit for Token verifier to check the expiration time of the token. Hopefully things can be clear and we can discuss more back to #162

mythi: that is not necessary since jsonwebtoken handles that [automatically when using the default `Validation`](https://github.com/confidential-containers/trustee/blob/96ddb44308a7351baa83f6f755350babe285578c/kbs/src/token/jwk.rs#L213). It's also possible to make that explicitly.

Xynnn007: Oh yes. nice. Let me delete the last commit

mythi (approved):
LGTM, thanks!

Issue #353, opened 2024-03-21
Adopt EAR in Attestation Service


fitzthum:
Following @thomas-fossati's excellent presentation about EAR in the community meeting, I think we should move towards replacing/reformating our attestation token to use EAR. There are a few pieces that we'll need to think about.

Currently the AS returns a JWT with these claims inside of it. Code for token issueing is [here](https://github.com/confidential-containers/trustee/blob/main/attestation-service/attestation-service/src/token/simple.rs#L91).
```
let token_claims = json!({
    "tee": to_variant_name(&tee)?,
     "evaluation-reports": policies,
      "tcb-status": flattened_claims,
      "reference-data": reference_data_map,
      "customized_claims": {
          "init_data": init_data_claims,
           "runtime_data": runtime_data_claims,
      },
});
```

Part of the question is how we can map these claims onto EAR. I don't have a thorough knowledge of the EAR format yet, so I don't have a lot to say about this. I do wonder if we should drop the `reference-data` field. I am also not sure if there is something like the `TEE` field. I guess EAR is supposed to be platform independent, but it is reasonable for a KBS to have policies based on this info (i.e. don't release secrets to a guest with the sample attester).

There are two deeper questions that I am concerned about. First, how do we produce the EAR trustworthiness claims. Currently evidence validation in the AS is based on OPA. Each verifier plugin extracts a claims map (also stored in the `tcb-status` field of attestation token) which is fed into the OPA engine. The output of the policy evaluation is stored in the `policies` field. Note that we support evaluation of multiple different policies. I'm not sure if we should continue with that or how to work it into the EAR format.

The bigger issue is that we currently have very loose requirements for what the AS validation produces. The OPA policies can essentially return anything. It is up to the client to synchronize their AS policies with their resource policies in the KBS. This is quite different from the EAR approach where the format of the attestation results are more strongly structured. I'm not sure how we should account for this. Maybe we can provide OPA policies that transform claims maps into trustworthiness claims. Maybe we should get rid of the OPA engine in the AS entirely. This is related to some of the discussion [here](https://github.com/confidential-containers/trustee/issues/350).
 
[...]

PR #516, opened 2024-09-27
EAR Support

fitzthum:
This is the very beginning of this PR. I still need to fix some issues with the underlying crate, add support for verifying tokens in the KBS, and fixup all the tests, examples, and default values.

See commit messages for more info.

Fixes: #353 

Progress so far: 

- [x] Miscellaneous Fixups
- [x] Implement EAR token generation in the AS
- [x] Implement EAR verification in KBS
- [x] Update docker compose 
- [x] Update k8s (hard to test fully without building the images, but should be smooth when we do)
- [x] Fix integration tests
- [x] FIxup RVPS
- [x] Fixup policy_ids
- [x] Improve default policy to validate TCB of all platforms (well, for a couple of platforms. the rest will be added by others hopefully)
- [x] Update docs and examples


fitzthum:
This is ready for review.

We still have some issues to resolve in the underlying crate before we can merge.

* https://github.com/veraison/rust-ear/issues/19 - without this we can't create extensions, which we aren't using anyway yet. we could probably go ahead without a fix, but we have to pin the crate to a hash right before they added extensions or it won't compile (and we will need more commits on top of this for the other issues)
* https://github.com/veraison/rust-ear/issues/20 - without this the tokens never expire. interestingly guest-components is fine with this. the code works fine, but it's a security problem.
* https://github.com/veraison/rust-ear/issues/21 - without this JSON serializing some values into EAR (null ones) causes everything to explode. this may even be triggered by the TDX or SGX claims, which would be bad.

attestation-service/src/lib.rs
```diff
@@ -295,7 +258,8 @@ fn parse_data(
 ) -> Result<(Option<Vec<u8>>, Value)> {
     match data {
         Some(value) => match value {
-            Data::Raw(raw) => Ok((Some(raw), Value::Null)),
+            // Ear RawValue does not support NULL, so use an empty string
+            Data::Raw(raw) => Ok((Some(raw), Value::String("".to_string()))),
```
Xynnn007, line 262: Would this cause ambiguity. As the user might not give input RuntimeData/InitData to AS (null), or the data is an empty string ( `""`). In the former case, verifier will not check the binding of data against the value inside evidence. While for the latter case, the verfier will calculate the digest of `""` and compare it with the claim inside tee evidence.

When we see a `""` value in ear, how could we distinguish it as a non-given runtime data, or a `""` runtime data?

Xynnn007 (commented):
Thanks Tobin. Nice work about code, tests and documents. I am excited to see that the design that leaves the trusted anchor calculation to users.

I have not dived into the code too deeply, but only for a quick catch for the architecture design. 

This PR abondons former AS token format. Maybe we could leave the token plugins still. Probably the difficulty is where should OPA lie. My suggestion is to make OPA a common lib for both legacy AS token and EAR.

In this way, the token type could be configured in the launch toml of AS. Such as
```toml
[token]
type = "ear"
# other configs for ear
```
or
```toml
type = "simple"
# other configs for simple
```

This can be implemented by
```rust
#[serde(tag = "type")]
pub enum AttestationTokenConfig {
    Ear(EarConfig),

    Simple(SimpleConfig),
}

pub type AttestationTokenBroker = Arc<dyn AttestationTokenBrokerTrait + Send + Sync>;

impl TryFrom<AttestationTokenConfig> for AttestationTokenBroker {
    type Error = Error;

    fn try_from(value: AttestationTokenConfig) -> Result<Self> {
        match value {
            AttestationTokenConfig::Ear(config) => {
                let ear_broker = Arc::new(EarBroker::new(config)?);
                Ok(ear_broker as _)
            }
            AttestationTokenConfig::Simple(config) => {
                let simple_broker = Arc::new(SimpleBroker::new(config)?);
                Ok(simple_broker as _)
            }
        }
    }
}
```
This would give enough flexibility to downstream users to define their different token format, although they might not be standard as EAR. But for CoCo, we can support EAR by default.

jialez0:
> This PR abondons former AS token format. Maybe we could leave the token plugins still. Probably the difficulty is where should OPA lie. My suggestion is to make OPA a common lib for both legacy AS token and EAR.
> 
> In this way, the token type could be configured in the launch toml of AS. Such as
> 
> ```toml
> [token]
> type = "ear"
> # other configs for ear
> ```
> 
> or
> 
> ```toml
> type = "simple"
> # other configs for simple
> ```
> 
> This can be implemented by
> 
> ```rust
> #[serde(tag = "type")]
> pub enum AttestationTokenConfig {
>     Ear(EarConfig),
> 
>     Simple(SimpleConfig),
> }
> 
> pub type AttestationTokenBroker = Arc<dyn AttestationTokenBrokerTrait + Send + Sync>;
> 
> impl TryFrom<AttestationTokenConfig> for AttestationTokenBroker {
>     type Error = Error;
> 
>     fn try_from(value: AttestationTokenConfig) -> Result<Self> {
>         match value {
>             AttestationTokenConfig::Ear(config) => {
>                 let ear_broker = Arc::new(EarBroker::new(config)?);
>                 Ok(ear_broker as _)
>             }
>             AttestationTokenConfig::Simple(config) => {
>                 let simple_broker = Arc::new(SimpleBroker::new(config)?);
>                 Ok(simple_broker as _)
>             }
>         }
>     }
> }
> ```
> 
> This would give enough flexibility to downstream users to define their different token format, although they might not be standard as EAR. But for CoCo, we can support EAR by default.

I concur that the format for generating tokens should be modular. We can default to supporting EAR, but simultaneously, retain the previous simple AS token mode as a plugin option. This approach will provide ample room for customization to the users of this project.

Cargo.toml
```diff
@@ -26,6 +26,8 @@ cfg-if = "1.0.0"
 chrono = "0.4.19"
 clap = { version = "4", features = ["derive"] }
 config = "0.13.3"
+#ear = "0.2.0"
```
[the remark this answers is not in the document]
fitzthum: This is because the latest commits to rust-ear broke the crate (Appraisal isn't `send` anymore`). Before we merge this, the crate will need to be fixed and i will fix the version.

attestation-service/src/lib.rs
fitzthum: This is a bug in the rust-ear crate imo that will hopefully be fixed. I made an issue about it there. https://github.com/veraison/rust-ear/issues/21

fitzthum:
@Xynnn007 @jialez0 You are both right that this PR makes our token provisioning less generic, in fact it's no longer generic at all. There are some reasons for this.

First, with EAR tokens the policy is fundamentally tied to the token. Our CoCo token supports just about any policy claims, but with EAR we need to generate an Appraisal. This means that 1) the policy is specific to the type of token we use. 2) A bunch of the code surrounding the policy is specific to the type of token we are using.

The first thing isn't a huge problem. It's not a great user experience to require totally different policies for different tokens, but at least that is configurable at runtime. The second thing is a bit more of a problem. There is code in both the policy engine and `lib.rs` that is specific to the type of policy that we are using. For instance, the policy engine parses the claims differently, and `lib.rs` composes the token token differently. Supporting generic tokens would require changing some abstractions and adding complexity.

One option would be to move most of the logic into the token generation code. The policy engine would always process all the claims and return the full results. The results would be provided as-is to the token broker. The token broker would also get the TCB claims and any other pieces of the token. I don't really like broadening the scope of the token broker like this. If we want to keep the token broker abstraction the same, we would probably have to add a bunch of conditional logic to the policy engine and `lib.rs`, which also doesn't seem like a great idea.

It's possible to keep both tokens, but there would be costs. Keep in mind that we would need to create more tests, more examples, extend the docs, etc. Besides the costs, I actually think there are advantages to supporting fewer things. So far we have tried to make Trustee as generic as possible. In general I think this is good, but currently I think we are in danger of giving users too many ways to configure things without showing them any particular way that actually works end-to-end. I think part of the reason why we have some gaps with policies, the RVPS, etc is that we haven't been prescriptive enough. I think EAR tokens are a good fit because they are supposed to be generic, but they are more opinionated. 

Anyway, this change wasn't an accident, but I understand that it is kind of scary. I am open to trying to support both, but my question for you is what is the value of keeping the CoCo token?

fitzthum:
Another way we could potentially compromise is by keeping the CoCo token but populating the policy results with an EAR-like Appraisal. I'm not sure if this really makes sense tho.

Xynnn007:
Hi @fitzthum 
> Anyway, this change wasn't an accident, but I understand that it is kind of scary. I am open to trying to support both, but my question for you is what is the value of keeping the CoCo token?

Actually we are trying to define some own token format in downstream pre-productions benefits from the original plugin design.

Let's go back to your worries. The core issues are 1. token generation code and 2. user interface.

For 1,
I tried to look into the code about the design of current policy module. Some ideas:
- Change the API of policy engine. That takes policy id, data and input as parameters, just as OPA defines. We could have only regorus as backend and abondon the extensibility for policy engine (I have a strong hunch we need to do this because so far there is no indication that we will have a strategic replacement other than rego)
- Change the `AttestationTokenBroker`'s `issue` API to receive parameter `reference_data_map`, `tcb_claims` and `policy_id`. Then the module itself decides how to interact with regorus.
- To help KBS recognize the concrete token type, we could add another return value of AS's [evaluate function](https://github.com/confidential-containers/trustee/pull/516/files#diff-41b55ad1748c6bf85c3d9db9000f0703eb76c4f3afd4fd236bbb70ebdaf33494R170). 
For 2,
The [current config of AS](https://github.com/confidential-containers/trustee/blob/main/kbs/config/as-config.json) looks like
```json
{
    "work_dir": "/opt/confidential-containers/attestation-service",
    "policy_engine": "opa",
    "rvps_config": {
	    "remote_addr":"http://rvps:50003"
    },
    "attestation_token_broker": "Simple",
    "attestation_token_config": {
        "duration_min": 5
    }
}
```

After some change, it would be 
```json
{
    "work_dir": "/opt/confidential-containers/attestation-service",
    "policy_engine": "opa",
    "rvps_config": {
	    "remote_addr":"http://rvps:50003"
    },
    "attestation_token_broker": {
        "type": "XXX",
        ...// Other options
    }
    "attestation_token_config": {
        "duration_min": 5
    }
}
```

The default one we could set as EAR, together with all other e2e examples. Although we provide "too" many ways to run trustee, I believe most users will directly use the default configuration of our e2e test (probably EAR or Simple token), which needs to be ensured to run normally by us. Some advanced users might want do extension, then the value of extensibility values.

fitzthum:
> Actually we are trying to define some own token format in downstream pre-productions benefits from the original plugin design.

Ok. This is a valid consideration. Ideally we could converge on one token, but I'm not sure what your requirements are and they might not be public. Keep in mind that the EAR token is in some ways very similar to our CoCo token. For instance, they are both JWT. The EAR token stores a public key in a similar way but with a different path. Consumers can somewhat ignore parsing the AR4SI Trust Claims by just looking at the status field of the Appraisal.

> We could have only regorus as backend and abondon the extensibility for policy engine 

I think we should try to keep the policy engine modular. My opinion of rego/opa is declining and I would love to provide another option (as long as we don't invent it ourselves).


Btw, this PR also switches from evaluating multiple policies with the AS, to only taking one policy id. If we keep support for the CoCo token, can it just take one policy or does it need to do multiple?

Xynnn007:
> Btw, this PR also switches from evaluating multiple policies with the AS, to only taking one policy id. If we keep support for the CoCo token, can it just take one policy or does it need to do multiple?

One policy is ok for now.

fitzthum:
ok @Xynnn007 I have added two more commits which demonstrate how we could support both EAR and the simple tokens. IMO these make the code significantly worse. The commits are a bit rough. The tests won't pass, but it should give a pretty good idea of what is required. I will clean it up if we decide to go this way. PTAL and see what you think. Also consider if there is any way we can get rid of the simple token requirement.

Xynnn007:
> ok @Xynnn007 I have added two more commits which demonstrate how we could support both EAR and the simple tokens. IMO these make the code significantly worse. The commits are a bit rough. The tests won't pass, but it should give a pretty good idea of what is required. I will clean it up if we decide to go this way. PTAL and see what you think. Also consider if there is any way we can get rid of the simple token requirement.

Yup. I tried to support multiple token formats on top of current implementation. https://github.com/fitzthum/kbs/pull/1

There are also some trade-offs and decisions of design.

Xynnn007:
> ok @Xynnn007 I have added two more commits which demonstrate how we could support both EAR and the simple tokens. IMO these make the code significantly worse. The commits are a bit rough. The tests won't pass, but it should give a pretty good idea of what is required. I will clean it up if we decide to go this way. PTAL and see what you think. Also consider if there is any way we can get rid of the simple token requirement.

Yup. Inspired by your design, I did some refactoring to support multiple token formats. The problems rose when designing this so some trade-offs are also included. Please take a look upon https://github.com/fitzthum/kbs/pull/1. It might not be the perfect version

fitzthum:
Ok, I've taken the changes that @Xynnn007 mentioned above and squashed them together with my last two commits. I have also rebased. A couple of changes were needed to catch the recent PR changing the token verifier in the kbs. Tomorrow I will fixup the tests, but at least it builds now and the unit tests pass locally. I will also fixup the unit test for the opa engine that we currently have disabled.

fitzthum:
I will rebase this again next week. All of the underlying issues with the EAR crate have been fixed and updated here. PTAL @Xynnn007 @mkulke @mythi et al

I'd like to either merge this soonish or abandon it and continue working on fixing some issues with RVPS and other stuff that I noticed while working on this.

attestation-service/Cargo.toml
```diff
@@ -41,9 +41,11 @@ async-trait.workspace = true
 base64.workspace = true
 cfg-if.workspace = true
 clap = { workspace = true, optional = true }
+ear.workspace = true
 env_logger = { workspace = true, optional = true }
 futures = "0.3.17"
 hex.workspace = true
+jsonwebtoken = "9.3.0"
```
Xynnn007, line 48: `jsonwebtoken` is imported in workspace Cargo.toml. We can directly `jsonwebtoken.workspace = true` here

attestation-service/Cargo.toml
```diff
@@ -59,6 +61,7 @@ serde_variant = "0.1.2"
 sha2.workspace = true
 shadow-rs.workspace = true
 strum.workspace = true
+tempfile = "3.13.0"
```
Xynnn007, line 64: Also imported

```suggestion
tempfile.workspace = true
```

attestation-service/config.json
```diff
@@ -1,12 +1,11 @@
...
     "work_dir": "/var/lib/attestation-service/",
-    "policy_engine": "opa",
     "rvps_config": {
         "store_type": "LocalFs",
         "remote_addr": ""
     },
-    "attestation_token_broker": "Simple",
-    "attestation_token_config": {
+    "attestation_token_broker": {
+        "type": "Simple",
```
Xynnn007, line 8: Do we need to mark this as `Ear` to align with what we changed in `README.md`? s.t.
 > If the policy is not updated, the AS will use the [default policy](src/token/ear_default_policy.rego).

attestation-service/docs/policy.md
```diff
@@ -1,6 +1,29 @@
 # Attestation Policy (Developing)
 
 CoCo AS provides a flexible policy support based on Rego to facilitate the customized verification rules.
+The policies are used to general an EAR appraisal, which is part of the EAR token returned by the
```
Xynnn007, line 4: We might need to claim that different token type needs different policy formattings. The added part is for Ear

attestation-service/src/bin/restful/mod.rs
```diff
@@ -139,13 +139,6 @@ pub async fn attestation(
         }
     };
 
-    let policy_ids = if request.policy_ids.is_empty() {
-        info!("no policy specified, use `default`");
-        vec!["default".into()]
-    } else {
-        request.policy_ids
-    };
-
```
Xynnn007, line 148: Noticed that https://github.com/confidential-containers/trustee/pull/516/files#diff-a4f0031656042cef2f9c6da58178868fb7f80de2d2f094d33cc38196135e2d1aR199 we will raise an error if no policy is selected. Do we need to keep this part of code?

attestation-service/src/lib.rs
```diff
@@ -165,9 +152,8 @@ impl AttestationService {
     ///   will not be performed.
     /// - `hash_algorithm`: The hash algorithm that is used to calculate the digest of `runtime_data` and
     ///   `init_data`.
-    /// - `policy_ids`: The policy ids that used to check this evidence. Any check fails against a policy will
-    ///   not cause this function to return error. The result check against every policy will be included inside
-    ///   the finally Token returned by CoCo-AS.
+    /// - `policy_id`: The id of the policy that will be used to evaluate the claims.
```
Xynnn007, line 155:
```suggestion
    /// - `policy_ids`: The ids of the policies that will be used to evaluate the claims.
```

attestation-service/src/token/ear_broker.rs
```diff
@@ -0,0 +1,496 @@
...
+    pub build_name: String,
+
+    /// The Profile that describes the EAR token
+    /// Default: `tag:github.com,2024:confidential-containers/Trustee`
+    #[serde(default = "default_profile")]
+    pub profile_name: String,
+
+    /// Configuration for signing the EAR
+    /// If this is not specified, the EAR
+    /// will be signed with an ephemeral private key.
```
Xynnn007, line 88: Please add
```
#[serde(default = "Option::default")]
```

attestation-service/src/token/ear_broker.rs
```diff
@@ -0,0 +1,496 @@
...
+    async fn issue(
+        &self,
+        tcb_claims: TeeEvidenceParsedClaim,
+        policy_ids: Vec<String>,
+        init_data_claims: serde_json::Value,
+        runtime_data_claims: serde_json::Value,
+        reference_data_map: HashMap<String, Vec<String>>,
+        tee: Tee,
+    ) -> Result<String> {
+
```
Xynnn007, line 178: Let's remove this empty line here

attestation-service/src/token/ear_broker.rs
```diff
@@ -0,0 +1,496 @@
+// Copyright (c) 2023 Alibaba Cloud
```
Xynnn007, line 1: I think it should be IBM :p

attestation-service/src/token/ear_broker.rs
```diff
@@ -0,0 +1,496 @@
...
+    "hardware",
+    "runtime_opaque",
+    "storage_opaque",
+    "sourced_data",
+];
+
+#[derive(Deserialize, Debug, Clone, PartialEq)]
+pub struct TokenSignerConfig {
+    pub key_path: String,
+    pub cert_url: Option<String>,
```
Xynnn007, line 51: Please add
```
#[serde(default = "Option::default")]
```

This is for parse config when `key_path` is give while `cert_url` is not given. The same for following

attestation-service/src/token/ear_broker.rs
```diff
@@ -0,0 +1,496 @@
...
+
+    // PEM format certificate chain.
+    pub cert_path: Option<String>,
```
Xynnn007, line 54: Please add
```
#[serde(default = "Option::default")]
```

attestation-service/src/token/ear_broker.rs
```diff
@@ -0,0 +1,496 @@
...
+impl EarAttestationTokenBroker {
+    pub fn new(config: Configuration) -> Result<Self> {
+        let policy_engine = PolicyEngineType::OPA.to_policy_engine(
+            Path::new(&config.policy_dir),
+            include_str!("ear_default_policy.rego"),
+        )?;
+
+        let private_key_bytes = match &config.signer {
+            Some(signer) => std::fs::read(&signer.key_path)?,
+            None => generate_ec_keys()?.0,
```
Xynnn007, line 156: Do we need to `info!()` here to tell signing key is generated randomly?

attestation-service/src/token/ear_broker.rs
```diff
@@ -0,0 +1,496 @@
...
+    config: Configuration,
+    private_key_bytes: Vec<u8>,
+    policy_engine: Arc<dyn PolicyEngine>,
+}
+
+impl EarAttestationTokenBroker {
+    pub fn new(config: Configuration) -> Result<Self> {
+        let policy_engine = PolicyEngineType::OPA.to_policy_engine(
+            Path::new(&config.policy_dir),
+            include_str!("ear_default_policy.rego"),
```
Xynnn007, line 151: Do we need to tell a key of `default` policy is included by `info!()` or `debug!()`?

attestation-service/src/token/simple.rs
```diff
@@ -14,30 +20,106 @@ use openssl::{
...
 const SIMPLE_TOKEN_ALG: &str = "RS384";
 
+const DEFAULT_POLICY_DIR: &str = concatcp!(DEFAULT_TOKEN_WORK_DIR, "/simple/policies");
+
+const RULES: &str = "allow";
+
+#[derive(Deserialize, Debug, Clone, PartialEq)]
+pub struct TokenSignerConfig {
+    pub key_path: String,
+    pub cert_url: Option<String>,
```
Xynnn007, line 47: Please add
```rust
#[serde(default = "Option::default")]
```

attestation-service/src/token/simple.rs
```diff
@@ -14,30 +20,106 @@ use openssl::{
...
+
+    // PEM format certificate chain.
+    pub cert_path: Option<String>,
```
Xynnn007, line 50: Please add
```rust
#[serde(default = "Option::default")]
```

attestation-service/src/token/simple.rs
```diff
@@ -14,30 +20,106 @@ use openssl::{
...
+    pub duration_min: i64,
+
+    /// the issuer of the token
+    #[serde(default = "default_issuer_name")]
+    pub issuer_name: String,
+
+    /// Configuration for signing the token.
+    /// If this is not specified, the token
+    /// will be signed with an ephemeral private key.
+    pub signer: Option<TokenSignerConfig>,
```
Xynnn007, line 67: Please add
```rust
#[serde(default = "Option::default")]
```

attestation-service/src/token/ear_broker.rs
```diff
@@ -0,0 +1,496 @@
...
+                build: self.config.build_name.clone(),
+                developer: self.config.developer_name.clone(),
+            },
+            raw_evidence: None,
+            nonce: None,
+            submods,
+            extensions,
+        };
+
+        let signed_ear = ear.sign_jwt_pem(Algorithm::ES256, &self.private_key_bytes)?;
```
Xynnn007, line 259: We now supports to embed `jwk` part or `kid` part into the JWT with https://github.com/veraison/rust-ear/pull/29 / https://docs.rs/ear/0.3.0/ear/struct.Ear.html#method.sign_jwt_pem_with_header. Thus we can embed the public key part of `self.private_key_bytes`. In this way

1. if KBS side set `insecure_key = true`. This public jwk can be used to verify the integrity of the jwt.
2. if KBS side set `insecure_key = false`. Some extra trusted root certificate should be set on KBS config side and this public jwk can be used to verify the integrity of the jwt

kbs/src/token/mod.rs
```diff
@@ -2,21 +2,41 @@
...
 pub const TOKEN_TEE_PUBKEY_PATH_COCO: &str = "/customized_claims/runtime_data/tee-pubkey";
+pub const TOKEN_TEE_PUBKEY_PATH_EAR: &str =
+    "/submods/cpu/ear.veraison.annotated-evidence/runtime_data_claims/tee-pubkey";
 
-#[derive(Deserialize, Debug, Clone, PartialEq, Default)]
+#[derive(Deserialize, Default, Debug, Clone, EnumString)]
+pub enum AttestationTokenVerifierType {
+    Jwk,
+    #[default]
+    Ear,
```
Xynnn007, line 30: Please see the previous comment of signing token part of AS' EAR token side. Now EAR token generation will result in a same format of JWT that Simple/ITA uses. Thus we do not the extra token verifier kind but they can share a common one.

Xynnn007 (commented):
Some more nits and let's see what if a rebase comes

attestation-service/src/token/ear_default_policy.rego
```diff
@@ -0,0 +1,132 @@
...
+# configurations (like debug_allowed) are set correctly.
+else := 3 if {
+	input.snp.policy_debug_allowed == 0
+	input.snp.policy_migrate_ma == 0
+}
+
+##### TDX TODO
+##### AZ SNP TODO
+##### AZ TDX TODO
+##### SE TODO
```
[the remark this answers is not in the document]
fitzthum: We will still get an EAR token but it will contain a contraindicated trust vector. Since we don't actually check the policy result in our existing attestation token, this shouldn't matter too much in the short term.

I am planning to create a follow-up PR that adds tests that actually check the TCB and that use more interesting KBS policies. Hopefully we can collaboratively iterate on the AS policy after that with some of the HW experts.

fitzthum:
Rebased. The tests probably won't completely work but I will fix them up tomorrow. Thanks for some god initial feedback.

attestation-service/src/token/ear_broker.rs
fitzthum: oh didn't see that they fixed that. i will take a look.

fitzthum:
Build failures with kbs-client fixed in https://github.com/confidential-containers/trustee/pull/586

attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -87,93 +52,93 @@ impl OPA {
...
+
         Ok(res)
     }
 
-    async fn set_policy(&mut self, policy_id: String, policy: String) -> Result<(), RegoError> {
+    async fn set_policy(&self, policy_id: String, policy: String) -> Result<(), PolicyError> {
         let policy_bytes = base64::engine::general_purpose::URL_SAFE_NO_PAD
             .decode(policy)
-            .map_err(RegoError::Base64DecodeFailed)?;
+            .map_err(PolicyError::Base64DecodeFailed)?;
```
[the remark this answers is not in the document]
fitzthum: Ok you we're right about there being a driveby commit. I can drop this one if you want, but I've added the from for this variant at least.

kbs/test/config/kbs.toml
```diff
@@ -5,7 +5,8 @@ private_key = "./work/https.key"
 certificate = "./work/https.crt"
 
 [attestation_token_config]
-trusted_certs_paths = ["./work/token-cert.pem"]
```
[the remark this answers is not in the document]
fitzthum: Yes, but I will shortly update this PR to use the new custom JWT support in rust-ear, which should allow us to use the same certs for simple and ear.

attestation-service/src/token/ear_broker.rs
fitzthum: Ok, I am using the new JWK support and I got rid of the EAR token verifier. Seems to work.

There is a little bit of repeated code to make the JWKs. I wanted to reuse the `pubkey_jwks` function for EAR and CoCo tokens, but there are a few differences that made it slightly tricky. EAR uses EC not RSA, for instance, and the rust-ear crate expects a jsonwebtoken struct rather than a json string. I think we can probably converge these more in a future PR, but for now I kept them separate.

mkulke (approved):

attestation-service/src/config.rs
```diff
@@ -79,8 +62,7 @@ impl TryFrom<&Path> for Config {
     ///            "store_config": {},
     ///            "remote_addr": ""
     ///        },
-    ///        "attestation_token_broker": "Simple",
-    ///        "attestation_token_config": {
+    ///        "attestation_token_broker": {
```
Xynnn007, line 65:
```suggestion
    ///        "attestation_token_broker": {
    ///            "type": "Simple",
```

kbs/src/token/ear.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2023 by Alibaba.
```
Xynnn007, line 1: Should this be IBM?

attestation-service/src/token/ear_broker.rs
Xynnn007: Ok. Could you help to mark a TODO comment or an issue to record this?

Xynnn007 (approved):
Thanks @fitzthum ! Last comments

fitzthum:
Found a small issue. Fixing. We should be good to go v soon

kbs/test/config/kbs.toml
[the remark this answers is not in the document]
fitzthum: Ok now that we use the JWK verifier for both, I switched back to the cert chain and token key that we were already using. I just had to change the token key to be EC.

attestation-service/src/token/ear_broker.rs
fitzthum: Now we have https://github.com/confidential-containers/trustee/issues/598 and I added a todo.

fitzthum:
Ok, we're back. I think this is good to go. I will merge tomorrowish if there is no more feedback.

One thing to note is that now that we are using the JWK verifier for Ear tokens, we can set `insecure_key` to true. This wasn't possible with EAR until the changes to support the JWT. I would prefer to have `insecure_key` always be set to false, but for this PR I now have it matching what we do for the simple tokens since they use the same implementation. Setting up the cert chain is really orthogonal to using EAR, so let's keep tracking that in https://github.com/confidential-containers/trustee/issues/598.

Don't worry the Makefile test does check that the full cert chain verification works with Ear.

fitzthum:
Time to merge. I will keep a close eye on the CI and make sure I didn't leave any configs behind. Tomorrow I will post some issues with next steps for improving the policy and making tests that actually check the TCB.

PR #521, opened 2024-10-01
kbs: Simplify deployments, split out IBM SE

stevenhorsman:
The current "s390x" overlay support is currently very SE specific, whereas deploying on non-SE s390x with the sample KBS is still an important scenario for our testing without specialised hardware.

Let's separate out the ibmse overlay(s), and update deploy-kbs.sh to help make the s390x non-SE scenario easier.


stevenhorsman:
I've tested this on s390x (non-SE):
```
trustee/kbs/config/kubernetes# ./deploy-kbs.sh
namespace/coco-tenant created
configmap/kbs-config-68g582hg7m created
configmap/policy-config-55tfg2kc22 created
secret/kbs-auth-public-key-h64bbgc98k created
secret/keys-4cbmd8h422 created
service/kbs created
deployment.apps/kbs created
root@sh-libvirt-s390x:~/go/src/github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/test/trustee/kbs/config/kubernetes# kubectl get pods -A
NAMESPACE                        NAME                                               READY   STATUS             RESTARTS       AGE
coco-tenant                      kbs-65f467957d-m86qc                               1/1     Running            0              6s
```
I don't have an SE LPAR, so @BbolroC can you test it on SE when you get a chance please? Thanks!

kbs/config/kubernetes/deploy-kbs.sh
```diff
@@ -22,18 +25,13 @@ kbs_cert="${k8s_cnf_dir}/base/kbs.pem"
     openssl pkey -in "${k8s_cnf_dir}/base/kbs.key" -pubout -out "${kbs_cert}"
 }
 
-if [ "${ARCH}" == "s390x" ]; then
-    if [ -n "${IBM_SE_CREDS_DIR:-}" ]; then
+if [ "${OVERLAY}" == "ibmse" ]; then
```
mythi, line 28:
```suggestion
if [ "${OVERLAY}" == "ibm-se" ]; then
```

mythi:
> The current "s390x" overlay support is currently very SE specific, whereas deploying on non-SE s390x

What are the arch specific settings in these deployments? It feels to me this `s390x` vs `x86_64` vs `common` just makes things more complex than they should be and they should be removed altogether.

stevenhorsman:
> > The current "s390x" overlay support is currently very SE specific, whereas deploying on non-SE s390x
> 
> What are the arch specific settings in these deployments? It feels to me this `s390x` vs `x86_64` vs `common` just makes things more complex than they should be and they should be removed altogether.

That's fair - it looks like s390x doesn't need the same patch.yaml as x86, but otherwise they are very similar. The thought was to keep them together for flexibility, but I'm happy to try merging and having `common` and `ibm-se` if you'd prefer?

mythi:
> it looks like s390x doesn't need the same patch.yaml as x86,

AFAUI, this just means `workload_key/key.bin` Secret is not available on `s390x`.

> try merging and having `common` and `ibm-se`

we can treat `ibm-se` and an independent overlay and restore `overlays` back to where it was. This would also make "Expose KBS using Ingress" instructions working again.

stevenhorsman:
> we can treat `ibm-se` and an independent overlay and restore `overlays` back to where it was. This would also make "Expose KBS using Ingress" instructions working again.

Hi Mikko, just to try and clarify what you are suggesting as I'm not sure what "back to where it was" is referencing. Is you suggestion that the "normal" deployment is like:
- kubernetes
  - overlays
    - ingress.yaml
    - kustomization.yaml (that points to ../base, lists the patch.yml and the secretGenerator) 
    - patch.yaml (with the keys volumes mounts/volume)
 thereby removing the `common` and `x86_64` overlays entirely

Then we'd have the ibm-se overlay still in `kubernetes/overlays/ibm-se` with:
- kustomization.yaml (that is largely a duplicate, but points to pvc.yaml as well as ../../base, patch.yml and the secretGenerator) 
- patch.yaml, pv.yaml and pvc.yaml as they are in this PR?

Sorry if I've misunderstood, I'm not great with kustomize!

kbs/config/kubernetes/deploy-kbs.sh
stevenhorsman: Fixed - thanks for the spot!

mythi:
> Is you suggestion that the "normal" deployment is like:

yes, this was exactly my thinking. "back to where it was" is referring to the point where we did not have any of that `common/x86_64/s390x`.

stevenhorsman:
Ok, updated now and re-tested on s390x:
```
/trustee/kbs/config/kubernetes# ./deploy-kbs.sh
namespace/coco-tenant created
configmap/kbs-config-68g582hg7m created
configmap/policy-config-55tfg2kc22 created
secret/kbs-auth-public-key-h64bbgc98k created
secret/keys-ft5bc4tbth created
service/kbs created
deployment.apps/kbs created
root@sh-libvirt-s390x:~/go/src/github.com/confidential-containers/cloud-api-adaptor/src/cloud-api-adaptor/test/trustee/kbs/config/kubernetes# kubectl get pods -n coco-tenant
NAME                   READY   STATUS    RESTARTS   AGE
kbs-6fc45c745d-z6hxf   1/1     Running   0          10s
```

mythi (commented):
pls double check `ita` and `custom_pccs` for `x86_64` specific paths

other than that, LGTM.

stevenhorsman:
> pls double check `ita` and `custom_pccs` for `x86_64` specific paths
> 
> other than that, LGTM.

Sorry about this - whilst I was working on the refactor I added a mental note to fix them too, but it got paged out, so thanks for keeping me on it!

kbs/config/kubernetes/custom_pccs/kustomization.yaml
```diff
@@ -3,15 +3,15 @@ kind: Kustomization
...
+      kind: Deployment
+      name: kbs
 
 configMapGenerator:
-- files:
-  - sgx_default_qcnl.conf
-  name: dcap-attestation-conf
+  - files:
+      - sgx_default_qcnl.conf
+    name: dcap-attestation-conf
```
mythi, line 17: do you know what triggers these whitespaces here?

stevenhorsman: Argh - sorry the linter in my editor auto-formats the yamls, which I missed on the latest changes

mythi (approved):
thanks!

stevenhorsman:
> thanks!

Thanks for you review advice and patience!

kbs/config/kubernetes/deploy-kbs.sh
```diff
@@ -22,18 +21,15 @@ kbs_cert="${k8s_cnf_dir}/base/kbs.pem"
     openssl pkey -in "${k8s_cnf_dir}/base/kbs.key" -pubout -out "${kbs_cert}"
 }
 
-if [ "${ARCH}" == "s390x" ]; then
-    if [ -n "${IBM_SE_CREDS_DIR:-}" ]; then
+if [ "$(uname -m)" == "s390x" ] && [ -n "${IBM_SE_CREDS_DIR:-}" ]; then
```
fitzthum, line 24: It might be slightly nicer to split this conditional into two checks and warn people if they are on `s390x` but don't have `IBM_SE_CREDS_DIR` set.

fitzthum (approved):
LGTM. One little suggestion.

You might also think about adding a note or two to the docs about using s390x. There is one mention of setting an env var, but I don't see anything else about what users should do. Maybe they don't need to do anything else but still could be useful to note that. Maybe [here](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/cluster.md) or [here](https://github.com/confidential-containers/trustee/blob/main/kbs/quickstart.md).

kbs/config/kubernetes/deploy-kbs.sh
stevenhorsman: I'm not sure about this. Running on s390x without IBM_SE_CREDS_DIR is the point of this PR, so I'm not sure what we are warning against? I could add an info message to let people know the set-up?

stevenhorsman: I've added an echo about ibm-se being used, I'm not sure if that helps?

stevenhorsman:
> You might also think about adding a note or two to the docs about using s390x. There is one mention of setting an env var, but I don't see anything else about what users should do. Maybe they don't need to do anything else but still could be useful to note that. Maybe [here](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/cluster.md) or [here](https://github.com/confidential-containers/trustee/blob/main/kbs/quickstart.md).

The IBM SE deployment, with an explanation of `IBM_SE_CREDS_DIR` is documented in the Kubernetes deployment readme: https://github.com/confidential-containers/trustee/blob/main/kbs/config/kubernetes/README.md#deploy-kbs. Maybe that could be linked elsewhere?

kbs/config/kubernetes/deploy-kbs.sh
fitzthum: ah I see. well the echo can't hurt I guess.

fitzthum:
> The IBM SE deployment, with an explanation of IBM_SE_CREDS_DIR is documented in the Kubernetes deployment readme: https://github.com/confidential-containers/trustee/blob/main/kbs/config/kubernetes/README.md#deploy-kbs. Maybe that could be linked elsewhere?

That's a good start. It's fine to put it off to another PR, but that doc doesn't mention what this dir is or when/why it should be setup. I assume the idea is that you need it if you want to verify an SE guest but not if you just want to run Trustee on s390x.

stevenhorsman:
> That's a good start. It's fine to put it off to another PR, but that doc doesn't mention what this dir is or when/why it should be setup. I assume the idea is that you need it if you want to verify an SE guest but not if you just want to run Trustee on s390x.

Yeah, that's not something I'm an expert on, but it is specifically for SE, so I can try and make that clearer in the doc.

PR #522, opened 2024-10-02
CI: Fix artifact names in kbs e2e test

mkulke:
With the latest bump in the upload/download github actions, there has been a change we need to accomodate. we need to specify the artifacts in a single workflow run:

https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes


fitzthum (approved):
LGTM

PR #523, opened 2024-10-07
docs: fix doctest error

fitzthum:
For doctests, there is no keyword `no-run`.
Doctests recognizes `no_run`, but this still compiles the code, which will fail in this case.

We could fixup these examples with some hidden lines so that they compile, but since we don't really use rustdoc currently, let's just have the compiler ignore these examples.


Xynnn007 (approved):
Thanks Tobin. This is not a code that needs to be compiled so I agree with the idea.

Issue #519, opened 2024-09-30
Support multiple token verifiers simultaneously


fitzthum:
Today we have to select which token verifier we want to use at configuration time. The KBS can't handle connections from multiple guests that got their tokens from different types of verifiers.

This probably isn't too big of a deal, but worth noting.

fitzthum:
Also note that different tokens store the tee pubkey in different places. We have the logic to extract this in the attestation code, which is not very generic. We might want to move this logic into the token verifier code.

Xynnn007:
Nice idea. Luckily current ASes generate tokens all in JWT format, e.g. CoCo-AS Simple token, ITA Token, and potential EAR. So the difference would be the path of the tee public key of the JWT claim body. Most codes are the same.

Xynnn007:
The question is how would KBS detect the concrete kind of the JWT token runtimely. Probably the `iss` field of the JWT? But CoCoAS could specify [random issuer name](https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/token/mod.rs#L51) when launching.

fitzthum:
This doesn't seem necessary these days. We only have JWT attestation tokens.

PR #524, opened 2024-10-09
KBS: combine CoCo Token and Jwk Token verifier

Xynnn007:
This patch is part of #514.

ITA token and CoCo Token are both JWTs. They both need a JWK to verify the JWT. The difference is the way to gather the JWK.

This commit combined the two logic, and add two ways to get the JWK.
1. From the configured JwkSet when launching KBS
2. From the JWT's Header's jwk field.

The two ways will check the jwk endorsement in different ways. The first way is to configure the trusted JwkSet from the config. The second way is to configure the trusted CA in config. Then get the public key cert chain from Jwk's x5c field. The both ways are also supported in this patch.

The difference between CoCo Token and ITA Token are where Tee Public key is embedded inside the token, this could be configured in the launch toml of KBS, because it is hard to detect during runtime.

Rust does not provide a mature crate to verify cert chain, thus openssl is used in this patch. We also abondon rustls and openssl feature of KBS because openssl is by default used. Then we use openssl by default to make the code base simpler.

Close https://github.com/confidential-containers/trustee/issues/486

cc @mythi


LiuPeng-running:
Good good good.

Xynnn007:
This commit also changes some logic of previous JWK of CoCoAS, so ptal @jialez0

kbs/src/token/mod.rs
```diff
@@ -2,54 +2,86 @@
...
     #[serde(default)]
-    pub trusted_certs_paths: Vec<String>,
+    pub trusted_jwk_sets: Vec<String>,
+
+    /// Whether a JWK that directly comes from the JWT token is allowed to verify
+    /// the signature. This is insecure as it will not check the endorsement of
+    /// the JWK. If this option is set to false, the JWK will be looked up from
+    /// the key store configured during launching the KBS with kid field in the JWT,
+    /// or be checked against the configured trusted CA certs.
+    #[serde(default = "bool::default")]
```
fitzthum, line 48: Is the default true or false?

fitzthum (approved):
LGTM

Wdyt @mythi

kbs/src/token/mod.rs
Xynnn007: false

mythi:
My earlier comment in the original PR was that my preference is to not expose "ITA" as a token type to the user and that seems to be in this PR still. I was wondering whether we could just ask users to provide the list of trust anchors and try to parse the provided `file://` and OpenID config paths.

The only reason for "token type" to still exist seems to be the `tee-pubkey` search path in the claim. Would it make sense to have the _paths_ configurable instead of having them implicit via the token type? The search paths config could be a list to support tokens with different paths simultaneously.

Xynnn007:
> I was wondering whether we could just ask users to provide the list of trust anchors and try to parse the provided `file://` and OpenID config paths.

Yes. This PR supports the original ITA way, s.t. provode `file://` to speficy jwkset, or openid url for jwksets. The modification is the configuration name in the config file, from `trusted_certs_paths` to `trusted_jwk_sets`.

Another option is `trusted_certs_paths`. It's for trusted root certs to verify the `jwk`/`x5c` field within the JWT header.



> The only reason for "token type" to still exist seems to be the `tee-pubkey` search path in the claim. Would it make sense to have the _paths_ configurable instead of having them implicit via the token type? The search paths config could be a list to support tokens with different paths simultaneously.

Nice suggestion. The pro is flexibility while the con is extra complexity to remember the paths. I personally prefer the "token type" way as the user only needs to care about token type rather than claims inside that. I think you are caring about the name `ITA` to be exposed to users, so what do you prefer to name it?

mythi:
> The modification is the configuration name in the config file, from `trusted_certs_paths` to `trusted_jwk_sets`.
> 
> Another option is `trusted_certs_paths`. It's for trusted root certs to verify the `jwk`/`x5c` field within the JWT header.

I was not very clear earlier. My comment was that can we have just one config entry. It would be easier for the user. It should be possible to handle `["file:///foo/coco.pem", "file:///foo/ita.jwks"]`.

> The pro is flexibility while the con is extra complexity to remember the paths

True, OTOH, we can always make known paths as the default and only provide the mechanism to add supplemental paths.

mythi:
Would it be OK to aim fixing #519 with PR too?

Xynnn007:
> Would it be OK to aim fixing #519 with PR too?

Probably not. Currently it is hard for KBS to detect the JWT type (CoCoAS/ITA or else) runtimely. See https://github.com/confidential-containers/trustee/issues/519#issuecomment-2398902746

Xynnn007:
> True, OTOH, we can always make known paths as the default and only provide the mechanism to add supplemental paths.

Do you mean that we have two config items
- `type`: specify the token type, s.t. `CoCo` or `ITA`
- `tee_publickey_path`: if this parameter is given, ignore `type` and try to extract tee key from this path

mythi:
> > Would it be OK to aim fixing #519 with PR too?
> 
> Probably not. Currently it is hard for KBS to detect the JWT type (CoCoAS/ITA or else) runtimely. See [#519 (comment)](https://github.com/confidential-containers/trustee/issues/519#issuecomment-2398902746)

Could `kid`/x5*` header fields be used to map to what trust anchors we have configured. ITA token has `kid`and we use it to search the corresponding JWK in the JwkSet.

> Do you mean that we have two config items

Just one because `type` isn't used for anything else than just for mapping to tee-pubkey paths. So I'm thinking could we drop `type` completely, have some default tee-pubkey paths and allow users to configure supplemental `tee_publickey_path`s.

Xynnn007:
> Could `kid`/x5*`header fields be used to map to what trust anchors we have configured. ITA token has`kid`and we use it to search the corresponding JWK in the JwkSet.

Yes it can and it is implemented in this PR. What #519 aims is to support different public key claim path.



> Just one because `type` isn't used for anything else than just for mapping to tee-pubkey paths. So I'm thinking could we drop `type` completely, have some default tee-pubkey paths and allow users to configure supplemental `tee_publickey_path`s.

Ok. Could you provide an example for this in your mind?

mythi:
> > Could `kid`/x5*`header fields be used to map to what trust anchors we have configured. ITA token has`kid`and we use it to search the corresponding JWK in the JwkSet.
> 
> Yes it can and it is implemented in this PR. What #519 aims is to support different public key claim path.

Right, sorry. I had missed that and got side tracked when you wrote that JWT detection cannot be done easily.

> > Just one because `type` isn't used for anything else than just for mapping to tee-pubkey paths. So I'm thinking could we drop `type` completely, have some default tee-pubkey paths and allow users to configure supplemental `tee_publickey_path`s.
> 
> Ok. Could you provide an example for this in your mind?

Instead of `match self.token_type` we could just loop all configured/known paths until `claim.pointer(path)` returns `Ok`. For starters this could be just the two we know today (i.e., `TOKEN_TEE_PUBKEY_PATH_COCO` and `TOKEN_TEE_PUBKEY_PATH_ITA`) but maybe we want to add a config entry that allows users to add more paths to search.

Xynnn007:
@mythi Adds the change that you suggested. I think it is good to have some internal defined paths together with user defined extra paths.

mythi:
> @mythi Adds the change that you suggested. I think it is good to have some internal defined paths together with user defined extra paths.

thanks! I personally feel it's better this way and I guess this now meets #519 also. My remaining concern is whether we need separate user facing config entries for `trusted_certs_paths` and `trusted_jwk_sets`.

Xynnn007:
Yes as they are different things. One is jwkset and the other is root pem certs

kbs/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -287,8 +283,10 @@ impl Attest for IntelTrustAuthority {
 impl IntelTrustAuthority {
     pub async fn new(config: IntelTrustAuthorityConfig) -> Result<Self> {
         let token_verifier = JwkAttestationTokenVerifier::new(&AttestationTokenVerifierConfig {
-            attestation_token_type: AttestationTokenVerifierType::Jwk,
-            trusted_certs_paths: vec![config.certs_file.clone()],
+            extra_teekey_paths: vec![],
+            trusted_certs_paths: vec![],
+            trusted_jwk_sets: vec![config.certs_file.clone()],
+            insecure_key: true,
```
mythi, line 289:
```suggestion
            insecure_key: false,
```

mythi:
> Well as they are different things. One is jwkset and the other is root pem certs

If the take this path, we might need a third one for "OpenID Configuration endpoints" which is what the default ITA config prefers. But my point was that some generic config entry would probably keep things simple for the users.

kbs/src/attestation/intel_trust_authority/mod.rs
Xynnn007: Let me fix this. Btw, do you think we need to rename `certs_file` to `jwkset` or something else?

Xynnn007:
> If the take this path, we might need a third one for "OpenID Configuration endpoints" which is what the default ITA config prefers. But my point was that some generic config entry would probably keep things simple for the users.

Well. There are actually three items as you said
1. JwkSet from local
2. JwkSet to get from OpenID endpoints
3. Root certs to verify JWKs

The 1 and 2 are in good design currently I think, by `https://` or `path://` prefix to distinguish
The 3 is hard to merge into 1 or 2, as it is a PEM root cert not a JWK.

Before this patch the config item `trusted_certs_paths` is actually used in a wrong way, at least not a good name, because it is JwkSet for ITA rather than certs.

mythi:
> Before this patch the config item `trusted_certs_paths` is actually used in a wrong way, at least not a good name, because it is JwkSet for ITA rather than certs.

FWIW, ITA documentation calls these certificates: "To verify the token signature, first download the Intel Trust Authority signing certificates from the portal. " and that's why I though reusing `trusted_certs_paths` is fine.

Would there be any better name describing both root certs and jwksets?

Xynnn007:
> FWIW, ITA documentation calls these certificates: "To verify the token signature, first download the Intel Trust Authority signing certificates from the portal. " and that's why I though reusing `trusted_certs_paths` is fine.

Well, it is somehow tricky. The things download from "https://portal.trustauthority.intel.com/certs" are actually public keys with its `x5c`s in JWK format. This is the same as `file://xxx` for `trusted_jwk_sets`.

like
```json
{
      "alg": "PS384",
      "e": "AQAB",
      "kid": "c68e5a7b3aec51a3ff384fe30799d465d55b16d57ef88d05f6067559d9156cd3d5302f89fe665d1fb9a6be4602be11d8",
      "kty": "RSA",
      "n": "yE07D7FRSXLsswdeK7h22kw-Xv2K2r4NFoefWElZ6FWmLvCcd27wGEczNeKrE91SWczPtR279tTasQN6_v8qsswC5rCGYlrRWvE0vuPoUXezlV4PX0tCJJJmxWtXFXW0dChWvR1j-_viOItfR8jrybV2-DyVBgGX1ad4BLJJseglPXcofhnKYcG9gp8J2zPFqs1tu6jTW-He3Xw7ZeQNq0n4ZfrRBM3GEYVVsWGlTlqVidMhbvMXSQgz1x2QjyPC2mSUrT-JyA2xTm84Mv_Lmz6FpHXsjXMyPKCUVUf8LSTAiw3UsHa-7QGUW51hh9lZsbWkdSfwGUGxjrcMNEwYo3KvcF8f9Cv1_bkla396poQhtTIHuV478PobzsCfdkbCF5CfwZN31KbqyD9o9pVyzmmQUmOikIZuiSPRnfIU_P8duM5F6yvxQPITZf1RhOPBNYLiOJge7C89OmsM46UKtAYNTieBH-J8oWUUWfAX3pO38bKIzNwHDelSbaeHterJ",
      "x5c": [
            "MIIE/DCCA2SgAwIBAgIBAjANBgkqhkiG9w0BAQ0FADBlMQswCQYDVQQGEwJVUzELMAkGA1UECAwCQ0ExGjAYBgNVBAoMEUludGVsIENvcnBvcmF0aW9uMS0wKwYDVQQDDCRJbnRlbCBUcnVzdCBBdXRob3JpdHkgQVRTIFNpZ25pbmcgQ0EwHhcNMjQwOTA5MTAzMzQ3WhcNMjUwOTA5MTAzMzQ3WjBwMQswCQYDVQQGEwJVUzELMAkGA1UECAwCQ0ExGjAYBgNVBAoMEUludGVsIENvcnBvcmF0aW9uMTgwNgYDVQQDDC9JbnRlbCBUcnVzdCBBdXRob3JpdHkgQXR0ZXN0YXRpb24gVG9rZW4gU2lnbmluZzCCAaIwDQYJKoZIhvcNAQEBBQADggGPADCCAYoCggGBAMhNOw+xUUly7LMHXiu4dtpMPl79itq+DRaHn1hJWehVpi7wnHdu8BhHMzXiqxPdUlnMz7Udu/bU2rEDev7/KrLMAuawhmJa0VrxNL7j6FF3s5VeD19LQiSSZsVrVxV1tHQoVr0dY/v74jiLX0fI68m1dvg8lQYBl9WneASySbHoJT13KH4ZymHBvYKfCdszxarNbbuo01vh3t18O2XkDatJ+GX60QTNxhGFVbFhpU5alYnTIW7zF0kIM9cdkI8jwtpklK0/icgNsU5vODL/y5s+haR17I1zMjyglFVH/C0kwIsN1LB2vu0BlFudYYfZWbG1pHUn8BlBsY63DDRMGKNyr3BfH/Qr9f25JWt/eqaEIbUyB7leO/D6G87An3ZGwheQn8GTd9Sm6sg/aPaVcs5pkFJjopCGbokj0Z3yFPz/HbjOResr8UDyE2X9UYTjwTWC4jiYHuwvPTprDOOlCrQGDU4ngR/ifKFlFFnwF96Tt/GyiMzcBw3pUm2nh7XqyQIDAQABo4GrMIGoMAwGA1UdEwEB/wQCMAAwHQYDVR0OBBYEFGXRi1GAPlb8fBXUcGQrN5P0nrQAMB8GA1UdIwQYMBaAFCRX9lEHLr6HXZtQaFKogfoiHFl5MAsGA1UdDwQEAwIE8DBLBgNVHR8ERDBCMECgPqA8hjpodHRwczovL3BvcnRhbC50cnVzdGF1dGhvcml0eS5pbnRlbC5jb20vY3JsL2F0cy1jYS1jcmwuZGVyMA0GCSqGSIb3DQEBDQUAA4IBgQAGg5OCRX0QWAoJDhP4bdB2ECu/0WLS1XZHO0eWk7tL+358QAr1mf4Xyw6Y5RLK763Vrr62TeBuUFzex6CpmTp3z3dB8oWHL8VPTcsvnYZFSyL313/iKfemnaVZY7wCaUwI0Hc7UT/sIGY/gmBsDMCIJRIhDjSnx+oW0a6xDVMoevRM5P6jJIJHxNaM+4bV11+uNSQ9kysEXQBDyYaP9KtFlgKrY7YcPlnUwLNfjAB6/9CsyXxRw8YF3wjVly1zT27KSL5YcSKACu4Jzkwlz/m3GMkEfOV9y3IZEsiuplKniuljBOdcD73HdDNZzmgbEy8OHlbjar1FRcdHHbSE+NN+ZoWpzSx7p+q43/x1SSxift6SZGk7nykGJDD4vq5PYwk+ia/1xzKS3+2xsVS/z9Hz3NMMAlXjCUEbMn4vwQ7BpY3PtYPL5J/RhHt0plyqRj3db4jaTUPd/FTinHZBMYm4IdcxzIjuboTYYu1Q7fpCgGo7zT9xUSPjQNYjj2bfkRA=",
            "MIIFCjCCA3KgAwIBAgIBATANBgkqhkiG9w0BAQ0FADB0MSYwJAYDVQQDDB1JbnRlbCBUcnVzdCBBdXRob3JpdHkgUm9vdCBDQTELMAkGA1UEBhMCVVMxCzAJBgNVBAgMAkNBMRQwEgYDVQQHDAtTYW50YSBDbGFyYTEaMBgGA1UECgwRSW50ZWwgQ29ycG9yYXRpb24wHhcNMjMwOTEyMTExMTQ5WhcNMzYxMjMwMTExMTQ5WjBlMQswCQYDVQQGEwJVUzELMAkGA1UECAwCQ0ExGjAYBgNVBAoMEUludGVsIENvcnBvcmF0aW9uMS0wKwYDVQQDDCRJbnRlbCBUcnVzdCBBdXRob3JpdHkgQVRTIFNpZ25pbmcgQ0EwggGiMA0GCSqGSIb3DQEBAQUAA4IBjwAwggGKAoIBgQCof1PJ6PFnZ5TOyxBP8H7kyBxsAopUcMZtJAIdLZV+L+5DVMvH6E/hT4+7XX5SGYkb0R+XquyBz0PayzVrs71k8nL0MrwBIKLuIWQEcmTLC5/18Njf7QxBDZ3+uFrTOcfYcfYpfTNl2v/RvjEM6+KuDhxqLlH//buRO9eALngQQDqHq7pigrB9vVoOtpdng5Az4kjlDOOmdrNNigpEP4u4sQsqcAkSUFbofTXk8OiWtWClL6ItjosedwcabcdXSkAhf/T0QfYCcRQBOhSIblveZbaWgVXKb4S+HlM1Ft/QEtHNpuldlyI/s7+0ISAzVM8vRZU27EuPpQBUseNIrH2+DXTtpop13tozOl64o7VJmB7mwi+Zqv31NT0BucvMUdeC/bg2RSIKlV6RRomUTKtMFo3RpBi3K7+GUMbiq5GNQBNece294wHDhtgA+Bjg59IIxsHx3O9PmTgGAxmx8qAN2e9FPObTNWIHijfue6D0RkbolJd1/BGgAFcmF3pZy+cCAwEAAaOBtTCBsjASBgNVHRMBAf8ECDAGAQH/AgEAMB0GA1UdDgQWBBQkV/ZRBy6+h12bUGhSqIH6IhxZeTAfBgNVHSMEGDAWgBTzCwdViUpG9BjW2nyu+DI+d6gWETAOBgNVHQ8BAf8EBAMCAQYwTAYDVR0fBEUwQzBBoD+gPYY7aHR0cHM6Ly9wb3J0YWwudHJ1c3RhdXRob3JpdHkuaW50ZWwuY29tL2NybC9yb290LWNhLWNybC5kZXIwDQYJKoZIhvcNAQENBQADggGBADtWnJTjnCT6P5GZHS7Qz6MsrpM2IBwWcpnayTe+nV0CAqk4RJay7rupzq8gn8PllPXyWFoComrsT6HPZ80uh0JUIACOmNcO5UhwuRxML+EPmgpVVQJlz68AXf99Y1HaJxJ0aHkFSPr11XUOQ3S657QKee7RJijwcYu6rgfw6eVnYCGr7UD6SSW63D9nZLsa11v8GcIDWPdZVkyPnDVNJulAuWby/FQtZWAs4vCmxWpJYWoy303AVRzEBYoiyBRznWbed0ykyVU6TogLuezoxwH6jrZ7NeaFKrpbnD1YvI3JfP6EzPo1EqjpfumlVW99yY80mrHdr7FpIe9h9RL05utnYcoGt2VzbwN0H3ZXFPBwsBoioLX17xtSM7894w/rHdQV9wEMvxUT2Hmo+rRNu6lCQ3gDsLVXPvBd5rB3tnEY7wYu/uaLvHf01lq9/X9aTuISg63pFsqcb9oCS3hnx//b47/oHjo7yYCPhgKWHJdC5yiiv6U2NqQLeM9FtZIPuQ==",
            "MIIE2TCCA0GgAwIBAgIUZZX2XASGiYPzpZfwGNa7QFlK+3QwDQYJKoZIhvcNAQENBQAwdDEmMCQGA1UEAwwdSW50ZWwgVHJ1c3QgQXV0aG9yaXR5IFJvb3QgQ0ExCzAJBgNVBAYTAlVTMQswCQYDVQQIDAJDQTEUMBIGA1UEBwwLU2FudGEgQ2xhcmExGjAYBgNVBAoMEUludGVsIENvcnBvcmF0aW9uMB4XDTIzMDkxMjExMTE0OFoXDTQ5MTIzMDExMTE0OFowdDEmMCQGA1UEAwwdSW50ZWwgVHJ1c3QgQXV0aG9yaXR5IFJvb3QgQ0ExCzAJBgNVBAYTAlVTMQswCQYDVQQIDAJDQTEUMBIGA1UEBwwLU2FudGEgQ2xhcmExGjAYBgNVBAoMEUludGVsIENvcnBvcmF0aW9uMIIBojANBgkqhkiG9w0BAQEFAAOCAY8AMIIBigKCAYEAuQ1g1Q4MtnvTRhYITKVou5CDvB5FpfjBe8ssUKaDILwkgJlGNv66IWPQ9vhIekZQjuvHbUj9yPcrM7Hva7h7Ehlo/Sm7ZhY8AgZfGFTRJjvNU+vg/BTr5vuqUu/a54eewyzMcmxwmWhK/4cGQs7spFB346jpjSgOHgk5PgJ39PgEr5UL9SvJ1LFRuCNxZTdzyLe7K8cWvEnwGkR2RJpK9pYgzfnAWy8J0djdAKaoQxt8TOE/IwafG/0ujTeuNbzo+3wxeF6SGz56MimE1+KgraPpULaeX2tAL9lUz+ECMetNLbAyqHQwxN1jQZ/3VgpQ8qqh7Cyo4rEjpja29iRtOihBYlW0/X6TxOG1LLSuGo/N9CcSW6EzjsC1Bzakzjs4OD+JGaqqvc255p8URTxZSRSJr1xtimZK+BJoCHECsrCLvCC5UmFfTQrxkeGJ+OIDMkQFgidUw2K7kYXe9k7glKw9yyVf9C0hhqBFfPD8r+CXEA9m2u1tvR4NGMuoegMxAgMBAAGjYzBhMA8GA1UdEwEB/wQFMAMBAf8wHQYDVR0OBBYEFPMLB1WJSkb0GNbafK74Mj53qBYRMB8GA1UdIwQYMBaAFPMLB1WJSkb0GNbafK74Mj53qBYRMA4GA1UdDwEB/wQEAwIBBjANBgkqhkiG9w0BAQ0FAAOCAYEALTVDjOJL3XVuK25K6v/hd4z5vZOzhHK6ifnCjLWKu+CkPbKQMMMYMbYTpqm6gEcSr28dPMOF1IWG4xcJX9oJ6b3mDWj3NQNBomBI8fq0hRgz9hKF8wpFpU3UzXtyBLLL18BTCQeJecb6pwrpASVZNmR/fAv8fbR9wWdTBVFHmIrg1wvU1u2WqYSdAgQGke+WdWFIvZCKKMVAB0jX2kPHBQnAkLF3pRaVyNT4I2MCRB8cW2fbSCIARBeryiIHyGCKnDGkDK+dvPxMJ9eMidPbBQBp5t6jxoicg9X8Gw9MeOboOEOB1sIAd0S25V9fzIwwn6j61K4d2VkLf4ZnDa2VKTgmv6NFMynu+JlHVYhQ0yin+dfD63XJvBLcfLrJwK10lsVMX7xv3dB4P+DBYHtWPwrjE28j6+IjCIupuarzDcahBTbRZIAAW4eWKrA3MPQVyGikcdnciFN7kL12EtHhrSBL2jTzsScWvatPqqzIeNxDCywNEyjtGYLnLBnZnbSP"
      ]
}
```

But what we want to set `trusted_certs_paths` are certificates in PEM, like
```
-----BEGIN CERTIFICATE-----
MIIDTTCCAjWgAwIBAgIUTLtjGAVcMh37IncRgMdtPFdLcaswDQYJKoZIhvcNAQEL
BQAwNjENMAsGA1UECgwEQ05DRjENMAsGA1UECwwEQ29DbzEWMBQGA1UEAwwNS0JT
LXRlc3Qtcm9vdDAeFw0yNDEwMTYwODU1MDRaFw0zNDEwMTQwODU1MDRaMDYxDTAL
BgNVBAoMBENOQ0YxDTALBgNVBAsMBENvQ28xFjAUBgNVBAMMDUtCUy10ZXN0LXJv
b3QwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQC7UidBWbWG3zvVenS7
vaR6krAPD+9ysuk2WTj1wk1v29itgRnLOujHMv6D+DSm5QVVS+qh9MQt3jN0nM1c
sOWzO1jHDBARIG77IzpDq6RbZDofP8o5qvk+/ASScCErJTn4Fn72oLv7/+ZzPCsU
xRw9f1yVdlOU7mSORRNyO2gFVP/hYaGKYoS9ooIm4lDwsJXEh5PPCNO9+nkPitiK
VUGniGRRIrqV+HuI0QdmKMyggUbdC/41eChyGznQSJ6zEUUhcN6wIOddQEz/nLVG
2HmJdrvwDwjImQPEOk+6m8IyxY3LMNUEcFhlvJu6zJ1vOXRzTQxDBYj1z2MbFJpl
tWVvAgMBAAGjUzBRMB0GA1UdDgQWBBR1oZBCXudvu5Ce3yXI798lQei03zAfBgNV
HSMEGDAWgBR1oZBCXudvu5Ce3yXI798lQei03zAPBgNVHRMBAf8EBTADAQH/MA0G
CSqGSIb3DQEBCwUAA4IBAQBXib58Qy1L+1YZawOsXYt0Pibri5sTJQ/XzCKsoTh3
i264q2fo91jQUhJgQ0Tmgqce4j8DI9dGxT2pq8EBdwsi6OmYp30eFZeVEkh3U2dA
jcjcq33PJg/1e0mCkaKhCv27Lu6ks6VSv+Xmsp5iuXbkE+mUfSfdf6siccFYvJtt
EwLZr5ViLvHJO8A+yWUU6Jox9v6A5r+/dqHljlf32JezMCaXEkqbZbmNK7amH0zm
UnBTF/8FdUG3Qop55XjQzVOaBrhjhx5zwpZZBoeJ4a+Hjrh+ebm2yqvN1nsaEIoU
G3HYuWHSt97IcYHqz5qbrHwgb+oe5i0tavklwQxf/A5i
-----END CERTIFICATE-----
```

The two things also work differently in code. The former one will be used to look up jwk by given `kid` in a token, while the latter will use the certificate to check the endorsement of the `jwk` in a token.

mythi:
I agree that their content is very different. My only comment has been that for the user they are still both "token verification collateral" which is why I also asked:

> Would there be any better name describing both root certs and jwksets?

If other reviewers think making them separate config entries is the right thing to do, I can live with that.

Xynnn007:
To make it clear. The current token verifier config looks like
```toml
...
[attestation_token_config]
# Path of root certificate used to verify the trustworthy of `x5c` extension in the JWT
trusted_certs_paths = ["/path/to/trusted_cacert.pem"]

# URL (`path://` or `https://`) of the trusted JWK that can be indexed by `kid` in
# JWT Header.
trusted_jwk_sets = ["/url/to/trusted_jwk_set"]

# For Attestation Services like CoCo-AS, the public key to verify the JWT will be given
# in the token's `jwk` field (with or without the public key cert chain `x5c`).
#
# - If this flag is set to `true`, KBS will ignore to verify the trustworthy of the `jwk`.
# - If this flag is set to `false`, KBS will look up its `trusted_certs_paths` and the `x5c`
# field to verify the trustworthy of the `jwk`.
insecure_key = false
```

fitzthum:
> If other reviewers think making them separate config entries is the right thing to do, I can live with that.

I see no real benefit in having fewer config entries, especially if doing so could cause confusion. If we are dealing with different types of stuff we might as well make different optional config entries and add some logic to make sure one of them is set.

mythi:
> I see no real benefit in having fewer config entries, especially if doing so could cause confusion.

I was mainly thinking user experience but ok. For some users it can be confusing that data downloaded from `https://portal.trustauthority.intel.com/certs` don't belong to `trusted_certs_paths`.

jialez0 (approved):
LGTM.

mythi:
This PR was not ready for merging. For example, you still have `insecure_key: true` in place.

Xynnn007:
> This PR was not ready for merging. For example, you still have `insecure_key: true` in place.

The configuration keeps the behavior same as before. Before this commit, the insecure key option is always true locally.

I think we can fix this in future PR.

mythi:
> Before this commit, the insecure key option is always true logically although there is not such a config item.

This was not the case for `attestation_token_type = "Jwk" since it unconditionally expected to have `JwkSet` non-empty.

Xynnn007:
> This was not the case for `attestation_token_type = "Jwk"` since it unconditionally expected to have `JwkSet` non-empty.

Well. For old JWK (what ITA uses), the `insecure_key` is not relevant to that, because ITA has a `kid` field in JWT. The `kid` is used to look up previously configured `jwkset`. The `insecure_key` only works on tokens like CoCo's, where the JWK itself is included inside the JWT.

The docs upon `insecure_key` is
```
# For Attestation Services like CoCo-AS, the public key to verify the JWT will be given
# in the token's `jwk` field (with or without the public key cert chain `x5c`).
#
# - If this flag is set to `true`, KBS will ignore to verify the trustworthy of the `jwk`.
# - If this flag is set to `false`, KBS will look up its `trusted_certs_paths` and the `x5c`
# field to verify the trustworthy of the `jwk`.
```

This is partially why two config items seem better to me during our previous conversion.

mythi:
Fair enough, the functionality seems fine. My original comment was related to unresolved conversations that were not addressed.

kbs/src/attestation/intel_trust_authority/mod.rs
Xynnn007: @mythi Ok. Let me response here. Actually for ITA the configure item here both are ok as it will not change the behavior. When I re-read the part of code, I have a question whether we need `token_verifier` object here inside ITA backend plugin code. I am afraid that we do not.

The logic here is 
- a user requests ITA for a token (and we do not need to care about the token trustiworthy as we just call the API of ITA)
- the user request KBS for resource (this time we should check the token trustiworthy)

wdyt?

mythi: I had thought about the same but decided to keep it for now. We may have future feature additions like the faithful verification that further checks the token and it might be useful to do it here.

PR #531, opened 2024-10-15
Ci: Install Rust via `actions-rust-lang/setup-rust-toolchain@v1` rather than `rustup`

Xynnn007:
The original way to install rust dependency will not check if rust is already installed, thus it would require `rustup` be installed on the ci machine already while self-hosted CI machine may not.

We use `actions-rust-lang/setup-rust-toolchain@v1` to install rust because `actions-rs` has not been maintained for over 1 year and archived.

### DIFF
--- a/.github/workflows/as-e2e.yml
+++ b/.github/workflows/as-e2e.yml
@@ -35,10 +35,10 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Install Rust toolchain (${{ env.RUSTC_VERSION }})
-      run: |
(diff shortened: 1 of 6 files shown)


mythi:
@Xynnn007 See also: https://github.com/confidential-containers/confidential-containers/issues/230#issuecomment-2393222060

Xynnn007:
@BbolroC @mythi Hi, the PR is ready for review for now.

mkulke (approved):

mythi (approved):

PR #532, opened 2024-10-15
verifier: verify init_data in az-*-vtpm verifiers

mkulke:
This is in preperation for the upcoming init_data feature.

On the attester equivalent we have settled for sha256#PCR8 to measure init_data into. This register is usually occupied by GRUB, since we use UKI images for peerpods this is vacant.

If this turns out to be too restrictive, we can add a verifier specific configuration option later.

Note: We cannot merely compare the hash values, since PCR registers are extend-only. PCR8 will be filled with 0s initially and extended with the init_data digest. For a given init_data the PCR will contain:

```bash
sha256sum -b <(cat \
    <(dd if=/dev/zero of=/dev/stdout bs=32 count=1) \
    <(xxd -r -ps < /run/peerpod/initdata.digest) \
)
1+0 records in
1+0 records out
32 bytes copied, 4.7e-05 s, 681 kB/s
bddaccb9c52249e97a31baea61b7d91be8221a16e703d92148d04fb8e9c1dfdd */dev/fd/63
```

`bdda...` is the value we want to verify. We cannot perform a simple equality check, but have to replay the extend operation. that's also why the checks have to be somewhat local to a TEE verifier implementation.

An option would be an overrideable default on the Verifier Trait. If we later move the init_data verification to a more generic place, we would call a verifier's `verify_init_data()` impl.

### DIFF
--- a/deps/verifier/src/az_snp_vtpm/mod.rs
+++ b/deps/verifier/src/az_snp_vtpm/mod.rs
@@ -15,7 +15,7 @@ use az_snp_vtpm::hcl::HclReport;
 use az_snp_vtpm::report::AttestationReport;
 use az_snp_vtpm::vtpm::Quote;
 use az_snp_vtpm::vtpm::QuoteError;
-use log::{debug, warn};
+use log::debug;
 use openssl::pkey::PKey;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


Xynnn007 (approved):
LGTM

PR #533, opened 2024-10-15
resource: add PKCS11 resource back-end

fitzthum:
Some very basic PKCS11 support for resources. There are some limitations to what we can do with the current design. See commit message for more info. 

I will be OOO for the rest of the week, but I will add a test when I get back. This will probably need a rebase after the plugins PR goes in.


kbs/src/resource/pkcs11.rs
```diff
@@ -0,0 +1,106 @@
...
+use super::{Repository, ResourceDesc};
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct Pkcs11Config {
+    /// Path to the Pkcs11 module
+    module: String,
+
+    /// The index of the slot to be used
+    /// If not provided, the first slot will be used.
+    slot_index: Option<u8>,
```
Xynnn007, line 23: If we want this to be a config item, `String` or `i64` might be better?

Xynnn007: ok please ignore my previous comment. `u8` is good while it supports 0~255

kbs/src/resource/pkcs11.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    /// The user pin for authenticating the session
+    pin: String,
+}
+
+pub struct Pkcs11Backend {
+    session: Arc<Mutex<Session>>,
```
Xynnn007, line 30: Do we need `Mutex` here to control race condition? I noticed that all methods of `Session` in `cryptoki` are declared `&self` rather than `&mut self`

fitzthum: I think the session struct itself is not thread safe. Without this I get 

    within `cryptoki::session::Session`, the trait `Sync` is not implemented for `*mut u32`, which is required by `Pkcs11Backend: Sync`
    
I'm not sure that this is the best way to resolve it. I don't know exactly what the properties of that field are so I went wiith a mutex.

Xynnn007: Makes sense to me. Based on this, I suggest `RwLock` to replace `Mutex` while `&self` operations will not be blocked.

bpradipt:
@fitzthum possible to add a README explaining how this may be used ?

fitzthum:
This PR has been rebased. 

I also fixed a few things in the Aliyun KMS that were broken in the plugin refactor PR and cleaned up the documentation.
@bpradipt there is now a section in the docs describing how to use this backend.

Xynnn007 (approved):

bpradipt (approved):
/lgtm
Thanks @fitzthum

PR #535, opened 2024-10-16
Derive Debug for Data and Enums

ssolit:
Hey, it would be nice to be able to debug these types and make serialization easier. I hope you would consider merging this.

### DIFF
--- a/attestation-service/src/lib.rs
+++ b/attestation-service/src/lib.rs
@@ -30,7 +30,7 @@ use verifier::{InitDataHash, ReportData};
 use crate::utils::flatten_claims;
 
 /// Hash algorithms used to calculate runtime/init data binding
-#[derive(Display, EnumString, AsRefStr)]
+#[derive(Debug, Display, EnumString, AsRefStr)]
 pub enum HashAlgorithm {
     #[strum(ascii_case_insensitive)]
     Sha256,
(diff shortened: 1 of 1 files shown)


Xynnn007:
Hi @ssolit  The error here is likely to be accidental, and subsequent retries are based on the first cache, so the error continues. I tested it locally with no errors. Can you help force push this PR and use different commit information to trigger CI execution?

ssolit:
I believe I did what you requested. I rebased off of the latest dependabot commit in origin/main and mildly changed the commit message.

I can't see if this is going to work on my end because the workflows are skipped until there's an approving review. If it doesn't work again should I create a fresh PR?

ssolit:
It seems now "CoCo-AS Rust Suites / Check (pull_request)" is failing. However it succeeds when I run it locally

fitzthum:
Re-running. Let's see if the caching works out.

@Xynnn007 maybe we should take another look at how we are [comparing the errors](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/tdx/claims.rs#L559-L560) in the test. That may be a bit error-prone.

I had some trouble comparing errors in some other tests and introduced a weird [compare errors](https://github.com/ssolit/trustee/blob/main/kbs/src/policy_engine/opa/mod.rs#L79-L95) function. I'm not sure if there is a great generic way to do it.

Xynnn007:
Hi @ssolit , the error is not related to your PR I believe. I have put a PR for the CI error https://github.com/confidential-containers/trustee/pull/538

Let's take a rebase after that gets merged and hopefully all will be green

Xynnn007:
@ssolit hey, could you help to take a rebase of this PR? #538 fixed the CI error I think hopefully

fitzthum:
Thanks @ssolit

ssolit:
Thank you all for help getting the change in

PR #538, opened 2024-10-18
verifier/tdx: fix TdShimPlatformConfigInfo parse error comparation

Xynnn007:
The original unit tests use anyhow for error handling. The results are unstable because the debug format of the anyhow error might change due to different environments. This patch replace anyhow with thiserror without changing the test aim. This would give a more robust error type defination for comparation.

cc @jodh-intel @fitzthum

### DIFF
--- a/deps/verifier/src/tdx/claims.rs
+++ b/deps/verifier/src/tdx/claims.rs
@@ -49,6 +49,7 @@ use anyhow::*;
 use byteorder::{LittleEndian, ReadBytesExt};
 use log::{debug, warn};
 use serde_json::{Map, Value};
+use thiserror::Error;
 
 use crate::{eventlog::AAEventlog, tdx::quote::QuoteV5Body, TeeEvidenceParsedClaim};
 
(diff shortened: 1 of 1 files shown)


deps/verifier/src/tdx/claims.rs
```diff
@@ -239,8 +240,20 @@ fn parse_ccel(ccel: CcEventLog, ccel_map: &mut Map<String, Value>) -> Result<()>
...
 
-const ERR_INVALID_HEADER: &str = "invalid header";
-const ERR_NOT_ENOUGH_DATA: &str = "not enough data after header";
+#[derive(Error, Debug, PartialEq)]
+pub enum ParseError {
+    #[error("Failed to parse `Descriptor`")]
+    ParseDescriptor,
+
+    #[error("Failed to parse `InfoLength`")]
+    ParseInfoLength,
```
mythi, line 249: AFAIU, it's more like:
```suggestion
    ReadInfoLength(#[from] io::Error),
```

Xynnn007: `std::io:Error` does not support `PartialEq` thus we cannot use `#[from]` macro. This is possible as the error information is enough.

mkulke (approved):

mythi (approved):
LGTM

PR #550, opened 2024-10-30
Put `rvps` dependencies behind `rvps-builtin` feature flag

cdrappi:
First of all thank you for building and maintaining such a wonderful suite of libraries! CoCo has been immensely valuable for our team and we appreciate all of the work that the community has put into these projects

### Problem
Currently if you run build without the `rvps-builtin` feature, it fails:

`cargo build --no-default-features --features restful-bin,rvps-grpc`

This is because the `reference-value-provider-service` dependency is not included under this feature set (requires `rvps-builtin`), yet it is used outside of that flag

### Proposed solution
This change tucks usage of `rvps` behind that flag, so the above built command will complete successfully

### DIFF
--- a/attestation-service/src/rvps/mod.rs
+++ b/attestation-service/src/rvps/mod.rs
@@ -5,6 +5,7 @@
 
 use anyhow::Result;
 use log::{info, warn};
+#[cfg(feature = "rvps-builtin")]
 use reference_value_provider_service::config::{Config as RvpsCrateConfig, DEFAULT_STORAGE_TYPE};
 use serde::Deserialize;
 use serde_json::{json, Value};
@@ -31,7 +32,11 @@ pub mod grpc;
 pub mod builtin;
 
 fn default_store_type() -> String {
-    DEFAULT_STORAGE_TYPE.into()
+    #[cfg(feature = "rvps-builtin")]
+    {
+        return DEFAULT_STORAGE_TYPE.into()
+    }
+    return "LocalFs".into()
 }
 
 fn default_store_config() -> Value {
@@ -54,6 +59,7 @@ pub struct RvpsConfig {
     pub store_config: Value,
 }
 
+#[cfg(feature = "rvps-builtin")]
 impl From<RvpsConfig> for RvpsCrateConfig {
     fn from(val: RvpsConfig) -> RvpsCrateConfig {
         RvpsCrateConfig {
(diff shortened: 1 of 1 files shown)


attestation-service/src/rvps/mod.rs
```diff
@@ -31,7 +32,11 @@ pub mod grpc;
 pub mod builtin;
 
 fn default_store_type() -> String {
-    DEFAULT_STORAGE_TYPE.into()
+    #[cfg(feature = "rvps-builtin")]
+    {
+        return DEFAULT_STORAGE_TYPE.into()
+    }
+    return "LocalFs".into()
```
cdrappi, line 39: unclear what the correct thing to do is here

attestation-service/src/rvps/mod.rs
```diff
@@ -31,7 +32,14 @@ pub mod grpc;
...
 fn default_store_type() -> String {
-    DEFAULT_STORAGE_TYPE.into()
+    #[cfg(feature = "rvps-builtin")]
+    {
+        DEFAULT_STORAGE_TYPE.into()
+    }
+    #[cfg(not(feature = "rvps-builtin"))]
+    {
+        return "LocalFs".into()
+    }
```
cdrappi, line 42: unclear what the correct thing to do is here

fitzthum: hmm I think the best solution here would be to get rid of these default functions entirely and change the config struct to have Optional fields. Then we can use `unwrap_or` to swap in the default values as needed (i.e. inside of the RVPS module). That's slightly more work but I think it will untangle the dependencies.

fitzthum (commented):
Good spotting. The current implementation isn't quite right.

Xynnn007:
Thanks @cdrappi for raising this up. The current RVSP config is not implemented well as it does not have an explicit field to specify whether a local one or a remote one to use. Probably we need to tidy things up.

attestation-service/src/rvps/mod.rs
Xynnn007: The configuration part about RVPS seem not good here. If `remote_addr` is set, the `store_type` is actually not useful. It might be better to not have useless fields that can be set. Let me make a fix to cover this. I made a PR https://github.com/confidential-containers/trustee/pull/553

fitzthum: I would have settled for a simpler fix but #553 seems good at first glance.

Xynnn007:
Close as finished by #553

PR #553, opened 2024-11-01
AS: fix build with rvps features

Xynnn007:
Before this commit when we only enable `restful-bin,rvps-grpc` features, the attestation service cannot be built.

This patch fixes this issue. Also, this commit deletes the `rvps-builtin` feature as the code of rvps does not bring in any new dependencies.

Related configurations are updated to make it more robust, together with the documents.

### DIFF
--- a/attestation-service/Cargo.toml
+++ b/attestation-service/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [features]
-default = ["restful-bin", "rvps-grpc", "rvps-builtin"]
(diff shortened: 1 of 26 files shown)


Xynnn007:
Let's mark this as draft because there is some overlap to #514

Xynnn007:
cc @lmilleri @spotlesstofu this might influence operator side i am not sure

mkulke (approved):

lmilleri:
> cc @lmilleri @spotlesstofu this might influence operator side i am not sure

@Xynnn007 thank you for the heads-up!

Issue #456, opened 2024-08-02
Verifier bails instead of automatically fetching certificate from KDS


AdithyaKrishnan:
The mod.rs code in verifier currently prints a statement that it Could not find either VCEK or VLEK in cert chain instead of automatically fetching it from the KDS

mkulke:
> The mod.rs code in verifier currently prints a statement that it Could not find either VCEK or VLEK in cert chain instead of automatically fetching it from the KDS

Afaik KDS has a rate-limit, that would made it unreliable to use it by default. The VCEK/VLEK should be in the SNP-attester's evidence?

Doctor-love:
> Afaik KDS has a rate-limit, that would made it unreliable to use it by default. The VCEK/VLEK should be in the SNP-attester's evidence?

If I understand you correctly, the "extended attestation workflow" as described in [this issue comment](https://github.com/AMDESE/AMDSEV/issues/212#issuecomment-2064093286) is required in the current implementation of the verifier? Or is there some other way this is meant to work?

mkulke:
I think that's how it's supposed to work. The VCEK could be cached in multiple places (like a cloud's IMDS), but the public KDS service is occasionally not available, so I'd be cautious to build an architecture around it. We could still have a local memory cache on trustee for VCEKs maybe.

fitzthum:
Apparently the latest upstream host patches no longer support setting the extended report (which the verifier currently requires), so we should make sure we implement some kind of fallback. If the AS reaches out to the KDS, it should cache the responses to avoid doing it too often.

Issue #363, opened 2024-04-08
SNP: add fallback for cert-chain without extended report


fitzthum:
If the cert chain does not include the extended report, we should connect to the KDS to get it.

salmanyam:
@fitzthum We have also experienced a similar issue for AMD "genoa" platform where we found that the cert-chain is null in the attestation report. We communicated with the KDS to fetch the cert-chain for the verification in order to complete our use case. We can submit a pull request to enable this functionality. What do you think?

fitzthum:
Sure. Note that @niteeshkd has a [PR for Kata](https://github.com/kata-containers/kata-containers/pull/9806) to allow us to specify the cert chain from the host side.

I think we could also add support to the verifier to download the VCEK if it is not provided. Note that the ARK/ASK/ASVK are hard-coded into the verifier but we don't yet have the Genoa certs there. It should be easy to add them.

I think @AdithyaKrishnan might be working on something in this area, but if you already have the code you should post it.

larrydewey:
I think this would be an excellent feature. I will follow-up with @AdithyaKrishnan to see what the status of the Genoa certificates are.

hashimotor-ntt:
We hit the same Genoa issue.  Additionally, the chain KDS provides for Milan and Genoa omits the ASVK.  We'll follow the approaches by @salmanyam or @AdithyaKrishnan if they have already addressed those issues.  Or I might bring a workaround to ignore ASVK and adapt Genoa certs, instead of the fundamental fix of changing load_milan_cert_chain() to load_cert_chain() and adding the fallback.

salmanyam:
@hashimotor-ntt  We have a working solution for Genoa, and we trying to implement our solution universal, at least for both Genoa and Milan. We are currently testing the solution, and experiencing a minor issue on Milan. As soon as we fix the issue, we will submit a PR.

PR #555, opened 2024-11-04
verifier: Fetch VCEK cert from KDS instead of bailing

AdithyaKrishnan:
Fetch the VCEK cert from the KDS if it is absent in the cert chain instead of just printing a bail statement stating that the VCEK is not found.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -605,9 +605,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "1.4.0"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
+checksum = "0c4b4d0bd25bd0b74681c0ad21497610ce1b7c91b1022cd21c80c6fbdd9476b0"
 
 [[package]]
 name = "axum"
@@ -1138,9 +1138,9 @@ checksum = "12170080f3533d6f09a19f81596f836854d0fa4867dc32c8172b8474b4e9de61"
 
 [[package]]
 name = "colorchoice"
-version = "1.0.3"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b63caa9aa9397e2d9480a9b13673856c78d8ac123288526c37d7839f2a86990"
+checksum = "d3fd119d74b830634cea2a0f58bbd0d54540518a14397557951e79340abc28c0"
 
 [[package]]
 name = "config"
@@ -2863,9 +2863,9 @@ checksum = "0717cef1bc8b636c6e1c1bbdefc09e6322da8a9321966e8928ef80d20f7f770f"
 
 [[package]]
 name = "linux-raw-sys"
-version = "0.4.14"
+version = "0.4.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "78b3ae25bc7c8c38cec158d1f2757ee79e9b3740fbc7ccf0e59e4b08d793fa89"
+checksum = "01cda141df6706de531b6c46c3a33ecca755538219bd484262fa09410c13539c"
 
 [[package]]
 name = "local-channel"
@@ -3071,9 +3071,9 @@ dependencies = [
 
 [[package]]
 name = "num"
-version = "0.4.3"
+version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "35bd024e8b2ff75562e5f34e7f4905839deb4b22955ef5e73d2fea1b9813cb23"
+checksum = "3135b08af27d103b0a51f2ae0f8632117b7b185ccf931445affa8df530576a41"
 dependencies = [
  "num-bigint",
  "num-complex",
@@ -3085,10 +3085,11 @@ dependencies = [
 
 [[package]]
 name = "num-bigint"
-version = "0.4.6"
+version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a5e44f723f1133c9deac646763579fdb3ac745e418f2a7af9cd0c431da1f20b9"
+checksum = "608e7659b5c3d7cba262d894801b9ec9d00de989e8a82bd4bef91d08da45cdc0"
 dependencies = [
+ "autocfg",
  "num-integer",
  "num-traits",
 ]
@@ -3112,9 +3113,9 @@ dependencies = [
 
 [[package]]
 name = "num-complex"
-version = "0.4.6"
+version = "0.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "73f88a1307638156682bada9d7604135552957b7818057dcef22705b4d509495"
+checksum = "23c6602fda94a57c990fe0df199a035d83576b496aa29f4e634a8ac6004e68a6"
 dependencies = [
  "num-traits",
 ]
@@ -3147,9 +3148,9 @@ dependencies = [
 
 [[package]]
 name = "num-iter"
-version = "0.1.45"
+version = "0.1.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1429034a0490724d0075ebb2bc9e875d6503c3cf69e235a8941aa757d83ef5bf"
+checksum = "d869c01cc0c455284163fd0092f1f93835385ccab5a98a0dcc497b2f8bf055a9"
 dependencies = [
  "autocfg",
  "num-integer",
@@ -3158,10 +3159,11 @@ dependencies = [
 
 [[package]]
 name = "num-rational"
-version = "0.4.2"
+version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f83d14da390562dca69fc84082e73e548e1ad308d24accdedd2720017cb37824"
+checksum = "0638a1c9d0a3c0914158145bc76cff373a75a627e6ecbfb71cbe6f453a5a19b0"
 dependencies = [
+ "autocfg",
  "num-bigint",
  "num-integer",
  "num-traits",
@@ -3169,9 +3171,9 @@ dependencies = [
 
 [[package]]
 name = "num-traits"
-version = "0.2.19"
+version = "0.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "071dfc062690e90b734c0b2273ce72ad0ffa95f0c74596bc250dcfd960262841"
+checksum = "da0df0e5185db44f69b44f26786fe401b6c293d1907744beaa7fa62b2e5a517a"
 dependencies = [
  "autocfg",
  "libm",
@@ -3278,9 +3280,9 @@ dependencies = [
 
 [[package]]
 name = "openssl-sys"
-version = "0.9.104"
+version = "0.9.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "45abf306cbf99debc8195b66b7346498d7b10c210de50418b5ccd7ceba08c741"
+checksum = "c597637d56fbc83893a35eb0dd04b2b8e7a50c91e64e9493e398b5df4fb45fa2"
 dependencies = [
  "cc",
  "libc",
@@ -4180,7 +4182,6 @@ dependencies = [
  "bytes",
  "cookie 0.18.1",
  "cookie_store",
- "futures-channel",
  "futures-core",
  "futures-util",
  "http 1.1.0",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2677,7 +2677,7 @@ dependencies = [
  "prost 0.12.6",
  "rand",
  "regorus",
- "reqwest 0.12.8",
+ "reqwest 0.12.9",
  "rsa 0.9.6",
  "rstest",
  "scc",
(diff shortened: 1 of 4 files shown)


deps/verifier/Cargo.toml
```diff
@@ -49,6 +49,7 @@ strum.workspace = true
 veraison-apiclient = { git = "https://github.com/chendave/rust-apiclient", branch = "token", optional = true }
 ear = { git = "https://github.com/veraison/rust-ear", rev = "43f7f480d09ea2ebc03137af8fbcd70fe3df3468", optional = true }
 x509-parser = { version = "0.14.0", optional = true }
+reqwest = { version="0.12.9", features = ["blocking"] }
```
fitzthum, line 52: I you are already specifying the needed version of `reqwest` in the global Cargo file, you can just say `request = workspace.true` here.

deps/verifier/src/snp/mod.rs
```diff
@@ -84,8 +85,10 @@ impl Verifier for Snp {
             cert_chain,
         } = serde_json::from_slice(evidence).context("Deserialize Quote failed.")?;
 
-        let Some(cert_chain) = cert_chain else {
-            bail!("Cert chain is unset");
+        let cert_chain = if let Some(chain) = cert_chain{
```
fitzthum, line 88: So the cert chain is type `Option<Vec<Cert>>` or something. This accounts for cases where the option is none, but what if the option is there but the vector is empty?

Ideally we should get the vcek even if there is a non-empty cert chain as long as the vcek/vlek isn't part of it.

deps/verifier/src/snp/mod.rs
```diff
@@ -307,6 +310,38 @@ fn get_common_name(cert: &x509::X509) -> Result<String> {
     Ok(e.data().as_utf8()?.to_string())
 }
 
+// Function to request vcek from KDS. Return vcek in der format.
+fn request_endorsement_key_kds(
```
fitzthum, line 314: nit: `request_endorsement_key_kds` -> `get_endorsement__key_from_kds`

deps/verifier/src/snp/mod.rs
```diff
@@ -307,6 +310,38 @@ fn get_common_name(cert: &x509::X509) -> Result<String> {
...
+    att_report: AttestationReport,
+) -> Result<Vec<CertTableEntry>, anyhow::Error> {
+    // KDS URL parameters
+    const KDS_CERT_SITE: &str = "https://kdsintf.amd.com";
+    const KDS_VCEK: &str = "/vcek/v1";
```
fitzthum, line 319: might be better to put these consts at the top of the file rather than inside this function

deps/verifier/src/snp/mod.rs
```diff
@@ -307,6 +310,38 @@ fn get_common_name(cert: &x509::X509) -> Result<String> {
...
+        att_report.reported_tcb.microcode
+    );
+    // VCEK in DER format
+    let vcek_rsp: ReqwestResponse = get(vcek_url).context("Unable to send request for VCEK")?;
+
+    match vcek_rsp.status() {
+        StatusCode::OK => {
+            let vcek_rsp_bytes: Vec<u8> =
+                vcek_rsp.bytes().context("Unable to parse VCEK")?.to_vec();
+            let key= CertTableEntry{cert_type: CertType::VCEK, data: vcek_rsp_bytes};
```
fitzthum, line 338: nit: `key=` -> `key =`. Also you probably want some spaces around the `{`. I'm surprised this survived `cargo fmt` check.

deps/verifier/src/snp/mod.rs
```diff
@@ -307,6 +310,38 @@ fn get_common_name(cert: &x509::X509) -> Result<String> {
     Ok(e.data().as_utf8()?.to_string())
 }
 
+// Function to request vcek from KDS. Return vcek in der format.
+fn request_endorsement_key_kds(
+    att_report: AttestationReport,
+) -> Result<Vec<CertTableEntry>, anyhow::Error> {
```
cclaudio, line 316: The function is specific to download the VCEK, so I think it could be rename to something more specific such as `fetch_vcek_from_kds()`

cclaudio (commented):
Just a few minor comments.

fitzthum (commented):
Nice PR. Looks like this will address #456 which is actually a duplicate of #363 

One thing to think about is that this could result in quite a few requests being made to the KDS. We should probably implement some kind of caching mechanism so that the verifier doesn't have to get the `vcek` for the same node again and again. We can probably do this in a follow-up. 

Also, we might want to add `vlek` support at some point although I suspect this will require specifying a few more config params.

Cargo.lock
cclaudio, line 1: It seems your Cargo.lock has more changes than required. You may need to restore it and apply the `request = worspace.true` change that Tobin suggested

deps/verifier/src/snp/mod.rs
cclaudio: Additionally, I think the function would be more self-contained if it returns `Result<Vec<u8>>`

AdithyaKrishnan: Changed. If appropriate, please mark this comment as Resolved.

deps/verifier/Cargo.toml
AdithyaKrishnan: Changed. If appropriate, please mark this comment as Resolved.

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: Per Claudio's suggestion, I've renamed it to fetch_vcek_from_kds().  If appropriate, please mark this comment as Resolved.

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: Fixed. If appropriate, please mark this comment as Resolved.

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: If appropriate, please mark this comment as Resolved.

Cargo.lock
AdithyaKrishnan: If appropriate, please mark this comment as Resolved.

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: > Additionally, I think the function would be more self-contained if it returns `Result<Vec<u8>>`

Tried this and got a type mismatch error

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: Could this be part of the future follow-up PR?

deps/verifier/src/snp/mod.rs
```diff
@@ -482,4 +518,4 @@ mod tests {
         let vendor_certs = load_milan_cert_chain().as_ref().unwrap();
         verify_report_signature(&attestation_report, &cert_chain, vendor_certs).unwrap_err();
     }
-}
+}
```
Xynnn007, line 521: this empty line can be reserved

Cargo.toml
```diff
@@ -36,7 +36,7 @@ jsonwebtoken = { version = "9", default-features = false }
 log = "0.4.17"
 prost = "0.12"
 regorus = { version = "0.1.5", default-features = false, features = ["regex", "base64", "time"] }
-reqwest = { version = "0.12", default-features = false, features = ["default-tls"] }
+reqwest = { version = "0.12", default-features = false, features = ["default-tls", "blocking"] }
```
Xynnn007, line 39: I noticed that the `reqwest` is used in `fetch_vcek_from_kds`. It is suggested to use `reqwest::get` (the async version) rather than blocking version, because the blocking version would hang on network I/O during which the process would wait.

deps/verifier/Cargo.toml
```diff
@@ -49,6 +49,7 @@ strum.workspace = true
 veraison-apiclient = { git = "https://github.com/chendave/rust-apiclient", branch = "token", optional = true }
 ear = { git = "https://github.com/veraison/rust-ear", rev = "43f7f480d09ea2ebc03137af8fbcd70fe3df3468", optional = true }
 x509-parser = { version = "0.14.0", optional = true }
+reqwest.workspace = true
```
Xynnn007, line 52: It is good to mark this dependency to feature `snp-verifier`. s.t.
```suggestion
reqwest = { optional = true, workspace = true }
```

and add `"request"` to previous `[features]` section
```toml
...
[features]
snp-verifier = [ "asn1-rs", "openssl", "sev", "x509-parser", "reqwest" ]
...
```

deps/verifier/src/snp/mod.rs
mkulke: there are more formatting/lint issues. we do have a problem in the build pipeline, the verifier crate is not linted/formatted.

mkulke: can you run

```bash
cargo fmt
cargo clippy -p verifier -- -D warnings
```

on your code in the mean time and fix the warnings?

deps/verifier/src/snp/mod.rs
fitzthum: We can leave check the cert types to a follow-up, but please add a case to this statement to catch an empty vector.

deps/verifier/src/snp/mod.rs
fitzthum: Btw I don't think you need `anyhow::Error` in in this return type.

deps/verifier/src/snp/mod.rs
```diff
@@ -307,6 +314,35 @@ fn get_common_name(cert: &x509::X509) -> Result<String> {
...
+    let vcek_rsp: ReqwestResponse = get(vcek_url).context("Unable to send request for VCEK")?;
+
+    match vcek_rsp.status() {
+        StatusCode::OK => {
+            let vcek_rsp_bytes: Vec<u8> =
+                vcek_rsp.bytes().context("Unable to parse VCEK")?.to_vec();
+            let key = CertTableEntry { cert_type: CertType::VCEK, data: vcek_rsp_bytes };
+            Ok(vec![key])
+        }
+        status => Err(anyhow::anyhow!("Unable to fetch VCEK from URL: {status:?}")),
```
fitzthum, line 342: `anyhow::anyhow!` -> `anyhow!`

deps/verifier/Cargo.toml
fitzthum: See Ding's suggestions about the Cargo file as well.

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: Fixed

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: Added the empty vector check

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: Ran both the commands and the latest commit reflects the same

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: Fixed

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: fixed

Cargo.toml
AdithyaKrishnan: I'm able to resolve most comments except this. Could we please have a conversation on slack?

deps/verifier/src/snp/mod.rs
```diff
@@ -84,8 +92,14 @@ impl Verifier for Snp {
...
-        let Some(cert_chain) = cert_chain else {
-            bail!("Cert chain is unset");
+        let cert_chain = if let Some(chain) = cert_chain {
+            if chain.is_empty() {
+                fetch_vcek_from_kds(report)?
+            } else {
+                chain
+            }
+        } else {
+            fetch_vcek_from_kds(report)?
```
fitzthum, line 102: I think you can simplify this by doing something like 

```
let cert_chain = match cert_chain {
    Some(chain) if !chain.is_empty() => chain,
    _ => fetch_vcek_from_kds(report)?
}
```
Or something like that

AdithyaKrishnan: Fixed!

deps/verifier/src/snp/mod.rs
```diff
@@ -307,6 +313,41 @@ fn get_common_name(cert: &x509::X509) -> Result<String> {
     Ok(e.data().as_utf8()?.to_string())
 }
 
+// Function to request vcek from KDS asynchronously. Return vcek in der format.
```
Xynnn007, line 316: We can change `//` to `///`. This would automatically [generate a document](https://doc.rust-lang.org/rust-by-example/meta/doc.html#doc-comments) about the function when we publish the crate to `crates.io`.

Xynnn007 (approved):
Thanks @AdithyaKrishnan ! LGTM

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: Do you want me to add description to all functions in the file?

AdithyaKrishnan:
> Thanks @AdithyaKrishnan ! LGTM. Please see @fitzthum and @cclaudio 's comments.

@fitzthum and @cclaudio  - I believe I've fixed all your concerns. Please let me know if there are any more to be resolved. Further, per Ding's suggestion, I've added a third commit that adds on to the documentation by providing descriptions of all functions.
Hoping this PR can be approved and merged at the earliest.

fitzthum (approved):
LGTM

cclaudio (approved):
LGTM.  Thanks @AdithyaKrishnan

PR #556, opened 2024-11-04
Dockerfile: fix capitalization

fitzthum:
Docker prefers that statements using FROM and AS have those two words in the same case and will even warn you about it if you don't.

See https://docs.docker.com/reference/build-checks/from-as-casing/


bpradipt (approved):
/lgtm

PR #558, opened 2024-11-05
Update kbs-tools README.md

EmmEff:
Reflect recent project restructuring.

### DIFF
--- a/tools/kbs-client/README.md
+++ b/tools/kbs-client/README.md
@@ -23,7 +23,7 @@ We have a community version of kbs-client on [Github ORAS](https://github.com/co
 Build the client binary with support to the default features as:
 
 ```shell
-make -C ../../ cli
+make -C ../../kbs cli
 ```
 
 By default the client is built with support to the sample attester, apart from the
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
Thanks for this!

tools/kbs-client/README.md
```diff
@@ -32,13 +32,13 @@ require fewer dependencies and so usually handy for CI) then you can pass the
 `sample_only` feature as:
 
 ```shell
-make -C ../../ cli CLI_FEATURES=sample_only
+make -C ../../kbs cli CLI_FEATURES=sample_only
```
mythi, line 35: the change itself is OK but the comment above about `sample_only` looks outdated so we could update that too while we're at it. 

Default features come with `sample_only`. This section here should probably give instructions how to build a TEE specific flavor with: `make -C ../../kbs cli ATTESTER=<selected attester, e.g., tdx-attester`.

fitzthum: The command in the diff is correct. Describing how to build with different attesters seems beyond the scope of this PR even though it's a pretty simple change.

fitzthum (approved):
LGTM

tools/kbs-client/README.md
mythi: Indeed. I did not check the config before adding my comment. My recollection was that it was sample-only by default.

PR #561, opened 2024-11-06
kbs: ITA: Allow policy IDs in config file

jodh-intel:
If one or more policy IDs are specified in the config file, add these
to the attestation request and implicitly set `policy_must_match=true`.

```toml
[attestation_service]
type = "intel_ta"

policy_ids = ["aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa"]
```

Alternatively, specify multiple policy IDs like this:

```toml
policy_ids = [
    "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa",
    "bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb",

    # ...

    "nnnnnnnn-nnnn-nnnn-nnnn-nnnnnnnnnnnn",
]
```


kbs/docs/config.md
```diff
@@ -144,13 +144,14 @@ attestation. The following properties can be set.
...
+|--------------------------|--------------|------------------------------------------------------------------------------------------|----------|---------|
+| `timeout`                | Integer      | The maximum time (in minutes) between RCAR handshake's `auth` and `attest` requests      | No       | 5       |
+| `base_url`               | String       | Intel Trust Authority API URL.                                                           | Yes      | -       |
+| `api_key`                | String       | Intel Trust Authority API key.                                                           | Yes      | -       |
+| `certs_file`             | String       | URL to an Intel Trust Authority portal or path to JWKS file used for token verification. | Yes      | -       |
+| `allow_unmatched_policy` | Boolean      | Determines whether to ignore the `policy_ids_unmatched` token claim.                     | No       | false   |
+| `policy_ids`             | String array | List of one or more quoted and comma-separated policy IDs.                               | No       | `[]`    |
```
fitzthum, line 154: It looks like these are only used for ITA at least in the current implementation so maybe the scope of these configs should be ITA?

fitzthum:
This PR seems straightforward enough but there has been a little bit of back-and-forth about where policies should come from and I think we will probably need something more sophisticated. We have thought about doing something with the init-data so that policies could be selected at guest configuration time.

This is probably a fine place to start but we should think a little about how this works and will work for different types of attesters and tokens.

kbs/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -102,6 +106,13 @@ impl Attest for IntelTrustAuthority {
         })
         .to_string();
 
+        let policy_ids = self.config.policy_ids.clone();
+
+        let policy_must_match = match policy_ids.len() {
+            0 => false,
+            _ => true,
+        };
```
Xynnn007, line 114:
```suggestion
        let policy_must_match = !policy_ids.is_empty();
```

jodh-intel: Updated.

kbs/docs/config.md
mythi: That is already covered: we have sections for AS specific config clags: "When `type` is set to `intel_ta`, KBS will try to connect a remote Intel TA service for attestation. The following properties can be set...."

kbs/docs/config.md
```diff
@@ -144,13 +144,14 @@ attestation. The following properties can be set.
 
 >gRPC CoCo AS is available only when `coco-as-grpc` feature is enabled.
```
mythi, line 145: There's a minor glitch in the doc that could be addressed in a separate commit in this PR:
```suggestion
>Intel Trust Authority AS is available only when `intel-trust-authority-as` feature is enabled.
```

jodh-intel: Fixed.

mythi (approved):
LGTM now and CI seems happy also

Xynnn007 (approved):
LGTM

fitzthum (approved):
We may need to rework this slightly down the road but seems fine for now.

PR #562, opened 2024-11-06
Bump CSV crate

fitzthum:
With the new version of this crate I think we might not need to use the `--locked` flag when we built. Should we remove it?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -570,7 +570,7 @@ dependencies = [
  "az-tdx-vtpm",
  "base64 0.22.1",
  "codicon",
- "csv-rs",
+ "csv-rs 0.1.0 (git+https://github.com/openanolis/csv-rs?rev=b74aa8c)",
  "hex",
  "hyper 0.14.30",
  "hyper-tls 0.5.0",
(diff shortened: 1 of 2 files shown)


Xynnn007:
> With the new version of this crate I think we might not need to use the `--locked` flag when we built. Should we remove it?

I think so. If any new error we met during CI, we can go back to fix it.

Xynnn007 (approved):

fitzthum:
Well the tests are fine without `--locked`. I've seen some mixed advice about whether `--locked` should be used. I lean towards removing it. We are still tracking the Cargo files so people can use `--locked` if they need.

@larrydewey I seem to recall that we talked about this at some point. Do you have an opinion here?

@mkulke any thoughts?

mkulke:
> Well the tests are fine without `--locked`. I've seen some mixed advice about whether `--locked` should be used. I lean towards removing it. We are still tracking the Cargo files so people can use `--locked` if they need.
> 
> @larrydewey I seem to recall that we talked about this at some point. Do you have an opinion here?
> 
> @mkulke any thoughts?

afaiu --locked makes sense for release builds. otherwise there might be a different dependency resolution during the release build on the CI, that you could not reproduce easily.

fitzthum:
@mkulke ok well I'll drop that commit for now and we can add it back later if needed.

mkulke (approved):

PR #564, opened 2024-11-07
KBS: Fix config default value and docs

Xynnn007:
There are two commits, please see the concrete commit message of both.


fitzthum (approved):
`LGTM`

PR #565, opened 2024-11-07
attestation-service: Allow conditional compilation of some features

tylerfanelli:
The Makefile is a bit strict on conditional compilation. Add some conditional verifier compilation in `Cargo.toml` and introduce a new configurable `FEATURES` variable to specify feature compilation other than default features.

PR #571, opened 2024-11-11
AS: Add support for selectable verifier suites

### LINKED ISSUE #568: s390x artifact build jobs failing
We're running into this [error](https://github.com/confidential-containers/trustee/actions/runs/11726091988/job/32663922127)

```
#14 101.2 error: failed to run custom build command for `tss-esapi-sys v0.5.0`
#14 101.2 
#14 101.2 Caused by:
#14 101.2   process didn't exit successfully: `/usr/src/attestation-service/target/release/build/tss-esapi-sys-e17dce3a146ba30a/build-script-build` (exit status: 101)
#14 101.2   --- stderr
#14 101.2   thread 'main' panicked at /usr/local/cargo/registry/src/index.crates.io-d11c229612889eed/tss-esapi-sys-0.5.0/build.rs:35:17:
#14 101.2   Compilation target (architecture, OS) tuple (s390x, linux) is not part of the supported tuples. Please compile with the "generate-bindings" feature or add support for your platform :)
#14 101.2   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
#14 101.2 warning: build failed, waiting for other jobs to finish...
#14 140.4 error: failed to compile `attestation-service v0.1.0 (/usr/src/attestation-service/attestation-service)`, intermediate artifacts can be found at `/usr/src/attestation-service/target`.
#14 140.4 To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
#14 ERROR: process "/bin/sh -c cargo install --path attestation-service --bin grpc-as --features grpc-bin --locked" did not complete successfully: exit code: 101
```

Seems like it may have started on https://github.com/confidential-containers/trustee/pull/565 but not sure if it's related yet

Xynnn007:
Proposed in #565 by @tylerfanelli

We soon have two needs when building an AS binary / docker image to decouple verifier dependencies.
1. Select specific verifier suites when building AS binary
2. Install specific verifier software stack due to the selected verifier suites when build docker images

This patch accomplish 1, by disabling all verifiers in CoCoAS basic code and providing a `VERIFIER` env for makefile to specify the verifier suites.

For 2, we add `make dockerfile` function which also follows `VERIFIER` env to install related verifier software stack.
By default the generation of dockerfiles will use `all-verifier`.

We add target platform detecting logic to the KBS crate cargo toml to determine the built-in AS features.

Also, the CI pipeline of CoCoAS is updated.

Close #568

The RVPS features are a little messy. It will be tidied in #553

cc @tylerfanelli @fitzthum @mkulke

### DIFF
--- a/.github/workflows/push-as-image-to-ghcr.yml
+++ b/.github/workflows/push-as-image-to-ghcr.yml
@@ -19,16 +19,45 @@ jobs:
           - coco-as-grpc
           - coco-as-restful
           - rvps
+        verifier:
+          - all-verifier
+          - se-verifier
         include:
           - docker_file: attestation-service/docker/as-grpc/Dockerfile
             tag: coco-as-grpc
             name: gRPC CoCo-AS
+            verifier: all-verifier
+            instance: ubuntu-latest
+          - docker_file: attestation-service/docker/as-grpc/Dockerfile
+            tag: coco-as-grpc
+            name: gRPC CoCo-AS (IBM SE)
+            verifier: se-verifier
+            instance: s390x
           - docker_file: attestation-service/docker/as-restful/Dockerfile
             tag: coco-as-restful
             name: RESTful CoCo-AS
+            verifier: all-verifier
+            instance: ubuntu-latest
+          - docker_file: attestation-service/docker/as-restful/Dockerfile
+            tag: coco-as-restful
+            name: RESTful CoCo-AS (IBM SE)
+            verifier: se-verifier
+            instance: s390x
           - docker_file: rvps/docker/Dockerfile
             tag: rvps
             name: RVPS
+            verifier: all-verifier
+            instance: ubuntu-latest
+          - docker_file: rvps/docker/Dockerfile
+            tag: rvps
+            name: RVPS
+            verifier: se-verifier
+            instance: s390x
+        exclude:
+          - instance: ubuntu-latest
+            verifier: se-verifier
+          - instance: s390x
+            verifier: all-verifier
     runs-on: ${{ matrix.instance }}
 
     steps:
@@ -50,6 +79,7 @@ jobs:
         commit_sha=${{ github.sha }}
         arch=$(uname -m)
         DOCKER_BUILDKIT=1 docker build -f "${{ matrix.docker_file }}" --push --build-arg ARCH="${arch}" \
+          --build-arg VERIFIER="${{ matrix.verifier }}" \
           -t "ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:${commit_sha}-${arch}" \
           -t "ghcr.io/confidential-containers/staged-images/${{ matrix.tag }}:latest-${arch}" .
 
--- a/attestation-service/Makefile
+++ b/attestation-service/Makefile
@@ -8,14 +8,24 @@ BIN_NAMES := grpc-as restful-as
 DEBUG ?=
 DESTDIR ?= $(PREFIX)/bin
 
-FEATURES ?=
+VERIFIER ?= all-verifier
 
-ifdef FEATURES
-	OPTIONAL_FEATURES := ,$(FEATURES)
-	default-features := --no-default-features
+RVPS_GRPC := true
+
+# TODO: Remove `RVPS_BUILTIN`
+# when https://github.com/confidential-containers/trustee/pull/553 gets merged
+# Here we also declare another variable `RVPS_FEATURES1` because a blank will
+# be added when doing '+=' operation in Makefile
+RVPS_BUILTIN := true
+
+ifeq ($(RVPS_GRPC), true)
+	RVPS_FEATURES1 := rvps-grpc
+endif
+
+ifeq ($(RVPS_BUILTIN), true)
+	RVPS_FEATURES := $(RVPS_FEATURES1),rvps-builtin
 else
-	OPTIONAL_FEATURES :=
-	default-features :=
+	RVPS_FEATURES := $(RVPS_FEATURES1)
 endif
 
 ifdef DEBUG
@@ -29,10 +39,10 @@ endif
 build: grpc-as restful-as
 
 grpc-as:
-	cargo build --bin grpc-as $(release) $(default-features) --features grpc-bin$(OPTIONAL_FEATURES)
+	cargo build --bin grpc-as $(release) --features grpc-bin,$(VERIFIER),$(RVPS_FEATURES)
 
 restful-as:
-	cargo build --bin restful-as $(release) $(default-features) --features restful-bin$(OPTIONAL_FEATURES)
+	cargo build --bin restful-as $(release) --features restful-bin,$(VERIFIER),$(RVPS_FEATURES)
 
 install:
 	for bin_name in $(BIN_NAMES); do \
--- a/attestation-service/docker/as-grpc/Dockerfile
+++ b/attestation-service/docker/as-grpc/Dockerfile
@@ -4,6 +4,7 @@
 
 FROM rust:latest AS builder
 ARG ARCH=x86_64
+ARG VERIFIER=all-verifier
 
 WORKDIR /usr/src/attestation-service
 COPY . .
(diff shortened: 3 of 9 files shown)


mkulke (commented):
I wonder whether we can get avoid the dockerfile templating: would something like this work:

```Dockerfile
# Define the build argument with a default value
ARG INSTALL_EXTRA_DEPS=false

# Use the argument in a conditional statement within the RUN command
RUN if [ "$INSTALL_EXTRA_DEPS" = "true" ]; then \
      echo "Installing extra dependencies" && \
      apt-get update && \
      apt-get install -y some-extra-package; \
    else \
      echo "Skipping extra dependencies"; \
    fi
```

Xynnn007:
@mkulke all work but not for `COPY` command for tss libraries [here](https://github.com/confidential-containers/trustee/blob/main/attestation-service/docker/as-grpc/Dockerfile#L39).

If we add extra `RUN rm ...` the image will still be huge because the original layer exists.

I did not have a good way for this. Do you have any ideas / Is it a reason?

attestation-service/Makefile
```diff
@@ -29,15 +28,24 @@ endif
 build: grpc-as restful-as
 
 grpc-as:
-	cargo build --bin grpc-as $(release) $(default-features) --features grpc-bin$(OPTIONAL_FEATURES)
+	cargo build --bin grpc-as $(release) $(default-features) --features grpc-bin,$(VERIFIER)
 
 restful-as:
-	cargo build --bin restful-as $(release) $(default-features) --features restful-bin$(OPTIONAL_FEATURES)
+	cargo build --bin restful-as $(release) $(default-features) --features restful-bin,$(VERIFIER)
```
tylerfanelli, line 34: You removed the definition of `default-features`. If `FEATURES` were specified, then `--no-default-features` was enabled as well. [`all-verifier` is a part of the default features](https://github.com/confidential-containers/trustee/blob/fafa3412bb34e93a7cbf50d51c4dca824fa771a4/attestation-service/Cargo.toml#L7), so without `--no-default-features` enabled, all of the verifiers will be enabled regardless of what is specified in `VERIFIERS`.

tylerfanelli (changes requested):

attestation-service/Makefile
```diff
@@ -8,15 +8,14 @@ BIN_NAMES := grpc-as restful-as
...
-FEATURES ?=
+VERIFIER ?= all-verifier
 
-ifdef FEATURES
-	OPTIONAL_FEATURES := ,$(FEATURES)
-	default-features := --no-default-features
-else
-	OPTIONAL_FEATURES :=
-	default-features :=
-endif
```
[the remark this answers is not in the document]
Xynnn007: Yea, I had thought about this by checking inside the [existing features](https://github.com/confidential-containers/trustee/blob/main/attestation-service/Cargo.toml#L7-L27). They are three types
1. verifiers: Specify which platform can be supported to verify
2. `rvps-*`: indicate how to integrate rvps. (grpc or builtin)
3. `*-bin`: `grpc` or `restful` extra dependencies of the binary

1 must be included to Makefile env.

3 is determined by the target binary to be built. If `grpc-bin` is enabled when building `restful-bin`, the binary size will not grow. Thus it is not needed to be included in the `Makefile` env.

What confuses is 2. There are two ways. The PR uses the relatively simplest way that by default both `rvps-grpc` and `rvps-builtin` are enabled. Another way is to add an extra flag to control the enablement of `rvps-grpc` and `rvps-builtin`. Seeing that `cargo build` could perform also any feature composition, I chose the way mentioned.

Wdyt?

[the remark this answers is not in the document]
tylerfanelli: > Yea, I had thought about this by checking inside the [existing features](https://github.com/confidential-containers/trustee/blob/main/attestation-service/Cargo.toml#L7-L27). They are three types
> 
> 1. verifiers: Specify which platform can be supported to verify

I think this is covered here. Either you specify the verifiers you want with` VERIFIER=` or it defaults to `all-verifier`. This sounds good to me.

> 2. `rvps-*`: indicate how to integrate rvps. (grpc or builtin)
> 3. `*-bin`: `grpc` or `restful` extra dependencies of the binary
> 
> 1 must be included to Makefile env.
> 
> 3 is determined by the target binary to be built. If `grpc-bin` is enabled when building `restful-bin`, the binary size will not grow. Thus it is not needed to be included in the `Makefile` env.
> 
> What confuses is 2. There are two ways. The PR uses the relatively simplest way that by default both `rvps-grpc` and `rvps-builtin` are enabled. Another way is to add an extra flag to control the enablement of `rvps-grpc` and `rvps-builtin`. Seeing that `cargo build` could perform also any feature composition, I chose the way mentioned for ease.
> 
> Wdyt?

There's no reason someone would want both **rvps-grpc** and **rvps-builtin**, correct? You could add a `RVPS=` variable, in which a user would specify:

```
RVPS=grpc make...
RVPS=builtin make...
```

And it defaults to one of those options...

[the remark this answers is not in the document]
Xynnn007: > There's no reason someone would want both **rvps-grpc** and **rvps-builtin**, correct? They're mutually exclusive. You could add a `RVPS=` variable, in which a user would specify:

This is only the build time choosing. The concrete behavior whether `grpc` or `builtin` would be used is determined by [the config file](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/config.md#rvps-configuration) **at runtime**. I know that 
in customized scenarios, broad-spectrum support is not required, only one-to-one runtime requirements are supported.

Another flag is also good to me. I still want to listen to your ideas with my comments : )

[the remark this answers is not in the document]
Xynnn007: I have a plan to merge `rvps-builtin` feature into AS code and deprecate the feature as 1) make the configs dependency relationship clear and 2) it does not imports many new crates in #553. 

Then, based on your suggestion a flag named `GRPC_RVPS` can be added to Makefile to indicate whether `rvps_grpc` is enabled when building `restful-as`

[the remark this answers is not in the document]
tylerfanelli: Great, let me know when this is ready for another review.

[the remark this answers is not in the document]
Xynnn007: Thanks. I have pushed the changes with a TODO indicating what is going on after this PR

attestation-service/docker/as-grpc/Dockerfile
```diff
@@ -17,26 +18,37 @@ RUN if [ "${ARCH}" = "x86_64" ]; then curl -L https://download.01.org/intel-sgx/
     apt-get update && apt-get install -y libsgx-dcap-quote-verify-dev; fi
 
 # Build and Install gRPC attestation-service
-RUN cargo install --path attestation-service --bin grpc-as --features grpc-bin --locked
+RUN cargo install --path attestation-service --bin grpc-as --features grpc-bin,${VERIFIER} --locked && \
+    strip /usr/local/cargo/bin/grpc-as
```
mkulke, line 22: why do we strip the binary?

attestation-service/docker/as-restful/Dockerfile
```diff
@@ -17,25 +18,35 @@ RUN if [ "${ARCH}" = "x86_64" ]; then curl -L https://download.01.org/intel-sgx/
     apt-get update && apt-get install -y libsgx-dcap-quote-verify-dev; fi
 
 # Build and Install RESTful attestation-service
-RUN cargo install --path attestation-service --bin restful-as --features restful-bin --locked
+RUN cargo install --path attestation-service --bin restful-as --features restful-bin,${VERIFIER} --locked && \
+    strip /usr/local/cargo/bin/restful-as
```
mkulke, line 22: why do we strip the binary?

.github/workflows/push-as-image-to-ghcr.yml
```diff
@@ -19,16 +19,40 @@ jobs:
           - coco-as-grpc
           - coco-as-restful
           - rvps
+        verifier:
+          - all-verifier
```
mkulke, line 23: `all-verifier` should probably be plural `all-verifiers`, or just `all`?

deps/verifier/src/lib.rs
```diff
@@ -181,7 +181,7 @@ pub trait Verifier {
 }
 
 /// Padding or truncate the given data slice to the given `len` bytes.
-fn regularize_data(data: &[u8], len: usize, data_name: &str, arch: &str) -> Vec<u8> {
+pub fn regularize_data(data: &[u8], len: usize, data_name: &str, arch: &str) -> Vec<u8> {
```
mkulke, line 184: is this related to this PR?

attestation-service/docker/as-grpc/Dockerfile
Xynnn007: Yea to make the binary smaller as more layers are added to the dockerfile

attestation-service/docker/as-restful/Dockerfile
Xynnn007: Yea to make the binary smaller as more layers are added to the dockerfile

.github/workflows/push-as-image-to-ghcr.yml
Xynnn007: This is a typo for a long time and exists in `verifier` and `attestation-service`. like https://github.com/confidential-containers/trustee/blob/main/deps/verifier/Cargo.toml#L8

deps/verifier/src/lib.rs
Xynnn007: Yes as I delete the `all-attester` from default feature, by default nothing will import this code so the lint would warn. By making it `pub` the lint can be ok

attestation-service/docker/as-grpc/Dockerfile
mkulke: hmm, symbols are useful/required for debugging. by default, in an unconstrained environment I wouldn't strip release binaries. If we still want to do that, we can [put it in Cargo.toml](https://doc.rust-lang.org/cargo/reference/profiles.html#strip), in which we can pick more granular size optimizations and it's a bit less obscure.

But again, for a webservice, I wouldn't optimize for size without a good reason to do so.

Xynnn007: Ok. So let me delete the strip here and if needs we can optimize the size later

mkulke (approved):

Xynnn007:
hi @tylerfanelli ptal if it looks good to you

Xynnn007:
Well. Based on [suggestion](https://cloud-native.slack.com/archives/C039JSH0807/p1731576810749239?thread_ts=1731574255.883349&cid=C039JSH0807) from @mythi , let me add another commit to update docker build CI to run on also s390x runner. cc @BbolroC

.github/workflows/push-as-image-to-ghcr.yml
```diff
@@ -19,16 +19,40 @@ jobs:
...
+          - se-verifier
         include:
```
BbolroC, line 25: We should add the following:

```diff
+        exclude:
+          - instance: ubuntu-latest
+            verifier: se-verifier
+          - instance: s390x
+            verifier: all-verifier
```

BbolroC (changes requested):
Please, update what I've mentioned. Otherwise, the workflow `push-as-image-to-ghcr.yaml` will trigger unnecessary 6 failing jobs.

.github/workflows/push-as-image-to-ghcr.yml
Xynnn007: Oh, thanks for the TIPs! I forced-push the commit, please take a look if it is right

BbolroC (approved):
LGTM thanks @Xynnn007 !

fitzthum (approved):
LGTM. Let's merge this so we can try to sneak it into the Kata release

PR #575, opened 2024-11-13
kbs: Avoid logging sensitive information

fidencio:
In a few parts of our code, we'd end up logging sensitive information when the KBS is running in "debug" mode.

A simple way to avoid doing that is taking advantage of the "derivative" crate, and simply omit the sensitive fields, which is the approach taken.

### DIFF
--- a/kbs/src/bin/kbs.rs
+++ b/kbs/src/bin/kbs.rs
@@ -23,7 +23,7 @@ async fn main() -> Result<()> {
     info!("Using config file {}", cli.config_file);
     let kbs_config = KbsConfig::try_from(Path::new(&cli.config_file))?;
 
-    debug!("Config: {:#?}", kbs_config);
+    debug!("Config (sensitive fields are omitted): {:#?}", kbs_config);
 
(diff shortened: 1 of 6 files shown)


kbs/src/bin/kbs.rs
```diff
@@ -23,7 +23,7 @@ async fn main() -> Result<()> {
     info!("Using config file {}", cli.config_file);
     let kbs_config = KbsConfig::try_from(Path::new(&cli.config_file))?;
 
-    debug!("Config: {:#?}", kbs_config);
+    debug!("Config (senstive fields are omitted): {:#?}", kbs_config);
```
fitzthum, line 26: nit: `senstive` -> `sensitive`.

fitzthum (approved):
one little spelling issue otherwise lgtm

mythi (approved):
+1

kbs/src/bin/kbs.rs
fidencio: Ooops, thanks, fixed now.

PR #576, opened 2024-11-15
gha: Add image build check for s390x

BbolroC:
Issue #568 highlights the need for an image build check for s390x, similar to what we have for x86_64. 
This PR addresses the need by:
- Extracting the image build steps from the push-{kbs,as}-image-to-ghcr workflow into separate workflows
- Configuring these workflows to run on PR events or after merging a PR

Notable change:
- On merge, the workflow now pushes `ghcr.io/confidential-containers/staged-images/rhel-ubi`

The changes on PR events have been verified with:
- KBS: https://github.com/BbolroC/trustee/actions/runs/11840113329
- AS: https://github.com/BbolroC/trustee/actions/runs/11839493749

The changes on merge needs to be verified after merging this PR. I will keep an eye on how it goes and bring things up immediately when something goes wrong. Thanks.

### DIFF
--- a/.github/workflows/as-docker-build.yml
+++ b/.github/workflows/as-docker-build.yml
@@ -1,4 +1,4 @@
-name: AS & RVPS Container Image Build
+name: AS/RVPS Container Image Build
 on:
   push:
     branches:
@@ -15,25 +15,7 @@ on:
   create:
 
 jobs:
-  basic_ci:
+  check_as_image_build:
     if: github.event_name == 'pull_request' || github.event_name == 'push'
-    name: Check
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-
-    steps:
-    - name: Code checkout
-      uses: actions/checkout@v4
-    
-    - name: Build gRPC AS Container Image
-      run: |
-        DOCKER_BUILDKIT=1 docker build -t attestation-service:latest . -f attestation-service/docker/as-grpc/Dockerfile
-    
-    - name: Build RESTful AS Container Image
-      run: |
-        DOCKER_BUILDKIT=1 docker build -t attestation-service:latest . -f attestation-service/docker/as-restful/Dockerfile
-
-    - name: Build RVPS Container Image
-      run: |
-        Docker_BUILDKIT=1 docker build -t rvps:latest . -f rvps/docker/Dockerfile
+    uses: ./.github/workflows/build-as-image.yml
+    secrets: inherit
--- a/.github/workflows/build-as-image.yml
+++ b/.github/workflows/build-as-image.yml
@@ -1,0 +1,85 @@
+name: Build CoCo AS/RVPS Image
+
+on:
+  workflow_call:
+    inputs:
+      build_option:
+        description: 'Build option for the image'
+        type: string
+        required: false
+
+jobs:
+  build_as_image:
+    strategy:
+      fail-fast: false
+      matrix:
+        instance:
+          - ubuntu-latest
+          - s390x
+        tag:
+          - coco-as-grpc
+          - coco-as-restful
+          - rvps
+        verifier:
+          - all-verifier
+          - se-verifier
+        include:
+          - docker_file: attestation-service/docker/as-grpc/Dockerfile
+            tag: coco-as-grpc
+            name: gRPC CoCo-AS
+            verifier: all-verifier
+            instance: ubuntu-latest
+          - docker_file: attestation-service/docker/as-grpc/Dockerfile
+            tag: coco-as-grpc
+            name: gRPC CoCo-AS (IBM SE)
+            verifier: se-verifier
+            instance: s390x
+          - docker_file: attestation-service/docker/as-restful/Dockerfile
+            tag: coco-as-restful
+            name: RESTful CoCo-AS
+            verifier: all-verifier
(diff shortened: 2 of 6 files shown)


.github/workflows/build-as-image.yml
```diff
@@ -0,0 +1,84 @@
+name: Build CoCo AS/RVPS Image
```
Xynnn007, line 1: The name here is different from `as-docker-build.yml`'s. What would this cause? idk do we need to make them the same?

Xynnn007 (approved):
Thanks @BbolroC ! LGTM, only a name thing I am not quite sure

.github/workflows/build-as-image.yml
BbolroC: Hmm.. To my best knowledge, a title of the called workflow does not appear on an action section. I prefer to having a different title for each workflow. 😉

.github/workflows/as-docker-build.yml
```diff
@@ -15,25 +15,9 @@ on:
...
-      run: |
-        DOCKER_BUILDKIT=1 docker build -t attestation-service:latest . -f attestation-service/docker/as-restful/Dockerfile
-
-    - name: Build RVPS Container Image
-      run: |
-        Docker_BUILDKIT=1 docker build -t rvps:latest . -f rvps/docker/Dockerfile
\ No newline at end of file
+    uses: ./.github/workflows/build-as-image.yml
+    with:
+      build_option:
```
mkulke, line 22: Can we make this optional on the build-as-image workflow, so we don't have to specify an empty build_option?

.github/workflows/build-as-image.yml
```diff
@@ -0,0 +1,84 @@
...
+          - ubuntu-latest
+          - s390x
+        tag:
+          - coco-as-grpc
+          - coco-as-restful
+          - rvps
+        verifier:
+          - all-verifier
+          - se-verifier
+        include:
```
mkulke, line 25: this looks like we're enumerating every combination, doesn't it work like this?

since we have 1:1 mapping from name to docker_file + tag

and we also have a 1:1 mapping from instance to verifier

```yaml
matrix:
  name:
  - gRPC CoCo-AS
  - ...
  instance:
  - ubuntu-latest
  - s390x
  include:
  - name: gRPC CoCo-AS
    docker_file: attestation-service/docker/as-grpc/Dockerfile
    tag: coco-as-grpc
  - name: ...
 - instance: s390x
   verifier: se-verifier
 - instance: ubuntu-latest
   verifier: all-verifier   
```

.github/workflows/kbs-docker-build.yml
```diff
@@ -6,16 +6,8 @@ on:
...
-    - name: Build KBS Container Image
-      run: |
-        DOCKER_BUILDKIT=1 docker build -t kbs:coco-as . -f kbs/docker/Dockerfile; \
-        DOCKER_BUILDKIT=1 docker build -t kbs:coco-as-grpc . -f kbs/docker/coco-as-grpc/Dockerfile; \
-        DOCKER_BUILDKIT=1 docker build -t kbs:coco-as-rhel-ubi . -f kbs/docker/rhel-ubi/Dockerfile; \
-        DOCKER_BUILDKIT=1 docker build -t kbs:coco-as-ita . -f kbs/docker/intel-trust-authority/Dockerfile
+  check_kbs_image_build:
+    uses: ./.github/workflows/build-kbs-image.yml
+    with:
+      build_option:
```
mkulke, line 12: same here, let's make the build_option not required, so we can skip this

.github/workflows/as-docker-build.yml
BbolroC: Yes, we could make it optional. I will update it.

.github/workflows/kbs-docker-build.yml
BbolroC: Yes.

.github/workflows/build-as-image.yml
BbolroC: It works like https://github.com/BbolroC/trustee/actions/runs/11839493749

BbolroC: And your suggestion does not work as expected because:
- These explicitly added `include` combinations do not negate the `exclude` entries
- This still allows all possible combinations of instance and verifier to remain valid, whereas the current configuration explicitly removes some

mkulke: yeah the idea is that it'll produce the same output, but with less repitition

you wouldn't need an exclude section because "verifier" is not part of the matrix, the only matrix is:

$instance * $name (so 2 * 3 options)

the "include" statement just adds additional attributes to valid combinations

mkulke: since this code is just moved around in this PR we can leave it in place here

mkulke (approved):

BbolroC:
It seems that two workflows (e.g., one triggered on merge and another on PR events) are being executed when the PR is merged. This results in a duplicate image build step, similar to what we observed in the guest-components repository.

I will create a follow-up PR similar to the following to deduplicate the image build process:
- https://github.com/confidential-containers/guest-components/pull/644
- https://github.com/confidential-containers/guest-components/pull/648

Thanks.

PR #577, opened 2024-11-15
AS: fix restful-as dockerfile

Xynnn007:
VERIFIER is not claimed as an ARG thus not take effection in the following building steps.


BbolroC (approved):
LGTM, thanks @Xynnn007 !

PR #580, opened 2024-11-15
CI: consolidate matrix configurations

mkulke:
The current matrix configurations are more like an manual enumeration of build options expressed in verbose matrix/include/exclude statement.

The PR changes this into a simple matrix of the build and then use include to add required auxiliary fields to each of the valid matrix combinations.

(the kbs workflow already follows the same scheme)


BbolroC (approved):
It looks much nicer and more readable, thanks @mkulke !

fitzthum (approved):
LGTM

PR #581, opened 2024-11-15
gha: Skip duplicate image build on merge

BbolroC:
The following workflows are triggered twice during a merge:

- build-kbs-image.yml
- build-as-image.yml

For self-hosted runners, which often have a limited number of available runners, this duplicate build is unnecessary and may cause CI bottlenecks.

This PR ensures the workflow of interest is skipped when `github.event_name` is `push`, which occurs during a merge.

### DIFF
--- a/.github/workflows/as-docker-build.yml
+++ b/.github/workflows/as-docker-build.yml
@@ -1,12 +1,5 @@
 name: AS/RVPS Container Image Build
 on:
-  push:
-    branches:
-      - "main"
-    paths:
-      - 'attestation-service/**'
-      - '.github/workflows/as-docker-build.yml'
-      - 'Cargo.toml'
   pull_request:
     paths:
       - 'attestation-service/**'
@@ -16,6 +9,5 @@ on:
 
 jobs:
   check_as_image_build:
-    if: github.event_name == 'pull_request' || github.event_name == 'push'
     uses: ./.github/workflows/build-as-image.yml
     secrets: inherit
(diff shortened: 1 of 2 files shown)


BbolroC:
Workflow examples for the duplicated build:

1. KBS
- https://github.com/confidential-containers/trustee/actions/runs/11855132932
- https://github.com/confidential-containers/trustee/actions/runs/11855132934

2. AS
- https://github.com/confidential-containers/trustee/actions/runs/11855132938
- https://github.com/confidential-containers/trustee/actions/runs/11855132944

.github/workflows/as-docker-build.yml
```diff
@@ -16,6 +16,6 @@ on:
 
 jobs:
   check_as_image_build:
-    if: github.event_name == 'pull_request' || github.event_name == 'push'
+    if: github.event_name != 'push'
```
mkulke, line 19: why don't we remove `push.branches` form the workflow instead?

.github/workflows/kbs-docker-build.yml
```diff
@@ -7,5 +7,6 @@ on:
 
 jobs:
   check_kbs_image_build:
+    if: github.event_name != 'push'
```
mkulke, line 10: see comment above

.github/workflows/as-docker-build.yml
BbolroC: I guess it would have the same effect. Let's give it a try! Thanks for the comment. 😉

.github/workflows/kbs-docker-build.yml
BbolroC: yep!

fitzthum (approved):
LGTM

PR #539, opened 2024-10-19
Add nebula_ca plugin

cclaudio:
This PR adds the nebula_ca plugin.

## How to build and start the plugin

Build trustee with the 'nebula-ca-plugin' cargo feature enabled and build a KBS image that has the nebula-cert binary
```bash
docker compose build --build-arg NEBULA_CA_PLUGIN=true --no-cache
```
Add the lines below to `kbs/config/docker-compose/kbs-config.toml`
```toml
    [[plugins]]
    name = "nebula-ca"
```
Start trustee
```bash
docker compose up
```
## How to test the credential service

Ultimately, the plugin will be called from the CDH (https://github.com/confidential-containers/guest-components/pull/763). In the meantime, we can build and use the kbs-client I have in the branch https://github.com/cclaudio/trustee/tree/nebula-ca-plugin-test

```bash
$ git clone https://github.com/cclaudio/trustee.git -b nebula-ca-plugin-test
$ cd trustee/kbs && make cli ATTESTER=snp-attester
$ sudo make install-cli
```

Request a Nebula credential

```bash
$ kbs-client config --auth-private-key kbs/config/private.key set-resource-policy --policy-file  kbs/sample_policies/allow_all.rego
$ kbs-client get-resource --plugin-name "nebula-ca" --resource-path "credential?name=podA&ip=10.9.8.2/21" | base64 -d
```
That should return a CredentialServiceOut structure like:
```json
{
[...]

PR #585, opened 2024-11-19
Add keyflux plugin

salmanyam:
This plugin generates credentials (keys and certificates) for both the API proxy server (required for
kata-containers/kata-containers#9159 and
kata-containers/kata-containers#9752) and the workload owner. This plugin also delivers the credentials to a sandbox (i.e., confidential PODs or VMs), specifically to the kata agent to initiate the SplitAPI proxy server so that a workload owner can communicate with the proxy server using a secure tunnel.

The IPv4 address, name, and the ID of the sandbox (i.e., pod) must be provided in the query string to obtain the credentials from the KBS.

After receiving the credential request, the splitapi plugin will create a key pair for the server and client and sign them using the self-signed CA. The generated credentials are stored in a hashmap with a unique key for each sandbox based on its name, ID, and IP address. SplitAPI plugin responds to a request from the sandbox by sending the server specific credentials (key, cert) along with the CA certificate. A request from workload owner gets the client specific credentials (key, cert) and CA's certificate.

The splitapi plugin itself is not built or initialized by default. To build or initialize it, following steps need to be followed.

# How to build
1. Enable `splitapi-plugin` feature while building `kbs`.
2. Update the `kbs/config/kbs-config.toml` to enable the `splitapi` plugin.

```
$ make background-check-kbs POLICY_ENGINE=opa SPLITAPI_PLUGIN=true
$ cat >> kbs/config/kbs-config.toml << EOF
[[plugins]]
name = "pki_vault"
#plugin_dir = "/opt/confidential-containers/kbs/plugin/splitapi"
#cred_filename = "certificates.json"
[plugins.pkivault_cert_details]
country = "AA"
#state = "Default State"
#locality = "Default City"
organization = "Default Organization"
org_unit = "Default Unit"

[plugins.pkivault_cert_details.ca]
#common_name = "grpc-tls CA"
validity_days = 3650

[plugins.pkivault_cert_details.server]
#common_name = "server"
#validity_days = 180

[plugins.pkivault_cert_details.client]
common_name = "client"
validity_days = 180
```
 > **Note:** Only `name` is mandatory for the plugin configuration. The rest fields will get default values in case those are not configured in the `kbs-config.toml`.

# How to test
This plugin can be tested similarly to the `nebula-ca` plugin (#539).  We need the kbs-client patch available in the branch https://github.com/cclaudio/trustee/tree/nebula-ca-plugin-test to build the kbs-client and test the plugin.

Once the `kbs-client` is built, use the `kbs-client` to make a request to KBS for generating and providing the server specific credentials.

```
$ cd kbs && make cli ATTESTER=snp-attester
$ sudo make install-cli
$ kbs-client --url http://127.0.0.1:8080 get-resource --plugin-name "pki_vault" --resource-path "credential?id=3367353&ip=60.11.12.48&name=pod33" | base64 -d
```
KBS should return the server specific credentials such as server key, server certificates, and the certificate of the CA.


bpradipt:
> This plugin generates credentials (keys and certificates) for both the API proxy server (required for [kata-containers/kata-containers#9159](https://github.com/kata-containers/kata-containers/pull/9159) and [kata-containers/kata-containers#9752](https://github.com/kata-containers/kata-containers/pull/9752)) and the workload owner. This plugin also delivers the credentials to a sandbox (i.e., confidential PODs or VMs), specifically to the kata agent to initiate the SplitAPI proxy server so that a workload owner can communicate with the proxy server using a secure tunnel.
> 
> The IPv4 address, name, and the ID of the sandbox must be provided in the query string to obtain the credential resources from the kbs.
> 
> After receiving the credential request, the splitapi plugin will create a key pair for the server and client and sign them using the self-signed CA. The generated ca.crt, server.crt, and server.key are stored in a directory specific to the sandbox (the caller) and returned to the caller. In addition, ca.key, client.key, and client.crt are also generated and stored to that particular directory specific to the sandbox (i.e., caller).
> 
> During the credential generation, a sandbox directory mapper creates a unique directory specific to the sandbox (i.e., the caller). The mapper creates the unique directory using the sandbox parameters passed in the query string. A mapping file is also maintained to store the mapping between the sandbox name and the unique directory created for the sandbox.
> 
> The splitapi plugin itself is not initialized by default. To initialize it, you need to add 'splitapi' in the kbs-config.toml.

A generic question. Would it make sense to have a generic plugin to create key-pair for a sandbox and not club it with splitapi?
There are other use cases for per sandbox (pod or VM) keys especially in the peer-pods case.
cc @yoheiueda @davidhadas

fitzthum:
> A generic question. Would it make sense to have a generic plugin to create key-pair for a sandbox and not club it with splitapi?

Some kind of PKI keypair would be very useful.

salmanyam:
> A generic question. Would it make sense to have a generic plugin to create key-pair for a sandbox and not club it with splitapi?
> There are other use cases for per sandbox (pod or VM) keys especially in the peer-pods case.

We are good. I think it would be a good idea to have the key generation as a basic service needed by other plugins. We can definitely consider refactoring and making it as a basic service. In that case, the splitapi plugin will utilize this basic service.

yoheiueda:
A generic PKI approach will be very helpful for our peer pod use cases.

I I understand correctly, the current proposal is to run a key generation service in a trustee, but a PKI approach sounds to me  a signing service in a trustee.  

Trustee has a private key for CA, and accepts key signing requests from clients. A client in a sandbox generates a key pair, and send a key signing request with the generated public key. The CA in Trustee issues a certificate by singing the received public key using the CA private key, and distribute the issued certificate. I think this is how the current PKI for TLS certificates works, and the sandbox does not need to reveal its private key to Trustee.

I think this will work If the purpose of the certificate generation is TLS secure tunneling.

This is just out of my curiosity, and  a key generation service will still very helpful for our peer pod use case.

Another point is that the current proposal generates RSA keys. To support other encryption mechanisms such as WireGuard and SSH, it is helpful to support different types of keys such as Ed25529. WireGuard uses Curve25529 keys, and Ed25529 is also prefabbed for SSH these days.

kbs/src/plugins/implementations/splitapi/manager.rs
```diff
@@ -0,0 +1,163 @@
...
+        if let Some(mapper) = mapper_guard.as_mut() {
+            let sandbox_dir_info: SandboxDirectoryInfo;
+
+            if let Some(existing_dir) = mapper.get_directory(&params.name) {
+            
+                log::info!("Found existing directory: {:?}", existing_dir.sandbox_dir());
+                sandbox_dir_info = existing_dir.clone();
+
+                //TODO: check if the credentails are already in there
+                // send the existing credentials if they are not expired
```
fitzthum, line 122: When you create the certs, just store the expiration time alongside them so you can easily see if they are expired.

kbs/src/plugins/implementations/splitapi/mapper.rs
fitzthum, line 1: I think it's a mistake to store all this stuff on the filesystem.

Here's what I would do. Create a struct that contains all the state that represents a keypair. OpenSSL has structs for all the things you need. Store this struct in a dictionary keyed by the connection id. Wrap the struct in a RwLock. If you need some kind of persistence you can serialize/desrialize the dictionary as needed.

I don't see any reason to be writing things to the filesystem and I think this entire mapper concept can be replaced by something from std, which decreases your complexity significantly.

kbs/src/plugins/implementations/splitapi/manager.rs
```diff
@@ -0,0 +1,163 @@
...
+    async fn get_server_credential(&self, params: &SandboxParams) -> Result<Vec<u8>> {
+        // Try locking the sandbox directory mapper
+        let mut mapper_guard = self.mapper.lock().map_err(|e| {
+            anyhow!("Failed to lock sandbox directory mapper: {}", e)
+        })?;
+
+        if let Some(mapper) = mapper_guard.as_mut() {
+            let sandbox_dir_info: SandboxDirectoryInfo;
+
+            if let Some(existing_dir) = mapper.get_directory(&params.name) {
```
fitzthum, line 116: These params are sent in other the network in plaintext. Would it be a problem if someone tampered with the name value in flight?

More fundamentally, why does this need to be stateful at all?

fitzthum (commented):
Made a few comments. Please run `cargo fmt` and `cargo clippy` to fix the CI.

kbs/src/plugins/implementations/splitapi/backend.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub struct SplitAPI {
+    pub backend: Arc<dyn SplitAPIBackend>,
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[serde(tag = "type")]
+pub enum SplitAPIConfig {
+    CertManager(manager::SplitAPIRepoDesc),
+}
```
Xynnn007, line 30: If we use `enum` here, we are hinting that potentially we could have other types of `SplitAPIConfig` besides `CertManager`. Do we? If not, I suggest to change `SplitAPIConfig` into `struct` and move members of `SplitAPIRepoDesc` directly into `SplitAPIConfig`

kbs/src/plugins/implementations/splitapi/manager.rs
```diff
@@ -0,0 +1,163 @@
...
+
+
+pub const DEFAULT_PLUGIN_DIR: &str = "/opt/confidential-containers/kbs/plugin/splitapi";
+pub const SANDBOX_DIRECTORY_MAPPING_FILENAME: &str = "sandbox-credential-mapping.json";
+
+
+// Use lazy_static to initialize the SANDBOX_DIRECTORY_MANAGER only once
+lazy_static! {
+    static ref SANDBOX_DIRECTORY_MAPPER: Arc<Mutex<Option<SandboxDirectoryMapper>>> = Arc::new(Mutex::new(None));
+}
```
Xynnn007, line 27: Could we move this static variable directly into `CertManager` as a member? s.t.
```rust
pub struct CertManager {
    pub plugin_dir: String,
    pub mapping_filename: String,
    mapper: Arc<Mutex<SandboxDirectoryMapper>>,
}
```

Also, could we use [tokio::sync::Mutex](https://docs.rs/tokio/1.41.1/tokio/sync/struct.Mutex.html#examples) rather than `std::sync::Mutex`? The async version would promote CPU efficiency.

kbs/src/plugins/implementations/splitapi/manager.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl SplitAPIBackend for CertManager {
+    async fn get_server_credential(&self, params: &SandboxParams) -> Result<Vec<u8>> {
+        // Try locking the sandbox directory mapper
+        let mut mapper_guard = self.mapper.lock().map_err(|e| {
+            anyhow!("Failed to lock sandbox directory mapper: {}", e)
+        })?;
```
Xynnn007, line 111: `anyhow` provides an easier way for `map_err(|e| anyhow!(..., e))` like things
```rust
xxx.context("Failed to lock sandbox directory mapper")?;
```

salmanyam:
Hi @fitzthum and @Xynnn007, I have addressed your comments in the latest commit. Could you please take a look? Thank you!

kbs/src/plugins/implementations/splitapi/backend.rs
```diff
@@ -0,0 +1,83 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::Result;
+use serde::Deserialize;
+use std::{path::PathBuf, sync::Arc};
+
+use super::generator::CertificateDetails;
+use super::manager;
+
+pub const CREDENTIALS_BLOB_FILE: &str = "certificates.json";
```
fitzthum, line 12: You're using the term `credential` throughout the PR, but also the term certificates. It might be clearer to just use one especially on lines like this one.

kbs/src/plugins/implementations/splitapi/manager.rs
```diff
@@ -0,0 +1,148 @@
...
+        };
+
+        // Generate the credentials (keys and certs for ca, server, and client)
+        let credential_dir = tempdir_in(self.plugin_dir.as_path())?;
+        let generator = CredentialGenerator::new(&credential_dir)?;
+        let credentials = generator.generate(&self.certificate_details)?;
+
+        log::info!("Credentials are generated!");
+
+        // Aquire the write lock and write the credential into the hashmap
```
fitzthum, line 130: It's a little odd to be writing state in a method called `get...`

kbs/src/plugins/implementations/splitapi/manager.rs
```diff
@@ -0,0 +1,148 @@
...
+        let state = self.state.read().await;
+        let serialized = serde_json::to_string(&*state)?;
+        tokio::fs::write(path, serialized).await?;
+        Ok(())
+    }
+}
+
+#[async_trait::async_trait]
+impl SplitAPIBackend for CertManager {
+    async fn get_server_credential(&self, params: &SandboxParams) -> Result<Vec<u8>> {
```
fitzthum, line 108: What is the `server` here? These credentials are for the client, right? Maybe you're referring to some other server, like the SplitAPI server in the guest, but this is pretty confusing.

kbs/src/plugins/implementations/splitapi/manager.rs
```diff
@@ -0,0 +1,148 @@
...
+        Ok(Self {
+            plugin_dir,
+            certificate_details: cert_details.clone(),
+            credblob_file: cblob_file,
+            state: Arc::new(RwLock::new(HashMap::new())),
+            credential_loaded_from_file: AtomicBool::new(false),
+        })
+    }
+
+    async fn load_credentials(&self, key: &str) -> Option<Credentials> {
```
fitzthum, line 72: nit: you are alternating between `credentials` (plural) and `credential` singular in your method names. Probably best to pick one.

Also, this function should probably return `Result<Option<Credentials>>` given that you could have a failure to load the credentials (i.e. if the file doesn't exist or something) or you could legitimately return no credentials.

kbs/src/plugins/implementations/splitapi/manager.rs
```diff
@@ -0,0 +1,148 @@
...
+        // Check if the credential is not loaded. If not, load them
+        if !self.credential_loaded_from_file.load(Ordering::SeqCst) {
```
fitzthum, line 74: Can we just do this at initialization time and avoid keeping track of whether they've been loaded with every request?

fitzthum (commented):
Thanks for the updates. I think this is a big improvement. At some point you should squash your commits together. I made some comments.

kbs/src/plugins/implementations/splitapi/manager.rs
salmanyam: @fitzthum, Thank you very much for providing feedback.

The reason for doing the loading this way is because the plugin initialization code (in `plugin_manager.rs`), that initializes the plugins, is synchronous, but the `load_credentials` function is async. Since the `async` function cannot be used inside a `sync` function, I cannot directly call the `load_credentials` function when the `splitapi` plugin is initialized. There might be two ways to solve it:

1. changing the `load_credentials` function to be synchronous, or
2. making the plugin initialization code to be asynchronous

I am not sure if there's a better way to call an async function inside a sync function.

fitzthum: Either one seems fine. Probably easier to make this sync. You're just reading from a file so that should be doable. I think you can make a new `tokio::sync::arc` without async.

bpradipt:
> > A generic question. Would it make sense to have a generic plugin to create key-pair for a sandbox and not club it with splitapi?
> 
> Some kind of PKI plugin would be very useful.

@salmanyam any plans to revisit this PR from the pov of creating a generic PKI plugin ?

salmanyam:
> > > A generic question. Would it make sense to have a generic plugin to create key-pair for a sandbox and not club it with splitapi?
> > 
> > 
> > Some kind of PKI plugin would be very useful.
> 
> @salmanyam any plans to revisit this PR from the pov of creating a generic PKI plugin ?

Hi @bpradipt, yes, I almost finished working on that aspect of the PR. The redesign of the PR will not be tied with splitapi. I will update the PR within a couple days.

portersrc:
Hey @salmanyam, cool work! One comment related to the `cred_store`:

Do you really need to maintain state with this plugin? You do it in two places. One is a hashmap called `cred_store`; and then you also persist that hashmap to file. The key to that hashmap is this IP-name-podID that you're using. Does a client ever need to request its credentials more than once? (Even if the pod dies and needs to get credentials again, won't it start up with a new podID, in which case it would need new keys anyway?)

salmanyam:
> Hey @salmanyam, cool work! One comment related to the `cred_store`:
> 
> Do you really need to maintain state with this plugin? You do it in two places. One is a hashmap called `cred_store`; and then you also persist that hashmap to file. The key to that hashmap is this IP-name-podID that you're using. Does a client ever need to request its credentials more than once? (Even if the pod dies and needs to get credentials again, won't it start up with a new podID, in which case it would need new keys anyway?)

Thanks, @portersrc! You brought a valid point. We also discussed this point in our internal meeting. Your are right. The client (or pod or sandbox) requests its credential only one time. But we need the persistence state for the following reason. That is if trustee restarts after the pod initialization, in that case, pod (client or sandbox) gets its credential, but the workload owner lose the CA that signs the pod's credential. As a result, the workload owner cannot perform mutual authentication between the pod and itself as we need the same CA for mutual authentication. Hope I was able to clarify.

portersrc:
I see. Is this PR as-is supporting this, though, or do you have more changes planned? I noticed when a new set of credentials have to be created, a new `ca_cert` is created. (That is, you don't sign each new `server_cert` with a common `ca_cert`; rather, you generate a new `ca_cert` for every new `server_cert`.)

Not sure if the following is related: The client credentials, though they get stored in a hashmap and persisted, aren't externally accessible. Maybe the client credentials need to have an API endpoint, as well, if you want any mutual authentication with the `ca_cert` that's common between the 1 server and 1 client?

salmanyam:
> I see. Is this PR as-is supporting this, though, or do you have more changes planned? I noticed when a new set of credentials have to be created, a new `ca_cert` is created. (That is, you don't sign each new `server_cert` with a common `ca_cert`; rather, you generate a new `ca_cert` for every new `server_cert`.)
> 
> Not sure if the following is related: The client credentials, though they get stored in a hashmap and persisted, aren't externally accessible. Maybe the client credentials need to have an API endpoint, as well, if you want any mutual authentication with the `ca_cert` that's common between the 1 server and 1 client?

The PR does not completely support whatever I said in the previous comment. I am working on the admin (workload owner) APIs. The admin APIs endpoints will allow workload owner to fetch the `client` credentials for the mutual authentication. Yes, to simplify the design, we decided to use a separate CA for each pod. The pod credentials can be obtained through the same API (`get_resource`).

bpradipt:
> @fitzthum, I have addressed your comments and refactored and cleaned the code. I have renamed the plugin name to `pki-vault`. I am also working to add the admin APIs which I will update soon. I would greatly appreciate your feedback.
> 
> Hi @bpradipt, The plugin name is now `pki-vault`. Currently, it generates all the keys [CA, server (sandabox), client (owner or admin)] and then generates the certificates for server and client signed by the CA certificate. I am working to support functionality where the sandbox or client (owner or admin) will generate their own key-pair and send the certificate signing request (CSR) to the plugin to sign it by the same CA. One key aspect of the current design is mutual authentication, i.e., both the sandbox and owner confirm each other. BTW, can you please tell me your use case? That will help me greatly for generalizing the design.

Thanks for working on this @salmanyam .. Following are my use cases:
1. Mutual authentication between the user (owner) and the CoCo pod
2. Mutual authentication between between two CoCo pods 
3. Mutual authentication between between a workload in the trusted env and CoCo pod (similar to 1)
4. A server in the trusted env (having the CA cert from Trustee) verifying the client certs sent from CoCo pods

Ability to send a CSR to the plugin from the pod as well as an API to retrieve the ca cert will be important imho. The features can be iteratively implemented.

Can you add a design doc or readme to the PR. It'll help to review.

fitzthum:
I agree with most of the things mentioned here. It seems like it would be simpler to not maintain any state. I was picturing there being one root cert, corresponding to Trustee itself. It seems like in the current implementation different pods would have different root certs, which is interesting.

Also, I'm not exactly sure about the data that is being provided from the client. The pod id and such. This seems a little arbitrary and it doesn't seem like this information actually makes it into the cert anyway. Do we want the cert to be endorsing something about the guest? One possible approach would be to allow clients to specify any values they want as part of a query string. This query string would pass through the policy engine, so Trustee admins could add their own logic about what is allowed. Then maybe these claims would be added to the cert as an extension or something.

Ultimately I'm not sure exactly what PKI for Trustee should look like, so if you have a vision here, feel free to go that direction. I think a design doc would help understand the sort of use cases that you are looking at.

salmanyam:
> > @fitzthum, I have addressed your comments and refactored and cleaned the code. I have renamed the plugin name to `pki-vault`. I am also working to add the admin APIs which I will update soon. I would greatly appreciate your feedback.
> > Hi @bpradipt, The plugin name is now `pki-vault`. Currently, it generates all the keys [CA, server (sandabox), client (owner or admin)] and then generates the certificates for server and client signed by the CA certificate. I am working to support functionality where the sandbox or client (owner or admin) will generate their own key-pair and send the certificate signing request (CSR) to the plugin to sign it by the same CA. One key aspect of the current design is mutual authentication, i.e., both the sandbox and owner confirm each other. BTW, can you please tell me your use case? That will help me greatly for generalizing the design.
> 
> Thanks for working on this @salmanyam .. Following are my use cases:
> 
> 1. Mutual authentication between the user (owner) and the CoCo pod
> 2. Mutual authentication between between two CoCo pods
> 3. Mutual authentication between between a workload in the trusted env and CoCo pod (similar to 1)
> 4. A server in the trusted env (having the CA cert from Trustee) verifying the client certs sent from CoCo pods
> 
> Ability to send a CSR to the plugin from the pod as well as an API to retrieve the ca cert will be important imho. The features can be iteratively implemented.
> 
> Can you add a design doc or readme to the PR. It'll help to review.

Hi @bpradipt, thank you for providing the use cases. It will help me greatly to design the admin (workload owner) APIs. I am working on the owner side APIs and design doc. I will update soon.

bpradipt:
@salmanyam any updates on this ?

salmanyam:
> @salmanyam any updates on this ?

@bpradipt Yes, I have made progress (both in code and documentation). I will update what I have so far by today.

salmanyam:
Hi @bpradipt and @fitzthum, I have made progress to this work as it was sidelined for a few weeks due to other work. I have also started the documentation. Could you please take a look? Thank you!

kbs/docs/pki_vault.md
```diff
@@ -0,0 +1,186 @@
+# PKI Vault plugin
+
+This plugin currently generates credentials (keys and certificates) for a server running inside the confidential VM (aka sandbox) and for the workload owner (who acts as a client for the server). The current design of the plugin prioritizes the mutual authentication between the server and the client. Such design is necessary for the SplitAPI (kata-containers/kata-containers#9159 and
+kata-containers/kata-containers#9752) and peer-pods.
+
+This plugin also delivers the server-specific credentials to the sandbox (i.e., confidential PODs or VMs), specifically to the kata agent to initiate the server. The workload owner can communicate with the server using a secure tunnel.
```
fitzthum: Probably no need to mention the Kata Agent here. That's not really in scope of Trustee.

kbs/docs/pki_vault.md
```diff
@@ -0,0 +1,186 @@
...
+
+The server-specific credentials can be obtained throught the `get-resource` APIs by specifying the `plugin-name`. Currently, the plugin requires that the sandbox or `kata-agent` sends the IPv4 address, name, and the ID of the sandbox (i.e., pod) as part of the query string to obtain the credentials from the KBS. 
+
+After receiving the credential request, the `pki_vault` plugin will create a CA, a key pair for the server and another key pair for client, and sign them using the self-signed CA. Currently, the generated credentials are stored in a hashmap with a unique key for each sandbox based on its name, ID, and IP address. But we expect this design will be changed in future. PKI Vault plugin responds to a request from the sandbox by sending the server specific credentials (key, cert) along with the CA certificate. A request from workload owner gets the client specific credentials (key, cert) and CA's certificate.
```
fitzthum: I'm still a little unclear on why the credentials are stored. Why not just generated a new certificate? What do you expect to change about the design in the future?

kbs/docs/pki_vault.md
```diff
@@ -0,0 +1,186 @@
...
+```toml
+[[plugins]]
+name = "pki_vault"
+```
+The following addition details can be set to the configuration file.
+```toml
+[[plugins]]
+name = "pki_vault"
+#plugin_dir = "/opt/confidential-containers/kbs/plugin/splitapi"
+#cred_filename = "certificates.json"
```
fitzthum: Why these commented-out lines?

kbs/docs/pki_vault.md
```diff
@@ -0,0 +1,186 @@
...
+
+# Runtime services
+
+All runtime services for both the server and client supported are described in the following sections.
+
+## Credentials service for server (i.e., pod or sandbox)
+
+Request the credentials for initiating a server inside a pod or sandbox.
+
+Only `GET` request is supported, e.g. `GET /kbs/v0/pki_vault/credentials?id=3367&ip=60.11.12.89&name=pod51`. Current the `GET` takes `id`, `ip`, and `name` parameters, but expect this parameters to be changed in the future design for supporting more generic use cases.
```
fitzthum: nit: `Current` -> `Currently`.

Yeah it's a little unclear why `id`, `ip`, and `name` are the required fields. You could just allow any query string to be given and create a cert with the values in various extensions. The plugin requests pass through the policy, so it might be best to have the policy, which is highly configurable, be the place where we have logic about checking the parameters.

kbs/docs/pki_vault.md
```diff
@@ -0,0 +1,186 @@
...
+
+```rust
+pub struct CredentialsOut {
+    pub key: Vec<u8>, // Key created
+    pub cert: Vec<u8>, // Self-signed certificate created
+    pub ca_cert: Vec<u8>, // CA certificate
+}
+```
+
+
```
fitzthum: nit: no need for these empty lines

kbs/docs/pki_vault.md
```diff
@@ -0,0 +1,186 @@
...
+Only `GET` request is supported, e.g. `GET /kbs/v0/pki_vault/credentials?id=3367&ip=60.11.12.89&name=pod51`. Current the `GET` takes `id`, `ip`, and `name` parameters, but expect this parameters to be changed in the future design for supporting more generic use cases.
+
+The request takes parameters via URL query string. All parameters supported are described in the table below. Note currently, all parameters are required.
+
+| Property            | Type   | Required | Description             | Default | Example                                   |
+|---------------------|--------|----------|-------------------------|---------|-------------------------------------------|
+| `name`              | String | Yes      | Name of the pod |         | `credentials?id=3367&ip=60.11.12.89&name=pod51` |
+| `ip`                | String | Yes      | IPv4 address of a pod to assign to the certificate |         | `credentials?id=3367&ip=60.11.12.89&name=pod51` |
+| `id`          | String | Yes       | Pod ID |  | `credentials?id=3367&ip=60.11.12.89&name=pod51` |
```
fitzthum: The example column is a little confusing to me. It's the same example for every row? Also it doesn't look like there is anything in the default column? Does the name field have to be a pod name or could it be any sort of name?

kbs/docs/pki_vault.md
```diff
@@ -0,0 +1,186 @@
...
+All APIs supported are described in the table below.
+
+| API            |  Description             | Example                                   |
+|---------------------|-------------------------|-------------------------------------------|
+| `list_pods`              | To get the list of pods that have server credentials created| `kbs/v0/pki_vault/list_pods` |
+| `get_client_credentials`              | To get the client credentials for a pod | `/kbs/v0/pki_vault/get_client_credentials?id=3367&ip=60.11.12.89&name=pod51` |
+
+The request will be processed only if the request is authenticated, otherwise an error is returned. In the current design, the credentials for the client already already exists in the KBS as they have been already created as part of the response of server credential request, so the plugin simply returns the existing client credentials.
+
+Once the request is processed, the following structure is returned in JSON format.
```
fitzthum: It might be better to give these examples as JSON since that is the interface users will see.

kbs/docs/pki_vault.md
```diff
@@ -0,0 +1,186 @@
...
+    pub cert: Vec<u8>, // Self-signed certificate created
+    pub ca_cert: Vec<u8>, // CA certificate
+}
+```
+
+
+# How to test (client or owner-side code)
+In order to test the `pki-vault` plugin, we need to enable the plugin support in the `kbs_protocol` inside `attestation-agent` of the `guest-components` respository. The plugin support has been enabled in the `pki-vault-test` branch of following `guest-components` repository.
+
+`https://github.com/salmanyam/guest-components/tree/pki-vault-test`
```
fitzthum: This sort of thing probably shouldn't go into the actual docs. Ideally we should just have an integration test in the PR. The docs are aimed a little more at users than developers. 

We do have a gap in guest components. We will need plugin support there. I think @cclaudio and @portersrc have also run into this problem.

kbs/src/plugins/implementations/pki_vault.rs
```diff
@@ -0,0 +1,590 @@
...
+                        let credentials = self.get_server_credentials(&params).await?;
+
+                        Ok(credentials)
+                    }
+                    "list_pods" => {
+                        let pods = self.list_pods().await?;
+
+                        Ok(pods)
+                    }
+                    "get_client_credentials" => {
```
fitzthum: I think having `get_` here is redundant and not very restful.

kbs/src/plugins/implementations/pki_vault.rs
```diff
@@ -0,0 +1,590 @@
...
+        _query: &str,
+        path: &str,
+        method: &Method,
+    ) -> Result<bool> {
+        let sub_path = path
+            .strip_prefix('/')
+            .context("accessed path is illegal, should start with `/`")?;
+
+        if method.as_str() == "GET" {
+            if sub_path != "credentials" {
```
fitzthum: It seems odd to have the same endpoint require both `auth` and `encryption` (next function). I might have this backwards, but don't you want `get_client_credentials` to be the one that requires auth?

kbs/src/plugins/implementations/pki_vault.rs
```diff
@@ -0,0 +1,590 @@
...
+                return Ok(true);
+            }
+        }
+
+        Ok(false)
+    }
+
+    /// Whether the body needs to be encrypted via TEE key pair.
+    /// If returns `Ok(true)`, the KBS server will encrypt the whole body
+    /// with TEE key pair and use KBS protocol's Response format.
```
fitzthum: Probably should change this from the sample comment to a comment that explains why we are encrypting various endpoints.

fitzthum (commented):
Nice update. Mainly commented on the doc. Code looks tighter. An integration test would be nice. I think we are definitely moving in the right direction here.

kbs/docs/plugins/pki_vault.md
```diff
@@ -0,0 +1,147 @@
...
+
+```toml
+[[plugins]]
+name = "pki_vault"
+```
+The following addition details can be set to the configuration file.
+```toml
+[[plugins]]
+name = "pki_vault"
+plugin_dir = "/opt/confidential-containers/kbs/plugin/splitapi"
```
bpradipt: nit: the dir name splitapi doesn't align with the intended functionality here.

kbs/docs/plugins/pki_vault.md
```diff
@@ -0,0 +1,147 @@
...
+
+3. Start trustee
+
+```bash
+sudo ../target/release/kbs --config-file ./config/kbs-config.toml
+```
+# Design choices
+There can be three key design choices one can consider to configure PKI Vault plugin. As of now, PKI Vault only supports the separte CA per Pod approach with a non-persistent credential storage.
+
+1. **Separate CA per Pod VM**:
```
bpradipt: With this model, mTLS between two CoCo pods can't be used right ?

salmanyam:
Hi @fitzthum and @Xynnn007 

I have significantly restructured the plugin according to your suggestions. I have also tested the plugin with the latest plugin-calling changes made by @Xynnn007. Could you please take a look at the latest changes? I have also rewritten the readme doc for the plugin. I have renamed the plugin name from `pki_vault` to `keyflux` to have the name aligned with what the plugin does.

## Key changes are the following.

- **Plugin name**: Changed from `pki_vault` to `keyflux`
- **Configuration structure**: Updated to use nested TOML sections
- **API paths**: Changed from `/kbs/v0/pki_vault/*` to `/kbs/v0/keyflux/*` (unauthenticated) and `/kbs/v0/keyflux/*` (authenticated)
- **Additional secret types**: Added support for Ed25519 and RSA keys beyond TLS and symmetric keys
- **Flexible query parameters**: Configurable required parameters instead of hardcoded `token`, `name`, `ip`

Please take a look the readme (kbs/docs/plugins/keyflux.md) for more details

fitzthum:
Thanks for the updates @salmanyam I'll review on monday

kbs/docs/plugins/keyflux.md
```diff
@@ -0,0 +1,373 @@
...
+- `validity_days`: Certificate validity period in days (default: 3650)
+
+**Query Configuration** (`plugins.keyflux.query`):
+- `required`: List of required query parameters (default: `["id"]`)
+- `spec.required`: Whether the `spec` parameter is mandatory (default: `true`)
+
+**Limits Configuration** (`plugins.keyflux.limits`):
+- `symmetric_key_size`: Size of symmetric keys in bytes (default: 32)
+- `rsa_bits`: RSA key size in bits (default: 2048)
+- `allow_types`: Allowed secret types (default: `["tls", "symmetric", "ed25519", "rsa"]`)
```
Xynnn007: Do we have a list of `allow_types`?

kbs/src/plugins/implementations/keyflux.rs
```diff
@@ -0,0 +1,867 @@
...
+pub struct SpecConfig {
+    #[serde(default)]
+    pub required: bool,
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+pub struct LimitsConfig {
+    pub symmetric_key_size: usize,
+    pub rsa_bits: usize,
+    pub allow_types: Vec<String>,
```
Xynnn007: Looks like this field is never used?

kbs/docs/plugins/keyflux.md
Xynnn007: Oh I see the full list are the four. Let's explicitly show it

kbs/src/plugins/implementations/keyflux.rs
```diff
@@ -0,0 +1,867 @@
...
+        Self {
+            keyflux: KeyFluxSection {
+                //secrets: vec![],
+                ca: TlsCertDetails::default(),
+                query: QueryConfig::default(),
+                limits: LimitsConfig::default(),
+            }
+        }
+    }
+}
```
Xynnn007: This can be simplifiled by a `default` macro upon `KeyFluxPluginConfig`

kbs/src/plugins/implementations/keyflux.rs
```diff
@@ -0,0 +1,867 @@
...
+            spec_store: Arc::new(RwLock::new(HashMap::new())),
+
+            server_cert_config_store: Arc::new(RwLock::new(HashMap::new())),
+            client_cert_config_store: Arc::new(RwLock::new(HashMap::new())),
+
+            ca_store: Arc::new(RwLock::new(HashMap::new())),
+
+            ed25519_store: Arc::new(RwLock::new(HashMap::new())),
+            rsa_store: Arc::new(RwLock::new(HashMap::new())),
+            symkey_store: Arc::new(RwLock::new(HashMap::new())),
```
Xynnn007: We can use `tokio::sync::Rwlock` than `std`'s here. As the whole runtime is in the async

kbs/src/plugins/implementations/keyflux.rs
```diff
@@ -0,0 +1,867 @@
...
+    }
+
+    fn parse_spec(spec: &str) -> Result<HashMap<SecretType, usize>> {
+        let mut result = HashMap::new();
+
+        for item in spec.split(';') {
+            let parts: Vec<&str> = item.split(':').collect();
+
+            if parts.len() != 2 {
+                return Err(anyhow!("Invalid spec format: {}", item));
```
Xynnn007:
```suggestion
                bail!("Invalid spec format: {}", item);
```

kbs/src/plugins/implementations/keyflux.rs
```diff
@@ -0,0 +1,867 @@
...
+            }
+
+            let secret_type = match parts[0] {
+                "tls" => SecretType::Tls,
+                "sym" | "symmetric" => SecretType::Symmetric,
+                "pkey" | "ed25519" => SecretType::Ed25519,
+                "rsa" => SecretType::Rsa,
+                other => return Err(anyhow!("Unknown secret type: {}", other)),
```
Xynnn007:
```suggestion
                other => bail!("Unknown secret type: {}", other),
```

kbs/src/plugins/implementations/keyflux.rs
```diff
@@ -0,0 +1,867 @@
...
+            };
+
+            let count: usize = parts[1]
+                .parse()
+                .map_err(|_| anyhow!("Invalid count for {}", parts[0]))?;
+
+            if count == 0 {
+                return Err(anyhow!("Count must be > 0 for {}", parts[0]));
```
Xynnn007:
```suggestion
                bail!("Count must be > 0 for {}", parts[0]);
```

kbs/src/plugins/implementations/keyflux.rs
```diff
@@ -0,0 +1,867 @@
...
+    pub client_cert_config_store: Arc<RwLock<HashMap<String, TlsCertDetails>>>,
+
+    // Ed25519 pub key store for clients
+    pub ed25519_store: Arc<RwLock<HashMap<String, HashMap<String, Vec<u8>>>>>,
+
+    // Rsa pub key store for clients
+    pub rsa_store: Arc<RwLock<HashMap<String, HashMap<String, Vec<u8>>>>>,
+
+    // secret stored for clients
+    pub symkey_store: Arc<RwLock<HashMap<String, HashMap<String, Vec<u8>>>>>,
```
Xynnn007: Not the issue for this PR, but the KBS design. This HashMap will bring state to a single KBS instance and will be against https://github.com/confidential-containers/trustee/issues/1092

This is a design problem of KBS that a unified key value storage interface is not easily achieved. Let me mark this as a TODO for me.

kbs/docs/plugins/keyflux.md
```diff
@@ -0,0 +1,373 @@
...
+These APIs are called by pod VMs after successful attestation.
+
+### Get Credentials
+
+Request credentials for a pod VM by specifying the required secret types and counts.
+
+**Endpoint**: `GET /kbs/v0/keyflux/credentials`
+
+**Query Parameters**:
+- `id` (required): Unique identifier for the pod
```
fitzthum: It might be good to note that this pod id is simply whatever the guest chooses. It isn't bound to the true pod id, or is it?

kbs/docs/plugins/keyflux.md
```diff
@@ -0,0 +1,373 @@
...
+**Supported Secret Types**:
+- `tls`: TLS certificate and private key
+- `sym` or `symmetric`: Symmetric encryption key
+- `ed25519` or `pkey`: Ed25519 key pair
+- `rsa`: RSA key pair
+
+**Example Request**:
+
+```bash
+../target/release/kbs-client \
```
fitzthum: Slightly confusing example. You talk about this being used by pods, but you show the KBS client, which should not be used from within a pod. Maybe you want to show with the ASR rest API instead.

More generally, I'm not sure that this doc should be talking about pods. Trustee can be used outside of confidential containers. It seems like this plugin would also work with VMs. So maybe you want to get rid of the terms pod and sandbox entirely and give an example of getting credentials either via the rest API for coco or the kbspclient in a VM.

kbs/docs/plugins/keyflux.md
```diff
@@ -0,0 +1,373 @@
...
+    --url http://localhost:8090 \
+    get-resource --path "credentials?id=pod123&spec=tls:1;sym:2;ed25519:1"
```
fitzthum: fwiw I don't really see the value in being able to request multiple keys at once. This makes things more complicated all around. If a user needs multiple keys they could just make multiple requests. I guess it is fine to keep this, tho.

kbs/docs/plugins/keyflux.md
```diff
@@ -0,0 +1,373 @@
...
+Retrieve a list of all pod identifiers that have credentials stored.
+
+**Endpoint**: `POST /kbs/v0/keyflux/list_pods`
+
+**Example Request**:
+
+```bash
+../target/release/kbs-client \
+    --url http://localhost:8090 \
+    config --auth-private-key config/private.key \
```
fitzthum: See recent changes to auth. We use a token here instead now.

kbs/docs/plugins/keyflux.md
```diff
@@ -0,0 +1,373 @@
...
+5. **Workload owner retrieves client credentials**:
+   ```bash
+   kbs-client --url http://localhost:8090 \
+       config --auth-private-key config/private.key \
+       get-client-creds --query "id=pod123"
+   ```
+
+6. **Establish mutual TLS connection** between pod VM (server) and workload owner (client) using the credentials
+
+## Security Considerations
```
fitzthum: this section is unnecessary imo

kbs/docs/plugins/keyflux.md
```diff
@@ -0,0 +1,373 @@
...
+
+- **Per-Pod CA Isolation**: Each pod has its own CA, limiting the blast radius of a compromise
+- **Non-Persistent Storage**: Credentials are stored in memory only and are lost on restart
+- **Attestation Required**: Pod VMs must pass attestation before receiving credentials
+- **Authentication Required**: Client APIs require KBS private key authentication
+- **Automatic Encryption**: Server credentials are automatically encrypted with TEE public key
+
+## Troubleshooting
```
fitzthum: not sure this is needed either. The first issue is obvious. For the others, just add a bit of the trouble shooting info to the error message itself.

kbs/src/plugins/implementations/keyflux.rs
```diff
@@ -0,0 +1,867 @@
...
+        self.common_name = name.into();
+        self
+    }
+}
+
+impl Default for KeyFluxPluginConfig {
+    fn default() -> Self {
+        Self {
+            keyflux: KeyFluxSection {
+                //secrets: vec![],
```
fitzthum: remove if not needed

fitzthum (commented):
I think this is an improvement. Made a few comments.

Btw, where does the term keyflux come from. I see there is a company and a rust crate already called keyflux, but they seem unrelated. If this is a new term you are coming up with, I might opt for something more descriptive.

salmanyam:
Hi @Xynnn007 and @fitzthum,

Sorry for the delay. I tried to address all comments. I also made some design and code quality improvements. Could you please take a look? Thank you!

@Xynnn007
- Replaced std::sync::RwLock with tokio::sync::RwLock throughout, converting all lock acquisitions to .read().await / .write().await. This avoids blocking the async runtime on lock contention.
- Removed the hand-written Default for KeyFluxPluginConfig
- Simplified the logic by removing the feature that generates multiple secrets at a time. Now, we generate one secret at a time which is much simpler
- Simplified update_cert_details — replaced the redundant and_modify(...).or_insert(...) pattern with a single insert call.

@fitzthum 
- Removed the unused constants DEFAULT_SERVER_VALIDITY_DAYS, DEFAULT_CLIENT_VALIDITY_DAYS, and CA_DEFAULT_COMMON_NAME.
- Removed the Security Considerations section from the documentation.
- Removed the Troubleshooting section from the documentation 
- Replaced pod/sandbox-centric language in the documentation with VM-centric language. The plugin is not specific to Kubernetes pods.
- Replaced the kbs-client examples for server-side credential requests with the KBS REST API form, since kbs-client should not be used from inside a confidential VM.
- Added a note that identity parameter values are supplied by the guest itself and are not verified against any external source — policy enforcement is delegated to the KBS resource policy.
- Updated config documentation to reflect the current field names (spec_required, supported_types).

kbs/config/kbs-config.toml
```diff
@@ -54,3 +54,26 @@ storage_backend_type = "kvstorage"
...
+[plugins.secretprov.query]
+# Identity params joined to build the store key, e.g. "pod-abc_default"
+required = ["name", "ns"]
+# Require secret_name + secret_type on every request
+spec_required = true
+
+[plugins.secretprov.limits]
+symmetric_key_size = 32
+rsa_bits = 2048
+supported_types = ["tls", "symmetric", "ed25519", "rsa", "p256"]
```
Xynnn007, line 79: Let's not add the configs here as it's used for docker-compose deployments and will not take effect.

salmanyam: Sounds good. Will remove this.

kbs/src/plugins/implementations/secretprov.rs
```diff
@@ -0,0 +1,876 @@
...
+use serde::{Deserialize, Serialize};
+
+use super::super::plugin_manager::ClientPlugin;
+
+/// Default values used when certificate fields are not set in the plugin config.
+pub const DEFAULT_COUNTRY: &str = "AA";
+pub const DEFAULT_STATE: &str = "Default State";
+pub const DEFAULT_LOCALITY: &str = "Default City";
+pub const DEFAULT_ORGANIZATION: &str = "Default Organization";
+pub const DEFAULT_ORG_UNIT: &str = "Default Unit";
```
fitzthum, line 30: maybe these defaults should mention coco/trustee

kbs/src/plugins/implementations/secretprov.rs
```diff
@@ -0,0 +1,876 @@
...
+pub const DEFAULT_CA_VALIDITY_DAYS: u32 = 3650;
```
fitzthum, line 31: Seems a little high. These certs are one-per-guest, right?

kbs/src/plugins/implementations/secretprov.rs
```diff
@@ -0,0 +1,876 @@
...
+    pub spec_required: bool,
+}
+
+fn default_true() -> bool {
+    true
+}
+
+/// Resource limits applied during secret generation.
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+pub struct LimitsConfig {
```
fitzthum, line 66: These aren't really limits. They're just settings.

kbs/src/plugins/implementations/secretprov.rs
```diff
@@ -0,0 +1,876 @@
...
+
+        Ok((secret_name, secret_type, spec_sub_key))
+    }
+
+    // ---- Endpoint handlers ----------------------------------------------
+
+    /// Handles `GET /credentials`.
+    ///
+    /// Generates a fresh key pair on every call and overwrites any previously
+    /// stored public material for the same (identity-key, spec-sub-key) pair.
```
fitzthum, line 605: It's interesting that you're overwriting an existing key. Since the identity is derived from the query string, which is unmeasured, could a third party overwrite someone else's entry?

kbs/src/plugins/implementations/secretprov.rs
```diff
@@ -0,0 +1,876 @@
...
+            ca_config: config.secretprov.ca,
+            query_config: config.secretprov.query,
+            limits_config: config.secretprov.limits,
+
+            // Outer key: identity string built from query.required params (e.g. "pod-abc_default")
+            // Inner key: spec sub-key built from secret_name:secret_type (e.g. "grpc:tls")
+            store: Arc::new(RwLock::new(HashMap::new())),
+
+            server_cert_config_store: Arc::new(RwLock::new(HashMap::new())),
+            client_cert_config_store: Arc::new(RwLock::new(HashMap::new())),
```
fitzthum, line 504: Well, at this point we should really be using the kvstorage backend for this.

kbs/src/plugins/implementations/secretprov.rs
```diff
@@ -0,0 +1,876 @@
...
+/// Controls how incoming query strings are interpreted.
+///
+/// `required` lists the params whose values are joined (with `_`) to form the
+/// per-identity store key.  For example, `["name", "ns"]` with a request of
+/// `?name=pod-abc&ns=default` produces the key `"pod-abc_default"`.
+///
+/// `spec_required` gates whether `secret_name` and `secret_type` must appear
+/// on every request (default: true).
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+pub struct QueryConfig {
```
fitzthum, line 50: Do you really need to have a configurable mapping for this? Seems like it would be simpler to just have one set of required args, at least in the first iteration.

kbs/src/plugins/implementations/secretprov.rs
```diff
@@ -0,0 +1,876 @@
...
+    /// Length of generated symmetric keys in bytes.
+    pub symmetric_key_size: usize,
+    /// Bit length for generated RSA key pairs.
+    pub rsa_bits: usize,
+    /// Allowlist of secret types. Requests for any other type are rejected.
+    pub supported_types: Vec<String>,
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Default)]
+pub struct SecretProvSection {
```
fitzthum, line 78: nit: `section` is kind of a vague term.

fitzthum (commented):
A couple comments, but this is looking pretty good.

tbh I'm not sure `SecretProv` is the best name either. All of the plugins provision secrets. This one is more like a keypair generator?

kbs/src/plugins/implementations/secretprov.rs
salmanyam: yes, that's a possibility. Do you have any suggestions or pointers here?

salmanyam:
All these comments make sense. I will address them asap.

kbs/src/plugins/implementations/secretprov.rs
fitzthum: Well you could get these parameters from init-data rather than from the query string. This would be bound to the Tee instance.

If you don't overwrite the connections, you risk mixing public keys between guests.

You could potentially stick with the current approach, but you probably want to document the limitations. Speaking of which, this can't really support certs that expire, right?

salmanyam: I like the init-data approach. I would like to give it a try. Regarding the certs expiration, yes, I just realized that. Let me think here and come up what is better.

fitzthum: Check out https://github.com/confidential-containers/trustee/pull/1305 regarding init-data. That plugin is relatively similar to this one, but it expects an out-of-band registration operation and it's more tailored to LUKs keys.

salmanyam:
Hi @fitzthum and @Xynnn007,

I have addressed all the comments. I rename the plguin to `credgen`. I am also open for a suggestion. Could you please take a look? Thank you!

Resolved:
1. Removed credgen config block from kbs-config.toml
2. Updated default cert subject fields to reference Confidential Containers / Trustee
3. Reduced CA validity from 3650 to 365 days; end-entity certs default to 90 days
4. Renamed LimitsConfig -> SettingsConfig and TOML section limits -> settings
5. Removed configurable QueryConfig; identity fields hardcoded to ["name", "ns"]
6. Renamed CredGenSection -> CredGenConfig
7. VM identity now derived from TEE-measured init-data instead of the query string
8. CA expiry fixed: fresh CA generated on every GET /credentials and stored in the entry
9. Migrated credential store from in-memory HashMap to the KBS kvstorage backend

PR #586, opened 2024-11-19
cargo: bump guest-components to fix kbs-client

fitzthum:
We recently bumped the version of csv-rs in guest-components and Trustee, but we didn't bump the version of guest-components that trustee uses, which leads building the kbs-client with a version of openssl that is not compatible with the older guest-components.

Update guest-components so we don't have this problem.

Also need to bump tempfile in Trustee to avoid a versioning snafu.


Xynnn007 (approved):

Cargo.toml
```diff
@@ -29,7 +29,7 @@ config = "0.13.3"
 env_logger = "0.10.0"
 hex = "0.4.3"
 jwt-simple = "0.11"
-kbs_protocol = { git = "https://github.com/confidential-containers/guest-components.git", tag="v0.10.0", default-features = false }
+kbs_protocol = { git = "https://github.com/confidential-containers/guest-components.git", rev="e6999a3c0fd877dae9e68ea78b8b483062db32b8", default-features = false }
 kbs-types = "0.7.0"
 kms = { git = "https://github.com/confidential-containers/guest-components.git", tag="v0.10.0", default-features = false }
```
mythi, line 34:
```suggestion
kms = { git = "https://github.com/confidential-containers/guest-components.git", rev="e6999a3c0fd877dae9e68ea78b8b483062db32b8", default-features = false }
```

fitzthum: good catch

Issue #589, opened 2024-11-21
verifier/snp: check for attestation report version causing verifier to fail


AdithyaKrishnan:
The check for the attestation report version is causing an error in the verifier due to a [recent update ](https://github.com/google/go-sev-guest/commit/f3a65377d1e69e05eb44c2d791423df910422233) changing the version and structure of the attestation report. A stakeholder from the confidential community brought up this concern. 

```
ERROR kbs::http::error] Attestation failed: Verifier evaluate failed: Unexpected report version
```

There are 2 proposals to fix this issue:

1. Skip the check for the report version
2. Change the check from ```report.version != 2``` to ```report.version != 3``` or ```report.version >= 2```

AdithyaKrishnan:
Fixed by Trustee [PR#590](https://github.com/confidential-containers/trustee/pull/590)

PR #590, opened 2024-11-21
verifier: Change logic to check the attestation report version

AdithyaKrishnan:
Fixes Issue [#589](https://github.com/confidential-containers/trustee/issues/589)

Change the check condition to handle multiple attestation report versions.

### DIFF
--- a/deps/verifier/src/snp/mod.rs
+++ b/deps/verifier/src/snp/mod.rs
@@ -37,6 +37,10 @@ const LOADER_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .1);
 const KDS_CERT_SITE: &str = "https://kdsintf.amd.com";
 const KDS_VCEK: &str = "/vcek/v1";
 
+/// Attestation report versions supported 
+const REPORT_VERSION_MIN: u32 = 2;
+const REPORT_VERSION_MAX: u32 = 3;
+
 #[derive(Debug)]
 pub struct Snp {
     vendor_certs: VendorCertificates,
@@ -104,8 +108,9 @@ impl Verifier for Snp {
 
         verify_report_signature(&report, &cert_chain, &self.vendor_certs)?;
 
-        if report.version != 2 {
-            return Err(anyhow!("Unexpected report version"));
+        // See Trustee Issue#589 https://github.com/confidential-containers/trustee/issues/589
+        if report.version < REPORT_VERSION_MIN || report.version > REPORT_VERSION_MAX {
+            return Err(anyhow!("Unexpected attestation report version. Check SNP Firmware ABI specification"));
         }
 
         if report.vmpl != 0 {


deps/verifier/src/snp/mod.rs
```diff
@@ -37,6 +37,10 @@ const LOADER_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .1);
 const KDS_CERT_SITE: &str = "https://kdsintf.amd.com";
 const KDS_VCEK: &str = "/vcek/v1";
 
+// Attestation Report Parameters
+const REPORT_VERSION_MIN_SUPPORT: u32 = 2;
+const REPORT_VERSION_MAX_SUPPORT: u32 = 3;
```
cclaudio, line 42: nit: keep SUPPORT only in the comment
```rust
// Attestation report version supported
const REPORT_VERSION_MIN: u32 = 2;
const REPORT_VERSION_MAX: u32 = 3;
```

deps/verifier/src/snp/mod.rs
```diff
@@ -104,7 +108,8 @@ impl Verifier for Snp {
 
         verify_report_signature(&report, &cert_chain, &self.vendor_certs)?;
 
-        if report.version != 2 {
+        // See Trustee Issue#589 https://github.com/confidential-containers/trustee/issues/589
+        if report.version >= REPORT_VERSION_MIN_SUPPORT && report.version <= REPORT_VERSION_MAX_SUPPORT {
```
cclaudio, line 112: I think you mean `if version < MIN || version > MAX `

cclaudio (changes requested):

deeglaze:
Do you plan on updating `fetch_vcek_from_kds` to determine which kds product endpoint to try based on the CPUID information in the attestation report? That plus the fact that KDS gets the ProductName wrong in its VCEK cert extensions (for complicated reasons that should be fixed for Turin and later) is the reason why they changed the attestation report.

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: Oops, my bad. Thanks!

AdithyaKrishnan:
> Do you plan on updating `fetch_vcek_from_kds` to determine which kds product endpoint to try based on the CPUID information in the attestation report? That plus the fact that KDS gets the ProductName wrong in its VCEK cert extensions (for complicated reasons that should be fixed for Turin and later) is the reason why they changed the attestation report.

Not in this PR but might have a new PR in the future to do the same.

cclaudio (approved):
LGTM

deps/verifier/src/snp/mod.rs
```diff
@@ -37,6 +37,10 @@ const LOADER_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .1);
 const KDS_CERT_SITE: &str = "https://kdsintf.amd.com";
 const KDS_VCEK: &str = "/vcek/v1";
 
+// Attestation report versions supported
```
Xynnn007, line 40: The note here looks very good to be an inline document

```suggestion
/// Attestation report versions supported 
```

deps/verifier/src/snp/mod.rs
```diff
@@ -104,7 +108,8 @@ impl Verifier for Snp {
 
         verify_report_signature(&report, &cert_chain, &self.vendor_certs)?;
 
-        if report.version != 2 {
+        // See Trustee Issue#589 https://github.com/confidential-containers/trustee/issues/589
+        if report.version < REPORT_VERSION_MIN || report.version > REPORT_VERSION_MAX {
             return Err(anyhow!("Unexpected report version"));
```
Xynnn007, line 113: Could we also add some context including `SNP` to the error information here? Because the information
```
ERROR kbs::http::error] Attestation failed: Verifier evaluate failed: Unexpected report version
```

is somehow confusing as it could be KBS protocol's error.

Xynnn007 (approved):
LGTM. Thanks! @AdithyaKrishnan . One thing to think about the error report

bpradipt (approved):
/lgtm

fitzthum (approved):
LGTM. See Ding's comments

deps/verifier/src/snp/mod.rs
AdithyaKrishnan: Fixed. Thanks Ding!

PR #597, opened 2024-11-25
KBS: Update KBS protocol to 0.2.0 to fix JWE format due to RFC7516

### LINKED ISSUE #583: JWE encryption is missing base64url-encoded protected header as AEAD
### Describe the bug

The JWE protected_header is not protected AEAD as expected by the [RFC for JWE](https://datatracker.ietf.org/doc/html/rfc7516#section-3.3).

### How to reproduce

n/a

### CoCo version information

all

Snp

_No response_

Xynnn007:
Fixes #583.

This patch does a bunch of fixes per RFC 7516.
1. AEAD Auth Tag is now expcilitly included inside the `tag` part.
2. Add new JWE algorithms `RSA OAEP` and `ECDH-ES+A256KW`
3. Mark `RSA PKCS#1 v1.5 padding scheme` as deprecated as it is not recommended.
4. Add unit tests based on `josekit` to check the compatibility of JWE used by `Response`.
5. Update rust toolchain to `1.83.0`

depends on
https://github.com/virtee/kbs-types/pull/53


Xynnn007:
Ok the test error happened because the guest-components side code is old. I just put another commit that references my repo to include the commit that includes the AEAD update.

Cargo.toml
```diff
@@ -29,14 +29,20 @@ config = "0.13.3"
 env_logger = "0.10.0"
 hex = "0.4.3"
 jwt-simple = "0.11"
-kbs_protocol = { git = "https://github.com/confidential-containers/guest-components.git", rev="e6999a3c0fd877dae9e68ea78b8b483062db32b8", default-features = false }
+kbs_protocol = { git = "https://github.com/Xynnn007/guest-components.git", rev = "2bfe3559bf273ddc835817498843f49c85359bbf", default-features = false }
```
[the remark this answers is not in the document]
mythi: I originally envisioned the opposite order because with Trustee changes available first, GC can test against `:latest` image that has the necessary changes available.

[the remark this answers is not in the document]
Xynnn007: Oh, ok I misunderstood your point. As now we imports GC code to cover e2e test of kbs. We could only choose the GC reference to repo rather than CoCo's for this PR to make the e2e test pass. Or, it will refer to the old version of GC code thus failed like [this](https://github.com/confidential-containers/trustee/actions/runs/12002528508/job/33454729423) due to different decryption logic of JWE.

So there are two ways
1. merge KBS side code firstly with gc git reference of personal development branch. Then merge GC side. Then go back to KBS side to change the gc reference to CoCo's
2. merge GC side code firstly with kbs image built from personal development code of kbs. Then merge kbs side. Then go back to gc to change back the image reference.


I think either is good to me.


What is not pretty clear to me are
1. Do we need to update minor number of kbs protocol for this patch? I guess yes. But the tricky thing is that the document of protocol is not bad, but the code.
2. Do we need to make a change log file for historical kbs protocol updates?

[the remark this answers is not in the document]
mkulke: > 1. Do we need to update minor number of kbs protocol for this patch? I guess yes. But the tricky thing is that the document of protocol is not bad, but the code.

I think yes, since this is pretty close to a schema change since tag was always an empty string. 

> 2. Do we need to make a change log file for historical kbs protocol updates?

Maybe it's sufficient to have a good commit message that will then end up in a generated changelog

> Do we want to move kbs_protocol crate to trustee, or combine GC and trustee repos in some future?

If we move the kbs_protocol crate to this repo, we wouldn't have a dependency from trustee to guest-components anymore? If that's the case, I think that would be a way to break the circular dependency, if I don't miss anything.

then let's maybe move the crate (in a separate PR) before we merge this PR?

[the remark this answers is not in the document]
mythi: > I originally envisioned the opposite order because with Trustee changes available first, GC can test against `kbs:latest` image that has the necessary changes available.

Actually, sorry. I checked the `kbs-types` bump case and it was indeed like you suggested: GC was merged first.

This is not only about keeping CI happy but also to notify users "both sides need updating" (which is important now as Trustee and GC can have their own releases). Without a `ProtocolVersion` error, users might just see an error without a clue what's wrong. Only `patch` version bump would be sufficient to make progress.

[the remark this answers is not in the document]
Xynnn007: > If we move the kbs_protocol crate to this repo, we wouldn't have a dependency from trustee to guest-components anymore? If that's the case, I think that would be a way to break the circular dependency, if I don't miss anything.

We still needs to import `attester`, `resource_uri` and `crypto` from GC. Whether these dependencies will cause problems is still open to question. If we need a quick fix on `attester` to make AA work, it would make us to have two PRs -- one in GC and one in trustee.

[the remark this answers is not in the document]
mkulke: right, I think that's too much of a refactoring then. I suspect with the kbs-client moving into dedicated set of tools (potentially into another repo), it's maybe better to wait how this will look like.

Xynnn007:
cc @deeglaze

Xynnn007:
Test locally and passed. Let's wait for https://github.com/virtee/kbs-types/pull/45 to be merged and update the `kbs-types` crate rev.

kbs/docs/kbs_attestation_protocol.md
```diff
@@ -227,6 +229,12 @@ Encryption algorithm used to encrypt the output of the KBS service API.
 The output of the KBS service API. It must be encrypted with the KBS-generated
 ephemeral key.
 
+- `aad` (Required if AEAD is used)
```
Xynnn007, line 232: Hi @tylerfanelli I added the `aad` and `tag` parts to this document. They all follow [RFC7516](https://datatracker.ietf.org/doc/html/rfc7516#appendix-A.2.3)

tylerfanelli:
> Test locally and passed. Let's wait for [virtee/kbs-types#45](https://github.com/virtee/kbs-types/pull/45) to be merged and update the `kbs-types` crate rev.

Released [here](https://crates.io/crates/kbs-types/0.9.0).

Cargo.toml
```diff
@@ -33,7 +33,7 @@ jwt-simple = { version = "0.12", default-features = false, features = [
     "pure-rust",
 ] }
 kbs_protocol = { git = "https://github.com/confidential-containers/guest-components.git", rev = "e6999a3c0fd877dae9e68ea78b8b483062db32b8", default-features = false }
-kbs-types = "0.7.0"
+kbs-types = { git = "https://github.com/virtee/kbs-types.git", rev = "a704036" }
```
mkulke, line 36: looks like `ProtectedHeader::generate_aad()` is not released yet. Will we do another kbs-types release before we merge this?

mkulke (commented):
should we maybe verify the validity of the kbs token by consuming it in another JWE implementation?

I was trying to use jwcrypto from python to decrypt a the payload in a token it failed w/ `jwcrypto.jwe.InvalidJWEData: No recipient matched the provided key['Failed: [InvalidTag()]']` I haven't really looked into it yet, though.

Cargo.toml
Xynnn007: I think it is ok to use a rev here. Once we have another release we can update then.

Xynnn007:
> should we maybe verify the validity of the kbs token by consuming it in another JWE implementation?

Nice catch. Let me add some unit tests. Btw I found some bugs. let me fix.

mkulke:
> > should we maybe verify the validity of the kbs token by consuming it in another JWE implementation?
> 
> Nice catch. Let me add some unit tests. Btw I found some bugs. let me fix.

yeah, we should definitely add a unit test. I have looked at that initially, however the code probably needs to be reworked a bit to allow injecting rng's with a fixed seed to test against a fixture.

kbs/src/jwe.rs
```diff
@@ -43,23 +53,65 @@ pub fn jwe(tee_pub_key: TeePubKey, payload_data: Vec<u8>) -> Result<Response> {
...
+
+    use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
+    use josekit::jwe::{alg::rsaes::RsaesJweAlgorithm::Rsa1_5, JweContext, JweHeader};
+    use kbs_types::TeePubKey;
+    use openssl::rsa::Rsa;
+
+    use super::jwe;
+
+    #[test]
+    fn jwe_compability() {
```
Xynnn007, line 82: @mkulke Following your nice idea, I added this unit test and found some potential bugs of serialization of jwe. Now fixed in upstream and this unit test passes locally on my side. 

cc @tylerfanelli with https://github.com/virtee/kbs-types/pull/53 we could make the response fully compatible with `josekit` crate which is rust crate that facilitates `jwe`.

mkulke:
image

ohhh, sorry that was unintended. will undo. I used the gh cli and I wasn't that "update-branch" would update the remote. interesting 😅

Xynnn007:
> edit: actually, @Xynnn007 that's a branch in your fork, I probably shouldn't mess any more with it. can you just reset to [7f00472](https://github.com/confidential-containers/trustee/commit/7f00472a01813e4a593666341e2a679758f67fbe) and push again?

Ok. No worries : )

Xynnn007:
locally test. The minimum version to build this is `1.80.0`. Given that trustee is not built in kata stacks, we might be able to use newer rust toolchain in ci?

mkulke:
also, we get a build warning now, similar to jwcrypto, RSA1_5 seems to be deprecated in josekit:

> warning: use of deprecated variant `josekit::jwe::RSA1_5`: This algorithm is no longer
recommended.

is this something that should concern us? can we change our TEE key algo, it might be a good point in time since we're about to have a breaking protocol change anyway

mkulke (approved):
Ok this looks better now. using key + token from a test run:

```python
>>> jwe_token = jwe.JWE(algs=["RSA1_5", "A256GCM"])
>>> jwe_token.deserialize(jwe_rust_json)
>>> jwe_token.decrypt(key)
>>> jwe_token.plaintext
b'this is a test data'
```

However, we also have to specifically allow "RSA1_5" here, otherwise, you'll get an "Algorithm not allowed" error.

Xynnn007:
> also, we get a build warning now, similar to jwcrypto, RSA1_5 seems to be deprecated in josekit:

This is my plan. To use ecdsa keys in this PR, too

mkulke:
> > also, we get a build warning now, similar to jwcrypto, RSA1_5 seems to be deprecated in josekit:
> 
> This is my plan. To use ecdsa keys in this PR, too

hmm, I wouldn't switch to EC for TEE keys without more consideration. I think for attestation-agents that cannot use openssl, we'd have to rely on some crates with unclear maturity status.

deeglaze:
> > > also, we get a build warning now, similar to jwcrypto, RSA1_5 seems to be deprecated in josekit:
> > 
> > 
> > This is my plan. To use ecdsa keys in this PR, too
> 
> hmm, I wouldn't switch to EC for TEE keys without more consideration.

All the technologies depend on ECC already, and it’s much faster to generate keys at boot for ECC. The coconut-svsm project will be switching to ecc from rsa.

mkulke:
> All the technologies depend on ECC already, and it’s much faster to generate keys at boot for ECC. The coconut-svsm project will be switching to ecc from rsa.

I guess that's fair. I remember asking about this at one point and there was a limitation for some TEE that would make switching the TEE keypair to EC hard. I don't remember details though and the concern might not be valid anymore. We can just go ahead, and we'll notice problems on a CI somewhere.

deeglaze:
> also, we get a build warning now, similar to jwcrypto, RSA1_5 seems to be deprecated in josekit:
> 
> > warning: use of deprecated variant `josekit::jwe::RSA1_5`: This algorithm is no longer
> > recommended.
> 
> is this something that should concern us, can we change our TEE key algo? it might be a good point in time since we're about to have a breaking protocol change anyway

Please. I’d advocate for a couple changes actually. KBS ought to distinguish between keys and secrets to allow for different handling. For example, we would prefer to store a key encryption key (aes-256-kw or kwp) wrapped by an HSM key and use attestation as part of authorization to rewrap the key to a key derived from a shared secret between the TEE and HSM. This is the COSE algorithm ECDH-ES-HKDF-256. The salt for the hkdf can be a parameter. It does mean sending more data to kbs, but it allows kbs to be less stateful, and it allows keys to be hsm protected in an attestable way, provided you have key creation attestations for the wrapping key. The key material is only visible to the HSM and the TEE. You’ll  want to ensure through attestation that the TEE key forwarded from the verifier to the HSM is only the one from the TEE to make sure the verifier isn’t sneakily rewrapping the key to its own key too, but we’ve known that verifiers should be attested.

Initial provisioning of keys from first boot becomes a requirement.

When you have keys protected in their own way, you can use the appropriate algorithms for key wrapping in transit. For secrets like rollback counters, you can use the current JWE, since there’s generally no service-specific non-extraction requirement for secrets the way there are with keys.

Xynnn007:
@deeglaze I am not sure I got your point well. I suppose that you are referring to the secret storage that we call "resource backend". Now it supports [HSM](https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/implementations/resource/pkcs11.rs).

To update the algorithm from RSAv1.5 is also a workaround I think. A better solution might be to involve DH Key Agreement to generate an ephermeral symmetric key to encrypt the dataflow every request to KBS. This could be done by leveraging existing TLS stacks.

mkulke:
> Please. I’d advocate for a couple changes actually. KBS ought to distinguish between keys and secrets to allow for different handling. For example, we would prefer to store a key encryption key (aes-256-kw or kwp) wrapped by an HSM key and use attestation as part of authorization to rewrap the key to a key derived from a shared secret between the TEE and HSM. This is the COSE algorithm ECDH-ES-HKDF-256. The salt for the hkdf can be a parameter. It does mean sending more data to kbs, but it allows kbs to be less stateful, and it allows keys to be hsm protected in an attestable way, provided you have key creation attestations for the wrapping key. The key material is only visible to the HSM and the TEE. You’ll want to ensure through attestation that the TEE key forwarded from the verifier to the HSM is only the one from the TEE to make sure the verifier isn’t sneakily rewrapping the key to its own key too, but we’ve known that verifiers should be attested.
> 
> Initial provisioning of keys from first boot becomes a requirement.
> 
> When you have keys protected in their own way, you can use the appropriate algorithms for key wrapping in transit. For secrets like rollback counters, you can use the current JWE, since there’s generally no service-specific non-extraction requirement for secrets the way there are with keys.

thx, it might be good to copy your thoughts to a discrete issue on this repo (verbatim is fine), since we probably won't fully address it in this PR (i hope 😅) and it's easier to track this way. I understand there are venues to decouple kbs from the handling of key material, so I feel there might be a conceptual overlap with sealed secrets/KMS/HSM backends and it's probably good to understand which role those facilities would play in the suggested scheme. There's also a dedicated [Trustee Sync](https://docs.google.com/document/d/1js1GVnKwzNJX3HUPI4q6nVRepmtwA14iw-yuLRSYyoY), which would be a great place to present those ideas.

Xynnn007:
@deeglaze Thanks for the suggestion! If interested, please feel free to join our bi-weekly meeting mentioned by Magnus to share your great ideas.

I replaced RSA PKCS1v15 with `ECDH-ES+A256KW` (Concat KDF is used as KDF) on curve P256.

I still worry about the perfect forward secrecy that seems not to be fully conquered by JWE. Thus I am considering to promote it in future with existing TLS suites to improve this.

Xynnn007:
I did some change upon the PR description. Compability tests are passed with `josekit` crate. Now we can wait for upstream PR to get merged and then I rebase the dep.

fitzthum (approved):
LGTM. Thanks for taking care of this

tylerfanelli:
@Xynnn007  Is there any update on this? Changes still being made, or anything needed from `kbs_types`?

Xynnn007:
> @Xynnn007 Is there any update on this? Changes still being made, or anything needed from `kbs_types`?

I have rebased the code and let's see CI on this PR and https://github.com/confidential-containers/guest-components/pull/820. If both succeed, nothing else needed from `kbs-types`.

Xynnn007:
@tylerfanelli Looks fine now. Will you do a release for kbs-types?

tylerfanelli:
> @tylerfanelli Looks fine now. Will you do a release for kbs-types?

Sure, [released](https://crates.io/crates/kbs-types/0.10.0).

kbs/src/jwe.rs
```diff
@@ -43,23 +121,277 @@ pub fn jwe(tee_pub_key: TeePubKey, payload_data: Vec<u8>) -> Result<Response> {
...
-    {
-       "alg": RSA_ALGORITHM.to_string(),
-       "enc": AES_GCM_256_ALGORITHM.to_string(),
-    });
+    Ok(Response {
+        protected,
+        encrypted_key,
+        iv: iv.into(),
+        ciphertext: payload_data,
+        aad: None,
```
huoqifeng, line 134: set aad to None is deliberated? why?

kbs/src/jwe.rs
```diff
@@ -43,23 +121,277 @@ pub fn jwe(tee_pub_key: TeePubKey, payload_data: Vec<u8>) -> Result<Response> {
...
-            .context("serde protected_header failed")?,
-        encrypted_key: URL_SAFE_NO_PAD.encode(wrapped_sym_key),
-        iv: URL_SAFE_NO_PAD.encode(iv),
-        ciphertext: URL_SAFE_NO_PAD.encode(encrypted_payload_data),
-        tag: "".to_string(),
+        protected,
+        encrypted_key,
+        iv: iv.into(),
+        ciphertext: payload_data,
+        aad: None,
```
huoqifeng, line 235: set aad to None is deliberated? why?

huoqifeng (commented):
Thanks @Xynnn007 to fix the JWE problem, I left some comments.

kbs/src/jwe.rs
huoqifeng: oh, checking the client side code in https://github.com/confidential-containers/guest-components/pull/820/files `let aad = response.protected.generate_aad()?;`, I guess aad is generated but not passed. Got.

huoqifeng (approved):
LGTM

kbs/src/jwe.rs
Xynnn007: Yes. It's specified in https://datatracker.ietf.org/doc/html/rfc7516#section-5.1 step 14

fitzthum (approved):
Looks good. One small nit. I will merge the GC PR so we can rebase this one.

Xynnn007:
CI error can be fixed by https://github.com/confidential-containers/trustee/pull/716

Xynnn007:
took a rebase.

mythi:
983a5c7 commit message mentions some lint fixed but there are none. Did they morph into some other commit?

fitzthum:
Last chance for feedback before this is merged @confidential-containers/trustee-maintainers

PR #604, opened 2024-11-28
KBS/AS | Fix configuration items and update the docs

Xynnn007:
cc @fitzthum

### DIFF
--- a/kbs/test_data/configs/coco-as-builtin-3.toml
+++ b/kbs/test_data/configs/coco-as-builtin-3.toml
@@ -7,7 +7,6 @@ insecure_key = false
 [attestation_service]
 type = "coco_as_builtin"
 work_dir = "/opt/confidential-containers/attestation-service"
-policy_engine = "opa"
 timeout = 5
 
 [attestation_service.attestation_token_broker]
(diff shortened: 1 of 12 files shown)


kbs/test_data/configs/coco-as-builtin-3.toml
```diff
@@ -2,15 +2,18 @@
...
 [attestation_service]
 type = "coco_as_builtin"
 work_dir = "/opt/confidential-containers/attestation-service"
 policy_engine = "opa"
+timeout = 5
 
 [attestation_service.attestation_token_broker]
 type = "Simple"
 duration_min = 5
+policy_dir = "/opt/confidential-containers/attestation-service/simple-policys"
```
fitzthum, line 15: same as above

Xynnn007:
> Do we not hit this test in the CI?

The CI cover this but somehow false negative. When serializing with the serde library, undefined fields, such as `policy_engine`, will be ignored if we do not explicitly restrict them in the code. Considering that users may refer to the configuration file in the unit test, this patch deletes those "fields that are not valid and ignored by the code logic".

Besides, about the unit tests for configuration reading, we define the default value explicitly inside the code and do not add related items inside the config files. Thus the tests pass which is expected. What we changed in the patch is to explicitly add the items also inside the config files, although they are default values.

mkulke (approved):

PR #606, opened 2024-11-29
verifier: add judge logic for quote and eventlog string

Xynnn007:
This patch adds a check logic before trying to base decoding TDX/SGX quote and AAEL/CCEL.

This would do help to raise a meaningful error
message when an evidence with empty Quote is given to verifier module.

Also, when AAEL/CCEL is existed inside the evidence but without any value, the deserialization and parse will also be skipped.


mkulke (approved):

mythi:
@Xynnn007 thanks for this!

PR #610, opened 2024-12-03
token: fix simple token broker

fitzthum:
Fix bug where wrong claims were added to the token.

Also, rename claims to flattened_claims to be slightly clearer.


Xynnn007 (approved):
Nice. I test locally and passed. cc @lmilleri

mkulke (approved):

lmilleri:
It works for me as well with the trustee-operator. Thank you!

PR #612, opened 2024-12-04
chore: Update jwt-simple dependency

Xynnn007:
Some other related packages are also updated in Cargo.lock by `cargo update -p jwt-simple`

Manually for #608


Xynnn007:
I have tried locally only with `pure-rust` the dep can be built, or there will be link issues.

PR #617, opened 2024-12-09
Added checks for AS and KBS policy setting

jialez0:
In the current KBS and AS set policy interfaces, there is no validity check for policies, which can result in policies with syntax errors being successfully set and causing errors when subsequently calling the evaluation API. This is not helpful for policy setters to promptly identify errors. This PR adds policy validity checks to the KBS and AS set policy interfaces to avoid the aforementioned issues.


fitzthum (approved):
LGTM

Great idea. Thanks @jialez0

Xynnn007 (approved):

PR #619, opened 2024-12-10
Setup integration tests

fitzthum:
Creates a framework for writing Trustee integration tests in Rust.

Put your tests in `integration-tests/tests/integration.rs` (or a similar file) and they will run when you do `cargo test` at the top level. 

I've added two simple integration tests. They aren't anything beyond what we already have in other tests, but they add a structure that should help us to create a lot more. Before I get too carried away, please take a look at the setup.

Note that running the KBS in the background was trickier than expected. I think I found a pretty good way to do it in the end.

### DIFF
--- a/attestation-service/src/token/simple_default_policy.rego
+++ b/attestation-service/src/token/simple_default_policy.rego
@@ -28,10 +28,11 @@
 package policy
 
 import future.keywords.every
+import future.keywords.if
 
-default allow = false
+default allow := false
 
-allow {
+allow if {
 	every k, v in input {
 		# `judge_field`: Traverse each key value pair in the input and make policy judgments on it.
 		#
@@ -44,7 +45,7 @@ allow {
 	}
 }
 
-judge_field(input_key, input_value) {
+judge_field(input_key, input_value) if {
 	has_key(data.reference, input_key)
 	reference_value := data.reference[input_key]
 
@@ -57,16 +58,16 @@ judge_field(input_key, input_value) {
 	match_value(reference_value, input_value)
 }
 
-judge_field(input_key, input_value) {
+judge_field(input_key, input_value) if {
 	not has_key(data.reference, input_key)
 }
 
-match_value(reference_value, input_value) {
+match_value(reference_value, input_value) if {
 	not is_array(reference_value)
 	input_value == reference_value
 }
 
-match_value(reference_value, input_value) {
+match_value(reference_value, input_value) if {
 	is_array(reference_value)
 
 	# `array_include`: judge the input value with the values in the array.
@@ -78,16 +79,16 @@ match_value(reference_value, input_value) {
 	array_include(reference_value, input_value)
 }
 
-array_include(reference_value_array, input_value) {
+array_include(reference_value_array, input_value) if {
 	reference_value_array == []
 }
 
-array_include(reference_value_array, input_value) {
+array_include(reference_value_array, input_value) if {
 	reference_value_array != []
 	some i
 	reference_value_array[i] == input_value
 }
 
-has_key(m, k) {
+has_key(m, k) if {
 	_ = m[k]
 }
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "integration-tests"
+version.workspace = true
+authors.workspace = true
+description.workspace = true
+documentation.workspace = true
+edition.workspace = true
+
+[dependencies] 
+kbs = { path = "../kbs" }
+attestation-service = { path = "../attestation-service" }
+kbs-client = { path = "../tools/kbs-client" }
+
+actix-web.workspace = true
+actix-rt = "2.10.0"
+anyhow.workspace = true
+env_logger.workspace = true
+log.workspace = true
+openssl.workspace = true
+rstest.workspace = true
+serde_json.workspace = true
+serial_test.workspace = true
+tempfile.workspace = true
+tokio.workspace = true
--- a/integration-tests/tests/integration.rs
+++ b/integration-tests/tests/integration.rs
@@ -1,0 +1,218 @@
+// Copyright (c) 2024 by IBM.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use kbs::admin::config::AdminConfig;
+use kbs::attestation::config::{AttestationConfig, AttestationServiceConfig};
+use kbs::config::HttpServerConfig;
+use kbs::config::KbsConfig;
+use kbs::policy_engine::PolicyEngineConfig;
+use kbs::token::AttestationTokenVerifierConfig;
+use kbs::ApiServer;
+
+use kbs::plugins::{
+    implementations::{resource::local_fs::LocalFsRepoDesc, RepositoryConfig},
+    PluginsConfig,
+};
+
+use attestation_service::{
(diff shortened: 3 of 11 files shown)


kbs/src/api_server.rs
```diff
@@ -129,6 +129,51 @@ impl ApiServer {
             .await
             .map_err(|e| Error::HTTPFailed { source: e.into() })
     }
+
+    /// Setup API server
+    pub fn server(self) -> Result<actix_web::dev::Server> {
```
mkulke, line 134: this seems to share a lot of logic with `serve()` in the same file, can we share the code?

Xynnn007:
Looks great! One question before getting into the code:

why not we directly have an `tests` directory under `kbs` to have these integration tests, like what we do in [image-rs](https://github.com/confidential-containers/guest-components/tree/main/image-rs/tests)

A quick glance is that way we need to add `kbs_client` to `kbs`'s `[dev-dependencies]` section in `Cargo.toml`.

This way, we could avoid another sub crate.

fitzthum:
> why not we directly have an tests directory under kbs to have these integration tests, like what we do in [image-rs](https://github.com/confidential-containers/guest-components/tree/main/image-rs/tests)

I think this would work. I see the integration tests as having a broader scope than just the KBS, so I made a new package, but if that adds overhead I could move it. Generally I think that since the repo merge, we've had  some stuff in individual packages that should really be in the top of the repo (like docs, perhaps).

integration-tests/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+actix-rt = "2.10.0"
+anyhow = "1.0.94"
+env_logger = "0.11.5"
+log = "0.4.22"
+openssl = "0.10.68"
+rstest.workspace = true
+serde_json = "1.0.133"
+serial_test = { version = "3.2.0", features = ["async"]}
+tempfile = "3.14.0"
+tokio = { version = "1.42.0", features = ["full"] }
```
Xynnn007, line 24: Can we try to mark them `xxx.workspace = true` to use same dep version with workspace as possible as we can?

integration-tests/tests/integration.rs
```diff
@@ -0,0 +1,222 @@
...
+    package policy
+    allow = true
+";
+
+const DENY_ALL_POLICY: &str = "
+    package policy
+    allow = false
+";
+
+#[allow(dead_code)]
```
Xynnn007, line 48: Why do we need this lint macro?

integration-tests/tests/integration.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    async fn get_secret(&self, secret_path: String) -> Result<Vec<u8>> {
+        info!("TEST: Getting Secret");
+        let resource_bytes =
+            kbs_client::get_resource_with_attestation(KBS_URL, &secret_path, None, vec![]).await?;
+
+        Ok(resource_bytes)
+    }
+
+    fn wait(&self) {
```
Xynnn007, line 204: what is the reason that we do not use async version here?

Xynnn007 (commented):
I think it would make sense to have a separate crate if we want to add more tests. One question is that we are having some e2e test
- docker-compose ci
- e2e test [here](https://github.com/confidential-containers/trustee/tree/main/kbs/test)
- this PR

Do we need to delete some duplicated ones?

integration-tests/tests/integration.rs
fitzthum: The `Custom` variant is never instantiated. I could move the dead code thing just to that variant.

fitzthum:
@Xynnn007 

> Do we need to delete some duplicated ones?

Since these integration tests use the crate directly, they don't test the binaries or the docker-compose configuration or the k8s configuration. I think it's fine to keep some simple tests around that make sure that those interfaces still work. We should do a little bit of cleanup and organization with our existing tests, but I think that's for another PR.

kbs/src/api_server.rs
fitzthum: Ok, I tried a bunch more things and I finally have something that works and avoids the duplication. PTAL

attestation-service/src/token/simple_default_policy.rego
```diff
@@ -28,10 +28,11 @@
 package policy
 
 import future.keywords.every
+import future.keywords.if
```
portersrc, line 31: what's going on here? was `if` not in the original simple-default-policy and should have been (so you added it)?

mkulke (approved):

kbs/Makefile
```diff
@@ -81,7 +81,7 @@ uninstall:
 	rm -rf $(INSTALL_DESTDIR)/kbs $(INSTALL_DESTDIR)/kbs-client $(INSTALL_DESTDIR)/issuer-kbs $(INSTALL_DESTDIR)/resource-kbs
 
 check:
-	cargo test -p kbs -p kbs-client
+	cargo test -p kbs -p kbs-client -p integration-tests
```
portersrc, line 84: This change will get hit [here](https://github.com/confidential-containers/trustee/blob/714b653f3253cb17718df56da9165b86503c0861/.github/workflows/kbs-rust.yml#L81) in our regular workflows, right? (If so, that's what we want, I think.)

fitzthum: Yeah. I considered adding another workflow for the integration tests, but I think it's best to keep it together. Note that this is just for checking the code itself, the actual tests will get run via the existing call to `cargo test`

attestation-service/src/token/simple_default_policy.rego
fitzthum: Our CI does a formatting check of the default policy using the OPA package. We don't pin the version of the package so sometimes it wants new things. Recently it has started complaining about not having `if` before blocks.

The versioning of OPA is a little squirrely but it hasn't been too big of an issue so far.

portersrc (approved):
lgtm

integration-tests/tests/integration.rs
fitzthum: Sure. I had to comment out some other stuff since we have the `set_policy` function.

fitzthum:
@Xynnn007 PTAL

integration-tests/tests/integration.rs
```diff
@@ -0,0 +1,218 @@
...
+        self.set_secret(SECRET_PATH.to_string(), SECRET_BYTES.as_ref().to_vec())
+            .await?;
+        self.set_policy(PolicyType::AllowAll).await?;
+
+        let secret = self.get_secret(SECRET_PATH.to_string()).await?;
+
+        assert_eq!(secret, SECRET_BYTES);
+        info!("TEST: test completed succesfully");
+
+        kbs_handle.stop(true).await;
```
Xynnn007, line 151: Here we don't need to explicitly kill KBS task, and we can simplify 136~139 lines to
```rust
actix_web::rt::spawn(async { api_server.serve().await });
```
This can avoid commit https://github.com/confidential-containers/trustee/pull/619/commits/c65676cb383115fc3785615eb0f545bbee29d72b. Because the `run` function is only called once in each test case, and each test case has another sub-`#[actix_rt:test]` macro, which is equivalent to creating a sub-asynchronous runtime, the whole sub-runtime will be automatically destroyed after the test case ends, which means that the KBS coroutine/task just created will also be destroyed.

Xynnn007 (approved):
We could do some code structure optimization later but it now looks good to me.


Sorry for the notifiaction mail has been drowned in my mailbox.

PR #620, opened 2024-12-10
ear: remove productId from default policy

fitzthum:
The sample attester does not report a fake productId field, but we have used productId in our tests since before EAR.

The EAR policy erroneously expects the evidence to contain a productId, this leads to generating contra-indicated tokens for EAR.

Remove this field and update the tests.

cc @lmilleri


Xynnn007 (approved):

huoqifeng (commented):
LGTM

huoqifeng (approved):

PR #623, opened 2024-12-12
KBS: use RSA-OAEP alg for JWE symkey encryption 

mkulke:
draft, since it depends on #597

The usage of RSA1_5 is deprecated:

https://www.ietf.org/archive/id/draft-madden-jose-deprecate-none-rsa15-00.html#section-1.2

We can switch to RSA-OAEP by using the padding in the openssl rsa impl. Once we switched to this implementation we need a small change on guest-components side to set the same alg.


Xynnn007:
FYI, I have added the RSA-OAEP in [this commit](https://github.com/confidential-containers/trustee/pull/597/commits/e49016a205effd24045c092945993bc73d25213d) also.

PR #626, opened 2024-12-13
attestation-service: implement *_policy() methods for EAR token broker

mythi:
without the changes, kbs-client "set-attestation-policy" gets "not supported" errors.

the changes are taken from 'simple' token broker. Both token brokers have a policy_engine instance created.

### DIFF
--- a/attestation-service/src/token/ear_broker.rs
+++ b/attestation-service/src/token/ear_broker.rs
@@ -299,6 +299,27 @@ impl AttestationTokenBroker for EarAttestationTokenBroker {
 
         Ok(signed_ear)
     }
+
+    async fn set_policy(&self, policy_id: String, policy: String) -> Result<()> {
+        self.policy_engine
+            .set_policy(policy_id, policy)
+            .await
+            .map_err(Error::from)
+    }
+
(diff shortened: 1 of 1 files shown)


mkulke (commented):
I assume this is setting the policy in memory for the process? so, the policy would be volatile and also potentially split-brained for multiple replicas?

it might make sense to do implement for consistency's sake. but IMO we should really re/consider how we want to handle state in KBS/AS and not expose stateful methods unless we know how we want to address with it.

If it's required for tests in the mean time, I could imagine that the policy storage is backed by a k8s configmap, a docker-compose mount or something.

mythi:
OPA policy engine seems to writes these to the disk and at least the docker-compose backed mount stores it persistently (on the AS side, not KBS).

> it might make sense to do implement for consistency's sake. but IMO we should really re/consider how we want to handle state in KBS/AS and not expose stateful methods unless we know how we want to address with it.

yeah makes sense. right now this is just to be able to experiment with EAR policies uploaded using `kbs-client config set-attestation-policy` the same way as with `simple` broker.

Xynnn007 (approved):

fitzthum:
maybe we should make an issue for tracking the statefulness stuff

PR #627, opened 2024-12-13
cluster: use latest staged images

mythi:
ad991b7 disabled image pull because that pointed to release images and often times they had conflicts with local files in the repository.

As we now have staged images published on merge, we can try pulling images without forcing local builds (or local .yml editing).

Also, Compose warns about 'version' being obsolete so drop that too, while we're at it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,10 +1,9 @@
-version: '3.2'
 services:
   kbs:
     build:
       context: .
       dockerfile: kbs/docker/coco-as-grpc/Dockerfile
-    #image: ghcr.io/confidential-containers/key-broker-service:latest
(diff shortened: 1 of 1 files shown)


fitzthum (commented):
I've gotten accustomed to using docker compose for development. What's the motivation for switching back to the images? Build time?

mythi:
`docker compose build` remains the option for local builds still. With this, there's flexibility to choose how images are prepared without having to edit the compose file.

fitzthum (approved):
Ok, the docker compose docs are a little unclear on the behavior here but I tested this and it seems to work fine.

mkulke (approved):

PR #633, opened 2024-12-16
Improve RVPS storage config

### LINKED ISSUE #629: Rework RVPS store config
The RVPS config currently looks like this 

```
pub struct Config {
    #[serde(default = "default_store_type")]
    pub store_type: String,

    #[serde(default = "default_store_config")]
    pub store_config: Value,
}
```

This doesn't really match how we've recently been creating configs inside of an enum and have `store_config` as an opaque JSON value doesn't seem ideal.

We should probably update this to have an enum containing a config, like we've started doing elsewhere in Trustee.

fitzthum:
Improves the RVPS config in two ways

1) Change the storage config to follow the pattern we are using across Trustee where we use an enum to contain the module type and the variant contains the appropriate config struct. Previously we were using two fields; one for the type and one for an opaque JSON structure.

2) Get rid of the server config. Previously we had a separate config struct for when the RVPS runs as a server. This duplicated a lot of config values and only added one more field, the address for the RVPS to server. To simplify things, I moved this to be a CLI arg and got rid of this extra config completely. There is a case to be made for keeping the server config, but I think the cli arg makes more sense and is similar to what we do with some other components.

I know there is a bit of overhead when changing the config, but I think this was somewhat inevitable as we begin to refine the RVPS. cc @confidential-containers/kbs-operator-maintainers 

Fixes #629


rvps/src/bin/rvps.rs
```diff
@@ -47,9 +56,9 @@ async fn main() -> Result<()> {
         Config::default()
     });
 
-    info!("Listen socket: {}", config.address);
+    info!("Listen socket: {}", &cli.address);
 
-    let socket = config.address.parse().context("parse socket addr failed")?;
+    let socket = &cli.address.parse().context("parse socket addr failed")?;
```
Xynnn007, line 61: we can delete the `&` here and thus do not need `*` on line 63

Xynnn007 (commented):
Thanks for the refactoring. Looks like the change upon `ReferenceValueStorageConfig` would make AS/KBS 's config have another subfield
```toml
[rvps.storage]
type = "..."
...
```

Could you help to add the sections to [AS config doc](https://github.com/confidential-containers/trustee/blob/main/attestation-service/docs/config.md#builtin-rvps) and [KBS config doc](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/config.md#builtin-rvps)? Also, if an unit test with explicit `rvps` section (e.g. [here](https://github.com/confidential-containers/trustee/blob/main/kbs/test_data/configs/coco-as-builtin-2.toml)) would be perfect.

docker-compose.yml
```diff
@@ -52,6 +52,11 @@ services:
     volumes:
       - ./kbs/data/reference-values:/opt/confidential-containers/attestation-service/reference_values:rw
       - ./kbs/config/rvps.json:/etc/rvps.json:rw
+    command: [
```
huoqifeng, line 55: Not sure other verifier, I think `SE` does have something related document here https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/se/README.md need be revised accordingly, maybe need a new PR to address this change in SE doc. CC @stevenhorsman @bpradipt

stevenhorsman: Thanks to the notice - cc @gauravkuredhat

fitzthum: oh yeah I didn't update any of the docs. I will add a commit to do that

fitzthum: updated

Xynnn007 (approved):

huoqifeng (approved):
LGTM thanks! @fitzthum

PR #640, opened 2024-12-20
Improve EAR policy

fitzthum:
There is no need to set claims for every platform and compare them with min. We can just set the claims directly and rely on short-circuiting to skip any platforms that don't have TCB claims set.

### DIFF
--- a/attestation-service/src/token/ear_default_policy.rego
+++ b/attestation-service/src/token/ear_default_policy.rego
@@ -84,7 +84,7 @@ sample_hardware := 2 if {
 ##### SNP
 snp_executables := 3 if {
 	# In the future, we might calculate this measurement here various components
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
Thanks for this. Do you think we need example policies for different platforms?

fitzthum:
> Do you think we need example policies for different platforms?

This policy is still intended to be used with all the platforms we support. Previously I defined intermediate claims such as `snp_hardware` and then produced one claim from all of the `*_hardware` claims. Now I just define `hardware` directly. If the evidence for the platform isn't there, the definition will be short-circuited.

We do still have TODOs at the bottom to add logic for other platforms. I am hoping that all our platforms will be defined here soon but I may need a bit of help with that from platform experts.

PR #645, opened 2024-12-26
chore: update to x509-parser 0.16.0 and asn1-rs 0.6.2

Xynnn007:
The two dependencies have version dependencies so we update them together.

Close #644


fitzthum (approved):
LGTM

PR #654, opened 2025-01-07
k8s: remove token signer from k8s config

fitzthum:
The Kata CI does not setup the attestation token signing keys. This causes the KBS to break when we bump the repo version and pick up this new config which had the signing key specified.

In the future we can change the CI to setup the signing keys (although we are already testing this in our Makefile test), but for now let's stick with the existing behavior.


fidencio (approved):
lgtm, thanks @fitzthum!

mythi:
> The Kata CI does not setup the attestation token signing keys

perhaps we should have some simple `kind` based k8s deployment tests similar to `kbs-docker-e2e.yml` to check the kubernetes setup works too.

PR #657, opened 2025-01-09
kbs: Bail if the jwk sets cannot be downloaded

fidencio:
Instead of just warning and proceeding, as done right now[0], let's bail as the current behaviour leads to an unusable KBS[1].

By bailing earlier at that point, when trustee pod is being deployed by the trustee-operator, we ensure that the pod will error out and kubernetes will take care of restarting the pod till its startup properly succeeds.

[0]:
```
[INFO  kbs] Using config file /etc/kbs-config/kbs-config.json
[WARN  kbs::token::jwk] error getting JWKS: SourceAccess("error sending request for url (https://portal.trustauthority.intel.com/.well-known/openid-configuration)")
[INFO  kbs] Starting HTTP server at [0.0.0.0:8080]
[WARN  kbs::token::jwk] error getting JWKS: SourceAccess("error sending request for url (https://portal.trustauthority.intel.com/.well-known/openid-configuration)")
[INFO  actix_server::builder] starting 56 workers
[INFO  actix_server::server] Tokio runtime found; starting in existing Tokio runtime
```

[1]:
```
[INFO  actix_web::middleware::logger] 10.128.0.32 "POST /kbs/v0/attest HTTP/1.1" 401 218 "-" "attestation-agent-kbs-client/0.1.0" 0.279838
[INFO  kbs::http::attest] Auth API called.
[INFO  actix_web::middleware::logger] 10.128.0.32 "POST /kbs/v0/auth HTTP/1.1" 200 108 "-" "attestation-agent-kbs-client/0.1.0" 0.000334
[INFO  kbs::http::attest] Attest API called.
[INFO  kbs::attestation::intel_trust_authority] POST attestation request ...
[ERROR kbs::http::error] Attestation failed: Failed to verify attestation token

    Caused by:
        Cannot verify token since trusted JWK Set is empty
```

### DIFF
--- a/kbs/src/token/jwk.rs
+++ b/kbs/src/token/jwk.rs
@@ -93,7 +93,7 @@ impl JwkAttestationTokenVerifier {
         for path in config.trusted_jwk_sets.iter() {
             match get_jwks_from_file_or_url(path).await {
                 Ok(mut jwkset) => trusted_jwk_sets.keys.append(&mut jwkset.keys),
-                Err(e) => log::warn!("error getting JWKS: {:?}", e),
+                Err(e) => bail!("error getting JWKS: {:?}", e),
             }
         }
 


kbs/src/token/jwk.rs
```diff
@@ -93,7 +93,7 @@ impl JwkAttestationTokenVerifier {
         for path in config.trusted_jwk_sets.iter() {
             match get_jwks_from_file_or_url(path).await {
                 Ok(mut jwkset) => trusted_jwk_sets.keys.append(&mut jwkset.keys),
-                Err(e) => log::warn!("error getting JWKS: {:?}", e),
+                Err(e) => bail!("error getting JWKS: {:?}", e),
```
fitzthum, line 96: I am open to this, but it might be too strict to bail if we have a single JWK path that can't be reached at startup. At this point we don't know if we need these. That isn't really determined until we try to verify a token, which is where we currently bail. Anyway I can understand bailing if something is provided that can't be reached, but on the other hand maybe someone wants a config where they specify a large set of paths and don't really care if some of them fail unless they're actually needed.

wdyt @Xynnn007

fidencio: I can adjust this to bail with all the paths failed, WDYT?

fidencio: Adjusted the code to only bail in case no JWK set was found.

Mind that the loop below the check will not be executed in case the JWK set is found, and Ok() would be incorrectly returned in that case.

fitzthum (approved):
LGTM

kbs/src/token/jwk.rs
Xynnn007: I think it is better to bail ASAP at the launch time. The typical use case is a user configures all jwk paths he will use, so if any of them cannot be read it should fail to launc. It might be confusing to have partial jwk path configs that cannot work and hard to debug.

bpradipt (approved):
/lgtm
Nice :-)

kbs/src/token/jwk.rs
mythi: > Anyway I can understand bailing if something is provided that can't be reached, but on the other hand maybe someone wants a config where they specify a large set of paths and don't really care if some of them fail unless they're actually needed.

that `warn()` has a bit of history because we used to have CoCo AS "trusted certs" and ITA "trusted JWK sets" combined in the same [config entry](https://github.com/confidential-containers/trustee/blob/68607d4300dda5a8ae948e2562fd06d09cbd7eca/kbs/src/token/mod.rs#L34-L39). Thanks to @Xynnn007's work these are now separated entries.

I think we want the behavior to be consistent between `trusted_jwk_sets` and `trusted_certs_paths`. The latter fails immediately if a file in the list cannot be read.

mkulke: +1 on bailing out early. If the user explicitly specifies a URL or a file which cannot be read, I'd expect the process to fail at launch instead of some attempt at graceful degradation.

fidencio: @Xynnn007, @mkulke,

I got it back to the version of bailing out early.

mkulke (approved):

mythi (approved):
LGTM too

bpradipt (approved):
/lgtm

PR #658, opened 2025-01-10
verifier: Rename user_data to report_data in SeAttestationClaims

BbolroC:
The EAR token broker does not insert the `report_data` for SE attestation claim because there is no matching field in `SeAttestationClaims`. The absence leads to `TokenVerifierError(NoTeePubKeyClaimFound)` after successful attestation.

As an interim solution, this PR renames the existing `user_data` to `report_data`, enabling the token broker to perform its task correctly.

### DIFF
--- a/deps/verifier/src/se/ibmse.rs
+++ b/deps/verifier/src/se/ibmse.rs
@@ -90,7 +90,7 @@ pub struct SeAttestationResponse {
 pub struct SeAttestationClaims {
     #[serde_as(as = "Hex")]
     cuid: ConfigUid,
-    user_data: String,
+    report_data: String,
     version: u32,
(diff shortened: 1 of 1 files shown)


mkulke (approved):

fitzthum (approved):
For some context, the EAR tokens will only expose the TEE public key if the verifier returns the `report_data` claim. This is by design. It means that for verifiers that don't check the binding of the report data, the token isn't very useful.

To work around this, we can add that claim to the output of the SE verifier. This is only a temporary workaround. The SE verifier will need to be improved to actually check the binding.

fitzthum:
Also, btw I am preparing a PR to fix a similar issue for SNP, where we don't report `report_data` even though we do check the binding.

PR #659, opened 2025-01-10
Snp add report claim

fitzthum:
The EAR broker expects the verifier to include `report_data` in the TCB claims as a signal that the verifier has checked the binding of the report data and the hw evidence.

Since we do check `report_data` in the SNP verifier, add the claim. Also add `init_data` claim for similar reasons.


fitzthum:
Need this for the release btw @Xynnn007 @mkulke

Xynnn007 (approved):

PR #664, opened 2025-01-16
Use Trustee-specific TCB Claims Extension

fitzthum:
As a simplification, we are currently storing the tcb claims in the annotated evidence field of the EAR Appraisal. This field is actually an extension defined by veraison, so we end up with a token that had a veraison field in it even though we don't really follow the veraison protocol/spec.

Fortunately some recent changes to the EAR crate allow us to stick the tcb claims into our own extension. That's what this PR does. See commit messages for details. 

This is going to be marked as a draft until the EAR crate has a release.


attestation-service/src/token/ear_broker.rs
```diff
@@ -245,6 +243,10 @@ impl AttestationTokenBroker for EarAttestationTokenBroker {
 
         let mut appraisal = Appraisal::new();
 
+        appraisal
+            .extensions
+            .register(TCB_CLAIMS_EXTENSION, -41121, ear::RawValueKind::Map)?;
```
tylerfanelli, line 248: Any specific purpose to `-41121`?

fitzthum: Just some random number in the private extensions space

Xynnn007:
There is some new information about EAR draft https://www.ietf.org/archive/id/draft-ietf-rats-ear-04.html#section-3.1-2.14.1

Now `ear_attester_claims` and `ear_verifier_claims` are appraisal-level things. Then we can put original evidence claims into `ear_attester_claims`; `initdata`, `report_data` and user defined things into `ear_verifier_claims`.

PR #666, opened 2025-01-17
ita: added processing event logs in SGX & TDX context

pawelpros:
Added processing base64 encoded CC event logs from tee evidence and map them to ITA json payload format


mythi:
we are going to have the option to submit the raw CCEL through the API so we can skip the JSON conversion. We'll reuse parts of this PR once that is available.

PR #667, opened 2025-01-17
ear: add TDX sample policy checks

mythi:
Something I drafted together with @jorgealmansa.

Populate the default EAR policy with an initial TDX policy configuration.

### DIFF
--- a/attestation-service/src/token/ear_default_policy.rego
+++ b/attestation-service/src/token/ear_default_policy.rego
@@ -86,7 +86,33 @@ else := 3 if {
 	input.snp.policy_migrate_ma == 0
 }
 
-##### TDX TODO
+##### TDX
+executables := 3 if {
+	# Check the kernel, initrd, and cmdline (including dmverity parameters) measurements
+	# TODO: add individual CCEL measurements from input.tdx.ccel instead
+	input.tdx.quote.body.rtmr_1 in data.reference.rtmr_1
+	input.tdx.quote.body.rtmr_2 in data.reference.rtmr_2
+}
+
+hardware := 2 if {
+	# Check the quote is a TDX quote signed by Intel SGX Quoting Enclave
+	input.tdx.quote.header.tee_type == "81000000"
+	input.tdx.quote.header.vendor_id == "939a7233f79c4ca9940a0db3957f0607"
+
+	# Check TDX Module version and its hash. Also check OVMF code hash.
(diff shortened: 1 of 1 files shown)


fitzthum (approved):
Awesome. I don't know enough about TDX to really confirm this but it looks like a big step forward.

mythi:
thanks. This should be enough to get started. I need to make a few adjustments once I get `tdx-verifier` improved a bit.

attestation-service/src/token/ear_default_policy.rego
```diff
@@ -86,7 +86,33 @@ else := 3 if {
...
+
+hardware := 2 if {
+	# Check the quote is a TDX quote signed by Intel SGX Quoting Enclave
+	input.tdx.quote.header.tee_type == "81000000"
+	input.tdx.quote.header.vendor_id == "939a7233f79c4ca9940a0db3957f0607"
+
+	# Check TDX Module version and its hash. Also check OVMF code hash.
+	input.tdx.quote.body.mr_seam in data.reference.mr_seam
+	input.tdx.quote.body.tcb_svn in data.reference.tcb_svn
+	input.tdx.quote.body.mr_td in data.reference.mr_td
```
Xynnn007, line 105: Following https://github.com/confidential-containers/trustee/blob/main/attestation-service/docs/parsed_claims.md#intel-tdx, seems that we can also add
`tdx.quote.body.mrsigner_seam` also?

Xynnn007 (approved):
LGTM. Looks like one important thing is to ensure the boot chain for tdx guest. If we use td-shim, we will have `tdx.ccel.kernel` in parsed claims. I am not sure what ovmf from [ubuntu upstream](https://github.com/canonical/tdx) will generate, needs a try.

mythi:
> I am not sure what ovmf from [ubuntu upstream](https://github.com/canonical/tdx) will generate, needs a try.

it does not work and that is one thing on my list two fix

attestation-service/src/token/ear_default_policy.rego
mythi: I left that out on purpose for now since that's all zeros for Intel's TDX Module. Do you think it would be good to add a check for zero here?

Xynnn007: As a sample policy, it is ok imo.

mythi: OK as it is now?

PR #671, opened 2025-01-21
Reorganized integration tests and add negative tests

fitzthum:
Main changes:

- Move the test logic out of the test harness so we can write more flexible tests using helpers provided by the harness.
- Separate the common testing code from the get_resource tests and move the common code into `src`.
- Add simple negative test

This is an incremental step towards more legit tests that have sophisticated policies and deal with reference values.


Xynnn007 (approved):

mkulke (approved):

fitzthum:
Had to throw in another random change to fix the link checker. Looks like jwt.io doesn't like people visiting it who aren't using a web browser.

PR #673, opened 2025-01-22
KBS | refactor the integration test modules

Xynnn007:
Let's wait until https://github.com/confidential-containers/trustee/pull/671 gets merged.


pmores (approved):
lgtm (To be clear and to prevent any future misunderstanding, I don't really consider this structure better than the original one but oh well I guess... ;-) Let's move on, this has been taking too much time already.)

Xynnn007:
> lgtm (To be clear and to prevent any future misunderstanding, I don't really consider this structure better than the original one but oh well I guess... ;-) Let's move on, this has been taking too much time already.)

Thanks for reviewing the PR! I get that you might prefer the original approach, and I appreciate you being open to the change. I just wanted to address the problem we discussed that the lifetime fixing for integration test, with referring to some [official examples](https://github.com/tikv/rust-prometheus/blob/master/examples/example_push.rs), but if you think this causes any issues, I'm happy to tweak it further.

I didn’t mean to drag this out—just wanted to save you some maintenance overhead down the road. If this ended up causing unhappiness instead, it would be my fault : (. Anyway, let's just move forward as you points out, to the cool thing of making metrics system better  : )

pmores:
I realise now that my previous message might have sounded a bit too stern and if so I'd like to apologise, that wasn't my intent.

To elaborate: I consider both approaches pretty much equivalent in the context of KBS.  I think there might be scenarios which could put this refactored architecture under some stress - like maybe if we wanted to create a dynamically-determined number of `ApiServer` instances with distinct metrics - but I also think they are rather unlikely to occur in KBS, and the rest is just a matter of taste. :-)  So when I said I don't see the refactor as better than the original I should have also at least said that I don't see it as worse either.

So again, no hard feelings, I do appreciate your change (it certainly looks much neater!) and my only remaining concern is not about the change itself but rather about this PR getting stuck in review as well.  I tried to drum up some support for merging this on slack already, I think it should actually be an easy review since as far as I can tell, you didn't change anything externally visible.

My apologies again for my original bluntness, I should have known better.

PR #674, opened 2025-01-23
tdx-verifier: eventlog: handle OVMF/efistub measurements correctly

mythi:
This PR is a stopgap for checking the OVMF/direct boot based measurements logged in the CCEL. It merely follows the original `MeasuredEntity` approach for finding the kernel/cmdline/initrd measurements from the logs.

We should look to improve this and make it more generic to allow policies to check for more of the events and to support for different boot flows as well.


deps/verifier/src/tdx/eventlog.rs
```diff
@@ -67,21 +85,73 @@ impl CcEventLog {
...
         for event_entry in self.cc_events.log.clone() {
-            if event_entry.event_desc.len() < event_desc_prefix.len() {
-                continue;
-            }
-            if &event_entry.event_desc[..event_desc_prefix.len()] == event_desc_prefix.as_slice() {
-                let digest = &event_entry.digests[0].digest;
-                return Some(hex::encode(digest));
+            match (entity, event_entry.event_type) {
+                (MeasuredEntity::TdvfKernel, ev)
+                    if ev.as_str() == "EV_EFI_BOOT_SERVICES_APPLICATION" =>
```
fitzthum, line 91: Can you put this into the match itself?

something like 
```rust
match (entity, event_entry.event_type.as_str())
    (MeasuredEntity::TdvfKernel, "EV_EFI_BOOT_SERVICES_APPLICATION") => {
```

deps/verifier/src/tdx/eventlog.rs
```diff
@@ -67,21 +85,73 @@ impl CcEventLog {
...
+                {
+                    let raw_bytes = &event_entry.event_desc[KERNEL_VENMEDIA_DEVPATH_OFFSET
+                        ..KERNEL_VENMEDIA_DEVPATH_OFFSET + 2 * entity.to_string().len()];
+
+                    match read_string(raw_bytes) {
+                        Ok(kernel) if kernel == entity.to_string() => {
+                            return Some(hex::encode(&event_entry.digests[0].digest));
+                        }
+                        Ok(s) => warn!("Unknown Vendor Media Device Path: {s}"),
```
fitzthum, line 100: This pattern with two `Ok` blocks looks a little odd to me. It might be simplest to first read the string from raw_bytes (and error if it fails) and then check if the string matches the entity.

fitzthum (commented):
I'm not an expert here but this seems ok. Made a couple comments about the style.

deps/verifier/src/tdx/eventlog.rs
mythi: much better, thanks

deps/verifier/src/tdx/eventlog.rs
mythi: Dropped the other `Ok` (for both kernel and event).

fitzthum (approved):
LGTM.

We probably want to give @Xynnn007 a chance to look at this.

deps/verifier/src/tdx/eventlog.rs
```diff
@@ -67,21 +85,74 @@ impl CcEventLog {
...
-            if event_entry.event_desc.len() < event_desc_prefix.len() {
-                continue;
-            }
-            if &event_entry.event_desc[..event_desc_prefix.len()] == event_desc_prefix.as_slice() {
-                let digest = &event_entry.digests[0].digest;
-                return Some(hex::encode(digest));
+            match (entity, event_entry.event_type.as_str()) {
+                (MeasuredEntity::TdvfKernel, "EV_EFI_BOOT_SERVICES_APPLICATION") => {
+                    let raw_bytes = &event_entry.event_desc[KERNEL_VENMEDIA_DEVPATH_OFFSET
+                        ..KERNEL_VENMEDIA_DEVPATH_OFFSET + 2 * entity.to_string().len()];
```
Xynnn007, line 92: Do we need the `to_string()` here?

deps/verifier/src/tdx/eventlog.rs
```diff
@@ -67,21 +85,74 @@ impl CcEventLog {
...
+
+                    match read_string(raw_bytes) {
+                        Ok(kernel) => {
+                            if kernel == entity.to_string() {
+                                return Some(hex::encode(&event_entry.digests[0].digest));
```
Xynnn007, line 97: Let's use `event_entry.digests.first()` here to avoid panic if `.digests` is empty

deps/verifier/src/tdx/eventlog.rs
```diff
@@ -67,21 +85,74 @@ impl CcEventLog {
...
-                let digest = &event_entry.digests[0].digest;
-                return Some(hex::encode(digest));
+            match (entity, event_entry.event_type.as_str()) {
+                (MeasuredEntity::TdvfKernel, "EV_EFI_BOOT_SERVICES_APPLICATION") => {
+                    let raw_bytes = &event_entry.event_desc[KERNEL_VENMEDIA_DEVPATH_OFFSET
+                        ..KERNEL_VENMEDIA_DEVPATH_OFFSET + 2 * entity.to_string().len()];
+
+                    match read_string(raw_bytes) {
+                        Ok(kernel) => {
+                            if kernel == entity.to_string() {
```
Xynnn007, line 96: let's add [AsRefStr](https://docs.rs/strum/0.26.3/strum/derive.AsRefStr.html) to `MeasuredEntity`. Then we can use
```suggestion
                            if kernel == entity.as_ref() {
```

here to avoid memory copy.

deps/verifier/src/tdx/eventlog.rs
```diff
@@ -67,21 +85,74 @@ impl CcEventLog {
...
+                    let event_size = (&event_entry.event_desc[offset..2 * offset])
+                        .read_u32::<LittleEndian>()
+                        .unwrap_or_default() as usize;
+
+                    // Read the tagged event after the event size
+                    match String::from_utf8(
+                        event_entry.event_desc[offset * 2..offset * 2 + event_size - 1].to_vec(),
+                    ) {
+                        Ok(event) => {
+                            if event == entity.to_string() {
```
Xynnn007, line 117: same suggestion as before.

Xynnn007 (commented):
LGTM. Only some nits.

deps/verifier/src/tdx/eventlog.rs
mythi: done.

deps/verifier/src/tdx/eventlog.rs
mythi: done.

Xynnn007 (approved):

mythi:
rebased + pushed to get the e2e-kbs CI retriggered

PR #675, opened 2025-01-23
kbs: ITA: Documentation update.

szymon-klimek:
Documentation of parameters `policy_ids` and `allow_unmatched_policy` updated to be more descriptive.

### DIFF
--- a/kbs/docs/config.md
+++ b/kbs/docs/config.md
@@ -187,14 +187,14 @@ attestation. The following properties can be set.
 
 > Intel Trust Authority AS is available only when the `intel-trust-authority-as` feature is enabled.
 
-| Property                 | Type         | Description                                                                              | Required | Default |
-|--------------------------|--------------|------------------------------------------------------------------------------------------|----------|---------|
-| `timeout`                | Integer      | The maximum time (in minutes) between RCAR handshake's `auth` and `attest` requests      | No       | 5       |
-| `base_url`               | String       | Intel Trust Authority API URL.                                                           | Yes      | -       |
-| `api_key`                | String       | Intel Trust Authority API key.                                                           | Yes      | -       |
-| `certs_file`             | String       | URL to an Intel Trust Authority portal or path to JWKS file used for token verification. | Yes      | -       |
-| `allow_unmatched_policy` | Boolean      | If set and `policy_ids` specified, unset the `request.policy_must_match` setting         | No       | false   |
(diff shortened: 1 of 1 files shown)


fitzthum (commented):
I guess the DCO is complaining because your sign-off doesn't match the name in the "from" field of your patch (which is lowercase). Can you tweak that please. Otherwise seems ok. 

cc @mythi

szymon-klimek:
> I guess the DCO is complaining because your sign-off doesn't match the name in the "from" field of your patch (which is lowercase). Can you tweak that please. Otherwise seems ok.
> 
> cc @mythi

After contacting @mythi I revoked the changes and only updated documentation.

kbs/docs/config.md
```diff
@@ -187,14 +187,14 @@ attestation. The following properties can be set.
...
+| `base_url`               | String       | Intel Trust Authority API URL.                                                                                                                                             | Yes      | -       |
+| `api_key`                | String       | Intel Trust Authority API key.                                                                                                                                             | Yes      | -       |
+| `certs_file`             | String       | URL to an Intel Trust Authority portal or path to JWKS file used for token verification.                                                                                   | Yes      | -       |
+| `policy_ids`             | String array | Quoted and comma-separated list of policy IDs defined in ITA portal.                                                                                                       | No       | `[]`    |
+| `allow_unmatched_policy` | Boolean      | If `policy_ids` is not specified, parameter is ignored and ignores result of policy matching. <br/> Otherwise, define whether result of policy matching should be ignored. | No       | false   |
```
mythi, line 197:
```suggestion
| `allow_unmatched_policy` | Boolean      | If `policy_ids` is empty, this setting is ignored. By default, the listed policies must always match. Enabling this flag relaxes that requirement. | No       | false   |
```

szymon-klimek: In my opinion `By default` here is misleading, beacuse by default `policy_ids` are empty

so the true default behavior is that policies does not have to match

fitzthum: How about "Whether policy matching is required. If no `policy_ids` are specified, policy matching is not checked."

mythi: Souunds good to me. FWIW, we are also considering changing the flag name (the flag was used for a slightly different purpose originally) to follow the API conventions (e.g, `policy_must_match`).

mythi (approved):
LGTM

PR #668, opened 2025-01-17
Fixes inspired by Clippy

fitzthum:
I went a little further on the middle one and removed the `ExtractorsImpl` trait.

PR #676, opened 2025-01-23
Minor refactor of RVPS

fitzthum:
These changes aren't strictly necessary, but I think it makes things nicer and more similar to our other projects.

* Consolidate `native.rs` and `lib.rs`. `native.rs` isn't quite the right name for the file because it contains the main logic of the RVPS which is used in built-in mode or remote mode.
* Rename `core` to `Rvps`. Core is a bit vague, especially if you're using this as a crate, so let's just call it Rvps.


mythi:
LGTM. Github complains that 5bd1c68 isn't an atomic change but that's probably OK. The RVPS change from #668 would fit in this PR too.

rvps/src/bin/server/mod.rs
```diff
@@ -16,11 +16,11 @@ use crate::rvps_api::{
 };
 
 pub struct RVPSServer {
-    rvps: Arc<Mutex<Core>>,
+    rvps: Arc<Mutex<Rvps>>,
```
Xynnn007, line 19: I am considering change the `Mutex` here to `RwLock`. Let's do it in a seperate PR.

Xynnn007 (approved):
LGTM

PR #679, opened 2025-01-28
`deps/verifier`: Add constructor for SnpEvidence

tylerfanelli:


fitzthum (approved):
LGTM

I wouldn't necessarily expect a client to import anything from the verifier. In the AA we have a matching struct in the attester. It's not ideal having to synchronize these manually, but it breaks a potential dependency loop. Either way, doesn't hurt to add this.

tylerfanelli:
Perhaps its worth an addition to kbs_types? Just so you don't have to sync manually.

Just a suggestion, this is fine too.

fitzthum:
Yeah interestingly there is an `SnpEvidence` struct in KBS types, but it's different from the one we use. In theory that is a good way to handle this shared dependency tho.

tylerfanelli:
Yea, but I don't think it's used (or if it is, we can change it to match this)

mythi:
> In the AA we have a matching struct in the attester. It's not ideal having to synchronize these manually, but it breaks a potential dependency loop.

FWIW, this extra synchronization need has stopped me from submitting the PR to make `snp-attester` use TSM reports without depending on the `sev` crate. With base64 encoded raw blobs sent by `snp-attester` this synchronization isn't needed (after it's changed, of course).

deps/verifier/src/snp/mod.rs
```diff
@@ -1,5 +1,5 @@
 use anyhow::anyhow;
-use base64::{Engine, engine::general_purpose::STANDARD};
+use base64::{engine::general_purpose::STANDARD, Engine};
```
mythi, line 2: nit: comes from `cargo fmt`.

tylerfanelli: Included in `cargo fmt` commit.

PR #684, opened 2025-01-31
as, rvps: Documentation fixes and add ons, podman Containerfiles

tylerfanelli:


rvps/README.md
```diff
@@ -80,6 +80,17 @@ Run
 docker run -d -p 50003:50003 rvps --address 0.0.0.0:50003
 ```
 
+Or we can build RVPS as a podman image
```
mythi, line 83: The podman flow is different from "RVPS docker image" since it expects the binary is already available. It would be good to make it clear here.,

rvps/podman/Containerfile
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2025 by Red Hat.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+FROM fedora:latest
+
+COPY target/release/rvps /usr/local/bin/rvps
```
mythi, line 7: > Docker and podman handle build arguments differently. Podman does not
> allow redefinition of certain variables, notably BUILDPLATFORM.

It looks these `Containerfile`s are rather easy to maintain but I was wondering if it made more sense to revisit the arm build support so that podman could use the `Dockerfile`s directly?

rvps/README.md
```diff
@@ -172,13 +172,14 @@ Then it will say something like
 [2023-03-09T04:44:11Z INFO  rvps_client] Register provenance succeeded.
 ```
 
-Let's then query the reference value
+Let's then query the reference values
 ```bash
-rvps-tool query --name test-binary-1 --addr http://$RVPS_ADDR
+rvps-tool query --addr http://$RVPS_ADDR
```
fitzthum, line 177: Yeah I think I removed this when we changed the interface of the RVPS. Thanks for fixing this.

fitzthum (approved):
LGTM

I agree that we might want to share the Dockerfiles but we could potentially do that in a follow-up.

rvps/podman/Containerfile
tylerfanelli: Some of the Dockerfile (mainly to do with arm build support) is incompatible with podman. For instance, the [BUILDPLATFORM](https://github.com/confidential-containers/trustee/blob/5d7cdea587bfa2c5a2d7118f7918f425195820c5/rvps/docker/Dockerfile#L6) arg is handled differently in podman. If some of that arm support is modified/removed altogether, than podman could likely use it.

I'm not sure how that Dockerfile is being used though, so didn't want to modify it. What do you think?

mythi: I think the case here is just that podman was not tested when the arm support was added but other than that there's nothing special about them. I tend to agree with @fitzthum that those can be fixed in a follow-up PR (with `Containerfile` removals) but the downside is that it'll break users who have move to use `Containerfile`s with podman.

seungukshin: Trustee's dockerfiles don't redefine any variables. It set only default value which is supported by podman also.

I've tested current dockerfiles with podman, and found follwoing problems.
1. Podman cannot find rust image
    * Full path like `docker.io/library/rust:latest` should be used instead of `rust:latest`
1. Podman ignores `--platform=$BUILDPLATFORM` from `FROM`
    * To generate an arm64 binary, docker uses rust corss compilation on top of amd64.
    * To generate an arm64 binary, podman uses arm64 emulation, so building performace is degraded than docker. But the correct binary is generated.
    * This affect building the other architecutre on top of the specific architecture. It doesn't affect native building.

I'll create a PR to fix 1 to unblock this PR shortly.
And now GitHub supports arm64 runner, so I'll create a separate PR to remove the cross compilation (removing `--platform=$BUILDPLATFORM`) later.

seungukshin: https://github.com/confidential-containers/trustee/pull/689 is created

fitzthum:
I've merged this. We may want to remove the separate podman files in the future soon, but fine for now.

rvps/podman/Containerfile
mythi: > And now GitHub supports arm64 runner, so I'll create a separate PR to remove the cross compilation (removing `--platform=$BUILDPLATFORM`) later.

Great, thanks!

PR #685, opened 2025-01-31
verifier: add parsing tdx td attributes and usage in policy

pawelpros:
Add parsing tdx td attributes from hex value and include them in claims report to be accessible in policies definition.


fitzthum (approved):
LGTM. Thanks

mythi (approved):
@pawelpros thanks!

PR #687, opened 2025-02-03
setup-opa action and Ubuntu 24.04 runners

mythi:


fitzthum (approved):
This is great. Thanks

PR #698, opened 2025-02-06
kbs/plugins: Add ID_KEY plugin

tylerfanelli:
The ID_KEY resource backend creates a new key tied to a Trustee server instance and mostly pertains to workloads that want a key unwrapping as a result of a successful attestation.

Admins can POST to the repository giving a base64 encoded byte vector as the resource tag. This resource tag will then be encrypted with the key and returned to the admin. Upon successful attestation, attestation clients can then send the encrypted data that was provisioned by the admin to the backend via a GET request. The resource tag in this case must be a base64 encoding of the encrypted data's bytes. The data will then be decrypted and sent back to the attestation client.

I will likely provide another PR in the near future allowing the server's encryption key to be derived from an HSM module for security purposes. While it won't be required, it will be preferred.

---

### Client use-case

SVSM will require attestation to establish trust and unlock persistent vTPM state to run CVMs. Virtual disks are made available to accommodate this and bootstrap persistent vTPMs. A block device will include:

- OS rootfs sealed with vTPM
- vTPM NVDATA (required for unsealing) encrypted with an "NVDATA Encryption Key" (NEK).
- NEK wrapped by a key on an attestation server.

```
┌────────────────────────────┐
│                            │
│  ┌──────────────────────┐  │
│  │vTPM-sealed rootfs    │  │
│  └──────────────────────┘  │
│  ┌──────────────────────┐  │
│  │Encrypted vTPM NVDATA │  │
│  │(wrapped by NEK)      │  │
│  └──────────────────────┘  │
│  ┌──────────────────────┐  │
│  │NEK (wrapped by key on│  │
│  │attestation server)   │  │
│  └──────────────────────┘  │
│                            │
│                Virtual disk│
└────────────────────────────┘
```

A simple unlocking and boot would be as follows:

- SVSM attests, gives encrypted NEK to attestation server, server unwraps it with ID-key.

- SVSM uses unwrapped NEK to decrypt vTPM NVDATA, sets up vTPM.

- SVSM boots OS with vTPM.

- OS initrd (systemd-cryptenroll) unseals and mounts rootfs.

With this chain, **an OS will not be able to boot without SVSM first attesting itself**.

### Admin set-up

Initializing and setup of the virtual disks should be done from the guest owner themselves. They should also be the ones in control (with admin privileges/credentials) of the Trustee server in which their CVMs will attest with. As such, the admin will:

- Initialize a "fresh" vTPM 
- Seal their OS's rootfs to the newly-created vTPM
- Create a new symmetric NEK
- Encrypt the vTPM's NVDATA with the newly-created NEK
- **Wrap the NEK with a key on an attestation server**

Once completed, the guest owner can supply this virtual disk to untrusted parties for running CVMs.

The highlighted portion pertains to this PR. Because the NEK (and corresponding key stored on attestation server that wraps it) is sensitive, only the KBS admin must be able to wrap NEKs that the KBS generates. I also plan to add support for deriving the wrap key from an HSM at some point, so an admin could just configure any KBS to use a key stored on the HSM as the wrap key.

### Using the existing resource backends

Recall how the virtual disks were presented to CVMs:

```
┌────────────────────────────┐
│                            │
│  ┌──────────────────────┐  │
│  │vTPM-sealed rootfs    │  │
│  └──────────────────────┘  │
│  ┌──────────────────────┐  │
│  │Encrypted vTPM NVDATA │  │
│  │(wrapped by NEK)      │  │
│  └──────────────────────┘  │
│  ┌──────────────────────┐  │
│  │NEK (wrapped by key on│  │
│  │attestation server)   │  │
│  └──────────────────────┘  │
│                            │
│                Virtual disk│
└────────────────────────────┘
```
As an alternative, we **could** store the NEK on the KBS as a retrievable resource. We could then provide the resource ID of the NEK so that the client would know which KBS resource to fetch once an attestation complete. The virtual disk would then look like this:

```
┌────────────────────────────┐
│                            │
│  ┌──────────────────────┐  │
│  │vTPM-sealed rootfs    │  │
│  └──────────────────────┘  │
│  ┌──────────────────────┐  │
│  │Encrypted vTPM NVDATA │  │
│  │(wrapped by NEK)      │  │
│  └──────────────────────┘  │
│  ┌──────────────────────┐  │
│  │KBS resource ID of NEK│  │
│  │                      │  │
│  └──────────────────────┘  │
│                            │
│                Virtual disk│
└────────────────────────────┘
```

With this, we could use the existing resource backends in Trustee to complete our attestation. However, there is a few principles that prevent us from doing this:

- Although the resource ID would be given in the virtual disk, there is nothing preventing a motivated bad actor from resources that pertain to other VMs, such as their NEKs. With some inside knowledge, an attacker could steal other VM NEKs and impersonate other VMs with their vTPM endorsement keys simply by knowing which resource ID corresponds to another VMs NEK.

- As we cannot guarantee that all clients using persistent vTPMs in this manner will use Trustee to attest, **the virtual disk format cannot tie itself to any Trustee-specific mechanism (like giving a resource ID tag of the NEK). The virtual disk format must be server-agnostic**. This applies for all attestation protocols, for that matter.

- As the virtual disk must be server-agnostic, **the result of an attestation should only be the unwrapping of the NEK**. No other information is supplied to the virtual disk.

### ID-key Plugin

Recall the point made above: **the result of an attestation should only be the unwrapping of the NEK**. That is exactly what this plugin provides.

- A way for admins to provision and encrypt NEKs
- An endpoint where CVMs can POST their NEKs for decryption

It allows for Trustee to be used for vTPM NVDATA encryption in a way it wasn't able to be used with the normal resource backend mechanisms. It still requires that an attested client present its attestation token when decrypting an NEK (much like when fetching a resource), but also allows the POSTing of plaintext NEK data.

### Preventing replay attacks

To prevent replay attacks (a third party snooping on the encrypted NEK in order to replay and get it decrypted), the ECDH mechanism for wrapping the encrypted NEK before sending it to the plugin is implemented. With this, every time an encrypted NEK is sent to the KBS, it is wrapped with an ephemeral EC key such that no third party can derive the original encrypted NEK and replay it.

### DIFF
--- a/kbs/docs/id-key.md
+++ b/kbs/docs/id-key.md
@@ -1,0 +1,84 @@
+# ID-key Plugin
+
+In addition to fetchable resources upon successful attestation, KBS also offers the ID-key plugin for encrypting and decrypting data with a key provisioned by the KBS. This key can be persisted across multiple reboots of CVMs and help tie decryption resources to a specific KBS instance.
+
+## Data Encryption
+
+As a guest owner would be responsible for encrypting persistent storage, they are the sole authority that is permitted to request resources be encrypted. The KBS is expected to be fully in control by the guest owner with credentials configured.
+
+## Flow
+
+### Provisioning encrypted data
+
+A credentialed guest owner is the only one authorized to request that keys be encrypted by the ID-key plugin. To do this, the guest owner will POST some plaintext data (created locally) to be encrypted. This plaintext data could be a LUKS passphrase, SecureBoot keys, etc. The actual purpose of the data is of no concern to the KBS.
+
+`POST /kbs/v0/id-key/{base64-plaintext-data}`
+
+
+```
+          ┌───────────┐        
+          │   POST    │        
+Owner ──► ├───────────┤ ──► KBS
+  ▲       │ Plaintext │        
+  │       │ data      │      │ 
+  │       └───────────┘      │ 
+  │                          │ 
+  │                          │ 
+  │       ┌───────────┐      │ 
+  └────── │ Encrypted │ ◄────┘ 
+          │ data      │        
+          └───────────┘        
+```
+
+With this, the guest owner is now free to configure persistent storage as they wish for their CVMs to have access to encrypted data when they boot. Upon successful attestation, this encrypted data will be presented to the KBS by the CVM for decryption.
+
+### Post-attestation, decrypting data
+
+With a CVM booted and attested, it is now assumed that the CVM was configured in such a way that it has access to readable storage that contains the encrypted data. The CVM can now send this data to the KBS for decryption.
+
+To prevent replay attacks, the CVM must wrap the encrypted data in a ECDH key cipher. To support this, the ID-key plugin presents an endpoint `ecdh-pub-sec1` that a VM can use to fetch the EC public key in order to wrap the encrypted data.
+
+`GET /kbs/v0/id-key/ecdh-pub-sec1`
+
+```
+          ┌────────┐
+          │  GET   │
+          ┤        │
+Client ──►│ EC     │ ──► KBS
+          │ public │
+  ▲       │ key    │      │
+  │       └────────┘      │
+  │                       │
+  │                       │
+  │  ┌─────────────────┐  │
+  │  │       EC        │  │
+  └──┤     public      │◄─┘
+     │       key       │
+     │  (sec1 encoded) │
+     └─────────────────┘
+```
+
+With the KBS's public EC key, the CVM can create its own ephemeral EC key and wrap the encrypted data with an ECDH/SHA256/AES256GCM cipher.
+
+The CVM will then present the ECDH-wrapped encrypted data along with all data needed for unwrapping (client EC public key and AESGCM iv). The KBS will use this data to unwrap the encrypted data and decrypt it.
+
+`GET /kbs/v0/id-key/{base64-wrapped-encrypted-data}?ecdh-pubkey={base64-client-ecdh-pubkey-sec1}&iv={base64-aesgcm-iv}`
+
+```                                            
+              ┌───────────────┐             
+              │      GET      │             
+              ├───────────────┤             
+    Client ──►│   Encrypted   │ ──► KBS     
+              │   Data        │             
+      ▲       │ (ECDH wrapped)│      │      
+      │       └───────────────┘      │      
+      │                              │      
+      │      ┌─────────────────┐     │      
+      │      │   Decrypted     │     │      
+      └───── │   Data          │ ◄───┘      
+             │  (TEE public    │            
+             │   key wrapped)  │            
+             └─────────────────┘            
+```
+
+As the decrypted secret is sensitive and will be passing through an untrusted host network, the secret must be wrapped by some data only known to the CVM. The KBS re-uses the TEE public key presented at attestation for this purpose. In this sense, a client reading the decrypted secret from the KBS is much like reading a resource from the KBS. A KBS `Response` wrapping the secret in the TEE public key will be returned to the CVM. The CVM will use the `Response` to unwrap the secret.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2880,6 +2880,7 @@ dependencies = [
  "cryptoki",
  "derivative",
  "env_logger 0.10.2",
+ "generic-array",
  "jsonwebtoken",
  "jwt-simple",
  "kbs-types",
@@ -2888,6 +2889,7 @@ dependencies = [
  "log",
  "mobc",
  "openssl",
+ "p384",
  "prost",
  "rand",
  "reference-value-provider-service",
@@ -2900,6 +2902,7 @@ dependencies = [
  "semver",
  "serde",
  "serde_json",
+ "sha2",
  "strum",
  "tempfile",
  "thiserror 2.0.3",
@@ -3514,9 +3517,9 @@ dependencies = [
 
 [[package]]
 name = "p384"
-version = "0.13.0"
+version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "70786f51bcc69f6a4c0360e063a4cac5419ef7c5cd5b3c99ad70f3be5ba79209"
+checksum = "fe42f1670a52a47d448f14b6a5c61dd78fce51856e68edaa38f7ae3a46b8d6b6"
 dependencies = [
  "ecdsa",
  "elliptic-curve",
--- a/kbs/Cargo.toml
+++ b/kbs/Cargo.toml
@@ -46,6 +46,7 @@ clap = { workspace = true, features = ["derive", "env"] }
 config.workspace = true
 cryptoki = { version = "0.8.0", optional = true }
 env_logger.workspace = true
+generic-array = "0.14.7"
 jsonwebtoken = { workspace = true, default-features = false }
 jwt-simple.workspace = true
 kbs-types.workspace = true
@@ -53,6 +54,7 @@ kms = { workspace = true, default-features = false }
 lazy_static = "1.4.0"
 log.workspace = true
 mobc = { version = "0.8.5", optional = true }
+p384 = { version = "0.13.1", features = ["arithmetic", "ecdh"] }
 prost = { workspace = true, optional = true }
 rand = "0.8.5"
 regex = "1.11.1"
@@ -63,6 +65,7 @@ scc = "2"
 semver = "1.0.16"
 serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
+sha2 = "0.10.8"
 strum.workspace = true
 thiserror.workspace = true
 time = { version = "0.3.37", features = ["std"] }
--- a/kbs/src/plugins/implementations/id_key.rs
+++ b/kbs/src/plugins/implementations/id_key.rs
@@ -1,0 +1,201 @@
+// Copyright (c) 2025 by Red Hat.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use super::super::plugin_manager::ClientPlugin;
+use actix_web::http::Method;
+use aes_gcm::{aead::Aead, Aes256Gcm, Key, KeyInit};
+use anyhow::{anyhow, bail, Context, Result};
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
+use generic_array::GenericArray;
+use openssl::{
+    pkey::Private,
+    rsa::{Padding, Rsa},
+};
+use p384::{ecdh, PublicKey, SecretKey};
+use serde::Deserialize;
+use sha2::Sha256;
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+pub struct IdKeyConfig;
+
+pub struct IdKey {
+    idkey_non_hsm: Rsa<Private>,
+    ecdh_secret: SecretKey,
+}
+
+impl TryFrom<IdKeyConfig> for IdKey {
+    type Error = anyhow::Error;
+
+    fn try_from(_value: IdKeyConfig) -> anyhow::Result<Self> {
+        Self::new()
+    }
+}
+
+#[async_trait::async_trait]
+impl ClientPlugin for IdKey {
+    async fn handle(
+        &self,
+        _body: &[u8],
+        query: &str,
+        path: &str,
+        method: &Method,
+    ) -> Result<Vec<u8>> {
+        let path = path
+            .strip_prefix('/')
+            .context("accessed path is illegal, should start with '/'")?;
+
+        match *method {
+            Method::POST => {
+                let plain = URL_SAFE_NO_PAD
+                    .decode(path)
+                    .context("invalid id-key path")?;
+
+                self.encrypt(plain)
+            }
+            Method::GET => match path {
+                "ecdh-pub-sec1" => Ok(self.ecdh_pubkey_sec1()),
+                _ => {
+                    let (public_key, iv) = ecdh_iv(query.to_string())?;
+
+                    let wrapped = URL_SAFE_NO_PAD.decode(path).context("invalid path")?;
+                    let encrypted = self.ecdh_unwrap(wrapped, public_key, iv)?;
+                    let decrypted = self.decrypt(encrypted)?;
+
+                    Ok(decrypted)
+                }
+            },
+            _ => bail!("Illegal HTTP method. Only supports `GET` and `POST`"),
+        }
+    }
+
+    async fn validate_auth(
+        &self,
+        _body: &[u8],
+        _query: &str,
+        _path: &str,
+        method: &Method,
+    ) -> Result<bool> {
+        match method {
+            &Method::POST => Ok(true),
+            _ => Ok(false),
+        }
+    }
+
+    async fn encrypted(
+        &self,
+        _body: &[u8],
+        _query: &str,
+        _path: &str,
+        method: &Method,
+    ) -> Result<bool> {
+        match *method {
+            Method::GET => Ok(true),
+            _ => Ok(false),
+        }
+    }
+}
+
+impl IdKey {
+    pub fn new() -> Result<Self> {
+        Ok(Self {
+            idkey_non_hsm: Rsa::generate(2048).context("unable to create ID key")?,
+            ecdh_secret: SecretKey::random(&mut rand::thread_rng()),
+        })
+    }
+
+    fn encrypt(&self, bytes: Vec<u8>) -> Result<Vec<u8>> {
+        let encrypted = {
+            let mut d = [0u8; 256];
+            let sz = self
+                .idkey_non_hsm
+                .public_encrypt(&bytes, &mut d, Padding::PKCS1)
+                .context("unable to decrypt key")?;
+
+            Vec::from(&d[..sz])
+        };
+
+        Ok(URL_SAFE_NO_PAD.encode(encrypted).into())
+    }
+
+    fn ecdh_pubkey_sec1(&self) -> Vec<u8> {
+        self.ecdh_secret.public_key().to_sec1_bytes().to_vec()
+    }
+
+    fn ecdh_unwrap(
+        &self,
+        wrapped: Vec<u8>,
+        ec_public_key: PublicKey,
+        iv: Vec<u8>,
+    ) -> Result<Vec<u8>> {
+        let shared = ecdh::diffie_hellman(
+            self.ecdh_secret.to_nonzero_scalar(),
+            ec_public_key.as_affine(),
+        );
+        let mut sha_bytes = [0u8; 32];
+
+        let hkdf = shared.extract::<Sha256>(None);
+        if hkdf.expand(&[], &mut sha_bytes).is_err() {
+            return Err(anyhow!("unable to get ECDH shared SHA hash"));
+        }
+
+        let key = Key::<Aes256Gcm>::from_slice(&sha_bytes);
+        let aes = Aes256Gcm::new(key);
+
+        let encrypted = match aes.decrypt(GenericArray::from_slice(iv.as_slice()), wrapped.as_ref())
+        {
+            Ok(e) => e,
+            Err(_) => {
+                return Err(anyhow!(
+                    "unable to unwrap encrypted secret with shared AES-GCM key"
+                ))
+            }
+        };
+
+        Ok(encrypted)
+    }
+
+    fn decrypt(&self, encrypted: Vec<u8>) -> Result<Vec<u8>> {
+        let mut d = [0u8; 256];
+        let sz = self
+            .idkey_non_hsm
+            .private_decrypt(&encrypted, &mut d, Padding::PKCS1)
+            .context("unable to decrypt key")?;
+
+        Ok(Vec::from(&d[..sz]))
+    }
+}
+
+fn ecdh_iv(query: String) -> Result<(PublicKey, Vec<u8>)> {
+    let subs: Vec<&str> = query.split('&').collect();
+    if subs.len() != 2 {
+        bail!("invalid query");
+    }
+
+    let public_key = {
+        let bytes = parse_val("ecdh-pubkey", &subs)?;
+        PublicKey::from_sec1_bytes(&bytes)
+            .context("public key cannot be derived from SEC1 bytes")?
+    };
+
+    let iv = parse_val("iv", &subs)?;
+
+    Ok((public_key, iv))
+}
+
+fn parse_val(key: &str, subs: &Vec<&str>) -> Result<Vec<u8>> {
+    for substr in subs {
+        let kv: Vec<&str> = substr.split('=').collect();
+        if kv.len() != 2 {
+            bail!("invalid query");
+        }
+
+        if kv[0] == key {
+            let bytes = URL_SAFE_NO_PAD.decode(kv[1]).context("invalid query")?;
+
+            return Ok(bytes);
+        }
+    }
+
+    Err(anyhow!("invalid query"))
+}
--- a/kbs/src/plugins/implementations/mod.rs
+++ b/kbs/src/plugins/implementations/mod.rs
@@ -5,5 +5,8 @@
 pub mod resource;
 pub mod sample;
 
+pub mod id_key;
+
+pub use id_key::{IdKey, IdKeyConfig};
 pub use resource::{RepositoryConfig, ResourceStorage};
 pub use sample::{Sample, SampleConfig};
--- a/kbs/src/plugins/plugin_manager.rs
+++ b/kbs/src/plugins/plugin_manager.rs
@@ -8,7 +8,7 @@ use actix_web::http::Method;
 use anyhow::{Context, Error, Result};
 use serde::Deserialize;
 
-use super::{sample, RepositoryConfig, ResourceStorage};
+use super::{sample, IdKey, IdKeyConfig, RepositoryConfig, ResourceStorage};
 
 type ClientPluginInstance = Arc<dyn ClientPlugin>;
 
@@ -59,6 +59,9 @@ pub enum PluginsConfig {
 
     #[serde(alias = "resource")]
     ResourceStorage(RepositoryConfig),
+
+    #[serde(alias = "id-key")]
+    IdKey(IdKeyConfig),
 }
 
 impl Display for PluginsConfig {
@@ -66,6 +69,7 @@ impl Display for PluginsConfig {
         match self {
             PluginsConfig::Sample(_) => f.write_str("sample"),
             PluginsConfig::ResourceStorage(_) => f.write_str("resource"),
+            PluginsConfig::IdKey(_) => f.write_str("id-key"),
         }
     }
 }
@@ -84,6 +88,10 @@ impl TryInto<ClientPluginInstance> for PluginsConfig {
                 let resource_storage = ResourceStorage::try_from(repository_config)
                     .context("Initialize 'Resource' plugin failed")?;
                 Arc::new(resource_storage) as _
+            }
+            PluginsConfig::IdKey(cfg) => {
+                let id_key = IdKey::try_from(cfg).context("Initialize 'ID_KEY' plugin failed")?;
+                Arc::new(id_key) as _
             }
         };
 


fitzthum:
I'll take a look at the code tomorrow.

On a high level, what does this give you that you can't get from an existing backend? Let's say I'm an admin. I post some bytes to the plugin and get back the wrapped bytes. I give those wrapped bytes to my guest. Then the guest can get the unwrapped version from Trustee. How is that different from an admin simply uploading a secret to Trustee and then giving the client the name of the secret?

Also, are you worried about replay attacks?

tylerfanelli:
> I'll take a look at the code tomorrow.
> 
> On a high level, what does this give you that you can't get from an existing backend? Let's say I'm an admin. I post some bytes to the plugin and get back the wrapped bytes. I give those wrapped bytes to my guest. Then the guest can get the unwrapped version from Trustee. How is that different from an admin simply uploading a secret to Trustee and then giving the client the name of the secret?
> 

The issue with this is that there's no way of preventing the client from fetching a secret that they're not supposed to. If there is more than one VM attempting to attest, there's no stopping them from requesting the secrets of another VM. This simplifies it by only allowing that VM's secret from fetched. We only care about one secret in this scenario, so that is what this offers.

> Also, are you worried about replay attacks?

Yep. A man-in-the-middle could sniff the URL and just replay to get the decrypted secret. This is functional but unsafe, I'm thinking of adding a ECDH exchange via the request's body.

tylerfanelli:
I've made a few modifications, and this is ready for review. Marking this as draft for now, as I'd like to add another commit introducing an ephemeral ECDH exchange to protect against relay attacks.

tylerfanelli:
This is now ready for review, but has a few blockers before it can be considered for merging. You will notice the final commit is prefixed by a `FIXME` tag. To facilitate the ECDH exchange, I've added a {de}serializable type to `kbs-types` for the data needed. The PR is here: https://github.com/virtee/kbs-types/pull/57

I've included the type (as well as the associated {de}serialization methods usually found in `kbs-types`) in the `id-key` module, but once the `kbs-types` PR is merged, it can be removed. However, the PR is also based on a later version of `kbs-types` that broke the API, so I'm also blocked by #597 (which updates the API and `kbs-types` version) as well. Once both of these are handled, this will be ready for merging.

@fitzthum When you have the chance, would you mind reviewing?

fitzthum:
> The issue with this is that there's no way of preventing the client from fetching a secret that they're not supposed to. If there is more than one VM attempting to attest, there's no stopping them from requesting the secrets of another VM. This simplifies it by only allowing that VM's secret available to it. We only care about one secret in this scenario.

The best mechanism for this kind of thing is probably init-data. With EAR we now have really graceful support for exposing the init-data to the policy. We still need some piping to get it from the KBS to the AS, but that will come soon.

Even with ECDH I'm a little unclear how this helps differentiate between two guests. How does the wrapped key get into the guest? If this is not provided via a measured channel, then can't a malicious host generate multiple guests with the same key, switch keys between guests, etc.

fitzthum:
Can you add a little description of this backend to this doc https://github.com/confidential-containers/trustee/blob/main/kbs/docs/resource_storage_backend.md

kbs/src/plugins/implementations/resource/backend.rs
```diff
@@ -17,7 +17,11 @@ type RepositoryInstance = Arc<dyn StorageBackend>;
 #[async_trait::async_trait]
 pub trait StorageBackend: Send + Sync {
     /// Read secret resource from repository.
-    async fn read_secret_resource(&self, resource_desc: ResourceDesc) -> Result<Vec<u8>>;
+    async fn read_secret_resource(
+        &self,
+        resource_desc: ResourceDesc,
+        data: &[u8],
```
fitzthum, line 23: This variable could be more specific. It's not just any data. It's always the request body, right? 

Also, I wonder if you wouldn't be better off just making your own plugin. Plugins already get the entire request body. Might be a better fit. You also probably wouldn't need the first commit.

tylerfanelli:
> > The issue with this is that there's no way of preventing the client from fetching a secret that they're not supposed to. If there is more than one VM attempting to attest, there's no stopping them from requesting the secrets of another VM. This simplifies it by only allowing that VM's secret available to it. We only care about one secret in this scenario.
> 
> The best mechanism for this kind of thing is probably init-data. With EAR we now have really graceful support for exposing the init-data to the policy. We still need some piping to get it from the KBS to the AS, but that will come soon.
> 

Init-data certainly helps with attestation reference values in the RVPS, but I'm not convinced it helps here. Keep in mind that any policy would need to have full understanding of what resources are available, what VM is entitled to a respective resource, etc. I want this to be done dynamically, therefore I can't rely on a policy to enforce this isolation.

> Even with ECDH I'm a little unclear how this helps differentiate between two guests.

The ECDH does nothing to differentiate between guests, it merely wraps the encrypted key (which *is* used to differentiate guests) to prevent replay attacks.

> How does the wrapped key get into the guest?

A virtual disk.

> If this is not provided via a measured channel, then can't a malicious host generate multiple guests with the same key, switch keys between guests, etc.

For my use-case, this resource backend cannot solve this problem standalone. For instance, generating multiple guests with the same key can be detected later on with a monotonic vTPM boot counter.

Switching keys between guests would be a denial of service, which I'm not concerned about since it doesn't leak any secrets or sensitive info.

tylerfanelli:
> Can you add a little description of this backend to this doc https://github.com/confidential-containers/trustee/blob/main/kbs/docs/resource_storage_backend.md

Sure.

kbs/src/plugins/implementations/resource/id_key.rs
```diff
@@ -0,0 +1,213 @@
...
+        if resource_desc.resource_type == *"ecdh-pub-sec1" {
+            return Ok(self.ecdh_pubkey_sec1());
+        }
+
+        if resource_desc.resource_type != *"nek" {
+            return Err(anyhow!("invalid resource type"));
+        }
+
+        let ecdh: IdKeyEcdh =
+            serde_json::from_slice(data).context("unable to deserialize request body")?;
```
fitzthum, line 52: I'm not sure this is going to work.

When making a resource request, the TEE public key is not in the request body. It's in the attestation token. Check out how we extract the public key [here](https://github.com/confidential-containers/trustee/blob/main/kbs/src/api_server.rs#L244) when we use it to encrypt the response.

kbs/src/plugins/implementations/resource/backend.rs
tylerfanelli: > This variable could be more specific. It's not just any data. It's always the request body, right?
> 

Yes, I can make sure to specify that.

> Also, I wonder if you wouldn't be better off just making your own plugin. Plugins already get the entire request body. Might be a better fit. You also probably wouldn't need the first commit.

Which would you prefer?

kbs/src/plugins/implementations/resource/id_key.rs
tylerfanelli: This isn't using the TEE public key. This is because I'm only supporting ECDH, and not leaving open the use of RSA like JWK's allow. Would you prefer it use JWK?

fitzthum: So you're expecting the request to have an additional key in it? That seems like an extension of the protocol.

tylerfanelli: I was under impression the protocol was a bit flexible with this. I'm using the protocol in an unconventional way anyway w/r/t needing to first fetch the server's ECDH public key before actually requesting the resource.

kbs/src/plugins/implementations/resource/backend.rs
fitzthum: Probably best to do a plugin; at the very least because you can enable multiple plugins at once, but you can only have one resource backend (maybe not relevant to your use case, but a nice property). Plugins have generally superseded backends. In fact, the resource thing is actually a plugin itself now that only exposes part of what it has to the backend.

kbs/src/plugins/implementations/resource/id_key.rs
fitzthum: There are no laws against it, but it's not totally ideal. This would fit better in another plugin. We do support query strings in the request. Any chance you could put the public key in there?

kbs/src/plugins/implementations/resource/backend.rs
tylerfanelli: Ok, I'll edit to make this its own plugin instead.

kbs/src/plugins/implementations/resource/id_key.rs
tylerfanelli: Yea, that could be done. I'd actually probably prefer that too, as it basically removes the need for the `kbs-types` PR I previously mentioned.

fitzthum:
> Switching keys between guests would be a denial of service, which I'm not concerned about since it doesn't leak any secrets or sensitive info.

The main guarantee that I see from this approach is that the KBS knows that it is releasing a secret to someone who has already in some way (maybe indirectly) been in contact with it. This is something, but I don't see how it differentiates one guest from another.

I know it's clunky to update the policy dynamically, but if you're limiting which guests can access which resources, won't the KBS have to be in the loop with that somehow? Additionally it seems like an identity is probably something you want to measure.

Either way, it doesn't hurt much to add a plugin. Let's see what Ding thinks once the changes are ready.

tylerfanelli:
> > Switching keys between guests would be a denial of service, which I'm not concerned about since it doesn't leak any secrets or sensitive info.
> 
> The main guarantee that I see from this approach is that the KBS knows that it is releasing a secret to someone who has already in some way (maybe indirectly) been in contact with it. This is something, but I don't see how it differentiates one guest from another.

When you say "indirectly" are you alluding to the attestation token that was provided from a successful attestation?

> 
> I know it's clunky to update the policy dynamically, but if you're limiting which guests can access which resources, won't the KBS have to be in the loop with that somehow? Additionally it seems like an identity is probably something you want to measure.

I don't see why the KBS needs to be in the loop. The randomness of the encryption key itself is sufficient for acting as an identity. In my case, we're verifying that a guest has attested its boot environment, and now wants to unlock some storage that its carrying locally. The beginning step of that unwrapping the storage encryption key that it has locally.

The dynamic nature of this makes it scalable. An admin can provision as many keys as needed for its VM fleet.

> 
> Either way, it doesn't hurt much to add a plugin. Let's see what Ding thinks once the changes are ready.

fitzthum:
> When you say "indirectly" are you alluding to the attestation token that was provided from a successful attestation?

I say indirectly because as soon as the encrypted key passes through the untrusted host it could be stolen and used by other guests. Anyone who shows up with that key and a valid enclave will get the secret.

> I don't see why the KBS needs to be in the loop. The randomness of the encryption key itself is sufficient for acting as an identity. In my case, we're verifying that a guest has attested its boot environment, and now wants to unlock some storage that it's carrying locally. The beginning step of that is unwrapping the storage encryption key that it also has locally.

Maybe you should spell out the full flow. I don't think we're on the same page here.

tylerfanelli:
> Maybe you should spell out the full flow. I don't think we're on the same page here.

Sure, I've added a commit showing the flow of how this plugin is expected to be used. It doesn't touch on some of the details I'll be adding in another patch series such as HSM support. I can amend the docs to include this info at that time.

Let me know if you have questions or concerns.

kbs/docs/id-key.md
```diff
@@ -0,0 +1,84 @@
...
+
+With this, the guest owner is now free to configure persistent storage as they wish for their CVMs to have access to encrypted data when they boot. Upon successful attestation, this encrypted data will be presented to the KBS by the CVM for decryption.
+
+### Post-attestation, decrypting data
+
+With a CVM booted and attested, it is now assumed that the CVM was configured in such a way that it has access to readable storage that contains the encrypted data. The CVM can now send this data to the KBS for decryption.
+
+To prevent replay attacks, the CVM must wrap the encrypted data in a ECDH key cipher. To support this, the ID-key plugin presents an endpoint `ecdh-pub-sec1` that a VM can use to fetch the EC public key in order to wrap the encrypted data.
+
+`GET /kbs/v0/id-key/ecdh-pub-sec1`
```
mkulke, line 41: `ecdh-pub-sec1` seems to be a magic string on the `/kbs/v0/id-key/{string}` route, should we move it to another hierarchy?

kbs/docs/id-key.md
```diff
@@ -0,0 +1,84 @@
+# ID-key Plugin
```
mkulke, line 1: what is the rationale behind the name? it's not obvious to me what an ID-Key is, maybe spell that out in the doc.

kbs/docs/id-key.md
```diff
@@ -0,0 +1,84 @@
...
+
+As a guest owner would be responsible for encrypting persistent storage, they are the sole authority that is permitted to request resources be encrypted. The KBS is expected to be fully in control by the guest owner with credentials configured.
+
+## Flow
+
+### Provisioning encrypted data
+
+A credentialed guest owner is the only one authorized to request that keys be encrypted by the ID-key plugin. To do this, the guest owner will POST some plaintext data (created locally) to be encrypted. This plaintext data could be a LUKS passphrase, SecureBoot keys, etc. The actual purpose of the data is of no concern to the KBS.
+
+`POST /kbs/v0/id-key/{base64-plaintext-data}`
```
mkulke, line 15: would `POST /kbs/v0/id-key` w/ the plaintext in the body make more sense?

kbs/docs/id-key.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Data Encryption
+
+As a guest owner would be responsible for encrypting persistent storage, they are the sole authority that is permitted to request resources be encrypted. The KBS is expected to be fully in control by the guest owner with credentials configured.
+
+## Flow
+
+### Provisioning encrypted data
+
+A credentialed guest owner is the only one authorized to request that keys be encrypted by the ID-key plugin. To do this, the guest owner will POST some plaintext data (created locally) to be encrypted. This plaintext data could be a LUKS passphrase, SecureBoot keys, etc. The actual purpose of the data is of no concern to the KBS.
```
mkulke, line 13: can you specify guest-owner? would the POST endpoint be called from within a CVM? or would a this be performed by someone who has admin access to Trustee?

mkulke (commented):
I also have problems understanding the rationale behind the plugin, also after reading the docs and the comments. there seems to be some conceptual overlap with sealed secrets with the addition of a resource creating endpoint.

afaik we attempt to group stateful operations withing the realm of an admin hierarchy, which is guarded by some form of auth. Can we do something similar with the POST endpoint here? 

Will this plugin work with stateless KBS deployments (i.e. you cannot perform write ops on the API)?

Would you be able to add a simple use-case and describe how the existing facilities are lacking and how the plugin will address this to the PR description?

tylerfanelli:
@fitzthum @mkulke 

Sorry for the delay.

I've updated the PR description outlining the use-case and how existing facilities are lacking. I'll address the individual review questions under their respective conversations.

kbs/docs/id-key.md
tylerfanelli: To prevent replay attacks, the encrypted data is wrapped with an ECDH cipher. To build this cipher, the client needs the server's EC public key. This endpoint allows for fetching that public key.

I chose SEC1 encoding because its a "standard" format for encoding EC public keys, found in _Elliptic-Curve-Point-to-Octet-String encoding described in [SEC 1: Elliptic Curve Cryptography (Version 2.0) section 2.3.3 (page 10)](http://www.secg.org/sec1-v2.pdf)_.

What exactly do you mean by "hierarchy"? I think the existing way it's done is fine, but am open to suggestions if you think another would be an improvement.

kbs/docs/id-key.md
tylerfanelli: It was sort-of a reference to @fitzthum's [The Mystery of the KBS Identity](https://gist.github.com/fitzthum/f3124f846e6ecbc066c53c6d31668979). Since the NEK was encrypted from a known key provisioned on a known-trusted source, the very fact that it could be decrypted indicated that the client was communicating with the "true" KBS server.

Not saying a **good** reference, just a reference :smile: . If you feel another name is more appropriate, I'm open to changing it.

kbs/docs/id-key.md
tylerfanelli: I use "guest owner" as the one who has admin access to Trustee, and will be running their "owned" CVMs on an untrusted platform.

kbs/docs/id-key.md
mkulke: but isn't the identity of the KBS settled by the KBS cert that we inject into a guest in init-data?

kbs/docs/id-key.md
mkulke: it's a bit unusual/counter-intuitive for rest-ish APIs to use payload as part of the resource uri in such a request.

kbs/docs/id-key.md
mkulke: ah, this is also just a question about API design. it's not optimal to have a string with special meaning on the same hiearchy as resource descriptions: like `/apple/count-apples` and `/apple/1`, `/apple/2`, etc....

mkulke:
> @fitzthum @mkulke
> 
> Sorry for the delay.
> 
> I've updated the PR description outlining the use-case and how existing facilities are lacking. I'll address the individual review questions under their respective conversations.

Thank you for providing extensive context, it wasn't obvious to me that this is supporting a non-CoCo use case, so we don't have init-data or sealed secrets facilities. I'll attempt to review this more thoroughly with that in mind (but it's not trivial, and I have limited cycles atm, I'd hope someone else can chime in: cc @confidential-containers/trustee-maintainers). 

I do have the feeling that the problem could/should either be generalized so it can be supported in the plain kbs API or, on the other end, that the requirements are so specific that they would warrant a `/coconut` plugin.

kbs/docs/id-key.md
tylerfanelli: Ok, I can make the change for this. Perhaps it can just be `/kbs/v0/id-key/register` with the plaintext in the body?

kbs/docs/id-key.md
tylerfanelli: Oh ok, I can modify for the public key to be fetched at:
`/kbs/v0/id-key/ecdh-pub-sec1`

While the key decryption is at an endpoint like:
`/kbs/v0/id-key/decrypt/...`

Would this be preferred?

fitzthum:
Ok. Thanks for adding the details about the use case. It helps to see the bigger picture.

I think you can use a standard resource request for this. You mention a few reasons why you didn't go that way.

> Although the resource ID would be given in the virtual disk, there is nothing preventing a motivated bad actor from resources that pertain to other VMs, such as their NEKs. With some inside knowledge, an attacker could steal other VM NEKs and impersonate other VMs with their vTPM endorsement keys simply by knowing which resource ID corresponds to another VMs NEK.

To pass an attestation check, the guest must be running the expected SVSM code. Via attestation we know that this guest will have an SVSM that correctly emulates a vTPM and safeguards its endorsement key. The SVSM is protected both from the host and the guest. Probably you would already agree that we aren't worried that the nvdata will be stolen or extricated regardless of whether we use id-key or standard resources.

The concern is that one guest might be able to "impersonate" another. To me this sounds like we're heading for the tricky territory of identity. There's a lot that could be said here, but the most important thing is that I don't think id-key resolves this. You mention that an attacker could impersonate another guest if they know the resource uri of the NEK. This is equivalent to an attacker knowing the wrapped id-key. Consider a malicious host that simply switches two virtual disks. The result would be exactly the same with id-key or standard resources. It might seem like it's easier to guess the name of a resource than an id-key, but this isn't something to bank on (and you could always randomly generate your resource ids).

In your design it looks like you are not measuring anything related to the virtual disk. It seems like all your guests will look exactly the same in terms of TCB. In that case, there is very little you can do to distinguish them confidentially. I don't know exactly how you can provide guarantees about the identity, but a good first step is to measure the virtual disk somehow. For instance, you could put a hash of the nvdata into the init-data.

Possibly a better option is to simply not worry about guest identity. If I start a guest with the wrong nvdata, it simply becomes a different guest. Sure maybe someone can unlock my root disk if they steal my nvdata, but my root disk has my ssh public key inside of it so it's no big deal. I think this approach is very reasonable and much simpler.


> As we cannot guarantee that all clients using persistent vTPMs in this manner will use Trustee to attest, the virtual disk format cannot tie itself to any Trustee-specific mechanism (like giving a resource ID tag of the NEK). The virtual disk format must be server-agnostic. This applies for all attestation protocols, for that matter.

I don't see how id-key is any more generic than resource uris. In fact the resource uri thing is just a key value store,  which seems easier for someone else to implement than id-key.

> As the virtual disk must be server-agnostic, the result of an attestation should only be the unwrapping of the NEK. No other information is supplied to the virtual disk.

I don't quite understand this point.

tylerfanelli:
> Ok. Thanks for adding the details about the use case. It helps to see the bigger picture.
> 
> I think you can use a standard resource request for this. You mention a few reasons why you didn't go that way.
> 
> > Although the resource ID would be given in the virtual disk, there is nothing preventing a motivated bad actor from resources that pertain to other VMs, such as their NEKs. With some inside knowledge, an attacker could steal other VM NEKs and impersonate other VMs with their vTPM endorsement keys simply by knowing which resource ID corresponds to another VMs NEK.
> 
> To pass an attestation check, the guest must be running the expected SVSM code. Via attestation we know that this guest will have an SVSM that correctly emulates a vTPM and safeguards its endorsement key.

It's not SVSM who the KBS is talking to. Since SVSM does not have network access, it uses a proxy on the untrusted host for network access.

> 
> The concern is that one guest might be able to "impersonate" another. To me this sounds like we're heading for the tricky territory of identity.

It's not that a **guest** can get another's NEK (at that point, what could it do with it?). It's that the untrusted proxy could get ahold of one of its VMs (the proxy serves all VMs running on the system) NEK.

> There's a lot that could be said here, but the most important thing is that I don't think id-key resolves this. You mention that an attacker could impersonate another guest if they know the resource uri of the NEK. This is equivalent to an attacker knowing the wrapped id-key.

Not quite. With the resource URI, the proxy knows the exact resource name of a guest's NEK (it literally is responsible for building the resource URL). It can then use another guest's token to replay that very same resource fetch. This is impossible with id-key, as the encrypted NEK is wrapped with an ephemeral EC key such that the proxy can never replay it.

> Consider a malicious host that simply switches two virtual disks. The result would be exactly the same with id-key or standard resources. It might seem like it's easier to guess the name of a resource than an id-key, but this isn't something to bank on (and you could always randomly generate your resource ids).

There's other mitigations for detecting this scenario, such as vTPM monotonic boot counters and rolling hashes of vTPM EKs on boot for some specific NVDATA. This is not being solved by id-key, but other mechanisms.

> 
> In your design it looks like you are not measuring anything related to the virtual disk. It seems like all your guests will look exactly the same in terms of TCB. In that case, there is very little you can do to distinguish them confidentially. I don't know exactly how you can provide guarantees about the identity, but a good first step is to measure the virtual disk somehow. For instance, you could put a hash of the nvdata into the init-data.
> 

The NEK is how the KBS distinguishes VMs to an extent. I do like the idea of hashing the NVDATA into the init-data, but I cannot always guarantee that HOST_DATA will be settable (on a cloud instance, for example).

> > As we cannot guarantee that all clients using persistent vTPMs in this manner will use Trustee to attest, the virtual disk format cannot tie itself to any Trustee-specific mechanism (like giving a resource ID tag of the NEK). The virtual disk format must be server-agnostic. This applies for all attestation protocols, for that matter.
> 
> I don't see how id-key is any more generic than resource uris. In fact the resource uri thing is just a key value store, which seems easier for someone else to implement than id-key.
> 

I see your point, I suppose the data in the "NEK" buffer could be arbitrary, and treated differently based on how the server is implemented. My above points still stand though.

fitzthum:
> It's not that a guest can get another's NEK (at that point, what could it do with it?). It's that the untrusted proxy could get ahold of one of its VM's (the proxy serves all VMs running on the system) NEK.

If the KBS protocol is terminated properly in the guest, the proxy is not relevant here.

> Not quite. With the resource URI, the proxy knows the exact resource name of a guest's NEK (it literally is responsible for building the resource URL). It can then use another guest's token to replay that very same resource fetch. This is impossible with id-key, as the encrypted NEK is wrapped with an ephemeral EC key such that the proxy can never replay it.

It's true that id-key protects the request as it travels from the guest to the KBS, but that really doesn't matter since the virtual disk is exposed to the host.

> There's other mitigations for detecting this scenario, such as vTPM monotonic boot counters and rolling hashes of vTPM EKs on boot for some specific NVDATA. This is not being solved by id-key.

If these approaches solve the impersonation problem, then you definitely don't need id-key.

> The NEK is how the KBS distinguishes VMs to an extent. I do like the idea of hashing the NVDATA into the init-data, but I cannot always guarantee that HOST_DATA will be settable (on a cloud instance, for example).

If the TCB doesn't change, you can't make any confidential guarantees here.

I feel like we're going in circles a bit. Is there an institutional blocker here? Like a review from another community or a client? Why not just start with simple resources until someone complains, if they do?

tylerfanelli:
> I feel like we're going in circles a bit. Is there an institutional blocker here? Like a review from another community or a client? Why not just start with simple resources until someone complains, if they do?

The idea that guests could request resources that "weren't for them" was a big concern. However, you mentioned randomly-generating resource IDs such that they wouldn't be "guessable". I hadn't thought of that before. 

Generate a UUID per NEK, and embedding that in the virtual disk such that no other guests (w/o the correct disk) could know what to request. I think that could work pretty securely with the existing resource implementation and id-key wouldn't be necessary. The "swapping disks" problem is another issue entirely.

Do you agree?

fitzthum:
> Do you agree?

Yeah, I think one thing that could really help would be to make some tooling that could do all this automatically. This would be on the front-end of things. It could just be a bash script or two (we could create an integrations or examples dir), but we also have an option to extend the kbs-client. For instance we could add a plugin interface to the client allowing people to create their own logic for registering resources and creating policies.

tylerfanelli:
One desired thing that id-key does provide is the ability to generate the encryption key from an HSM. There's the PKCS11 plugin, but that just takes a label of something in an HSM and returns it, rather than encrypting and decrypting secrets while the secret never leaves the HSM.

This is useful, because it means the secret decryption itself is not tied to a specific KBS instance. That's not available in id-key yet, but I can add it.

tylerfanelli:
> One desired thing that id-key does provide is the ability to generate the encryption key from an HSM. There's the PKCS11 plugin, but that just takes a label of something in an HSM and returns it (which AFAIU is not recommended), rather than encrypting and decrypting data in-place while the key never leaves the HSM.
> 
> This is useful, because it means the secret decryption itself is not tied to a specific KBS instance. That's not available in id-key yet, but I can add it.

tylerfanelli:
One desired thing that id-key does provide is the ability to generate the encryption key from an HSM. There's the PKCS11 plugin, but that just takes a label of something in an HSM and returns it (which AFAIU is not recommended), rather than encrypting and decrypting data in-place while the key never leaves the HSM.

This is useful, because it means the secret decryption itself is not tied to a specific KBS instance. That's not available in id-key yet, but I can add it.

fitzthum:
> One desired thing that id-key does provide is the ability to generate the encryption key from an HSM. There's the PKCS11 plugin, but that just takes a label of something in an HSM and returns it (which AFAIU is not recommended), rather than encrypting and decrypting data in-place while the key never leaves the HSM.

Yeah the PKCS11 backend is very basic. There are some rough plans to improve it but they are low priority for me atm. The PKCS11 backend was developed just before the plugin infrastructure existed. It is a resource backend instead, which creates some limitations. We can't really pass sophisticated parameters to it. At the moment it uses the generic secret mechanism to store secrets as blobs. This is a valid PKCS11 flow but it's not what most people think of when using an HSM.

For a resource backend, keywrapping doesn't add a whole lot of value. It would be one thing if we could unwrap a key provided from the guest (more on this in a second) but the resource backend and the KBS protocol don't really support this. The main thing we miss out on currently is that it's tricky for people to bring their own HSM with keys already in it if they don't happen to be using the right key type. Note that for confidential containers, we have another method of accessing an HSM from inside the guest (sealed secrets) that allows people to use keywrapping.

I think it could be valuable to add more sophisticated HSM support. I mentioned earlier the idea of a remote vTPM provided by Trustee. In some ways that's not so different from the id-key stuff, but it's a little more generic. If we do want to introduce something like that I think we should consider adding two-way encryption to the KBS protocol itself rather than providing it in a plugin. There are some open questions here, but I think it's a good area to explore.

tylerfanelli:
I'm closing this as I now believe it'd be more worthwhile to add more sophisticated HSM support. As there's restrictions for it being used strictly as a resource backend, I'll try to explore it as its own plugin.

It could perhaps replace the PKCS11 resource backend, but that's for others to decide.

fitzthum:
> Doesn't really require changes to the KBS protocol, as what is actually returned from a KBS attestation is undefined. What do you think?

I'm not sure the best way to implement tbh. Currently we allow plugins to specify if they require encryption or admin authentication. It could be cool to have another field where they can specify that they require requests to be encrypted. Then the KBS would take care of it.

PR #700, opened 2025-02-06
config: fix insecure_key parameter

fitzthum:
`insecure_api` is a parameter of the admin config whereas `insecure_key` is part of the attestation token config. We use the wrong field in one of our config files.


Xynnn007 (approved):
Thanks for the fix!

PR #704, opened 2025-02-10
verifier: add tcb_info status, advisory_ids and collateral_expiration_status to policy

pawelpros:
Added tcb_status, advisory_ids and collateral_expiration_status to claims report which will be accessible in policies definition.

### DIFF
--- a/attestation-service/src/token/ear_default_policy.rego
+++ b/attestation-service/src/token/ear_default_policy.rego
@@ -103,6 +103,22 @@ hardware := 2 if {
 	input.tdx.quote.body.mr_seam in data.reference.mr_seam
 	input.tdx.quote.body.tcb_svn in data.reference.tcb_svn
 	input.tdx.quote.body.mr_td in data.reference.mr_td
+	# Check TCB status
+	# input.tdx.tcb_status == "OK"
+
+	# Check collateral expiration status
+	# input.tdx.collateral_expiration_status == "0"
+
+	# Check against allowed advisory ids
+	# allowed_advisory_ids := {"INTEL-SA-00837"}
+	# attester_advisory_ids := {id | id := input.attester_advisory_ids[_]}
+	# object.subset(allowed_advisory_ids, attester_advisory_ids)
+
+	# Check against disallowed advisory ids
+	# disallowed_advisory_ids := {"INTEL-SA-00837"}
+	# attester_advisory_ids := {id | id := input.tdx.advisory_ids[_]} # convert array to set
+	# intersection := attester_advisory_ids & disallowed_advisory_ids
+	# count(intersection) == 0
 }
 
 configuration := 2 if {
--- a/deps/verifier/src/tdx/quote.rs
+++ b/deps/verifier/src/tdx/quote.rs
@@ -7,10 +7,13 @@ use qvl::{
     tee_qv_get_collateral, tee_supp_data_descriptor_t, tee_verify_quote,
 };
 use scroll::Pread;
+use std::ffi::CStr;
 use std::mem;
+use std::os::raw::c_char;
 use std::time::{Duration, SystemTime};
 
 use intel_tee_quote_verification_rs as qvl;
+use serde_json::{Map, Value};
 
 pub const QUOTE_HEADER_SIZE: usize = 48;
 
@@ -395,7 +398,7 @@ pub fn parse_tdx_quote(quote_bin: &[u8]) -> Result<Quote> {
     }
 }
 
-pub async fn ecdsa_quote_verification(quote: &[u8]) -> Result<()> {
+pub async fn ecdsa_quote_verification(quote: &[u8]) -> anyhow::Result<Map<String, Value>> {
     let mut supp_data: sgx_ql_qv_supplemental_t = Default::default();
     let mut supp_data_desc = tee_supp_data_descriptor_t {
         major_version: 0,
@@ -454,7 +457,6 @@ pub async fn ecdsa_quote_verification(quote: &[u8]) -> Result<()> {
     };
 
     // set current time. This is only for sample purposes, in production mode a trusted time should be used.
-    //
     let current_time = SystemTime::now()
         .duration_since(SystemTime::UNIX_EPOCH)
         .unwrap_or(Duration::ZERO)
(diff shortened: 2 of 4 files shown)


deps/verifier/src/tdx/quote.rs
```diff
@@ -563,5 +619,10 @@ mod tests {
         let quote_bin = fs::read(quote).unwrap();
         let res = ecdsa_quote_verification(quote_bin.as_slice()).await;
         assert!(res.is_ok(), "{res:?}");
+
+        let claims = serde_json::to_string_pretty(&res.unwrap())
+            .expect("Custom claims are available.");
+
+        let _ = fs::write(format!("{quote}.json"), claims);
```
Xynnn007, line 626: We might need to add the file to `.gitignore`?

Xynnn007 (commented):
Thanks for this! Only one comment

deps/verifier/src/tdx/quote.rs
```diff
@@ -506,7 +509,60 @@ pub async fn ecdsa_quote_verification(quote: &[u8]) -> Result<()> {
...
+    claims_map.insert(
+        "tcb_status".to_string(),
+        Value::String(tcb_status_desc.to_string()),
+    );
+    claims_map.insert("advisory_ids".to_string(), Value::Array(advisory_ids));
+
+    Ok(claims_map)
+}
+
+pub struct SgxQlQvResultWrapper(pub sgx_ql_qv_result_t);
```
mythi, line 529: would `impl fmt::Display for qvl::sgx_ql_qv_result_t` work?

deps/verifier/src/tdx/quote.rs
```diff
@@ -506,7 +509,60 @@ pub async fn ecdsa_quote_verification(quote: &[u8]) -> Result<()> {
...
+
+    let mut claims_map = Map::new();
+
+    claims_map.insert(
+        "tcb_status".to_string(),
+        Value::String(tcb_status_desc.to_string()),
+    );
+    claims_map.insert("advisory_ids".to_string(), Value::Array(advisory_ids));
+
+    Ok(claims_map)
```
mythi, line 526: `az-vtpm-tdx` uses this result too (in the future SGX too if we de-duplicate this function implementation). Is it OK for that verifier to change to a serde output?

deps/verifier/src/tdx/quote.rs
```diff
@@ -506,7 +509,60 @@ pub async fn ecdsa_quote_verification(quote: &[u8]) -> Result<()> {
...
     }
 
-    Ok(())
+    let tcb_status_desc = SgxQlQvResultWrapper(quote_verification_result).as_str();
+
+    let advisory_ids: Vec<Value> = get_sa_list(&supp_data.sa_list).iter()
+        .map(|s| Value::String(s.to_string()))
+        .collect();
+
+    let mut claims_map = Map::new();
```
mythi, line 518: we may want `collateral_expiration_status` as  a token field too

deps/verifier/src/tdx/quote.rs
mythi: I don't see where the output is used so I'd suggest we just drop the write.

deps/verifier/src/tdx/quote.rs
mythi: A: no

attestation-service/src/token/ear_default_policy.rego
```diff
@@ -103,6 +103,20 @@ hardware := 2 if {
 	input.tdx.quote.body.mr_seam in data.reference.mr_seam
 	input.tdx.quote.body.tcb_svn in data.reference.tcb_svn
 	input.tdx.quote.body.mr_td in data.reference.mr_td
+
+	# Check TCB status
+	# input.tdx.tcb_status == "OK"
+
+	# Check against allowed adivosry ids
```
fitzthum, line 110: nit: `adivosry` -> `advisory`

attestation-service/src/token/ear_default_policy.rego
```diff
@@ -103,6 +103,20 @@ hardware := 2 if {
...
+	# allowed_advisory_ids := {"INTEL-SA-00837"}
+	# attester_advisory_ids := {id | id := input.attester_advisory_ids[_]}
+	# object.subset(allowed_advisory_ids, attester_advisory_ids)
+
+	# Check against disallowed adivosry ids
```
fitzthum, line 115: nit: `adivosry` -> `advisory`

deps/verifier/src/tdx/quote.rs
pawelpros: I have replaced it with expected comparison - should do the trick

deps/verifier/src/tdx/quote.rs
pawelpros: Updated claims to include it

deps/verifier/src/tdx/quote.rs
pawelpros: Added custom claims also to az_tdx_vtpm

pawelpros: Currently az_vtpm is already extending claims using similar approach so should be ok in my opinion

deps/verifier/src/tdx/quote.rs
```diff
@@ -506,7 +509,65 @@ pub async fn ecdsa_quote_verification(quote: &[u8]) -> Result<()> {
...
+        "tcb_status".to_string(),
+        Value::String(tcb_status_desc.to_string()),
+    );
+    claims_map.insert(
+        "collateral_expiration_status".to_string(),
+        Value::String(collateral_expiration_status.to_string()),
+    );
+    claims_map.insert("advisory_ids".to_string(), Value::Array(advisory_ids));
+
+    Ok(claims_map)
```
mythi, line 531: we could move the new code to the match arm that covers OK and the non-terminal results and return from there, wdyt?

attestation-service/src/token/ear_default_policy.rego
```diff
@@ -103,6 +103,22 @@ hardware := 2 if {
 	input.tdx.quote.body.mr_seam in data.reference.mr_seam
 	input.tdx.quote.body.tcb_svn in data.reference.tcb_svn
 	input.tdx.quote.body.mr_td in data.reference.mr_td
+	# Check TCB status
```
mythi, line 106: move the policy updates to a separate commit.

pawelpros: moved

mkulke (approved):

mythi (approved):
LGTM

fitzthum (approved):
Thanks!

PR #709, opened 2025-02-14
verifier: combined duplicated dcap implementation

pawelpros:
Separated duplicated ecdsa quote verification code used for dcap quote data.


mythi (approved):
de-duplicating code makes sense, thanks

fitzthum (approved):
LGTM

Issue #708, opened 2025-02-13
Post-merge jobs failing


fitzthum:
I think the Azure e2e jobs have been failing ever since https://github.com/confidential-containers/trustee/pull/689

Error:
```
./client: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by ./client)
```
cc @seungukshin @mkulke

mythi:
> Can we change the ubuntu version for the build environemnt?

#687 moved to use 24.04 but it had been like that since `ubuntu-latest` had defaulted to it for some time. What's the workflow that builds the artifacts?

seungukshin:
@mythi , [kbs-e2e.yml](https://github.com/confidential-containers/trustee/blob/main/.github/workflows/kbs-e2e.yml)'s `build-binaries` job creates the artifact. It was runned on ubuntu-22.04 before [this commit](https://github.com/confidential-containers/trustee/commit/838e49c8ca158acaf4659e592a47b34fbf1ae55c)

fitzthum:
Building the binaries is here https://github.com/confidential-containers/trustee/blob/f02634470ea35d9a9c1d205b260045a9182170c1/.github/workflows/kbs-e2e.yml#L25

And the test runs on this https://github.com/confidential-containers/trustee/blob/f02634470ea35d9a9c1d205b260045a9182170c1/.github/workflows/kbs-e2e-azure-vtpm.yml#L63

This was working fine before #689 tho which is odd. I looked through the history again and it does seem like that's when the error showed up (although we did have a different problem after #687)

mkulke:
github is rotating their `-latest` runners slowly to 24.04 I think, so that we probably would have to look at the logs closely to see which lts ubuntu the artifacts were built on.

I assume the self-hosted pool can be configured to use ubuntu 24.04 images, however it's not great that we break compatibility with LTS distros that are still widely used, so I would suggest to the fix the builder images to 22.04 for the time being.

mythi:
> So maybe we should just downgrade those builder images in the CI

I'd be OK with that but I'd still like to understand the core problem. Why `make e2e-test` does not retrigger a new build if `test.tar.gz` isn't reusable?

mkulke:
> I'd be OK with that but I'd still like to understand the core problem. Why make e2e-test does not retrigger a new build if test.tar.gz isn't reusable?

[...]

PR #711, opened 2025-02-19
ci: build kbs-e2e binaries on Ubuntu 22.04

mythi:
Fixes: #708

az-vtpm-* runners are still 22.04 based which makes kbs-client to fail when built with the tss2 libraries and cached on 24.04 runners.

Move e2e-kbs build-binaries job back to 22.04 instead of bumping the az-vtpm-* runners to 24.04.

### DIFF
--- a/.github/workflows/kbs-e2e.yml
+++ b/.github/workflows/kbs-e2e.yml
@@ -8,7 +8,7 @@ on:
         required: true
       runs-on:
         type: string
-        default: '["ubuntu-24.04"]'
+        default: '["ubuntu-22.04"]'
         description: JSON representation of runner labels
       tarball:
         type: string
@@ -22,9 +22,10 @@ defaults:
 
 jobs:
   build-binaries:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
     env:
       RUSTC_VERSION: 1.80.0
+      OS_VERSION: ubuntu-22.04
     steps:
     - name: Download artifacts
       uses: actions/download-artifact@v4
@@ -38,6 +39,7 @@ jobs:
         toolchain: ${{ env.RUSTC_VERSION }}
         components: rustfmt, clippy
         rustflags: ""
+        cache: false
 
     - name: Set up rust build cache
       uses: actions/cache@v4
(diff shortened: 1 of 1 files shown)


mythi:
> Move e2e-kbs build-binaries job back to 22.04 i

need to figure out how to bust the cache properly.

mkulke (approved):

fitzthum:
We need something else here to merge? Clearing the cached stuff?

mythi:
> We need something else here to merge? Clearing the cached stuff?

I'm not exactly sure what is needed. It looks Az forces Sample attester to also downgrade because of the same `test.tar.gz` cached artifact.

fitzthum:
> I'm not exactly sure what is needed. It looks Az forces Sample attester to also downgrade because of the same test.tar.gz cached artifact.

@mkulke do you know how we get this passing?

mkulke:
> > I'm not exactly sure what is needed. It looks Az forces Sample attester to also downgrade because of the same test.tar.gz cached artifact.
> 
> @mkulke do you know how we get this passing?

I don't think it'll pass in this PR, since it's using a workflow from the main branch, no? (due to the pull-request-target workflow)

mythi:
> > We need something else here to merge? Clearing the cached stuff?
> 
> I'm not exactly sure what is needed. It looks Az forces Sample attester to also downgrade because of the same `test.tar.gz` cached artifact.

It looks we had double caching which made the sample attester e2e test to fail: build-bins errored when re-building on 22.04 with a cache from 24.04. It wasn't enough to bust `actions/cache@v4` cache since `actions-rust-lang/setup-rust-toolchain@v1` added its own entries. I've now disable the latter and made `actions/cache@v4` the primary cache with an OS version included in the key.

> I don't think it'll pass in this PR, since it's using a workflow from the main branch, no? (due to the pull-request-target workflow)

I've dropped the `e2e_test` label. It should be OK to merge and see if it improves the `main` tests (i.e., fixes #708 properly).

fitzthum (approved):
Ok, might as well merge this since it's broken anyway

mythi:
> Ok, might as well merge this since it's broken anyway

it looks it's working now

fitzthum:
Thanks for fixing this. I guess we should have merged it sooner.

PR #712, opened 2025-02-19
kbs-client: make set-resource verify resource path

pmores:
set-resource allowed resources to be created at invalid paths since there has been no validation of path so far.  Such a resource can't even be retrieved since get-resource does validate resource path.

This commit changes that by introducing the same path check into set-resource that is called from get-resource.  If path validation fails on set-resource kbs-client now aborts with the same error message as on get-resource.

### DIFF
--- a/tools/kbs-client/src/lib.rs
+++ b/tools/kbs-client/src/lib.rs
@@ -12,6 +12,7 @@ use kbs_protocol::evidence_provider::NativeEvidenceProvider;
 use kbs_protocol::token_provider::TestTokenProvider;
 use kbs_protocol::KbsClientBuilder;
 use kbs_protocol::KbsClientCapabilities;
+use kbs_protocol::ResourceUri;
 use serde::Serialize;
 
 const KBS_URL_PREFIX: &str = "kbs/v0";
@@ -216,6 +217,10 @@ pub async fn set_resource(
     let token = auth_private_key.sign(claims)?;
 
     let http_client = build_http_client(kbs_root_certs_pem)?;
+
+    // verify 'path'
+    let resource_kbs_uri = format!("kbs:///{path}");
+    serde_json::from_str::<ResourceUri>(&format!("\"{resource_kbs_uri}\""))?;
 
     let resource_url = format!("{}/{KBS_URL_PREFIX}/resource/{}", url, path);
     let res = http_client


fitzthum (approved):
LGTM

pmores:
I don't know my way around this code base yet so I can't directly answer that question, but if I run `set-resource` with a bad path without this change `kbs-client` responds with
```
Set resource success 
 resource: MTIzNDU2Nzg5MGFiY2RlCg==
```

and `kbs` logs

```
[2025-02-19T16:08:28Z INFO  actix_web::middleware::logger] 127.0.0.1 "POST /kbs/v0/resource/default/smazat HTTP/1.1" 200 0 "-" "kbs-client/0.1.0" 0.000672
```

So it seems to me that KBS doesn't check either.

fitzthum:
> So it seems to me that KBS doesn't check either.

Hm that's surprising. We convert the request into a `ResourceDescription` [here](https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/implementations/resource/mod.rs#L35) and that [TryFrom](https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/implementations/resource/backend.rs#L35-L59) method has some logic in it.

Maybe we can make a separate issue about this.

pmores:
It's in the `kbs` log message in my previous message (basically it has just 2 parts instead of 3).

pmores:
Also thanks for the pointers, I'm happy to investigate as an exercise. :-)

pmores:
Apparently what happens in the `try_from()` function if the path only has 2 parts is that the two parts end up in the "type" and "tag" fields of `captures` respectively.  The "repo" part remains `None` which is then replaced by "default" on assignment to `ResourceDesc::repository_name` in the return statement of the function.

fitzthum:
Ah yeah that sound familiar. I guess that is a feature, although it's slightly odd. Maybe we should align more closely with the resource uri requirements? Maybe we don't need an extra check? I don't have a strong opinion. Probably still good to have a check in the client.

@Xynnn007 might have more thoughts on this

pmores:
Yes it looks intentional but the asymmetry in `kbs-client` seems rather confusing to me.  Without this PR you're allowed to leave out the repo name on `set-resource` but you still have to spell it out explicitly on a corresponding `get-resource`.

Is the "default" repo name substitution in KBS actually useful?  I'd expect something like that in software that's heavily used from command line, to save typing essentially.  I don't know if that's the case with KBS though.

fitzthum:
I suspect it was just intended to make things more convenient. I don't think anything really relies on it. I wouldn't mind requiring the resource description to have three parts.

pmores:
As an alternative approach, it would apparently be _technically_ easy enough to modify `ResourceUri::try_from()` to tolerate two-part paths, however that's in guest-components and there might be other users which rely on it bailing if 3 parts aren't present for all I know.  But if feasible this would enable the repo part to be omitted both on get and set.

fitzthum:
I think the two-part paths might be more confusing than convenient but either way is fine with me.

Xynnn007:
The default value is defined in https://github.com/confidential-containers/trustee/blob/main/kbs/docs/kbs_attestation_protocol.md#secret-resource. 

Automatically setting default can really cause confusion when clients set resources. There are two ways,

1. Add a `info!` when Resource URI automatically completes default.
2. Add a `info!` for kbs-client when setting resources to explicitly tell the user what the URI of the currently successfully set resource is.

It seems that method 2 is more convenient for the current PR. But we can consider updating the protocol and deleting default, although this will cause some upward compatibility issues.

pmores:
If repository being optional is directly part of the protocol definition then it might be more appropriate to change `ResourceUri::try_from()` to reflect that.  I think I'll be able to propose that fix later today, we can then decide which one is better.

pmores:
I've proposed a `ResourceUri` fix in guest-components [PR #917](https://github.com/confidential-containers/guest-components/pull/917), please take a look.

Xynnn007:
Sry, I must have caused some ambiguity for my mix the concepts of `ResourcePath` and `KBS Resource URI`.

`ResourcePath` is a concept in KBS. It is the access path you demonstrated in this PR and it will be completed as `default/default/smazat` in the internal logic of KBS. Note that there will be another `default` [here](https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/implementations/resource/backend.rs#L53). That means when we set resource using `kbs-client`, it allows to be a two-parts path. It is only the behavior of KBS. 


`KBS Resource URI` is something like `kbs:///default/1/2`. It is defined [in the doc](https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/docs/KBS_URI.md#specification). Now the `TryFrom` [logic](https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/deps/resource_uri/src/lib.rs#L102) is good to require a `repo` field. 

So overall there might be two things to do
1. Add some log macro to [that part](https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/implementations/resource/backend.rs#L53) to say something like `"warn: the requested path does not have a 'repo' field, and automatically set to 'default'. ". or directly show the whole resource path to be set. 
2. Do nothing to `resource_uri` crate.

pmores:
Thanks for your explanation @Xynnn007 , it makes sense.  However I'm afraid I fail to see how the proposed steps fix the problem I'm trying to fix.

Let me summarise the context to make 100% sure we're all on the same page.  When a two-part path is submitted to `kbs-client`'s `set-resource` command it's tolerated since the missing repo part is substituted with "default" in 'kbs'.  However when the same two-part path is then submitted to `get-resource` it's rejected as badly formed.

One way to solve this asymmetry is to make both `get-resource` and `set-resource` not tolerate two-part paths.  This is what this PR proposes but it might be inappropriate seeing as two-part paths are explicitly allowed by kbs-protocol.

So then the other way would be to make both commands tolerate two-part paths which means fixing `get-resource` handling.  Now the reason why `get-resource` fails with a two-part path is that `kbs-client`'s [`get_resource_with_attestation()`](https://github.com/confidential-containers/trustee/blob/2a742bef5ee926c2fa9c3ded00c25cddfae1b15e/tools/kbs-client/src/lib.rs#L82) is called which then invokes `KbsClient::get_resource()` [here](https://github.com/confidential-containers/trustee/blob/2a742bef5ee926c2fa9c3ded00c25cddfae1b15e/tools/kbs-client/src/lib.rs#L101).  As part of constructing its argument which is of type `ResourceUri`, `ResourceUri::try_from(url::Url)` is ultimately called.  This is where the two-part resource path is rejected and why I proposed the `ResourceUri` fix (in #917 against guest-components).

If we don't touch the `resource-uri` crate, how do we fix this?

Xynnn007:
@pmores 
Okay, what I originally thought was that after the KBS resource was set, the log would explicitly indicate the client that a certain field was set to default. But now I think what you want is that client-tool can support symmetric path format of set and get.

I think we can update the api description section of KBS protocol and also deny the `unwrap_or("default")` logic, both in this PR straightly.

Coincidentally, there will be a bi-weekly meeting next Monday. If possible, we can discuss this matter at the meeting together.

pmores:
Closing this in favour of solution taken in PR #720.  Making the repository path mandatory has an equivalent effect wrt this PR but it takes action at a more appropriate level.

PR #725, opened 2025-02-26
rust: go back to rust 1.80.0

fitzthum:
Right after I merged #597 there was a request from downstream to try to have less churn in the Cargo version.

I don't think updating the Cargo version is actually required for that PR, so here's a commit backing out that change.

If we take this commit, it will mean that people building without the lockfile will have to do a little but of massaging to get things to build. Specifically, I did this.
```
cargo update home@0.5.11 --precise 0.5.5
cargo update native-tls@0.2.14 --precise 0.2.10
```

It's not ideal to have people need to mess around like this when building things, so maybe we should just stick with 1.83.0. Let's discuss this over on https://github.com/confidential-containers/confidential-containers/issues/258. I am making this PR to show some of the tradeoffs.


fitzthum:
(gotta fixup josekit as well to make tests pass)

mythi:
> (gotta fixup josekit as well to make tests pass)

josekit is the reason why guest-components bumped to 1.80.0. josekit v0.9.0 moved to `LazyLock` from standard library available in 1.80..

fitzthum:
> josekit is the reason why guest-components bumped to 1.80.0. josekit v0.9.0 moved to LazyLock from standard library available in 1.80..

Maybe I'll go to 1.80.0

fitzthum:
Ok, let's give it a go with 1.80.0. I think it might be reasonable to use that version across the project.

mkulke (approved):

PR #726, opened 2025-02-27
tdx: fix bitflags serde bug

fitzthum:
These changes are very simple, but they resolve a tricky bug.

It all started when I tried to bump the guest components to the latest version. This caused a terrible error.
```
281 | / bitflags! {
282 | |     #[derive(Debug, Clone)]
283 | |     struct TdAttributesFlags: u64 {
284 | |         const DEBUG            = 1 << 0;
...   |
289 | |     }
290 | | }
    | | ^
    | | |
    | |_expected associated type, found `anyhow::Error`
    |   expected `std::result::Result<InternalBitFlags, <D as sample::serde::Deserializer<'de>>::Error>` because of return type
    |
    = note: expected enum `std::result::Result<_, <D as sample::serde::Deserializer<'de>>::Error>`
               found enum `std::result::Result<_, anyhow::Error>`
    = note: this error originates in the macro `$crate::__impl_external_bitflags_serde` which comes from the expansion of the macro `bitflags` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider constraining the associated type `<D as sample::serde::Deserializer<'de>>::Error` to `anyhow::Error`
```

What's going on here? It has to do with the `serde` feature of the `bitflags` crate. You can isolate the problem a little bit by sticking with the current version of `guest-components` and turning on the `serde` feature. That will cause the same error. With `guest-components` we get this feature from an indirect dependency, but the root of the problem is that this TDX code cannot tolerate it for some reason.

Turns out the reason is that we imported everything from `Anyhow`. This includes re-importing `Error` which is used in the `bitflags` macro expansion.

To fix this, just import `Result`, which is the better practice anyway. Also, enable the `serde` feature in `bitflags` in Trustee to try to avoid introducing similar snafus in the future.

@ssolit also ran into this problem in https://github.com/confidential-containers/guest-components/issues/906 but was able to sidestep it.

cc @mythi @pawelpros


Cargo.lock
```diff
@@ -2243,11 +2246,11 @@ dependencies = [
 
 [[package]]
 name = "home"
-version = "0.5.11"
+version = "0.5.5"
```
mythi, line 2249: is `home` downgrade related or just a leftover from some other work?

mythi (commented):
> To fix this, just import Result, which is the better practice anyway. Also, enable the serde feature in bitflags in Trustee to try to avoid introducing similar snafus in the future.

Makes sense, thanks a lot for getting on top of the issue!

One question about `home` still.

mkulke (approved):
well spotted. * imports are indeed unfortunate, I was hoping the linter would complain about those by default, but apparently not. there is a clippy rule though: [wildcard_imports](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/wildcard_imports.rs)

Cargo.lock
fitzthum: unrelated. Was using the 1.80.0 environment that requires this. I will rebase.

PR #727, opened 2025-02-28
build: drop libtdx-attest

mythi:
kbs-client w/ tdx-attester does not depend on libtdx-attest anymore so drop that install dependency.


fitzthum (approved):
LGTM

PR #728, opened 2025-03-01
kbs: Add support for two-way encryption

tylerfanelli:
KBS offers a mechanism for plugin responses (such as resources) to be encrypted with the client's TEE public key. However, there is no mechanism to specify that requests' bodies themselves to be encrypted, which would be useful (for example, to prevent replay attacks in key unwrapping).

Plugins that would like to use two-way encryption need a public key from the KBS to encrypt the data being sent. Upon successful attestation, a RSA public/private key pair is created and public key is returned with the attestation token. For plugins that do not use two-way encryption, clients can ignore the key.

For plugins that would like to use two-way encryption, they can use the public key to encrypt their request bodies and protect data in transit.

Each plugin has the option to enable two-way encryption. As no plugin currently uses two-way encryption, all existing plugins disable the feature.

### DIFF
--- a/kbs/src/api_server.rs
+++ b/kbs/src/api_server.rs
@@ -9,6 +9,8 @@ use actix_web::{
 use actix_web_httpauth::headers::authorization::{Authorization, Bearer};
 use anyhow::Context;
 use log::info;
+#[cfg(feature = "as")]
+use openssl::rsa::Padding;
 
 use crate::{
     admin::Admin, config::KbsConfig, jwe::jwe, plugins::PluginManager, policy_engine::PolicyEngine,
@@ -202,7 +204,9 @@ pub(crate) async fn api(
                     plugin_name: plugin_name.to_string(),
                 })?;
 
-            let body = body.to_vec();
+            #[cfg(feature = "as")]
+            let mut body = body.to_vec();
+
             if plugin
                 .validate_auth(&body, query, additional_path, request.method())
                 .await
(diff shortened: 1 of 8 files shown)


fitzthum:
Ok, before we get too far down the road on this, one basic question. If we give the guest the public key of the KBS, can we use TLS for this? That will protect communication from the client to the KBS and it will ensure that the guest is talking to the expected KBS (just don't ask me how they get the kbs public key).

tylerfanelli:
I guess it depends on how the guest is provisioned in the first place. Is there some configuration file for guest's with CoCo?  For all of my use-cases there is some configuration info/file.

But that's for getting the public key. **Trusting** the public key given (from a guest perspective) is the unknown bit, right?

fitzthum:
We would generally use init-data to provision the kbs public key, so it would be measured, but we're headed straight for the ol' KBS public key paradox. Do you get anything from measuring the KBS public key when the KBS itself is validating the evidence? Using TLS here wouldn't really guarantee who you are talking to, just that the channel is protected, but maybe that is equivalent to your proposal?

tylerfanelli:
> We would generally use init-data to provision the kbs public key, so it would be measured, but we're headed straight for the ol' KBS public key paradox.

Yea, and I'm not much concerned with this problem.

> Do you get anything from measuring the KBS public key when the KBS itself is validating the evidence?

IMO, no.

> Using TLS here wouldn't really guarantee who you are talking to, just that the channel is protected, but maybe that is equivalent to your proposal?

Precisely. This doesn't solve anything that TLS couldn't. But if you don't get the authentication guarantees of TLS, is it worth adding?

tylerfanelli:
Updated to include trait implementations in PKCS11 plugin.

tylerfanelli:
Ping @fitzthum @Xynnn007 Any thoughts here? I don't think we get much with TLS. To me, this is a simpler approach. I'm wondering if anybody feels differently.

fitzthum:
I think generally we should start pushing people to use HTTPS all the time. Without HTTPS enabled I think there might be a risk of people steeling the session cookie used for the RCAR handshake. We should probably investigate this more thoroughly. Currently the biggest obstacle to using HTTPS seems to be that it's a little more annoying to setup. Maybe we should try to update our tooling so that the basic KBS deployments can have HTTPS setup by default.

PR #731, opened 2025-03-03
bumps shadow-rs to 1.0.1 and adjust build.rs

ssolit:
Hello, this bumps the version of the shadow-rs dependency, as the old one triggers security warnings from dependabot.


fitzthum:
Looks like you're using a newer version of Rust locally, which causes some problems with the Cargo.lock.

Can you try `rustup override set 1.80.0` (assuming you use rustup) and build again.

You may need to set a few dependencies with `cargo update foo@bar --precise baz`

Sorry for the inconvenience but we are trying to avoid constantly bumping the MSRV.

ssolit:
It looks like the version of shadow-rs being introduced has a chain of dependencies (tz-rs @ 0.7+, tzdb @ 0.7+, and tzdb_data @ 0.2.1) that require at least Rust 1.81.0.

I tried to update them with ```
cargo update -p tz-rs@0.7.0 --precise 0.6.14
cargo update -p tzdb@0.7.2 --precise 0.5.10
cargo update -p tzdb@0.7.2 --precise 0.6.1
cargo update -p tzdb_data@0.2.1 --precise 0.1.4
```

and get errors like ```
failed to select a version for the requirement `tz-rs = "^0.7.0"`
candidate versions found which didn't match: 0.6.14
location searched: crates.io index
required by package `tzdb v0.7.2`
    ... which satisfies dependency `tzdb = "^0.7.2"` (locked to 0.7.2) of package `shadow-rs v1.0.1`
    ... which satisfies dependency `shadow-rs = "^1.0.1"` (locked to 1.0.1) of package `attestation-service v0.1.0 (/home/azureuser/coco/ssolit_trustee/attestation-service)`
    ... which satisfies path dependency `attestation-service` (locked to 0.1.0) of package `integration-tests v0.1.0 (/home/azureuser/coco/ssolit_trustee/integration-tests)`
    ```
    
  This might be blocked until the mcrv is bumped from 1.80 to 1.81

Xynnn007:
Ok. Now it is 1.3.0 and let's close this

PR #734, opened 2025-03-06
Bump Maximum version of SNP Attestation Report.

DGonzalezVillal:
Version 4 of the attestation report has arrived into some people's machines, so we have to bump up the minimum version to allow them to continue to attest their coco environment.

Slack issue:
https://cloud-native.slack.com/archives/C039JSH0807/p1740776076921049

### DIFF
--- a/deps/verifier/src/snp/mod.rs
+++ b/deps/verifier/src/snp/mod.rs
@@ -51,7 +51,7 @@ const KDS_VCEK: &str = "/vcek/v1";
 
 /// Attestation report versions supported
 const REPORT_VERSION_MIN: u32 = 2;
-const REPORT_VERSION_MAX: u32 = 3;
+const REPORT_VERSION_MAX: u32 = 4;
 
 #[derive(Debug)]
 pub struct Snp {


fitzthum (commented):
Since we haven't updated to the latest virtee/sev code, what will happen if we actually get a v4 (or v3) report? Will it deserialize it with some fields missing or will it explode?

DGonzalezVillal:
> Since we haven't updated to the latest virtee/sev code, what will happen if we actually get a v4 (or v3) report? Will it deserialize it with some fields missing or will it explode?

So without the Virtee changes with the V3 of the report, only some fields won't be present but attestation and verification should still work. This has been confirmed internally so V3 reports should not be an issue. 

V4 of the report might be a different story although, and we might need to do some internal testing. V4 is introducing a change to TCB. The structure of the TCB will be different in V4 reports depending on your Processor model! 

So Genoa and Milan will retain the known TCB format, but Turin (and new gen Procs) will have the new TCB format. This may affect verification if you are doing any type of TCB verification, since the new TCB will have entries moved around to different places. You can find more info in the [SNP ABI](https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/specifications/56860.pdf).

The good news is that we just merged a big change to the SEV library (https://github.com/virtee/sev/pull/278) and after we are done with some testing, we will release a library crate that contains support for both V3 and V4 versions of the report.

I will then put a PR to do the proc generation certificate (add support for Genoa and Turin) and will make sure that the changes work fine with v3 and v4 of the report.

PR #735, opened 2025-03-07
kbs/plugins: Replace PKCS11 resource backend with its own plugin

tylerfanelli:
This PR replaces the pkcs11 resource backend with its own plugin. The motivation is to support other PKCS11  features such as key wrapping, but this also can serve to enable other features present in HSMs like digital signatures and authentication.

## Resource Storage

The resource storage backend functionality is preserved, storing and retrieving secrets were originally done by setting the storage backend in the resource plugin to be PKCS11. With this plugin, the original resource URL to store/fetch secrets:

`/kbs/v0/resource/{REPOSITORY}/{TYPE}/{TAG}`

can be replaced by:

`/kbs/v0/pkcs11/resource/{REPOSITORY}/{TYPE}/{TAG}`

and the behavior will remain the same.

## Key wrapping/unwrapping

Upon initialization of the plugin, a unique public/private keypair is generated for wrapping and unwrapping data with an HSM.

### Key wrapping

To wrap data, clients can send a `POST` request to the `kbs/v0/pkcs11/wrap-key` endpoint giving the plaintext data in the request body. The PKCS11 plugin will wrap the data with the public key and return it to client.

Key wrapping requires admin validation, as only admins should be able to wrap data for attestation clients.

### Key unwrapping

To unwrap data, clients can send a `GET` request to the `kbs/v0/pkcs11/wrap-key` endpoint giving the wrapped data in the request body. The PKCS11 plugin will unwrap the data with the private key and return it to client.


kbs/src/plugins/implementations/pkcs11.rs
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        let mut session = pkcs11.open_rw_session(slots[slot_index])?;
+        session.login(UserType::User, Some(&AuthPin::new(config.pin.clone())))?;
+
+        // Generate a UUID to for the wrapping keypair.
+        let wrapkey_id = Uuid::new_v4();
+
+        // Create the HSM wrapping keypair.
+        Pkcs11Backend::wrap_key_new(&mut session, &wrapkey_id)?;
```
fitzthum, line 62: In the future we may want to do this lazily to avoid using keyslots for people who don't need wrapping keys. Fine for now tho.

fitzthum (commented):
This looks generally good. There's probably a bunch of other PKCS11 features we can add in the future.

Can you update the docs a little bit. We probably want to remove [this](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/resource_storage_backend.md#pkcs11) and add something [here](https://github.com/confidential-containers/trustee/tree/main/kbs/docs/plugins).

Also, at some point it would be good to setup softhsm in the CI and un-ignore the tests, but we can do that later I guess.

tylerfanelli:
Thanks for the review. I've updated and addressed your suggestions, and I'm updating the docs now.

kbs/src/plugins/implementations/pkcs11.rs
tylerfanelli: Agreed. I also want to add a config option if the key already exists on the HSM. The user could just configure the plugin to use a certain keypair.

tylerfanelli:
@fitzthum would you mind re-reviewing?

kbs/src/plugins/implementations/pkcs11.rs
```diff
@@ -0,0 +1,332 @@
...
+use super::super::plugin_manager::ClientPlugin;
+
+#[derive(Derivative, Deserialize, Clone, PartialEq)]
+#[derivative(Debug)]
+pub struct Pkcs11Config {
+    /// Path to the PKCS11 module.
+    module: PathBuf,
+
+    /// The index of the slot to be used. If not provided, the first slot will be used.
+    slot_index: Option<u8>,
```
Xynnn007, line 30: We can have
```rust
#[serde(default)]
slot_index: u8
```

so it will be set to `0` if not provided during deserializing the config

kbs/src/plugins/implementations/pkcs11.rs
```diff
@@ -0,0 +1,332 @@
...
+
+        // check that object has a readable value attribute
+        let value_attribute = vec![AttributeType::Value];
+        let attribute_map = session.get_attribute_info_map(object, value_attribute.clone())?;
+        let Some(AttributeInfo::Available(_size)) = attribute_map.get(&AttributeType::Value) else {
+            bail!("Key does not have value attribute available.");
+        };
+
+        // get the value
+        let value = &session.get_attributes(object, &value_attribute)?[0];
```
Xynnn007, line 146: looks like we are directly indexing the result here `[0]`, this would cause panic if the returned vector is empty. Could we use an error handling like
```rust
let value = &session.get_attributes(object, &value_attribute)?.get(0).ok_or(anyhow!("empty attributes returned"))?;
```

kbs/src/plugins/implementations/pkcs11.rs
tylerfanelli: Fixed in latest revision.

fitzthum (approved):
Looks good. One nit

kbs/docs/plugins/pkcs11.md
```diff
@@ -0,0 +1,33 @@
+# PKCS11 plugin
+
+PKCS11 is a broad specification supporting many cryptographic operations, mainly applicable to Hardware Security Modules (HSMs). Trustee implements a subset of features that can be of use to attestated clients.
```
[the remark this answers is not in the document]
tylerfanelli: Must've been tired :) fixed.

Xynnn007 (approved):

kbs/src/plugins/implementations/pkcs11.rs
```diff
@@ -171,6 +186,87 @@ impl Pkcs11Backend {
...
+            .await
+            .find_objects(&pubkey_template)
+            .context("unable to find public wrap key in PKCS11 module")?;
+
+        let encrypted = self
+            .session
+            .lock()
+            .await
+            .encrypt(&Mechanism::RsaPkcs, pubkey.remove(0), body)
+            .context("unable to decrypt HTTP body with private wrap key")?;
```
pmores, line 244: Is the message a copy/paste typo?

kbs/docs/plugins/pkcs11.md
```diff
@@ -0,0 +1,33 @@
+# PKCS11 plugin
+
+PKCS11 is a broad specification supporting many cryptographic operations, mainly applicable to Hardware Security Modules (HSMs). Trustee implements a subset of features that can be of use to attested clients.
+
+## Resource Storage
+
+The PKCS11 plugin can store plaintext resources in an HSM. Since the KBS expects resources to be in plaintext, we store these resources in the HSM as secret keys of the generic secret type. This storage interface will provision resources to the HSM in the expected way when a user uploads a resource to the KBS. The user must simply specify the location of an initialized HSM slot. A user can fetch resources using the same semantics as a resource backend.
```
pmores, line 7: Just out of curiosity, why are resources stored as keys?  Wouldn't say private data objects be a better match in terms of semantics?  (Apologies if I'm making no sense, I don't have experience with PKCS#11.)

kbs/src/plugins/implementations/pkcs11.rs
tylerfanelli: Yes, good catch. Fixed in latest revision.

kbs/docs/plugins/pkcs11.md
tylerfanelli: IIUC although called "secret keys", they can't be used as "keys" for encryption, decryption, etc. They can be thought of a "secret objects" in that sense.

See [here](https://www.cryptsoft.com/pkcs11doc/v220/group__SEC__12__7__2__GENERIC__SECRET__KEY__OBJECTS.html).

fitzthum:
Hopefully we can follow this up to add some more features, testing, etc

tylerfanelli:
Some CI testing is on my radar now. Once that's done, we can unignore the test (or perhaps set them behind some feature that we'd expect only the CI to enable?)

PR #738, opened 2025-03-11
Arm CCA local verifier

thomas-fossati:
This PR adds a local verifier for Arm CCA evidence based on [veraison/rust-ccatoken](https://github.com/veraison/rust-ccatoken).

The existing remote verification using the Versaison API has been brought up to date with HTTPS support, discovery API, and the new EAR “CCA submods” arrangement.  In this mode, the Linaro public Veraison instance is used by default.

Some refactoring was necessary to support remote mode alongside local, which created some extra churn in the existing codebase.  Sorry!

The functionality (local and remote) has been end-to-end tested, and I believe it is complete.

I do plan to make some additional clean-ups (docs, refactoring, tooling) but those will be addressed in a subsequent set of PRs.


deps/verifier/src/cca/local.rs
```diff
@@ -0,0 +1,139 @@
...
+    // Check that the Realm token was correctly signed using the RAK and that
+    // the RAK was correctly attested.
+    if realm_tvec.instance_identity.tier() != TrustTier::Affirming {
+        bail!("RAK signature or RAK attestation could not be verified");
+    }
+
+    // Synthesize TCB claims the way EAR wants to report them:
+    // realm part
+    let realm_annotated_evidence =
+        realm_annotated_evidence(&e).context("syntesizing CCA Realm TCB claims-set")?;
```
fitzthum, line 57: nit: `syntesizing` -> `synthesizing`

deps/verifier/src/cca/remote.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    if ca_cert.is_some() {
+        cr_builder = cr_builder.with_root_certificate(ca_cert.clone().unwrap())
+    }
+
+    let cr = cr_builder
+        .build()
+        .context("Failed to start challenge-response API with the service")?;
+
+    let n = Nonce::Value(expected_report_data.clone());
```
fitzthum, line 68: Let me double-check what's going on here. Nonce could mean a couple different things. It could be a value that is used to make sure that the token we get from veraison is fresh. If this were the case I would expect it to be picked randomly here and not related to the `expected_report_data` although maybe you could reuse that value.

The `expected_report_data` is to be compared against the report data in the hardware evidence (idk the name for this with ARM) to make sure that the key + nonce use for the rcar handshake are bound to the hw evidence and can't be changed out.

It's a little unclear which of these two things is happening.

deps/verifier/src/cca/mod.rs
```diff
@@ -227,27 +210,28 @@ impl Verifier for CCA {
 ///    }
 /// }
 /// NOTE: each of the value are base64 encoded hex value.
-fn parse_cca_evidence(evidence_map: &BTreeMap<String, RawValue>) -> Result<Evidence> {
-    let mut evidence = Evidence::default();
-    let platfrom = evidence_map
-        .get("platform")
-        .context("get platform evidence from the cca evidence map")?;
-    let output = serde_json::to_string(platfrom)?;
+fn parse_cca_evidence(
```
fitzthum, line 213: This comment doesn't relate directly to this line, but it's important to know that Trustee now basically requires that verifiers return a `report_data` claim (or a secure channel won't be established) and encourages verifiers to return an `init_data` claim if they check the initdata (which you do). Adding these values to the claims signals that the verifier has checked these properties.

deps/verifier/test_data/cca/scripts/kbs-client.sh
fitzthum, line 1: We have a `kbs-client` tool already so we may want to call this something slightly different like `kbs-test-client` or something.

fitzthum (commented):
Nice. I only looked at half of this but I had a few comments.

One high-level question is whether we can integrate the remote verifier more directly. The KBS can take an EAR attestation token as input and do its thing without ever contacting the AS. You just gotta add the path of the tee-pubkey [here](https://github.com/confidential-containers/trustee/blob/bc26ac5acd8314ba34fe837d6886beb1ce384106/kbs/src/token/mod.rs#L79). So in theory we could easily support a passport model with veraison + KBS. 

This doesn't necessarily preclude also having a remote verifier. There are some advantages to the current approach, but would simplify things. Obviously we could remove some code, but also the local verifier wouldn't have to use an EAR token as an intermediary, which is slightly clunky. Food for thought, but feel free to stick to the current approach.

thomas-fossati:
> Nice. I only looked at half of this but I had a few comments.
> 
> One high-level question is whether we can integrate the remote verifier more directly. The KBS can take an EAR attestation token as input and do its thing without ever contacting the AS. You just gotta add the path of the tee-pubkey [here](https://github.com/confidential-containers/trustee/blob/bc26ac5acd8314ba34fe837d6886beb1ce384106/kbs/src/token/mod.rs#L79). So in theory we could easily support a passport model with veraison + KBS.
> 
> This doesn't necessarily preclude also having a remote verifier. There are some advantages to the current approach, but would simplify things. Obviously we could remove some code, but also the local verifier wouldn't have to use an EAR token as an intermediary, which is slightly clunky. Food for thought, but feel free to stick to the current approach.

Thanks very much, @fitzthum! As it is IETF week, my availability is somewhat limited. In the meantime, I wanted to ack your review and explain why my replies might be a bit sluggish.
Talk to you soon!

deps/verifier/src/cca/mod.rs
```diff
@@ -79,71 +84,49 @@ impl Verifier for CCA {
...
-        let dk = jwt::DecodingKey::from_ec_pem(public_key_pem.as_bytes())
-            .context("get the decoding key from the pem public key")?;
-        let plain_ear = Ear::from_jwt(result.as_str(), jwt::Algorithm::ES256, &dk)
-            .context("decrypt the ear with the decoding key")?;
-
-        let ear_nonce = plain_ear.nonce.context("get nonce from ear")?;
-        let nonce_byte = base64::engine::general_purpose::STANDARD
-            .decode(ear_nonce.to_string())
-            .context("decode nonce byte from ear")?;
+        info!("EAR: {:?}", ear);
```
Xynnn007, line 112: Could we mark this log `debug` level, or add some context if we want to keep `info!`

deps/verifier/src/cca/mod.rs
```diff
@@ -227,27 +210,28 @@ impl Verifier for CCA {
...
 
     Ok(evidence)
 }
 
-fn cca_generate_parsed_claim(tcb: Evidence) -> Result<TeeEvidenceParsedClaim> {
-    let v = serde_json::to_value(tcb).context("build json value from the cca evidence")?;
+fn cca_generate_parsed_claim(tcb: EvidenceClaimsSet) -> Result<TeeEvidenceParsedClaim> {
+    let v = serde_json::to_value(tcb).context("serializing CCA evidence claims into JSON")?;
+
+    info!("CCA claims-set: {v}");
```
Xynnn007, line 233: can we set this log level to `debug`?

deps/verifier/src/cca/remote.rs
```diff
@@ -0,0 +1,104 @@
...
+) -> Result<Ear, anyhow::Error> {
+    let config::CcaVerifierConfig::Remote { address, ca_cert } = config.cca_verifier else {
+        panic!("expecting remote configuration")
+    };
+
+    let mut disco_builder = DiscoveryBuilder::new().with_base_url(address.clone());
+
+    if ca_cert.is_some() {
+        disco_builder = disco_builder.with_root_certificate(ca_cert.clone().unwrap())
+    }
```
Xynnn007, line 41: We can use to avoid `unwrap` here

```rust
if let Some(ca_cert) = ca_cert {
    disco_builder = disco_builder.with_root_certificate(ca_cert.clone());
}
```

deps/verifier/src/cca/local.rs
```diff
@@ -0,0 +1,139 @@
...
+use ear::{Appraisal, Ear, RawValue, VerifierID};
+use ear::{Extensions, TrustTier};
+use log::debug;
+use std::collections::BTreeMap;
+use std::fs;
+
+pub fn verify(
+    config: Config,
+    token: &Vec<u8>,
+    _expected_report_data: &[u8],
```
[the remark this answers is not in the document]
Xynnn007: To add some context here, [like other platforms](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/tdx/mod.rs#L71-L77), the parameter `expected_report_data` is to used to compare with the `report_data/challenge` field when getting the evidence.

Probably we need to compare this parameter with the `challenge` field of cca token.

deps/verifier/test_data/cca/README.md
```diff
@@ -0,0 +1,85 @@
+Set `TRUSTEE_SRC` to the absolute path of the checked out trustee repo:
```
Xynnn007, line 1: We could put the README for cca directly under https://github.com/confidential-containers/trustee/tree/main/deps/verifier/src/cca like [ibmse does](https://github.com/confidential-containers/trustee/tree/main/deps/verifier/src/se)

deps/verifier/test_data/cca/conf/default.rego
```diff
@@ -0,0 +1,42 @@
...
+# This is the default, unless the RIM claim matches one of the configured
+# reference values.
+default executables := 33
+# The value 3 stands for
+# "Only a recognized genuine set of approved executables have
+#  been loaded during the boot process."
+# the RIM (realm initial measurement) must match
+executables := 3 if {
+	input.cca.realm["cca-realm-initial-measurement"] in data.reference["cca.realm.cca-realm-initial-measurement"]
+}
```
Xynnn007, line 42: Probably we could also add this part to default policy of CCA https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/token/ear_default_policy.rego

Xynnn007 (commented):
Thanks @thomas-fossati for this! I am happy that we will soon see CCA can be supported via local verification logic! Only some nits.

Also, please check with what @fitzthum mentioned about integrating KBS with veraison. Some reference links of Intel ITA 
https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/backend.rs#L131-L137

https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/intel_trust_authority/mod.rs#L97

deps/verifier/src/cca/remote.rs
```diff
@@ -0,0 +1,104 @@
...
+    debug!("server challenge: {:?}", nonce);
+    debug!("acceptable media types: {:#?}", accept);
+    // TODO: Get the CCA media type from the slice of `accept`.
+    Ok((token, MEDIA_TYPE.to_string()))
+}
+
+pub async fn verify(
+    config: Config,
+    token: &[u8],
+    expected_report_data: &Vec<u8>,
```
huoqifeng, line 31: I'm also curious why local and remote verifier don't have same algorithm. shouldn't the only difference is the method to get the certificates?

huoqifeng (commented):
Thank you very much to bring the CCA verifier! @thomas-fossati , I'm still looking at the code and may need more time to complete it, a general comments is shall the local and remote verify function measure same aspect?

deps/verifier/src/cca/remote.rs
thomas-fossati: TL; DR.  It’s the nonce in the token submitted for appraisal.

A slightly longer story: Veraison appraisal API has two “modes”: 

* Veraison is the challenger:
```http
POST /challenge-response/v1/newSession?nonceSize=64
```

* The relying party (in this case, Trustee) is the challenger:
```http
POST /challenge-response/v1/newSession?nonce=base64_encoded_value
```

In this case, we use the latter since the RP has already chosen the challenge (`expected_report_data`) and needs to tell Veraison which value to check in the submitted evidence.

deps/verifier/src/cca/remote.rs
thomas-fossati: The appraisal algorithm is (in theory) the same, except it’s hidden in two different black boxes:
1. For local, the rust-ccatoken evidence appraisal function,
2. For remote, the Veraison CCA appraisal plugin.

Given the same evidence payload, the resulting trustworthiness vector should be identical.

A slight difference is the freshness check, which Veraison handles as part of the appraisal, whereas in “local” mode it needs to be done manually.

Note: CCA platform verification does not use certificates (in the sense of X.509 certs).  Platform evidence is verified using a raw public key endorsed alongside the platform instance identifier.

deps/verifier/test_data/cca/conf/default.rego
thomas-fossati: This is a great suggestion.  However, I’d like to take some time to incubate it a little bit more before making it part of the global policy.

I’ll raise an issue for tracking this activity.

thomas-fossati: See #753

thomas-fossati:
> Nice. I only looked at half of this but I had a few comments.
> 
> One high-level question is whether we can integrate the remote verifier more directly. The KBS can take an EAR attestation token as input and do its thing without ever contacting the AS. You just gotta add the path of the tee-pubkey [here](https://github.com/confidential-containers/trustee/blob/bc26ac5acd8314ba34fe837d6886beb1ce384106/kbs/src/token/mod.rs#L79). So in theory we could easily support a passport model with veraison + KBS.
> 
> This doesn't necessarily preclude also having a remote verifier. There are some advantages to the current approach, but would simplify things. Obviously we could remove some code, but also the local verifier wouldn't have to use an EAR token as an intermediary, which is slightly clunky. Food for thought, but feel free to stick to the current approach.

Thanks for the suggestion @fitzthum - and @xynnn007 for the useful pointers.
I wasn’t aware of this possibility, which sounds very cool.
However, here I have only rearranged the pre-existing remote verifier, and I’d rather avoid any rearchitecting in the context of this PR.

I’ll raise an issue to do this separately.

thomas-fossati:
> […]
> I’ll raise an issue to do this separately.

See #754

deps/verifier/src/cca/mod.rs
thomas-fossati: Apologies. To make progress on this, I need to understand two things:

1. how to populate `report_data` and `init_data` (with the successfully matched values?), and
2. where to put those claims (in the `TeeEvidenceParsedClaim`?)

Thanks for your help!

thomas-fossati: And a third thing: what is the expected encoding? Base64 (padded?), base64url (padded?), Base16, other?

thomas-fossati: See my shot in the dark [here](https://github.com/confidential-containers/trustee/compare/e6c14169f06e9fe76abf20c52da379c3d3b1524e..e3a950565bf64888de48446eb83508b33374077d), which produces:

```json
“[...]"
"init_data": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==",
"init_data_claims": null,
"report_data": "LJz1gp5O3CVbqAWUDeJWiKVZJJYva4zqStzj6JowYnXOTHZfv7UL1fKiYekjx9j9AAAAAAAAAAAAAAAAAAAAAA==",
```

thomas-fossati:
@fitzthum @Xynnn007 @huoqifeng thanks very much for your comments and guidance.

I should have addressed all your comments, but please have a look and check if I have missed anything.

(Note that this PR now also includes a29a052, which is a fix on top of #742 to allow full roundtripping between EAR trustworthiness vector claims and OPA.)

deps/verifier/src/cca/mod.rs
Xynnn007: Hey @thomas-fossati . About these Qs
1. How to populate init/report data & 2.Where to place
This [commit](https://github.com/confidential-containers/trustee/compare/e6c14169f06e9fe76abf20c52da379c3d3b1524e..e3a950565bf64888de48446eb83508b33374077d) looks good!

3. Initdata/report data field Encoding.
I think you can use hex which all other platforms s.t. SGX/TDX/SNP/SE/vTPM are using.

deps/verifier/src/cca/remote.rs
```diff
@@ -0,0 +1,104 @@
...
+) -> Result<Ear, anyhow::Error> {
+    let config::CcaVerifierConfig::Remote { address, ca_cert } = config.cca_verifier else {
+        panic!("expecting remote configuration")
```
Xynnn007, line 34: The same panic logic as in `local.rs`

deps/verifier/src/cca/remote.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    if let Some(ref ca_cert) = ca_cert {
+        cr_builder = cr_builder.with_root_certificate(ca_cert.clone())
+    }
+
+    let cr = cr_builder
+        .build()
+        .context("Failed to start challenge-response API with the service")?;
+
+    let n = Nonce::Value(expected_report_data.clone());
```
Xynnn007, line 68: looks like [the logic](https://github.com/confidential-containers/trustee/pull/738/commits/ccf6f60ea5926e5ddcd9f2d645a544992a537e90#diff-580b74950e63c48036ae07b63e1bd7fff3f6fb61d479d60f8447cbb8975043d4R87-R98) also compares the realm token's challenge field. Is there any duplication or anything I missed?

btw, we need to consider the byte length of `expected_report_data` to keep it the same as [challenge](https://docs.rs/ccatoken/0.1.0/ccatoken/token/struct.Realm.html#structfield.challenge). (fwiw the TSM api of linux kernel for CCA will get the challenge into a realm's token, right?)

deps/verifier/test_data/cca-claims.json
```diff
@@ -30,5 +30,7 @@
     "cca-realm-hash-algo-id": "sha-256",
     "cca-realm-public-key": "BIL70TKptcOWh5+7FTQNkFCXjlXHnVJ5oroOlYVPN+IM0vZPO3K1cLvXc+7iznaEJe31Re2+if+v4OlrvUbicPIHlsRIuY2vRqdk0nRC5ubthPjOyBfm7ManHTo959Z+zQ==",
     "cca-realm-public-key-hash-algo-id": "sha-512"
-  }
+  },
+  "init_data": "QURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBREFEQURBRA==",
```
Xynnn007, line 34: let's make it with hex for `init_data` and `report_data` claims.

Xynnn007 (commented):
Thanks. Let's go some further!

deps/verifier/src/cca/local.rs
```diff
@@ -0,0 +1,143 @@
...
+
+pub fn verify(
+    config: Config,
+    token: &Vec<u8>,
+    expected_report_data: &Vec<u8>,
+) -> Result<Ear, anyhow::Error> {
+    debug!("using config: {:?}", config);
+
+    let config::CcaVerifierConfig::Local { ta_store, rv_store } = config.cca_verifier else {
+        panic!("expecting local configuration")
```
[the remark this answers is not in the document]
thomas-fossati: Yes, you are correct this is checking an invariant.  I am relatively new to Rust and used a Go-ism :-)  What I likely want to use here is `assert!()`, i.e., something that fires in a debug build but is disabled in release mode.  Would that work instead?

deps/verifier/src/cca/local.rs
```diff
@@ -0,0 +1,143 @@
...
+    };
+
+    let jta = fs::read_to_string(ta_store).context("loading TA store")?;
+    let jrv = fs::read_to_string(rv_store).context("loading RV store")?;
+
+    let mut tas: MemoTrustAnchorStore = Default::default();
+    tas.load_json(&jta)
```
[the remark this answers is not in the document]
thomas-fossati: If the user hasn’t provided a TA/RV store, they see:

> Verifier evaluate failed: loading TA store
> Caused by:
> No such file or directory (os error 2)"

If they have configured it incorrectly (e.g., wrong syntax — including the empty file condition), they see:

> Verifier evaluate failed: loading trust anchors from JSON store
> Caused by:
> Syntax error: EOF while parsing a value at line 1 column 0

If the TA/RV array is empty (`[]`), the failure will manifest at appraisal time:

> Verifier evaluate failed: RAK signature or RAK attestation could not be verified

Adding the store file name and some more context may provide a slightly smoother UX.

deps/verifier/src/cca/local.rs
```diff
@@ -0,0 +1,143 @@
...
+    let (platform_tvec, realm_tvec) = e.get_trust_vectors();
+
+    // Check that the Realm token was correctly signed using the RAK and that
+    // the RAK was correctly attested.
+    if realm_tvec.instance_identity.tier() != TrustTier::Affirming {
+        bail!("RAK signature or RAK attestation could not be verified");
+    }
+
+    // Check that the challenge in the Realm token matches the expected_report_data
+    if *expected_report_data != e.realm_claims.challenge {
```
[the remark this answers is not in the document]
thomas-fossati: Yes, that is done in mod.rs:100, before dispatching local/remote mode appraisal.

deps/verifier/src/cca/local.rs
```diff
@@ -0,0 +1,143 @@
...
+
+// TODO(tho) why not populating RealmClaims and serialize?
+fn realm_annotated_evidence(e: &Evidence) -> Result<BTreeMap<String, RawValue>, anyhow::Error> {
+    let pv = general_purpose::STANDARD.encode(e.realm_claims.perso);
+    let rim = general_purpose::STANDARD.encode(e.realm_claims.rim.clone());
+    let rem0 = general_purpose::STANDARD.encode(e.realm_claims.rem[0].clone());
+    let rem1 = general_purpose::STANDARD.encode(e.realm_claims.rem[1].clone());
+    let rem2 = general_purpose::STANDARD.encode(e.realm_claims.rem[2].clone());
+    let rem3 = general_purpose::STANDARD.encode(e.realm_claims.rem[3].clone());
+    let nonce = general_purpose::STANDARD.encode(e.realm_claims.challenge);
```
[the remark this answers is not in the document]
thomas-fossati: Thanks for the pointer.  I’ve added a note inline.

deps/verifier/src/cca/local.rs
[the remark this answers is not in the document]
Xynnn007: Oh I did not see that.

deps/verifier/src/cca/local.rs
```diff
@@ -0,0 +1,143 @@
...
+use log::debug;
+use serde_json::json;
+use std::collections::BTreeMap;
+use std::fs;
+
+pub fn verify(
+    config: Config,
+    token: &Vec<u8>,
+    expected_report_data: &Vec<u8>,
+) -> Result<Ear, anyhow::Error> {
```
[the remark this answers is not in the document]
thomas-fossati: Cute! Thanks for the hint.

deps/verifier/src/cca/local.rs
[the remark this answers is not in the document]
thomas-fossati: I understand your point. However, I’d like to avoid requiring the caller to know all the details about the callee configuration.  The tension here lies in maintaining a coherent interface in the face of evolution.  The risk is that if, for example, the remote verifier later requires additional authentication parameters, fallback Veraison endpoints, or anything else, these would need to be added as function parameters, which doesn’t scale well.

deps/verifier/src/cca/remote.rs
thomas-fossati: > looks like [the logic](https://github.com/confidential-containers/trustee/pull/738/commits/ccf6f60ea5926e5ddcd9f2d645a544992a537e90#diff-580b74950e63c48036ae07b63e1bd7fff3f6fb61d479d60f8447cbb8975043d4R87-R98) also compares the realm token's challenge field. Is there any duplication or anything I missed?

That check makes sure that the EAR we are getting back from Veraison is the one associated with the current appraisal session.  We reuse the same nonce across the whole chain, which may be the source of your confusion 😅 

> btw, we need to consider the byte length of `expected_report_data` to keep it the same as [challenge](https://docs.rs/ccatoken/0.1.0/ccatoken/token/struct.Realm.html#structfield.challenge).

Isn’t the call to `regularize_data` in `mod.rs:101` sufficient?

> (fwiw the TSM api of linux kernel for CCA will get the challenge into a realm's token, right?)

That’s correct. The relevant ABI aspects are documented [here](https://gitlab.arm.com/linux-arm/linux-cca/-/blame/cca-guest/v7/drivers/virt/coco/arm-cca-guest/arm-cca-guest.c?ref_type=heads#L79-L80).

deps/verifier/src/cca/local.rs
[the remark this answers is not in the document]
thomas-fossati: I opted for using `bail!` which is slightly softer than I’d like to, but at least it removes the excessive brutality of panicking.

thomas-fossati:
@Xynnn007 thanks again for your thoughtful review and the nice suggestions!

I should have addressed all your second round of comments - either in code or by providing a rationale for the choices being made.

huoqifeng (approved):
LGTM
thanks! @thomas-fossati

fitzthum:
I missed some of the review here due to vacation, but seems like we can merge this once it is rebased @thomas-fossati

thomas-fossati:
> I missed some of the review here due to vacation, but seems like we can merge this once it is rebased @thomas-fossati

Excellent, thanks everoyne for the help and great suggestions.

(I have done the rebase.)

fitzthum:
Hmm that check failure is my fault. Can be fixed with 

```
diff --git a/integration-tests/src/common.rs b/integration-tests/src/common.rs
index 8d712b8..bca40fd 100644
--- a/integration-tests/src/common.rs
+++ b/integration-tests/src/common.rs
@@ -23,7 +23,7 @@ use attestation_service::{
 
 use reference_value_provider_service::client as rvps_client;
 use reference_value_provider_service::config::Config as RVPSConfig;
-use reference_value_provider_service::rvps_api::reference_value_provider_service_server::ReferenceValueProviderServiceServer;
+use reference_value_provider_service::rvps_api::reference::reference_value_provider_service_server::ReferenceValueProviderServiceServer;
 use reference_value_provider_service::storage::{local_json, ReferenceValueStorageConfig};
 use reference_value_provider_service::{server::RvpsServer, Rvps};
```

I can make a PR to fix this or you can stick this into yours. What do you prefer?

PR #742, opened 2025-03-14
fix(as): make EAR TV names into acceptable OPA variable names

thomas-fossati:
EAR TV names are kebab-case, which is not acceptable for OPA variables.
This results in TV names with `-` to be ignored.

This fix transforms kebabs into snakes to make OPA happy.


fitzthum (approved):
LGTM

PR #743, opened 2025-03-14
fix(cli): conflict in argument short flags options for restful-as

rdjondo:
Issue:
Running the executable target `restful-as` panics with the following error:
```
$  cd ./attestation-service
$  cargo build --bin restful-as --features restful-bin
$  ../debug/release/restful-as
Short option names must be unique for each argument, but '-h' is in use by both 'https_pubkey_cert' and 'https_prikey'
```

This fix ensures that the executable target `restful-as` doesn't panic at runtime because the command has
by default conflicting short arguments.

The proposed short options take inspiration from the convention for the TPM2 commands: '-u' for the public key and '-r' for the private key (second letter in public and private). See  https://tpm2-tools.readthedocs.io/en/latest/man/tpm2_create.1/


fitzthum:
Can you add [sign-off](https://cert-manager.io/docs/contributing/sign-off/) to your commit message

Issue #637, opened 2024-12-18
RVPS Improvements Brainstorm


fitzthum:
We've discussed changing the RVPS in https://github.com/confidential-containers/trustee/issues/238 and https://github.com/confidential-containers/trustee/issues/350 and https://github.com/confidential-containers/trustee/issues/407 (cc @thomas-fossati, @Xynnn007, and @deeglaze), but we've yet to really figure out what we should do. I have a few concrete ideas.

First a little overview of what we have. There are three important parts of the RVPS. 

1) The interface to the attestation service where we use reference values to evaluate the AS policy. This API is pretty simple. It basically looks like this

    ```
    pub async fn get_digests(&self) -> Result<HashMap<String, Vec<String>>>
    ``` 
    The RVPS takes no input and returns a map of ids to vectors of reference values (usually hashes). We're assuming that for each id there could be multiple valid values.

2) The internal representation of reference values.

    ```
    pub struct ReferenceValue {
        #[serde(default = "default_version")]
        pub version: String,
        pub name: String,
        #[serde(deserialize_with = "primitive_date_time_from_str")]
        pub expiration: DateTime<Utc>,
        #[serde(rename = "hash-value")]
        pub hash_value: Vec<HashValuePair>,
    } 
    ```
    Note that the internal representation is assuming that the reference values are hashes. This might not be correct. A reference value could be a version number or the policy could even calculate hashes internally. The `HashValuePair` struct has info about the hashing algorithm that we never pass to the client. Maybe we don't need it?

3) Extractors produce the reference value struct above after verifying some input. Currently we only have a sample extractor and an in-toto extractor which uses go bindings. We also have a pre-processor which currently doesn't seem to do much and has a confusingly-named `Ware` interface.

**Improvements**
* **Add evidence groups/context** I'm planning to add an abstraction that allows us to group reference values. In some ways this will be simple. We can change the `get_digests` API to take a group name. The AS policy won't be aware of the groups, it will continue to operate the same way using the reference value map, but the AS will be able to switch the contents of the map based on some context. The tricky question is how the AS should decide what group of values to use. We can implement this as a second step based on a local config, an init-data field, or something else.
[...]

PR #747, opened 2025-03-19
rvps: Input INIT_DATA when fetching reference values

tylerfanelli:
For VMs with different reference value sets, there requires a method to distinguish them from one another. The main distinguishing "identifier" for VMs is the INIT_DATA.

Prepare the RVPS for using the INIT_DATA to distinguish different sets of reference values for VMs.

The RVPS currently does not use this data.


Xynnn007:
Requesting different reference values ​​based on different ids is a good design direction, but I think we should firstly define the following well which I do not see in the PR.

1. How to define the id corresponding to the RVPS reference value in the AS policy.
2. How to specify a specific AS policy in KBS remote attestation.

In addition, there was a similar design in history, which was later temporarily [deleted](https://github.com/confidential-containers/trustee/commit/bd7b1c8e44a26ddcee1b22d41e6cf7ef8d998ed7) for a better design in the future.

tylerfanelli:
> Requesting different reference values ​​based on different ids is a good design direction, but I think we should firstly define the following well which I do not see in the PR.
> 
> 1. How to define the id corresponding to the RVPS reference value in the AS policy.

Can you elaborate? The id is defined by the INIT_DATA, and it corresponds to a set of reference values. Maybe I'm misunderstanding the question though.

> 2. How reference values are organized and indexed by the RVPS inside.

Good question, and one worth some further discussion. Some type of optional mapping of reference values to INIT_DATAs perhaps?

> 3. How to specify a specific AS policy in KBS remote attestation.
> 
> In addition, there was a similar design in history, which was later temporarily [deleted](https://github.com/confidential-containers/trustee/commit/bd7b1c8e44a26ddcee1b22d41e6cf7ef8d998ed7) for a better design in the future.

Curious to hear your thoughts.

fitzthum:
I think we probably don't want to provide the entire init-data to the RVPS. I'd like to keep the RVPS as simple as possible and have the AS in charge of figuring out which reference values it wants. One thing I proposed in #637 was to introduce a reference value URI (similar to the resource URI). The AS could resolve one of the pats of this URI from the init-data. Unfortunately that proposal got a bit bogged down by issues with the OPA engine.

tylerfanelli:
> One thing I proposed in #637 was to introduce a reference value URI (similar to the resource URI). The AS could resolve one of the pats of this URI from the init-data.

At first glance, I quite like this idea.

fitzthum:
> are you currently working on that by chance? If not, I can try a solution.

It's on the back burner for me. Feel free to take a crack at it. This is related to another change we've been talking about, which is supporting both a push and a pull model for reference values. It would be really good if the RVPS didn't have to be pre-loaded with RVs, but could look them up as needed. Today we can't really support that because the policy itself doesn't express which values it needs. I had wanted to use rego extensions for this, but ran into some problems calling async code within them. See https://github.com/microsoft/regorus/issues/363. Anand suggested a workaround here (introduce a new policy claim that lists the reference values and evaluate the first) that should work but I got a bit sad about my plan not working and switched to other stuff.

PR #752, opened 2025-03-21
Support Verifying Guests with Multiple Devices

fitzthum:
See also https://github.com/confidential-containers/guest-components/pull/957
and https://github.com/virtee/kbs-types/pull/58

A few things to know about this piece of the puzzle: 

1) We have a different attestation policy for each device class. The evidence from each attester gets passed through the appropriate policy and then added to the attestation token. We use the device class again to name each submod in the attestation token.

2) I have done my best to account for other configurations, such as the simple tokens and ITA, that don't use EAR. These models don't really support multi-device attestation in this context yet, but they should continue to function.

3) Generally, for users of Trustee who don't care about multiple devices there are only a few small changes. The submod name for a single cpu has changed from `cpu` to `cpu0` and if you add a new attestation service policy for the cpu, it should end with `_cpu`. Otherwise, everything will be the same.


attestation-service/src/lib.rs
```diff
@@ -183,11 +190,20 @@ impl AttestationService {
...
-            .await
-            .map_err(|e| anyhow!("Verifier evaluate failed: {e:?}"))?;
-        info!("{:?} Verifier/endorsement check passed.", tee);
+        let mut claims_from_tee_evidence: Vec<(Tee, String, TeeEvidenceParsedClaim)> = Vec::new();
+        for tee in &tees {
+            let ev = evidence
+                .get(tee)
+                .ok_or(anyhow!("No evidence found for {:?}", tee))?;
+            let verifier = verifier::to_verifier(tee)?;
+            let claims = verifier
```
tylerfanelli, line 199: Is this usable at the moment? For instance, there are no GPU verifiers, correct?

kbs/src/attestation/backend.rs
```diff
@@ -187,9 +192,12 @@ impl AttestationService {
             );
         }
 
+        // For now just use the first TEE to get the challenge.
+        // Currently the only platform that requires its own challenge
+        // is s390x, which does not yet require attestation of devices.
         let challenge = self
             .inner
-            .generate_challenge(request.tee, request.extra_params.clone())
+            .generate_challenge(request.tees[0], request.extra_params.clone())
```
tylerfanelli, line 200: What if the first TEE is a non-CPU one? With no report data?

attestation-service/src/bin/restful/mod.rs
```diff
@@ -36,7 +36,7 @@ type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct AttestationRequest {
-    tee: String,
+    tees: Vec<String>,
```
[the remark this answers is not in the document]
fitzthum: Not a bad idea. I was just following what we currently have, but I could switch this. It would probably avoid some of the annoying juggling we are doing from string to TEE.

attestation-service/src/lib.rs
```diff
@@ -158,18 +158,25 @@ impl AttestationService {
...
-        tee: Tee,
+        tees: Vec<Tee>,
         runtime_data: Option<Data>,
         runtime_data_hash_algorithm: HashAlgorithm,
         init_data: Option<Data>,
         init_data_hash_algorithm: HashAlgorithm,
         policy_ids: Vec<String>,
     ) -> Result<String> {
-        let verifier = verifier::to_verifier(&tee)?;
+        // Tee => (DeviceClass, Evidence)
```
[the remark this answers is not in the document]
fitzthum: Yes. The device class is a little bit generic because I'm not sure exactly what taxonomy people will want to use, but the basic idea is that each device class will have its own attestation policy. There will be one store of reference values which can be referenced from all the policies. The EAR token we get at the end will have a TrustVector for each device, which will be identified generically as classN.

attestation-service/src/lib.rs
fitzthum: This PR introduces the sample_device verifier, but otherwise we don't have any additional functionality yet.

kbs/src/attestation/backend.rs
fitzthum: Yeah this should change to check if any of the TEEs are s390x.

attestation-service/src/lib.rs
```diff
@@ -158,18 +158,25 @@ impl AttestationService {
...
+            &String::from_utf8(evidence).context("Failed to parse evidence as string")?,
+        )
+        .context("Failed to deserialize evidence")?;
 
         let (report_data, runtime_data_claims) =
             parse_data(runtime_data, &runtime_data_hash_algorithm).context("parse runtime data")?;
 
+        // report data and init data should work without too many changes
+        // there is one set per guest
+        // but most non-cpu verifiers won't return a claim it
```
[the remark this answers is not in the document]
fitzthum: Oh I think this is left over from development. I can clean up.

attestation-service/src/bin/restful/mod.rs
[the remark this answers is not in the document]
Xynnn007: If we use enum we could run into some delima.

`String::new("tdx")` can be serialized to `"tdx"`
`Tee::Tdx` will be serialized to "\"tdx\""`, with extra `\"`

attestation-service/src/token/ear_broker.rs
```diff
@@ -234,42 +226,61 @@ impl AttestationTokenBroker for EarAttestationTokenBroker {
...
+            let tcb_claims_json = serde_json::to_string(&tcb_claims)?;
+
+            let rules = TrustVector::new()
+                .into_iter()
+                .map(|c| c.tag().to_string())
+                .collect();
+
+            // There is a policy for each tee class.
+            // The cpu tee class is loaded as the default.
+            let policy_id = format!("{}_{}", policy_ids[0], tee_class);
```
Xynnn007, line 252: let's check if `policy_ids` is not empty before indexing it

attestation-service/src/token/ear_broker.rs
```diff
@@ -234,42 +226,61 @@ impl AttestationTokenBroker for EarAttestationTokenBroker {
...
+                .evaluate(&reference_data, &tcb_claims_json, &policy_id, rules)
+                .await?;
+
+            for (k, v) in &policy_results.rules_result {
+                let claim_value = v.as_i8().context("Policy claim value not i8")?;
+
+                appraisal
+                    .trust_vector
+                    .mut_by_name(k)
+                    .unwrap()
```
Xynnn007, line 264: Noticed that there is an `unwrap()` in old version. Let's fix it now

attestation-service/src/token/ear_broker.rs
```diff
@@ -234,42 +226,61 @@ impl AttestationTokenBroker for EarAttestationTokenBroker {
...
+            if !appraisal.trust_vector.any_set() {
+                bail!("At least one policy claim must be set.");
+            }
+
+            appraisal.update_status_from_trust_vector();
+            appraisal.annotated_evidence = tcb_claims;
+            appraisal.policy_id = Some(policy_ids[0].clone());
+
+            tee_class_indices.entry(tee_class.clone()).or_insert(0);
+            let tee_class_index = *tee_class_indices.get_mut(&tee_class).unwrap();
```
Xynnn007, line 277: Let's avoid `unwrap()`s

kbs/src/attestation/backend.rs
Xynnn007, line 200: let's go back to the disscussion. Here it can be simplified by calling `platform/base/cpu` field's api only to cover `s390x` case

kbs/src/attestation/coco/grpc.rs
```diff
@@ -92,21 +92,26 @@ impl Attest for GrpcClientPool {
...
-        let tee = serde_json::to_string(&tee)
-            .context("CoCo AS client: serialize tee type failed.")?
-            .trim_end_matches('"')
-            .trim_start_matches('"')
-            .to_string();
+        let tees = tees
+            .iter()
+            .map(|t| {
+                serde_json::to_string(&t)
+                    .unwrap()
```
Xynnn007, line 107: let's get rid of `unwrap()`s

attestation-service/docs/policy.md
```diff
@@ -35,7 +35,12 @@ values provided by the RVPS.
 
 `input` are the TCB claims generated by the verifier.
 
-See the [default policy](../src/token/ear_default_policy.rego) for an example.
+See the [default policy](../src/token/ear_default_policy_cpu.rego) for an example.
+
+When using EAR tokens, the policy used to evaluate CPU evidence should end in `_cpu`.
```
Xynnn007, line 40: A naive idea is to have only one default policy. It could have different parts to filter all supported devices. This would help to avoid the complex policy set-ups for users.

Xynnn007 (commented):
Some basic coding comments. Back to the design, I found it could make current codebase simpler especially cpu device indexing with hierarchical design. wdyt?

Issue #670, opened 2025-01-21
GCP TDX attests failed in RHEL VM


yuxisun1217:
### Describe the bug

Attest GCP TDX VM failed in RHEL-9.6:
```
# kbs-client --url http://trusteeserver:8080 attest --tee-key-file test/tee_key.pem  > test/attestation_token
[2025-01-21T08:29:18Z WARN  attester::tdx] Read AA Eventlog failed: Os { code: 2, kind: NotFound, message: "No such file or directory" }
[2025-01-21T08:29:18Z WARN  kbs_protocol::client::rcar_client] RCAR handshake failed: RcarHandshake("KBS Server Internal Failed, Response: \"payload reached size limit\""), retry 1...
[2025-01-21T08:30:19Z WARN  kbs_protocol::client::rcar_client] RCAR handshake failed: RcarHandshake("error sending request for url (http://trusteeserver:8080/kbs/v0/auth)"), retry 2...
[2025-01-21T08:30:20Z WARN  attester::tdx] Read AA Eventlog failed: Os { code: 2, kind: NotFound, message: "No such file or directory" }
[2025-01-21T08:30:20Z WARN  kbs_protocol::client::rcar_client] RCAR handshake failed: RcarHandshake("KBS Server Internal Failed, Response: \"payload reached size limit\""), retry 3...
[2025-01-21T08:31:21Z WARN  kbs_protocol::client::rcar_client] RCAR handshake failed: RcarHandshake("error sending request for url (http://trusteeserver:8080/kbs/v0/auth)"), retry 4...
[2025-01-21T08:31:22Z WARN  attester::tdx] Read AA Eventlog failed: Os { code: 2, kind: NotFound, message: "No such file or directory" }
Error: RCAR handshake failed: Unable to get token. RCAR handshake retried 5 times. Final attempt failed with: RcarHandshake("KBS Server Internal Failed, Response: \"payload reached size limit\"")
```

Not see such issue in GCP SNP VM.

### How to reproduce

Create a RHEL-9.6 TDX VM in GCP:
```
gcloud compute instances create rhel-96-tdx-client --confidential-compute-type=TDX --machine-type=c3-standard-4 --maintenance-policy=TERMINATE --zone=us-central1-c --image=rhel-guest-image-9-6-20250116-5-x86-64 --image-project=virt-qe
```
And compile install all-attesters from source code. Then try to attest it:
```
[...]

PR #755, opened 2025-03-28
kbs: added request payload size config option

pawelpros:
Parametrized request payload size which now can be configured by user by providing value in mb.

Fixes: #670

### DIFF
--- a/integration-tests/src/common.rs
+++ b/integration-tests/src/common.rs
@@ -173,6 +173,7 @@ impl TestHarness {
                 private_key: None,
                 certificate: None,
                 insecure_http: true,
+                payload_request_size: 2,
             },
             admin: AdminConfig {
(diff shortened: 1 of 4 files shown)


mythi (approved):
LGTM. Perhaps this could be a config option too?

fitzthum (approved):
LGTM. 

Is there any downside to have the size limit be larger? Maybe we should set the default even higher?

Xynnn007 (approved):

pawelpros:
In my opinion there is no downside -> as we had 256kb as default for long time 2 MB should be enough for now to force users also to try compress or clean data without altering default config but as it's parametrized element -> it shouldn't be a problem to adjust it to specific needs

pawelpros:
This PR was tested on GCP with TDX ITA configuration

PR #759, opened 2025-04-04
Clean up some text in the README

jonner:
Just some minor improvements to readability while looking through the README.


Xynnn007 (approved):
Thanks for the fix!

fitzthum (approved):
Thanks

PR #763, opened 2025-04-08
kbs-client: add built-in policies

fitzthum:
Let's ship a couple of basic policies (allow-all, deny-all, and the default) in the KBS Client itself. This way users can easily set these policies without having to create a file.

In the future we could consider moving this functionality into the KBS itself, but for now let's avoid changing the API.

### DIFF
--- a/tools/kbs-client/src/main.rs
+++ b/tools/kbs-client/src/main.rs
@@ -111,8 +111,21 @@ enum ConfigCommands {
     /// Set resource policy
     SetResourcePolicy {
         /// Policy file path
-        #[clap(long, value_parser)]
-        policy_file: PathBuf,
+        #[clap(long, value_parser, group = "resource_policy")]
+        policy_file: Option<PathBuf>,
+
+        /// Use built-in policy that allows access to all resources
+        #[clap(long, action, group = "resource_policy")]
+        allow_all: bool,
+
+        /// Use built-in policy that does not allow access to any resources
+        #[clap(long, action, group = "resource_policy")]
+        deny_all: bool,
+
+        /// Use built-in default policy that allows access to all policies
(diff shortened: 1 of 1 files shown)


tools/kbs-client/src/main.rs
```diff
@@ -111,8 +111,18 @@ enum ConfigCommands {
...
+
+        /// Built-in policies
+        #[clap(long, action, group = "resource_policy")]
+        allow_all: bool,
+
+        #[clap(long, action, group = "resource_policy")]
+        deny_all: bool,
+
+        #[clap(long, action, group = "resource_policy")]
+        default: bool,
```
Xynnn007, line 125: Let's flourish the cmdline
```suggestion
        /// Use built-in policy: allow all accesses to KBS resource
        #[clap(long, action, group = "resource_policy")]
        allow_all: bool,
       
        /// Use built-in policy: deny all accesses to KBS resource
        #[clap(long, action, group = "resource_policy")]
        deny_all: bool,

        /// Use built-in policy: default policy (deny accesses from sample TEE)
        #[clap(long, action, group = "resource_policy")]
        default: bool,
```

Then the cmdline can be 
```
Set resource policy

Usage: kbs-client config --auth-private-key <AUTH_PRIVATE_KEY> set-resource-policy [OPTIONS]

Options:
      --policy-file <POLICY_FILE>  Policy file path
      --allow-all                  Use built-in policy: allow all accesses to KBS resource
      --deny-all                   Use built-in policy: deny all accesses to KBS resource
      --default                    Use built-in policy: default policy (deny accesses from sample TEE)
  -h, --help                       Print help
```

Xynnn007 (approved):
Thanks for this. Some personal preference

PR #770, opened 2025-04-17
Update eventlog.rs

gcoon151:
Ignore invalid events that can cause problems (as discussed on slack)

### DIFF
--- a/deps/verifier/src/tdx/eventlog.rs
+++ b/deps/verifier/src/tdx/eventlog.rs
@@ -86,7 +86,7 @@ impl CcEventLog {
     pub fn query_digest(&self, entity: MeasuredEntity) -> Option<String> {
         for event_entry in self.cc_events.log.clone() {
             match (entity, event_entry.event_type.as_str()) {
-                (MeasuredEntity::TdvfKernel, "EV_EFI_BOOT_SERVICES_APPLICATION") => {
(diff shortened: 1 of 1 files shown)


fitzthum:
Can you add a bit more info in the commit message and do you have a link to this slack message so I can check it out?

Let's let the linter run, but looks like you may have added some extra spaces.

gcoon151:
> Can you add a bit more info in the commit message and do you have a link to this slack thread so I can check it out?
> 
> Let's let the linter run, but looks like you may have added some extra spaces.

I think I fixed both the commit message and spacing.

Xynnn007:
Thanks for the fix. Could you share some test cases or the ccel binary file you met the error?

cc @mythi

gcoon151:
> Thanks for the fix. Could you share some test cases or the ccel binary file you met the error?
> 
> cc @mythi

Frank put a lot of the details including the event entry that caused it to panic in the CNCF #confidential-containers-peerpod slack channel here https://cloud-native.slack.com/archives/C04A2EJ70BX/p1743708184751959?thread_ts=1743444129.613059&cid=C04A2EJ70BX (the relevant part starts "There are several "EV_EFI..."

gcoon151:
Relevant comments from Mikko that I think answer what you're looking for, "Might be a bug in my code that parses CCEL with OVMF" ... "az-vtpm-tdx does not include the CC eventlog" 
... therefore eventlog.rs panics.

fitzthum (approved):
Looks fine. Note that the next line is trying to index into the buffer, so this is really just a bounds check. Possibly would be better to move onto its own line and log an error if we have this type of event but it is short. It sounds like there is some root cause still to be understood, but this seems reasonable. I would like to hear from @mythi tho

mythi:
I was working on a better fix for this. Let me get that PR asap

PR #771, opened 2025-04-17
tdx: change default qcnl configuration

fitzthum:
The current QCNL configuration has the collateral service commented out. This seems to cause problems for anyone who tries to use it.

Maybe there is some reason for this being set that i am not aware of, but this has been a big stumbling block for users.

@mythi

### DIFF
--- a/kbs/config/sgx_default_qcnl.conf
+++ b/kbs/config/sgx_default_qcnl.conf
@@ -2,7 +2,7 @@
   // *** ATTENTION : This file is in JSON format so the keys are case sensitive. Don't change them.
   
   //PCCS server address
-  "pccs_url": "https://localhost:8081/sgx/certification/v4/"
+  //"pccs_url": "https://localhost:8081/sgx/certification/v4/"
 
   // To accept insecure HTTPS certificate, set this option to false
   ,"use_secure_cert": true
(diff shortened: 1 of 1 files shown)


fitzthum:
Do we need to comment out that pccs config?

Xynnn007:
The minimal run-able one is https://github.com/confidential-containers/trustee/blob/8565e7fe11f6be8d5f744580e8cf1f1b8cadbde9/attestation-service/docs/sgx_default_qcnl.conf, and we do not need to set pccs config

fitzthum:
Ok, I commented out the PCCS config as well. I think that might expect the user to be running something locally, which we probably don't want for our simple deployments.

mythi (approved):
This is OK for TDX/SGX quote verification in the docker-compose setup. I hope users would still read the docs to find what might be best for their env.

kbs/config/sgx_default_qcnl.conf
```diff
@@ -2,15 +2,15 @@
   // *** ATTENTION : This file is in JSON format so the keys are case sensitive. Don't change them.
   
   //PCCS server address
-  "pccs_url": "https://localhost:8081/sgx/certification/v4/"
+  //"pccs_url": "https://localhost:8081/sgx/certification/v4/"
```
bpradipt, line 5: Shouldn't the pccs_url be also set to `"https://api.trustedservices.intel.com/sgx/certification/v4/"` ?
Afaicr, sometime back attestation didn't work without setting pccs_url. I will retest this by tomorrow but wanted to mention it here for @mythi to comment.

mythi: This should be fine if used for quote verification only. The QGS setting needs a valid pccs url

PR #774, opened 2025-04-22
ci: enable kbs integration / e2e sample tests on arm64

seungukshin:
Enable kbs integration / e2e sample tests on arm64
* kbs integration test: https://github.com/confidential-containers/trustee/blob/main/.github/workflows/kbs-rust.yml
* kbs e2e sample test: https://github.com/confidential-containers/trustee/blob/main/.github/workflows/kbs-e2e-sample.yml
* kbs docker e2e test: https://github.com/confidential-containers/trustee/blob/main/.github/workflows/kbs-docker-e2e.yml

Test results on the fork:
* kbs integration test: https://github.com/seungukshin/trustee/actions/runs/14595864727
* kbs e2e sample test: https://github.com/seungukshin/trustee/actions/runs/14595864726
* kbs docker e2e test: https://github.com/seungukshin/trustee/actions/runs/14595864719

### DIFF
--- a/.github/workflows/kbs-e2e-sample.yml
+++ b/.github/workflows/kbs-e2e-sample.yml
@@ -17,9 +17,19 @@ jobs:
       with:
         path: ./kbs.tar.gz
 
-  e2e-test:
+  e2e-test-amd64:
     needs: checkout
     uses: ./.github/workflows/kbs-e2e.yml
     with:
       tee: sample
+      runs-on: '["ubuntu-22.04"]'
       tarball: kbs.tar.gz
+
+  e2e-test-arm64:
+    needs: checkout
+    uses: ./.github/workflows/kbs-e2e.yml
+    with:
+      tee: sample
+      runs-on: '["ubuntu-22.04-arm"]'
+      tarball: kbs.tar.gz
+      kbs-client-features: "sample_only,cca-attester"
--- a/.github/workflows/kbs-docker-e2e.yml
+++ b/.github/workflows/kbs-docker-e2e.yml
@@ -11,7 +11,21 @@ env:
 
 jobs:
   e2e-test:
-    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        instance:
+          - ubuntu-24.04
+          - ubuntu-24.04-arm
+        include:
+          - instance: ubuntu-24.04
+            build_platform: linux/amd64
+            target_arch: x86_64
+            verifier: all-verifier
+          - instance: ubuntu-24.04-arm
+            build_platform: linux/arm64
+            target_arch: aarch64
+            verifier: cca-verifier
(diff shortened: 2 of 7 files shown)


Xynnn007 (approved):
Thanks @seungukshin  LGTM. Nice to see that arm is included into e2e test.

fitzthum (approved):
Seems good. Let's just keep an eye on it once we merge and make sure everything is working.

Xynnn007:
This error seems like the runner cannot find the make command. I checked [a successful execution](https://github.com/confidential-containers/trustee/actions/runs/14619670415/job/41016200590) some hours ago and the failed execution this time, and found that there are some differences in the runner's setup job stage. The failed version does not clearly indicate the image version. I suspect it may be related to the github runner image update.

Let me retrigger a historical successful one to ensure. 

see https://github.com/confidential-containers/trustee/actions/runs/14586301191/job/40912361147

.github/workflows/kbs-e2e-sample.yml
```diff
@@ -17,9 +17,19 @@ jobs:
...
       tee: sample
+      runs-on: '["ubuntu-22.04"]'
       tarball: kbs.tar.gz
+
+  e2e-test-arm64:
+    needs: checkout
+    uses: ./.github/workflows/kbs-e2e.yml
+    with:
+      tee: sample
+      runs-on: '["ubuntu-22.04-arm"]'
```
Xynnn007, line 33: The original runs-on format is `runs-on: ubuntu-22.04`. Here we have extra `[..]`, which would mean `runs-on: ["ubuntu-22.04"]`

Is it expected?

Also, the field https://github.com/confidential-containers/trustee/pull/774/files#diff-5369a77cb0b4db3d0426e02510e87a6d21f90d0158ef0aa25fc6d69bafd2133aL11 seems no use

seungukshin: Thanks @Xynnn007 for taking a look into this.

> The original runs-on format is runs-on: ubuntu-22.04. Here we have extra [..], which would mean runs-on: ["ubuntu-22.04"]
Is it expected?

the original code's runs-on of `build-binaries` is `ubuntu-22.04`, but the runs-on of `e2e-test` is `["ubuntu-22.04"]` (because it uses default value). the original author seems to consider passing multiple runners with this input. in this case `build-binaries` should be executed all these runners to generate appropriate binaries. For current usage which is passing only one runner, it can be changed to single input.

> Also, the field https://github.com/confidential-containers/trustee/pull/774/files#diff-5369a77cb0b4db3d0426e02510e87a6d21f90d0158ef0aa25fc6d69bafd2133aL11 seems no use

right, `default` value is not used anymore.

seungukshin:
I just created this PR (https://github.com/confidential-containers/trustee/pull/781) to fix this issue.

Issue #543, opened 2024-10-21
Add metrics endpoint


fitzthum:
Currently the only way to figure out the status of Trustee is to look at the logs. We should add a more sophisticated metrics infrastructure that can be connected into third-party systems.

mkulke:
Note that while prometheus support is a good idea, it will not supplement logs. metrics are about aggregate values (think: number of successful tdx verifications per minute), you shouldn't/cannot debug individual transactions with it (because of restrictions on the cardinality of metrics).

the closest to a supplement logs today are opentelemetry traces. sadly, this is still a bit rough for rust webservices today. but embracing structured logs via [tracing ](https://crates.io/crates/tracing) is a step in the right direction. e.g. there should be correlation ids to debug logical transactions (like an RCAR ceremony).

pmores:
Hi, I'm looking into adding Prometheus support.  I have a bunch of questions:

- where do we want Prometheus exporters added?  KBS?  Attestation service?  Both?  Any others?  More generally, is there any preexisting idea of the general structure of Prometheus support?
- is it acceptable to use a crate, probably [this one](https://crates.io/crates/prometheus) (it looks fairly well-maintained and widely-used), in an implementation?
- if so, is it OK to use it directly in trustee code, or does it need to be hidden behind an interface to ease re-implementing Prometheus support using a different crate if need be?  (Looking at trustee code it probably is but asking anyway to make sure.)
- could you advice a bunch of example specific metrics for each exporter to be added (preferably one on each data types or at least a counter, gauge and histogram) so that I can work with real-world use cases?

Thanks in advance!

pmores:
Excellent, thanks Tobin!  This looks like enough for me to start an implementation.

As for a metrics abstraction, I'd actually prefer to have one too but I think I'll start by deploying the prometheus crate directly and see how it turns out.  Maybe a neat abstract interface suggests itself, or maybe I find out that I'd be just tediously replicating the underlying crate interface one-to-one.  I'm planning to post a WIP PR as early as it makes sense so we can then make a final decision on this as a community.

spotlesstofu:
Following up @pmores foundational work, I'd like to propose which metrics to implement.

Some of these are already in the code but they're missing the suggested labels.

Metric Name | Prometheus Type | Description | Suggested Labels | Rationale/Prometheus Best Practice
-- | -- | -- | -- | --
[...]

PR #775, opened 2025-04-22
kbs: add skeleton prometheus endpoint

pmores:
This PR proposes what I feel is the most straightforward approach to implementing a Prometheus endpoint in KBS.  Please do check out the commit messages, admittedly it's a bit of wall of text but they try to outline the design choices I came across.  The TLDR; would be I guess
- are we fine with a singleton
- are we fine with storing metric object in their users, or do we prefer to have a centralised storage for them.

A KBS built from this branch should respond to a query like
```
curl http://127.0.0.1:12345/metrics
```
with something like
```
# HELP resource_reads_total KBS resource read count
# TYPE resource_reads_total counter
resource_reads_total{repository="pvl",tag="smazat",type="test"} 2
# HELP resource_writes_total KBS resource writes count
# TYPE resource_writes_total counter
resource_writes_total{repository="pvl",tag="smazat",type="test"} 1
```
after a `set-resource` and a bunch of `get-resource` requests.

I feel it would be helpful to flesh out our metrics design a bit at this stage since one lesson learned from this initial work seems to be that what precise metrics we want to gather, and their partitioning using labels, can have a major impact on what the right implementation strategy will be.

This PR implements right away two counters for resource accesses, both read and write, partitioned by repository, type and tag labels implemented as variable labels.  Based on my understanding this should be fine but since Prometheus guidelines to labels usage seem a bit confusing to me (along the lines of "do use labels but not too much") I'd greatly appreciate insights from anybody experienced in Prometheus metrics design.


kbs/src/plugins/implementations/resource/backend.rs
```diff
@@ -114,12 +151,28 @@ impl ResourceStorage {
...
             .await
     }
 
     pub(crate) async fn get_secret_resource(&self, resource_desc: ResourceDesc) -> Result<Vec<u8>> {
+        self.prometheus_metrics
+            .resource_reads_total
+            .with_label_values(&[
+                &resource_desc.repository_name,
+                &resource_desc.resource_type,
+                &resource_desc.resource_tag,
```
mkulke, line 173: note: prometheus has a cardinality issue, that means that you want to limit the amount of string values that a label can take. ideally they shouldn't be dynamic but just from a closed static set of values. see [here](https://prometheus.io/docs/practices/naming/) the secret resource path is very user-defined and hence dynamic

grafik

A good example for a closed set of values would be a TEE platform.

kbs/src/api_server.rs
```diff
@@ -115,6 +123,11 @@ impl ApiServer {
                             .route(web::get().to(api))
                             .route(web::post().to(api)),
                     )
+                    .service(
+                        web::resource("/metrics")
+                            .route(web::get().to(prometheus_metrics))
+                            .route(web::post().to(|| HttpResponse::MethodNotAllowed())),
```
mkulke, line 129: I don't think this is required, is it?

kbs/src/api_server.rs
```diff
@@ -100,7 +101,14 @@ impl ApiServer {
...
         );
 
-        let http_config = self.config.http_server.clone();
+        let mut http_config = self.config.http_server.clone();
+
+        // TODO: make the address/port configurable?
+        http_config.sockets.push(SocketAddr::new(
+            IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)),
+            12345,
+        ));
```
mkulke, line 110: why do we listen on another port? a /metrics route just sits alongside the other routes (like /health), no?

mkulke (commented):
afaik prometheus data is gathered as a singleton, so that's ok imo (not sure what you mean by "storing metrics objects in their users"). however, using the path segments as label values is not idiomatic I think. prometheus is supposed to gather metrics, not produce rich k/v data for data analysis (that can get costly quickly, and you can use structured logging or traces for that).

also, this looks pretty manual to me. I think the web framework that trustee is using is not the latest and greatest of rust-land, so maybe there is no good crate that'll add prometheus support and sets up a bunch of standard metrics, so you just have to add your own business metrics on top.

Usually you want to have common data like a histogram for response times, etc. Those should also follow some conventions (like `http_request_duration_seconds`), otherwise you'll create a headache for folks that want to add the service to standardized grafana dashboards. We shouldn't figure this out ourselves, a good idea is to take a look at an existing webservice with some reputation.

kbs/src/api_server.rs
pmores: Quite possibly, yes.  Some code I saw used it and other routes handle `web::post()` too so I thought I'd include it for completeness but I'm happy to remove it.

pmores:
> afaik prometheus data is gathered as a singleton, so that's ok imo 

Yes, metrics are a singleton by nature, my concern is more about general implementation issues with singletons (they tend to complicate unit testing etc.).  Some projects dislike singletons, I don't know yet what the stance of this one is so I'm pointing the singleton usage out.

> also, this looks pretty manual to me. I think the web framework that trustee is using is not the latest and greatest of rust-land, so maybe there is no good crate that'll add prometheus support and sets up a bunch of standard metrics, so you just have to add your own business metrics on top.

By "manual", do you mean how the counters are created and registered?  I guess as we add more metrics a pattern might emerge that could allow us to move the repeating code into `PrometheusExporter` (proposed in the initial commit of this series).

I haven't found a crate you mention.  There is a crate called prometheus_exporter which does add a set of standard (I guess) metrics but it also creates its own HTTP server.

> Usually you want to have common data like a histogram for response times, etc. Those should also follow some conventions (like http_request_duration_seconds), otherwise you'll create a headache for folks that want to add the service to standardized grafana dashboards. We shouldn't figure this out ourselves, a good idea is to take a look at an existing webservice with some reputation.

Yes, you're right about standard HTTP server metrics, we can lift them from somewhere as you say and will do that at some point.  My present concern though is more about KBS-specific metrics - but maybe none are needed?

kbs/src/api_server.rs
fitzthum: Are we thinking that this endpoint is trusted? The KBS endpoint is generally exposed to the internet so it could make sense to use another one if we don't want to expose things.

mkulke: if such a service is exposed to the internet it would usually sit behind an L7 proxy/api gw, which doesn't expose such routes

mkulke:
> I haven't found a crate you mention. There is a crate called prometheus_exporter which does add a set of standard (I guess) metrics but it also creates its own HTTP server.

at first glance [there are a couple](https://crates.io/search?q=actix-web%20prometheus). This looks like a good candidate: https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-instrumentation-actix-web (OTEL has standardised prom as otel metrics I think)

kbs/src/plugins/implementations/resource/backend.rs
mkulke: in my anecdotal experience high cardinality can indeed be severe (i.e. very costly in terms of perf and $$$) and you might not be a good service-citizen on a given platform team if you use dynamic labels. my experience is a few years old, so maybe that has been relaxed since.

one rule of thumb would be to avoid turning something that is inherently dynamic (like a user-provided hierarchy) into a label value. It's a valid use case of course to grab those dynamic data points for analysis, but you would use them in traces and structured log events, which are ingested and stored differently.

pmores:
To be clear, this PR focuses on Prometheus support, telemetry is out of scope.  To my knowledge, Prometheus and Open Telemetry are two separate things.  I do remember some discussion about Prometheus being able to ingest otel data but that's it, and that also seems irrelevant to our case.  If you have evidence to the contrary though, please do post, I guess if true it could even make our job easier.

If the above is the case it renders all of your search results but the first one pretty much irrelevant.  actix-web-prom does facilitate adding standart HTTP metrics to an actix_web HTTP server, however it clearly does nothing to solve the actual difficulty (where to store metric object instances) - note how the example code creates a single metric and passes it to the handler via `wrap()` which is what examples usually do because they don't mind that it doesn't really scale.  Considering this, the crate doesn't really seem to bring in much, we still need to decide how to solve the metrics storage problem.  When that's done, adding standard HTTP metrics honestly seems a matter of several dozen mostly trivial lines of Rust code. IMHO this wouldn't justify pulling in another dependency.

mkulke:
otel these days also includes metrics, which afaik is just prometheus (possibly w/ an option to push rather than scrape). Regarding storage, usually there's no problem, counters/histograms that populate the metrics endpoint are buffered globally in memory and if the service restarts there is no drama and counting starts from scratch. the state is kept in the prom db, normalized as timeseries data.

kbs/src/plugins/implementations/resource/backend.rs
pmores: > in my anecdotal experience high cardinality can indeed be severe (i.e. very costly in terms of perf and $$$) and you might not be a good service-citizen on a given platform team if you use dynamic labels. my experience is a few years old, so maybe that has been relaxed since.

That's good information, I wasn't clear on that, thanks.

> one rule of thumb would be to avoid turning something that is inherently dynamic (like a user-provided hierarchy) into a label value. It's a valid use case of course to grab those dynamic data points for analysis, but you would use them in traces and structured log events, which are ingested and stored differently.

I understand the guideline about dynamic values, however if you consider the best-case scenario I described above, i.e. each of repository, type and tag labels are guaranteed for external reasons to have two or three values each, would that still be unacceptable just because the values are dynamic (not known ahead of time)?

To be clear, I'm not arguing in favour of any specific setup, I'm just trying to figure out what the options are.  In fact, if we decide not to use labels at all, it would likely simplify the problem, possibly quite a bit.

pmores:
I took a closer look at the actix-web-opentelemetry crate you linked to above.  When it comes to working with actual metric objects (creating them, updating their values), it seems just to expose the prometheus crate I'm currently using, so much of this PR would probably look pretty much the same.

The only difference seems to be that it has a `PrometheusMetricsHandler` object that's used to keep the `Registry` and also seems to handle the `/metrics` endpoint automatically.  So this would allow us to get rid of the `/metrics` handler in `ApiServer` (not a huge win though, it's literally just a couple of lines), and seemingly - also potentially more importantly - also of the `PrometheusExporter` singleton.  I'm not sure though if this is good or bad. 

If we let our `Registry` disappear in a `PrometheusMetricsHandler` instance I don't see a way how to access it later from an arbitrary place in KBS code.  If that's the case then it would mean that we need to register all metrics at a single place (probably in `ApiServer`, before the server is started), and also figure out how to make them available to objects like `ResourceStorage` which need to update their values.  To avoid these issues I'm afraid I don't see any way other than to make a singleton (to access both registry and metrics) but then we're kind of back to square one I guess.

So unless I'm overlooking something I'm not really sure if the crate would bring in enough benefit tbh...

mkulke:
> So unless I'm overlooking something I'm not really sure if the crate would bring in enough benefit tbh...

yes, if there's not much that a crate provides we don't have to use it, I was assuming it'll auto-instrument all the http endpoints with conventional metrics like perf histograms etc, but the rust web ecosystem is not very mature yet, so maybe you have to hand-roll those...

I think the idiomatic use of the prometheus crate is pretty trivial:

```rust
lazy_static! {
    static ref HIGH_FIVE_COUNTER: IntCounter =
        register_int_counter!("highfives", "Number of high fives received").unwrap();
}
```

then in the route handlers

```rust
HIGH_FIVE_COUNTER.inc();
```

in the `/metrics` handler:

```rust
let mut buffer = Vec::new();
let encoder = TextEncoder::new();
let metric_families = prometheus::gather();
encoder.encode(&metric_families, &mut buffer).unwrap();
let output = String::from_utf8(buffer.clone()).unwrap();
```

pmores:
That's basically what this PR does.  The only difference is that the metric objects are stored in the user (in `ResourceStorage`, grouped in a local, `ResourceStorage`-specific `PrometheusMetrics` helper struct for additional cleanliness).  If we prefer representing them as lazy statics instead we can switch to that though personally I'd consider that unfortunate.

Your final block of code is literally the implementation of `PrometheusExporter::export_metrics()` (see the initial commit).

pmores:
As for the conventional HTTP server metrics, I couldn't find any actual standard, there's maybe just a de-facto standard, or just a convention.  This means that dashboards have to be prepared for some variation in metric sets anyway.

Also, the conventional set doesn't seem to be huge - apparently there's a counter for total requests, a bunch of histograms for duration and request/response sizes, maybe a gauge for in-progress requests.  So I wouldn't worry too much really about having to implement those ourselves.

Lastly and more broadly, how much do we want the KBS to look as a HTTP server from the outside, or is HTTP more of an implementation detail of the protocol the KBS uses?  I'm asking since some of the conventional metrics might not even be very useful in KBS (like perhaps the request or the response sizes), given that the KBS is more specialised in what it serves compared to a general HTTP server?

pmores:
I've added some conventional HTTP metrics to the `ApiServer` to give an idea how it would go.  Streamed request sizes are not accounted for at this time since I don't expect streamed requests to be important for the KBS.  Let me know if I'm wrong, I believe streamed requests could be handled, too, but at the cost of a more complex implementation.

I've appended the metrics code to `api_server.rs` at least for now, we could decide to move it into a separate file if appropriate.

The updated metrics output is now along the lines of
```
# HELP http_request_duration_seconds Distribution of request handling duration
# TYPE http_request_duration_seconds histogram
http_request_duration_seconds_bucket{le="0.0005"} 4
http_request_duration_seconds_bucket{le="0.001"} 6
http_request_duration_seconds_bucket{le="0.005"} 9
http_request_duration_seconds_bucket{le="0.01"} 9
http_request_duration_seconds_bucket{le="0.05"} 9
http_request_duration_seconds_bucket{le="0.5"} 9
http_request_duration_seconds_bucket{le="1"} 9
http_request_duration_seconds_bucket{le="+Inf"} 9
http_request_duration_seconds_sum 0.009319642
http_request_duration_seconds_count 9
# HELP http_request_size_bytes Distribution of request body sizes
# TYPE http_request_size_bytes histogram
http_request_size_bytes_bucket{le="32"} 1
http_request_size_bytes_bucket{le="128"} 3
http_request_size_bytes_bucket{le="512"} 5
http_request_size_bytes_bucket{le="2048"} 5
http_request_size_bytes_bucket{le="8192"} 5
http_request_size_bytes_bucket{le="+Inf"} 5
http_request_size_bytes_sum 776
http_request_size_bytes_count 5
# HELP http_requests_total Total HTTP requests count
# TYPE http_requests_total counter
http_requests_total 10
# HELP http_response_size_bytes Distribution of response body sizes
# TYPE http_response_size_bytes histogram
http_response_size_bytes_bucket{le="32"} 1
http_response_size_bytes_bucket{le="128"} 5
http_response_size_bytes_bucket{le="512"} 7
http_response_size_bytes_bucket{le="2048"} 9
http_response_size_bytes_bucket{le="8192"} 9
http_response_size_bytes_bucket{le="+Inf"} 9
http_response_size_bytes_sum 4056
http_response_size_bytes_count 9
# HELP resource_reads_total KBS resource read count
# TYPE resource_reads_total counter
resource_reads_total{repository="pvl",tag="smazat",type="test"} 2
# HELP resource_writes_total KBS resource write count
# TYPE resource_writes_total counter
resource_writes_total{repository="pvl",tag="smazat",type="test"} 1
```
after a single set-resource and two get-resource commands.

fitzthum:
Looks like most of the CI complaints here come from not having committed the most recent version of the lockfile.

pmores:
I've tried to include `Cargo.lock` changes that are evidently related with my changes but apparently it's not enough.  I'm hesitant committing all of my local `Cargo.lock` changes since they include a number of those which don't seem related to whatever I'm doing, and above all since I had to rollback version from 4 to 3 for some reason to be able even to compile, and it doesn't feel right to commit that.

I still don't quite understand why a generated file is to be tracked in a version control system in the first place so I'm not sure at the moment how to solve this.  Unless this is a known problem with a known solution I'd prefer to deal with `Cargo.lock` only once it's clear that this is going to be merged in the first place.

pmores:
I'd done `rustup override` to 1.81 before I started working on this PR:
```
$ rustc --version
rustc 1.81.0 (eeb90cda1 2024-09-04)
```
(I think there was a reason for getting 1.81 but I can't recall what it was. ;-))

So I'm actually on an even newer compiler than 1.80 - could that be the problem?

fitzthum:
> So I'm actually on an even newer compiler than 1.80 - could that be the problem?

Rust has recently switched from resolver 3 to 4. I forget the exact version where this changes, but if you are using a newer version with the lockfile in the repo or if you rebuild the lockfile with the new version push it to the CI which has the old version, you can run into issues. A bit of a catch with the idea of MSRV.

Hopefully you can get it to work with 1.80.0 by massaging the lockfile a bit. We don't really want to bump the msrv.

mkulke (commented):
looks ok to me in general, some comments:

- let's fix the build errors, switch to the repo toolchain with rustup if necessary, you can do this per repo.
- since we hand-craft the metrics endpoint, should we include a test?
- there are a couple of unwraps, if reasonable we should convert those into Errors
- afaik it's uncommon to expose /metrics on a different port (unless metrics are provided out-of-process, e.g in a sidecar). if we are concerned about leakage of sensitive information because someone would put the service on the internet as-is (which is a bad idea for most http apis), then we should not provide sensitive information on the metrics endpoint.
- I'm not convinced that user provided values should end up in labels, someone with a token and a 1-liner can dos the prom server.
- Note that access to /metrics is unauthorized, if the key hierarchy is sensitive, we should not leak it

pmores:
@mkulke Thanks for the comments, to address them:
- I'll fix the `Cargo.lock` problems obviously
- I'll look into adding some tests
- unless I'm overlooking something all `unwrap()`s but one are on `Mutex::lock()` return value where - if I understand correctly - they are pretty much idiomatic (since `lock()` should only return an error if the lock is poisoned).  The remaining one is to retrieve app data of type `ApiServer` from an `HttpRequest`.  My understanding is that this can only fail if we don't pass an `ApiServer` instance to `app_data()` on HTTP server setup, which we do.  So I'm inclined to leave the `unwrap()`s in.
- as for the port, I'm fine with either way but do we have a consensus?  So far it looks like one reviewer is in favour of the same port and one in favour of a separate one.
- considering labeling, I understand that the currently proposed approach is not a good one and of course I will change it, however I'm still not quite clear to what.  From my research it seems that endpoint is a fairly customary label which would probably translate to a single resource path label (not broken down to its components).  However, unlike an ordinary HTTP server where endpoints come from a static closed set (I suppose) this is still dynamic in KBS.  If this is important, or if the resource paths are considered sensitive and should not be exposed via metrics, we can't use any labelling based on resource accesses.  How should I change the labelling part?  A single label with a complete resource path, or no label at all?

mkulke:
> @mkulke Thanks for the comments, to address them:
> 
> * I'll fix the `Cargo.lock` problems obviously
> * I'll look into adding some tests
> * unless I'm overlooking something all `unwrap()`s but one are on `Mutex::lock()` return value where - if I understand correctly - they are pretty much idiomatic (since `lock()` should only return an error if the lock is poisoned).  The remaining one is to retrieve app data of type `ApiServer` from an `HttpRequest`.  My understanding is that this can only fail if we don't pass an `ApiServer` instance to `app_data()` on HTTP server setup, which we do.  So I'm inclined to leave the `unwrap()`s in.
> * as for the port, I'm fine with either way but do we have a consensus?  So far it looks like one reviewer is in favour of the same port and one in favour of a separate one.
> * considering labeling, I understand that the currently proposed approach is not a good one and of course I will change it, however I'm still not quite clear to what.  From my research it seems that endpoint is a fairly customary label which would probably translate to a single resource path label (not broken down to its components).  However, unlike an ordinary HTTP server where endpoints come from a static closed set (I suppose) this is still dynamic in KBS.  If this is important, or if the resource paths are considered sensitive and should not be exposed via metrics, we can't use any labelling based on resource accesses.  How should I change the labelling part?  A single label with a complete resource path, or no label at all?

yeah, those points are all nitpicky to a degree, I would not veto merging that as is, but someone else should take a look. especially with regards to leakage of the key hierarchy being a potential problem. if it is, we should just use a single path=":repository/:type/:tag" label

fitzthum:
The resource URIs are generally not protected. They are exposed in the KBS protocol itself and they're usually based on unencrypted config information. Using a different path for each resource exposes the resource names more generally, not just to the metrics endpoint.

We might want to make this clear to users via the website.

pmores:
I've pushed a new version:
- complete resource path is now used as a single label on resource access counters
- my complete local `Cargo.lock` is included (fingers crossed).  After downgrading my compiler from 1.81 to 1.80 I also had to downgrade the prometheus crate (that was actually the reason why I initially went with 1.81, now I recall :-))
- `/metrics` endpoint is now served on the same port as other endpoints
- also, some adjustments were made to the conventional HTTP server metrics collection (see the final commit), specifically requests lacking a `content-length` header are no longer ignored but are included as zero-sized, and non-KBS protocol requests (notably `/metrics` itself) are excluded.

I'm not sure how to go about adding tests though.  Neither of the objects this PR modifies is currently tested as far as I can tell (`ApiServer` nor `ResourceStorage`) and unit testing seems kind of tough due to coupling and/or complicated signatures (like the `prometheus_metrics_middleware()` function).  Also, testing against expected counter values might turn out brittle depending on which level testing is performed (e.g. setting a resource value is a single HTTP request but getting it is four requests IIRC).  If anybody has any specific idea in mind as to what to test and how please do share as I'm a bit at a loss.

pmores:
Honestly I have no idea what this CI wants from me.  I gave it all my `Cargo.lock` I had ;-), I don't see what else I can do for it.

kbs/src/api_server.rs
spotlesstofu: Same. There is no point in exposing metrics to the public by default.

pmores:
The remaining CI error is not related to this PR IMHO.

To make tests pass, specifically the get_resource integration tests which create and tear down `ApiServer` instances repeatedly, I had to add destructors to metric-holding structs that unregister the metrics.  That's fair enough, however I also observed that the structs are constructed once and then destructed many times on actix web server worker threads.  I'm aware that the application factory is passed to each worker thread and instantiated once per worker, however I still don't understand how it can be that there's a single constructor call and many destructor calls for an individual instance.

The solution was to make metric registration idempotent and unregistration kind of too (unregistering a metric that's not registered in the moment can't be specifically detected so a warning is printed but the program can continue).  This is fair enough I believe, however I'd really like to understand why `ApiServerMetrics`'s constructor and destructor calls don't match up.

kbs/src/api_server.rs
pmores: It's currently served on the same port as the KBS protocol as per this review request.  It's easy enough to change in the future anyway, should we make that call.

kbs/Cargo.toml
```diff
@@ -60,6 +60,8 @@ lazy_static = "1.4.0"
 log.workspace = true
 mobc = { version = "0.8.5", optional = true }
 p256 = { workspace = true, features = ["ecdh"] }
+#prometheus = "0.14.0"
```
fitzthum, line 63: can remove this

fitzthum (approved):
LGTM. One little nit. @mkulke are you good with this?

We've been ignoring the link checker lately although this week it seems to work better so you might push again.

kbs/src/prometheus_exporter.rs
```diff
@@ -0,0 +1,57 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use lazy_static::lazy_static;
+use prometheus::{core::Collector, TextEncoder};
+use std::sync::Mutex;
+
+lazy_static! {
+    pub static ref instance: Mutex<PrometheusExporter> = Mutex::new(PrometheusExporter::new());
```
Xynnn007, line 11: Let's use [tokio::sync::RwLock](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html) rather than `std::sync::Mutex`. 

It can improve performance and efficiency in two aspects: allowing requests at the same time and asynchronous IO.

kbs/src/api_server.rs
```diff
@@ -115,6 +123,11 @@ impl ApiServer {
                             .route(web::get().to(api))
                             .route(web::post().to(api)),
                     )
+                    .service(
+                        web::resource("/metrics")
+                            .route(web::get().to(prometheus_metrics))
```
Xynnn007, line 128: This function name is the same as the [struct member name](https://github.com/confidential-containers/trustee/pull/775/files). Could we rename it into something else like `prometheus_metrics_handler`?

Xynnn007 (commented):
Thanks for the new feature. I only have some comments about the code structure.

I realized that we now have several different ways to integrate `Collector`:
1. In API-Server, we embedded a structure ApiServerMetrics to detect some indicators of /kbs path access
2. In Resource access, we use PrometheusMetrics to monitor resource access indicators

Both structures have internal `Collector` registration logic. I wonder if we can abandon these two structures and converge all Collector initialization operations into the initialization function of the static variable `instance`. Then, the `prometheus_exporter.rs` module exposes several interfaces, namely (names are for explaination)
- log_resource_write(resource_desc)
- log_resource_read(resource_desc)
- log_request_duration(duration)
- log_request_size(size)
- log_new_request
- log_response_size(size)
- prometheus_metrix()

Let other modules call, so that the code design can avoid too much intrusion between logics and improve maintainability.

kbs/src/api_server.rs
pmores: Sure, the  name was modelled after the `api()` handler name but I'm happy to rename the function.

pmores:
> I wonder if we can abandon these two structures and converge all Collector initialization operations into the initialization function of the static variable instance.

That's an excellent question which I've been thinking about since I've started working on this.  Initially I too would have preferred having everything encapsulated within `PrometheusExporter` as you suggest however I got dissuaded by the amount of tedium required and I also liked the locality of the currently proposed structure - every object which needs Prometheus metrics declares, initialises, handles and deinitialises its metrics objects locally, whereas concentrating this stuff in `PrometheusExporter` feels like a dependency at a distance.

Later, fixing the integration tests highlighted one more potential downside of putting everything into `PrometheusExporter`, which is lifetime of metric objects.  By storing metric objects (`Collector` subtypes) in KBS objects that use them (so far `ApiServer` and `ResourceStorage`) their lifetime is automatically and properly tied to the owning object's lifetime which is great.  Once we move everything into `PrometheusExporter` we'll have to figure out when to deinitialise `Collector`s properly.

This might not be a huge problem if all `Collector`s are ultimately tied to `ApiServer`'s lifetime which might well be the case since the KBS does appear to me basically to be an `ApiServer` instance at this point.  However the moment we add metrics with a different lifetime we're in trouble I'm afraid.

In the end I don't have a strong opinion about this.  Personally I prefer the current decentralised architecture but if you want me to switch to putting everything into `PrometheusExporter` I'm happy to make the change.  Please just let me know.

Xynnn007 (approved):
@pmores OK, I think you are trying to make the life cycle of the measurement object consistent with the life cycle of the corresponding Collector. In my opinion, this is not a very mandatory requirement - because our current API Server object (and its related APi implementation module) can be considered to exist throughout the entire life cycle of KBS, and there is no dynamic plug-in. But if we use a centralized approach, it may involve quite a lot of changes to the integration test, which I think is unnecessary in this PR. We can fix it together with the integration test related code later. There are no other problems. Thanks for your great work!

pmores:
CI failures seem to be due to a failure to fetch ubuntu sgx packages so they don't seem related to the PR itself.

fitzthum:
I think the SGX CI stuff might be resolved now. There is a merge conflict so maybe fix that and push again.

fitzthum:
Ok, we ready to merge? There is some more work planned for this so I think it's a good time to get this in as a first step.

pmores:
I'm not aware of anything preventing a merge at the moment, and I'd welcome it since once these commits are on main it's a bit easier for me to start adding code implementing more metrics as discussed on #543.

Issue #529, opened 2024-10-14
RFC: Trustee CLI


fitzthum:
We should consider introducing a new tool, maybe called Trustee CLI, that supports a wide range of functions.

Here are some things it could do.

1. Everything that KBS Client does. We could extend KBS Client, but I think we might want to keep that tool small in scope.
2. RVPS provisioning. We have a separate tool to provision reference values. IMO this should be deleted and moved into another tool.
3. Provisioning Trustee itself. We have a script to deploy Trustee in k8s and a guide to use docker compose. Why not unify these with a tool that allows users to deploy Trustee with one command. The CLI would also generate the keypairs (which you still have to do manually when using docker compose). Since the CLI would generate the keys, it could also keep track of them for later when the user wants to configure Trustee (later commands wouldn't require specifying the certs).
4. Provisioning Trustee in CoCo. As an extension of the last point, the CLI could also allow a user to deploy Trustee as a CoCo workload with one command. This is a little bit trickier, but it's something we've wanted to support for a long time. The deployment command would be somewhat long running. It would start a pod with the needed runtime classes (and probably sealed secrets to provision various keypairs). The workload would be configured to use the CLI itself as its KBS. The CLI's start confidential cluster command would block until it receives an attestation, which it would somehow fulfill locally. We can discuss this in more detail separately. It's going to be a bit tricky, but I think we have to implement something like this eventually.
5. Encrypting images. We have a CoCo keyprovider tool in guest-components, but it's hard to find and somewhat hard to use. It's arguable that this more related to the workload and should not be part of a KBS tool, but I think it could make sense to add it, especially if it's going to automatically push keys to the KBS.
6. Sealing secrets. Again, we have a tool in guest components for this, but it's hard to find; very similar considerations as the above.
7. Plugin front-end. Our new plugin interface will require some support in admin-facing tooling. This could be as simple as letting an admin provide arbitrary `POST` requests, but we could also implement a front-end plugin interface that would allow different plugins to define their own logic for generating admin requests from commands. This probably isn't necessary in the short term.
8. Status. Currently the only way to figure out what is happening with Trustee and the guests that contact it is to stare at the logs. We should think about creating a more friendly interface. We could do this by parsing the logs into a nicer format or we could introduce more sophisticated interfaces to Trustee.

[...]

PR #776, opened 2025-04-22
Trustee unified CLI

spotlesstofu:
Hey! I've been trying to address #529 and bootstrap a CLI for Trustee.

### Aliases and plugins
I'm supporting aliases and plugins.
Aliases are when you link to the same executable with different names (à la busybox).
Plugins allow third parties to create a `trustee-something` executable and run it with
```
trustee something do-my-third-party-thing
```

### Structure / design
The design direction I took is to unify all the command lines in a single one.
I'm opening this pull as a draft while it's in early stages. Below there is a graphic of the structure I'm thinking about.

trustee-new-architecture drawio

### Next steps

I'll be updating this pull in the next days. This is what I have in mind right now:

- Implement more commands.
- Make the `run` command spawn all the components separately.
- Allow to `run` without a config file, using defaults + CLI arguments.
- Use functions that already exist in the codebase for the `keygen` command.

Harsh feedback is welcome 😁

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5696,6 +5696,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "trustee"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "clap",
+ "env_logger 0.10.2",
+ "log",
+]
+
+[[package]]
 name = "try-lock"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5703,6 +5703,7 @@ dependencies = [
  "clap",
  "env_logger 0.10.2",
  "log",
+ "openssl",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5702,8 +5702,10 @@ dependencies = [
  "anyhow",
  "clap",
  "env_logger 0.10.2",
+ "kbs",
  "log",
  "openssl",
+ "tokio",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
     "tools/kbs-client",
     "deps/verifier",
     "integration-tests",
+    "trustee",
 ]
 resolver = "2"
 
--- a/trustee/Cargo.toml
+++ b/trustee/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "trustee"
+version.workspace = true
+authors.workspace = true
+description.workspace = true
+documentation.workspace = true
+edition.workspace = true
+
+[features]
+default = ["aliases", "plugins"]
+aliases = []
+plugins = []
+
+[dependencies]
+anyhow.workspace = true
+clap.workspace = true
+env_logger.workspace = true
+log.workspace = true
--- a/trustee/Cargo.toml
+++ b/trustee/Cargo.toml
@@ -15,5 +15,7 @@ plugins = []
 anyhow.workspace = true
 clap.workspace = true
 env_logger.workspace = true
+kbs = { version = "0.1.0", path = "../kbs" }
 log.workspace = true
 openssl.workspace = true
+tokio.workspace = true
--- a/trustee/Cargo.toml
+++ b/trustee/Cargo.toml
@@ -16,3 +16,4 @@ anyhow.workspace = true
 clap.workspace = true
 env_logger.workspace = true
 log.workspace = true
+openssl.workspace = true
--- a/trustee/src/aliases.rs
+++ b/trustee/src/aliases.rs
@@ -1,0 +1,21 @@
+use std::env;
+use std::path::Path;
+
+/// get_exe_basename returns the name the executable is running with.
+fn get_exe_basename() -> Option<String> {
+    let exe_name = env::args().next()?;
+    let exe_basename = Path::new(&exe_name).file_name()?.to_str()?;
+    Some(exe_basename.to_string())
+}
+
+/// match_alias selects and runs a cli according to the executable name.
+pub fn match_alias() -> Result<(), String> {
+    let alias = get_exe_basename().unwrap_or_default();
+
+    match alias.as_str() {
+        "true" => {}
+        alias => return Err(format!("not a known alias: {alias}")),
+    }
+
+    Ok(())
+}
--- a/trustee/src/cli.rs
+++ b/trustee/src/cli.rs
@@ -1,0 +1,19 @@
+use anyhow::Result;
+use clap::{error::Error, Parser};
+
+/// Enum to represent different subcommands.
+#[derive(Debug, Parser)]
+enum Commands {}
+
+#[derive(Debug, Parser)]
+#[command(author, version, about, long_about = None)]
+struct Cli {
+    #[command(subcommand)]
+    command: Commands,
+}
+
+pub fn cli_default() -> Result<(), Error> {
+    let cli = Cli::try_parse()?;
+
+    Ok(())
+}
--- a/trustee/src/cli.rs
+++ b/trustee/src/cli.rs
@@ -1,6 +1,11 @@
+use std::path::Path;
+
 use anyhow::Result;
 use clap::{error::Error, Parser};
 use openssl::pkey::PKey;
+use tokio;
+
+use kbs::{ApiServer, KbsConfig};
 
 fn trustee_keygen(path: Option<String>) -> Result<()> {
     let private = PKey::generate_ed25519()?;
@@ -15,7 +20,15 @@ fn trustee_keygen(path: Option<String>) -> Result<()> {
     Ok(())
 }
 
-/// Enum to represent different subcommands.
+async fn trustee_run(config_file: &str) -> Result<()> {
+    let kbs_config = KbsConfig::try_from(Path::new(config_file))?;
+    let api_server = ApiServer::new(kbs_config).await?;
+    // TODO initialize the components separately
+    // and spawn each one within try_join
+    tokio::try_join!(api_server.server()?)?;
+    Ok(())
+}
+
 #[derive(Debug, Parser)]
 enum Commands {
     /// Generate a new key pair
@@ -23,6 +36,12 @@ enum Commands {
         /// Output file for the private key
         #[arg(long = "out")]
         out: Option<String>,
+    },
+    /// Launch Trustee
+    Run {
+        /// Configuration file
+        #[arg(long = "config")]
+        config_file: String,
     },
 }
 
@@ -33,11 +52,12 @@ struct Cli {
     command: Commands,
 }
 
-pub fn cli_default() -> Result<(), Error> {
+pub async fn cli_default() -> Result<(), Error> {
     let cli = Cli::try_parse()?;
 
     match cli.command {
         Commands::Keygen { out } => trustee_keygen(out).unwrap(),
+        Commands::Run { config_file } => trustee_run(&config_file).await.unwrap(),
     };
 
     Ok(())
--- a/trustee/src/cli.rs
+++ b/trustee/src/cli.rs
@@ -1,9 +1,30 @@
 use anyhow::Result;
 use clap::{error::Error, Parser};
+use openssl::pkey::PKey;
+
+fn trustee_keygen(path: Option<String>) -> Result<()> {
+    let private = PKey::generate_ed25519()?;
+    let public = private.public_key_to_pem()?;
+
+    if let Some(path) = path {
+        std::fs::write(&path, private.private_key_to_pem_pkcs8()?)?;
+        let public_path = format!("{path}.pub");
+        std::fs::write(&public_path, public)?;
+    }
+
+    Ok(())
+}
 
 /// Enum to represent different subcommands.
 #[derive(Debug, Parser)]
-enum Commands {}
+enum Commands {
+    /// Generate a new key pair
+    Keygen {
+        /// Output file for the private key
+        #[arg(long = "out")]
+        out: Option<String>,
+    },
+}
 
 #[derive(Debug, Parser)]
 #[command(author, version, about, long_about = None)]
@@ -14,6 +35,10 @@ struct Cli {
 
 pub fn cli_default() -> Result<(), Error> {
     let cli = Cli::try_parse()?;
+
+    match cli.command {
+        Commands::Keygen { out } => trustee_keygen(out).unwrap(),
+    };
 
     Ok(())
 }
--- a/trustee/src/main.rs
+++ b/trustee/src/main.rs
@@ -10,7 +10,8 @@ mod aliases;
 #[cfg(feature = "plugins")]
 mod plugins;
 
-fn main() {
+#[tokio::main]
+async fn main() {
     env_logger::init_from_env(env_logger::Env::new().default_filter_or("info"));
 
     #[cfg(feature = "aliases")]
@@ -21,7 +22,7 @@ fn main() {
         }
     }
 
-    let cli_error = match cli::cli_default() {
+    let cli_error = match cli::cli_default().await {
         Ok(_) => return,
         Err(e) => e,
     };
--- a/trustee/src/main.rs
+++ b/trustee/src/main.rs
@@ -1,0 +1,35 @@
+use std::env;
+
+use log::debug;
+
+mod cli;
+
+#[cfg(feature = "aliases")]
+mod aliases;
+
+#[cfg(feature = "plugins")]
+mod plugins;
+
+fn main() {
+    env_logger::init_from_env(env_logger::Env::new().default_filter_or("info"));
+
+    #[cfg(feature = "aliases")]
+    match aliases::match_alias() {
+        Ok(_) => return,
+        Err(e) => {
+            debug!("matching alias failed: {}", e)
+        }
+    }
+
+    let cli_error = match cli::cli_default() {
+        Ok(_) => return,
+        Err(e) => e,
+    };
+
+    #[cfg(feature = "plugins")]
+    if env::args().count() > 1 {
+        plugins::exec();
+    }
+
+    cli_error.exit()
+}
--- a/trustee/src/plugins.rs
+++ b/trustee/src/plugins.rs
@@ -1,0 +1,42 @@
+use std::env;
+use std::io::ErrorKind::{NotFound, PermissionDenied};
+use std::process::{exit, Command};
+
+use log::{debug, error};
+
+const PLUGIN_PREFIX: &str = "trustee";
+
+/// Spawns the plugin command and exits.
+///
+/// Returns only if the plugin is not found or it's not executable,
+/// to allow to continue the CLI flow.
+pub fn exec() {
+    let mut args_iter = env::args();
+    // skip arg0
+    args_iter.next();
+    let command = args_iter.next().expect("missing plugin command");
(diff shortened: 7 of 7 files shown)


fitzthum:
It's great to see this PR. A few high-level comments 

First, some simple things. I'd probably find a different location for the CLI code; maybe in the tools repo? Also, I wonder if we need the keygen command to be exposed on its own. Maybe we could just generate the key automatically if it doesn't already exist.

A bigger question is whether the CLI should spawn any other processes. I lean towards not doing that. I like how you are starting the KBS a a crate like we do in the CI. This is suited to the local Trustee approach (where you have a short-lived trustee running locally that can attest something else like another Trustee in an enclave). We should also think about if the CLI can hook into docker compose or the operator to deploy a long-lived Trustee.

spotlesstofu:
Pushed new commits.

Partially addressed @fitzthum comments:
- Moved the module to `tools/`.
- Now generating the admin auth key automatically if it's not set.

I also added a self-signed certificate generation and set a default policy, see commits.

I think exposing the keygen command is still useful. It enables to integrate it in a workflow that we are not yet supporting automatically. For example one may want to use it to create a secret on Kubernetes?

To do:
- Do not spawn processes.
- Use the trustee_home directory as a base for other file paths, like the attestation service work_dir.

These two I'd like to address in this pull request. Then I'll open new pulls to further extend the features of the tool.

tools/trustee/src/cli.rs
```diff
@@ -50,6 +83,9 @@ pub async fn cli_default() -> Result<(), Error> {
             let out = out.unwrap_or_else(|| trustee_home_dir.join("key"));
             trustee_keygen(&out).unwrap();
         }
+        Commands::Run { config_file } => {
+            trustee_run(config_file, &trustee_home_dir).await.unwrap();
```
pmores: Do we want to go with `unwrap()` here (and other commands as well)?  With `unwrap()` I might get
```
thread 'main' panicked at tools/trustee/src/cli.rs:126:18:
called `Result::unwrap()` on an `Err` value: refusing to overwrite file: "/home/pvl/.trustee/https_key.pem"
```
whereas if I replace the `unwrap()` with just a quick fix like this:
```
.map_err(|e| Error::raw(clap::error::ErrorKind::InvalidValue, format!("{}\n", e.to_string())))?
```
the same error is reported as
```
error: refusing to overwrite file: "/home/pvl/.trustee/https_key.pem"
```
which seems much clearer.  Granted, both contain the important information but frankly I initially didn't even notice it in the unwrap message. :-)

pmores: A broader question might be how we want to handle errors.  The fix I proposed above assumes we want to return `clap::Error` from `cli_default()` - but do we?  The only reason seems to be the `Cli::try_parse()?` call at the top of the function, command handlers apparently use `anyhow`.

Maybe we'd prefer to convert the `try_parse()` error to `anyhow` and then `cli_default()` returns anyhow errors, making it possible for command handlers to be called simply like
```
trustee_run(config_file, &trustee_home_dir).await?;
```
?

pmores:
Also, unless I revert the initial commit (but keep `derive(Default)` on `KbsConfig`) `KbsConfig` seems to be unable to read the sample config file in `kbs/config/kbs-config.toml` anymore due to
```
invalid config: missing field `sockets`
```
If `sockets` is supplied it complains about the next config member.  Basically all configuration items seem to have become mandatory.

spotlesstofu:
> ```
> invalid config: missing field `sockets`
> ```

So weird. If I add a `derive(Default)` it works fine.

It's like `try_deserialize` is ignoring the `impl Default` but not the `derive(Default)`.

spotlesstofu:
> It's like `try_deserialize` is ignoring the `impl Default` but not the `derive(Default)`.

My bad. I confused with `#[serde(default)]`. I'll push a fix.

spotlesstofu:
Comments addressed and rebased on main. I'd appreciate a review.

kbs/src/config.rs
```diff
@@ -82,14 +83,6 @@ impl TryFrom<&Path> for KbsConfig {
     /// `config` crate. See `KbsConfig` for schema information.
     fn try_from(config_path: &Path) -> Result<Self, Self::Error> {
         let c = Config::builder()
-            .set_default("admin.insecure_api", DEFAULT_INSECURE_API)?
-            .set_default("http_server.insecure_http", DEFAULT_INSECURE_HTTP)?
-            .set_default("http_server.sockets", vec![DEFAULT_SOCKET])?
```
fitzthum, line 87: Seems like we are currently setting the default socket to `127.0.0.1:8080` but is that still going to be the case with these changes? I see you added a default derive above, but idk if that is going to get the same value.

tools/trustee/src/aliases.rs
```diff
@@ -0,0 +1,21 @@
+use std::env;
+use std::path::Path;
+
+/// get_exe_basename returns the name the executable is running with.
+fn get_exe_basename() -> Option<String> {
+    let exe_name = env::args().next()?;
+    let exe_basename = Path::new(&exe_name).file_name()?.to_str()?;
+    Some(exe_basename.to_string())
```
fitzthum: Should this function return an option or a result? It seems like it will always return the binary name unless there is an error?

tools/trustee/src/aliases.rs
```diff
@@ -0,0 +1,21 @@
...
+}
+
+/// match_alias selects and runs a cli according to the executable name.
+pub fn match_alias() -> Result<(), String> {
+    let alias = get_exe_basename().unwrap_or_default();
+
+    match alias.as_str() {
+        "true" => {}
```
fitzthum: This is odd control flow. Maybe use an `if let` statement instead.

tools/trustee/src/aliases.rs
```diff
@@ -0,0 +1,21 @@
...
+
+/// get_exe_basename returns the name the executable is running with.
+fn get_exe_basename() -> Option<String> {
+    let exe_name = env::args().next()?;
+    let exe_basename = Path::new(&exe_name).file_name()?.to_str()?;
+    Some(exe_basename.to_string())
+}
+
+/// match_alias selects and runs a cli according to the executable name.
+pub fn match_alias() -> Result<(), String> {
```
fitzthum: Any reason not to use `anyhow` here?

tools/trustee/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+#[cfg(feature = "plugins")]
+mod plugins;
+
+#[tokio::main]
+async fn main() {
+    env_logger::init_from_env(env_logger::Env::new().default_filter_or("info"));
+
+    #[cfg(feature = "aliases")]
+    match aliases::match_alias() {
+        Ok(_) => return,
```
fitzthum: Maybe I don't understand the aliases feature at all but why would we return here?

tools/trustee/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+        Err(e) => {
+            debug!("matching alias failed: {}", e)
+        }
+    }
+
+    #[cfg(feature = "plugins")]
+    match std::env::args().nth(1) {
```
fitzthum: I don't really like reading the args manually like this. You are using clap in cli.rs but not here.

tools/trustee/src/cli.rs
```diff
@@ -0,0 +1,169 @@
...
+    if !trustee_home_dir.exists() {
+        std::fs::create_dir_all(&trustee_home_dir)?;
+    }
+
+    match cli.command {
+        Commands::Keygen { output_file: out } => {
+            let out = out.unwrap_or_else(|| trustee_home_dir.join("key"));
+            trustee_keygen(&out)?;
+        }
+        Commands::Run {
```
fitzthum: May want to make this enum variant more explicit in terms of what it is running.

fitzthum (commented):
A few comments. 

I still don't really understand the purpose of plugins or aliases. In the code, the relationship between `main.rs` and `cli.rs` is a little confusing. It seems like the CLI parsing should potentially be the top level.

kbs/src/config.rs
spotlesstofu: Yes. The `impl Default for HttpServerConfig` uses `DEFAULT_SOCKET` too:
https://github.com/confidential-containers/trustee/blob/8462025ed2d2a94281344e63405bbe7500bd4484/kbs/src/config.rs#L42

tools/trustee/src/aliases.rs
mkulke: there is `env::current_exe()` that already returns a PathBuf and on that you can use `.file_name()`

tools/trustee/src/aliases.rs
```diff
@@ -0,0 +1,21 @@
...
+    let alias = get_exe_basename().unwrap_or_default();
```
mkulke: I think we should always be able to get the basename of the executable and just propagate the error or abort here w/ unwrap or maybe `.expect("couldn't get executable's name");`

tools/trustee/src/aliases.rs
mkulke: I don't understand this either. why `"true"`? in which cases would the alias be set to "true"? and why are all other alias names invalid?

tools/trustee/src/cli.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+fn get_config(
+    config_file: Option<PathBuf>,
+    allow_all: bool,
+    trustee_home_dir: &Path,
+) -> Result<KbsConfig> {
+    let mut config = config_file
+        .map(|config_file| KbsConfig::try_from(config_file.as_path()).unwrap())
+        .unwrap_or_default();
```
mkulke: let's propagate the errors (and avoid the unwraps), if we have already return a `Result`.

tools/trustee/src/cli.rs
```diff
@@ -0,0 +1,169 @@
...
+            &config.policy_engine.policy_path,
+            include_bytes!("../../../kbs/sample_policies/deny_all.rego"),
+        )?;
+    }
+
+    debug!("Config: {:?}", config);
+    Ok(config)
+}
+
+async fn trustee_run(config: KbsConfig) -> Result<()> {
```
mkulke: is the rationale of this CLI to have a simple launcher for trustee that will implicitly create all the required resources?

is there an overlap with what "trustee-operator" wants to provide?

tools/trustee/src/aliases.rs
spotlesstofu: All the calls in the function return an option, so I figured I could pass that behavior on transparently?

tools/trustee/src/keys_certs.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    let public_pem = private_key.public_key_to_pem()?;
+    safe_write(&public_path, public_pem)?;
+
+    Ok(())
+}
+
+/// Generate a new key for authentication and write to the given file paths.
+pub(crate) fn write_new_auth_key_pair(private_path: &Path, public_path: &Path) -> Result<()> {
+    let private_key = PKey::generate_ed25519()?;
```
mkulke: maybe document the choice of the key algo somewhere (in the CLI help would be sufficient)

tools/trustee/src/keys_certs.rs
```diff
@@ -0,0 +1,124 @@
...
+
+/// Ensure a key pair for HTTPS exists in the given home directory.
+///
+/// Create the key files if they don't exist.
+/// Return the path to the private and public keys.
+pub(crate) fn ensure_https_key_pair(home_dir: &Path) -> Result<(PathBuf, PathBuf)> {
+    let (private_path, public_path) = get_key_paths(home_dir, "https_key");
+
+    if !private_path.try_exists()? {
+        let rsa = Rsa::generate(2048)?;
```
mkulke: Maybe document the choice for the key algo somewhere

tools/trustee/src/cli.rs
fitzthum: This is a bit different because it's running Trustee as a crate directly from the CLI process. This is intended to run locally potentially to bootstrap another Trustee running inside an enclave.

tools/trustee/src/aliases.rs
spotlesstofu: I added some comments in the file to explain how I thought this alias feature. Please have a look at those.

The `"true"` is a placeholder for other aliases to be added in the future. For the same reason, the `match` seems a better fit.

tools/trustee/src/main.rs
spotlesstofu: I added inline comments for this as well.

tools/trustee/src/main.rs
spotlesstofu: The thing with plugins is that they're dynamic. Like `kubectl` works.

For example. If I have a `trustee-nvidia` in my `$PATH`, calling `trustee nvidia ...` should redirect to `trustee-nvidia`.
We don't know this at compile time.

I could use the clap builder pattern, to define a subcommand for the plugins dymanically. I'm not sure it's worth it.

tools/trustee/src/cli.rs
spotlesstofu: What would you suggest? The code is looping on an enum that's defined above, I can't think of anything to make it clearer.

fitzthum:
Yeah, I would just drop the alias support for now to make this PR as simple as possible. Keep the code and we can add it back in if someone comes with the requirement.

For plugins I would prefer that we access any sub-programs via crates rather than as binaries. I don't have a strong opinion about what goes into this PR for plugins.

Otherwise seems good.

kbs/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -84,6 +84,7 @@ pub struct IntelTrustAuthorityConfig {
     pub api_key: String,
     pub certs_file: String,
     pub allow_unmatched_policy: Option<bool>,
+    #[serde(default)]
```
pmores, line 94: Is this change actually required?  It seems isolated from the rest of the commit.

spotlesstofu: Indeed. I removed it.

spotlesstofu: No. Removing it makes the test fail. I put it back.
```
invalid config: missing field `policy_ids`
```

pmores: Which test is that?  `cargo test` is failing for me in `test tests::parse_data_json_binding::case_3` and
`token::simple::tests::test_issue_simple_ephemeral_key`, and that's _regardless_ of this hunk...

spotlesstofu: [job/51133821636](https://github.com/confidential-containers/trustee/actions/runs/17977292295/job/51133821636)

To reproduce, I guess you have to enable the feature `intel-trust-authority-as`.

pmores: Alright, never mind then.

tools/trustee/src/cli.rs
```diff
@@ -0,0 +1,171 @@
...
+    trustee_home_dir: &Path,
+) -> Result<KbsConfig> {
+    let mut config = if let Some(path) = config_file {
+        KbsConfig::try_from(path.as_path())?
+    } else {
+        KbsConfig::default()
+    };
+
+    config.policy_engine.policy_path =
+        replace_base_dir(config.policy_engine.policy_path.as_path(), trustee_home_dir);
```
pmores: Is it guaranteed that the policy path - and other paths passed to `replace_base_dir()` later in this function - will have the base path that `replace_base_dir()` assumes?

Or is it the intended semantics that we redirect specifically and only `/opt/confidential-containers/*` paths to `trustee_home_dir` but leave any other paths alone?  If so I think it would warrant a short explanation in a code comment.

spotlesstofu: I added a comment to explain this, reproduced below.

```rust
/// replace_base_dir replaces the leading `/opt/confidential-containers/` in the path with a new base path.
///
/// This behavior is a compromise to set the base directory at runtime and workaround the hardcoded paths all around the codebase.
///  replace_base_dir will become obsolete when it's possible to set the base directory at runtime project-wide.
fn replace_base_dir(path: &Path, new_base: &Path) -> PathBuf {
    let old_base = "/opt/confidential-containers/";
    let suffix = if path.starts_with(old_base) {
```
https://github.com/confidential-containers/trustee/compare/d3fae6ba11e8d451e8511cc723e0f9f2fe251d1f..df251a09316d0c7bbe367927caeee150c13b9b9f

pmores:
Cool, you even managed to handle the default value of `--home` with the `default_value` annotation - I didn't think this was even possible for options of a type which is not `Display`. :-)

This seems much easier to make sense of, at least to me.  Now, I'm still not clear about the purpose of the `keygen` command.  I'm aware that this has been raised and discussed before but... looking at the code it appears that `keygen` could be used to generate the `auth` key pair though not the HTTPS key pair which uses a different algorithm.  However, the `run` command will generate it too, if it's missing.  What is the advantage of generating the `auth` key pair ahead of time using `keygen`?

spotlesstofu:
Yes, `keygen` generates the auth key.

pmores:
Right.  I understand the idea of shielding users from raw `openssl` commands and I know it can be easy to misuse them inadvertently.  But isn't that goal achieved already by having the `run` command generate `auth` if necessary?

Or, again, what is then the advantage of using `keygen` to generate `auth` explicitly?

spotlesstofu:
I think of these use cases for the `keygen` command:
1. Generate the key pair for a remote deployment (Kubernetes). Because of the security model for private-public key pairs, a remote deployment can't auto-generate the key. The admin can generate the key with `keygen` and provide the public key to the remote deployment. This is a stand-alone usage of `keygen` as a utility and doesn't involve the `run` command.
2. The user wants to make a custom config file and reference the key from there. Then the config file can be passed to the `run` command.

pmores:
Quite a bunch of paths are retargeted to trustee home dir in `run::get_config()` but not the resource repository path.  Even if I give `--home` my resources are stored under `/opt/confidential-containers/kbs/repository` (the  value of `dir_path` of the `LocalFs` plugin in my KBS config file).  Is that an omission, or intentional?

pmores (approved):
lgtm thanks @spotlesstofu for your patience! :-)

fitzthum (approved):
LGTM. 

There are some things to tweak but I think this is a good start.

tools/trustee/src/main.rs
```diff
@@ -0,0 +1,106 @@
...
+        } => run::trustee_run(&cli.home, config_file, allow_all).await,
+    }
+}
+
+/// Write a key to separate files: one for private, one for public.
+///
+/// Error if the file already exists.
+fn write_pem(private_path: &Path, public_path: &Path, private_key: &PKey<Private>) -> Result<()> {
+    // Write the private key to a file private to the owner.
+    {
```
mkulke: why the block scope? is it to flush f?

mkulke (approved):

tools/trustee/src/main.rs
spotlesstofu: It's not that needed, but I think it makes the code more readable.

mkulke: hmm, block scopes in Rust will materialize in the compiled code, e.g. objects will get drop()ed at he end of the block, which means files will be closed at this point. I don't think it's idiomatic in Rust to use blocks for pure visual grouping, we can use comments or functions for that.

spotlesstofu: I moved the block to a function as suggested.

PR #777, opened 2025-04-22
extractors: add SWID/RIM extractor

fitzthum:
Add a new extractor to the RVPS. This will pull reference values out of SWID/RIM-IM manifests. 

If you don't know those standards, consider yourself lucky. Basically some reference value manifests follow both the SWID and RIM-IM spec, which each define some tags and some bindings that can be used to describe a set of reference values. NVIDIA happens to use this format for some of their reference values, such as the vbios of the h100.

This extractor adds only the most basic support for SWID/RIM-IM. It should be iterated on in the future. Right now it does not validate the signatures of the manifest, so you'll need to trust the source of your reference values. It also only knows about certain types of SWID payloads and won't be able to handle all valid manifests. Still, this is a reasonable place to start.

For now Trustee has a push model for reference values. This means that the admin needs to go find the reference values and upload them. For an H100 this means doing something like this 

```
curl -s 'https://rim.attestation.nvidia.com/v1/rim/NV_GPU_VBIOS_1010_0200_882_960074001C -o rim.b64
provenance=$(cat ~/rim.b64)

cat << EOF > message
{
    "version" : "0.1.0",
    "type": "swid",
    "payload": "$provenance"
}
EOF

./rvps-tool register --path ./message --addr htp://127.0.0.1:50003
./rvps-tool query --addr htp://127.0.0.1:50003
```
Then, your RVPS should have all the reference values from the manifest. Hopefully this will be useful once we start writing policies for devices.


Xynnn007 (approved):
Nice PR. It is probably the most difficult partto bridge parsed claims and reference value carriers such as RIM/SWID with a unified ID. We need to support dynamic acquisition of values ​​in RVPS in AS policy, and for each reference value carrier, we need to specify how it is mapped to a unique RV id. I think this PR is a good start.

fitzthum:
Yeah while working on this I ran into some questions about how to resolve the reference value names. For instance, should the reference value names have the version of an artifact in them? It's important to track the versions of the reference values somewhere, but if the name includes the version, it makes the policy much more brittle. Adopting something more standardized (like the multi-part reference value URI proposal) could help with this. Anyway we can adjust as we start to understand all of this better.

PR #779, opened 2025-04-23
Add built-in affirming policy to KBS client

fitzthum:
Recently I added a couple built-in policies to the KBS Client. These simplify the workflow for admins who are setting basic policies. Unfortunately, I didn't think to add the policy that is probably most important; the affirming policy, which basically makes sure that the attestation policy passed.


Xynnn007 (approved):
lgtm

bpradipt (approved):

PR #780, opened 2025-04-23
policy: fix SNP policy

fitzthum:
The SNP policy references a claim called `launch_measurement`, but the claim is actually just `measurement`.

cc @bpradipt @ryansavino


Xynnn007 (approved):
LGTM. See https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/snp/mod.rs#L333

bpradipt (approved):

mythi (approved):

ryansavino (approved):
Thanks @fitzthum

PR #785, opened 2025-04-30
Add support for new EPYC generations and update VIRTEE SEV dependency

DGonzalezVillal:
Updated the VIRTEE SEV crate dependency in the verifier to version 6.0.0. This release adds support for newer versions of the AMD attestation report. Starting with version 3 of the report, we can identify the EPYC processor generation used by the host, allowing us to select the correct certificate chain for attestation verification.

Additional TCB fields introduced with the Turin generation are now supported. We’ve also added logic to request the VCEK from the KDS for all supported generations.

To prevent ambiguity during attestation, we have bumped the minimum supported attestation report version to 3. Reports from earlier versions lack the necessary information to determine the correct certificate chain.

We removed some of the previous custom verification logic and now use the built-in verification methods provided by the SEV crate. This allows us to offload maintenance of verification logic and simplifies future updates—just a dependency bump should suffice.

Lastly, we refactored the certificate chain logic. Previously, we carried two chains—one from stored certs and another from user-provided input. We’ve now simplified this to use a single chain: if the user provides certs via extended attestation, those are used; otherwise, the verifier defaults to the stored certificates.

Another note, for these changes to work on the stack we need to update guest components (I have those changes which I'll be PR tomorrow), and the azure_snp_vtpm crate also needs to be updated, so that we can add support for the azure verifier. Right now the Azure vtpm tests will fail since it's using 2 different versions of the attestation report. I can update this PR to fix the azure verifier once the crate is updated to use the right version of the VIRTEE SEV crate. 

Also the KBS-types just had a pr to support Turin, so that new release will probably need to be added to trustee at some point too.


deps/verifier/src/snp/mod.rs
```diff
@@ -44,57 +42,95 @@ const UCODE_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .8);
...
+
+impl FromStr for ProcType {
+    type Err = anyhow::Error;
+    fn from_str(input: &str) -> Result<ProcType, anyhow::Error> {
+        match input.to_lowercase().as_str() {
+            "milan" => Ok(ProcType::Milan),
+            "genoa" => Ok(ProcType::Genoa),
+            "turin" => Ok(ProcType::Turin),
+            _ => Err(anyhow::anyhow!("Processor type not found!")),
+        }
```
Xynnn007, line 122: We can use [strum](https://docs.rs/strum_macros/latest/strum_macros/derive.EnumString.html) crate to handle this

deps/verifier/src/snp/mod.rs
```diff
@@ -44,57 +42,95 @@ const UCODE_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .8);
...
-    ark: X509,
-    asvk: X509,
+impl fmt::Display for ProcType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            ProcType::Milan => write!(f, "Milan"),
+            ProcType::Genoa => write!(f, "Genoa"),
+            ProcType::Turin => write!(f, "Turin"),
+        }
+    }
```
Xynnn007, line 133: we can use [strum](https://docs.rs/strum_macros/latest/strum_macros/derive.Display.html) to handle this

deps/verifier/src/snp/mod.rs
```diff
@@ -111,32 +147,149 @@ impl Verifier for Snp {
...
+            std::str::from_utf8(evidence)?
+        ))?;
 
         // See Trustee Issue#589 https://github.com/confidential-containers/trustee/issues/589
-        if report.version < REPORT_VERSION_MIN || report.version > REPORT_VERSION_MAX {
+        // Version 3 minimum is needed to tell processor type in report
+        if report.version < REPORT_VERSION_MIN {
+            return Err(anyhow!(
+                "Attestation Report version is too old. Please update your firmware."
+            ));
```
Xynnn007, line 160: what about replace `return Err(anyhow!(...))`  with `bail!(...)`?

deps/verifier/src/snp/mod.rs
```diff
@@ -44,57 +42,95 @@ const UCODE_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .8);
...
 
         let vendor_certs = VendorCertificates {
-            ask: certs[0].clone(),
-            ark: certs[1].clone(),
-            asvk: certs[2].clone(),
+            ask: Certificate::from(certs[0].clone()),
+            ark: Certificate::from(certs[1].clone()),
+            asvk: Certificate::from(certs[2].clone()),
         };
+
```
Xynnn007, line 97: [OnceLock](https://doc.rust-lang.org/beta/std/sync/struct.OnceLock.html) will only be executed once in the entire life cycle. The `cert_authority` initialized here is related to the input parameters, which means that only one type of certificate chain will be initialized?

Xynnn007 (commented):
Thanks for the nice feature. Let's get it better

deps/verifier/src/snp/mod.rs
DGonzalezVillal: I replaced a few of the Err(anyhow!(...)) with bail!(...), do you want to replace all of them? Or when would you choose bail! over anyhow!

deps/verifier/src/snp/mod.rs
DGonzalezVillal: I see. So that means we want to initialize each chain separately then? That way you can attest any cert model that may be provided by different vms?

DGonzalezVillal: But in one life cycle we would only attest the one attestation report right? and one report is only attached to only one chain? So wouldn't we want to only initialize one chain? We are using the report to get the appropriate chain.

DGonzalezVillal: nvm, i think i see the problem. This value gets locked even if we call evaluate() several times.

deps/verifier/src/snp/mod.rs
fitzthum: `bail!` is equivalent to `return Err(anyhow!())`. Anywhere that you want to immediately return an error use `bail!`. There might be places where you just want to create an error without immediately returning (like inside some map statements) and you can stick with `anyhow!` there.

DGonzalezVillal:
@Xynnn007 @fitzthum Updated the commit addressing the changes.

I now initialize a hashmap that contains all the vendor certs and uses the proc_type as a key. I also bumped the dependency for the SEV library to the recent 6.1.0 release. The only reason for this is that in that release we have a couple bug fixes we caught later on in development. 

Please let me know what other comments or suggestions you may have.

deps/verifier/src/snp/mod.rs
```diff
@@ -44,57 +49,93 @@ const UCODE_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .8);
...
-    /// Returns an error if the certificate chain can not be loaded.
+    /// Creates a new `Snp` instance by loading the SNP certificate chain into a hash-map.
+    /// Returns an error if the certificate chains can not be loaded.
     pub fn new() -> Result<Self> {
-        let Result::Ok(vendor_certs) = load_milan_cert_chain() else {
-            bail!("Failed to load Milan cert chain");
+        let Result::Ok(vendor_certs) = load_cert_chain() else {
+            bail!("Failed to load stored SNP vendor certs");
         };
         let vendor_certs = vendor_certs.clone();
```
Xynnn007, line 74: I noticed that the `load_cert_chain()` function is called only once but will be `clone()` for every `Snp::new()`. This implies cert data memory copy every time a verification request comes but the data of certs is static and constant. 

Another way to design this is use a `static LazyLock` on top of the module to include the HashMap<HashMap<ProcType, VendorCertificates>> of the module and get rid of the `vendor_certs` member of `Snp`. Then try to directly reference the static variable wherever needs.

This is relative easy like
```rust
static CERT_CHAINS: LazyLock<HashMap<ProcType, VendorCertificates>> = LazyLock::new(|| {
    ....// initial the map and feel free to panic or unwrap as the cert files are statically included thus
    // will not cause unexpected errors
}
```

Xynnn007 (commented):
Thanks. There are some more comments. Btw, noticed that az-snp-vtpm imports the [load_milan_cert_chain](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/az_snp_vtpm/mod.rs#L8) function which is retired in this PRm, will the update break that?

deps/verifier/src/snp/mod.rs
```diff
@@ -1,26 +1,31 @@
...
+        guest::AttestationReport,
+        host::{CertTableEntry, CertType},
+    },
+};
+use std::{collections::HashMap, hash::Hash, result::Result::Ok, sync::OnceLock};
+use strum::{Display, IntoEnumIterator};
+use strum_macros::{EnumIter, EnumString};
 use x509_parser::prelude::*;
 
+extern crate serde;
```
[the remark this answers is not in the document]
DGonzalezVillal: In the original code, this was already written as `extern crate serde`.(https://github.com/confidential-containers/trustee/blob/d9eb5e0cb0aca97abe35b58908e061850ff60a51/deps/verifier/src/snp/mod.rs#L4) That's why I kept it that way, but my ide moved it down. I'll change it to `use serde` if you guys don't mind then.

DGonzalezVillal:
> Thanks. There are some more comments. Btw, noticed that az-snp-vtpm imports the [load_milan_cert_chain](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/az_snp_vtpm/mod.rs#L8) function which is retired in this PRm, will the update break that?

@Xynnn007 I’ve addressed those comments. The vendored certs are now initialized using a static reference to a LazyLock, and are no longer being set up with OnceLock.

Regarding your question about az-snp-vtpm: the version bump of the SEV crate will break functionality for az-snp-vtpm. In its current state, az-snp-vtpm will not work—not only because load_milan_cert_chain() was removed, but also because it still depends on the azure-snp crate, which in turn relies on an outdated version of the Virtee SEV library.

This has been discussed with @fitzthum. We have two paths forward:

Wait for the az-snp-vtpm crate to be updated to use the latest Virtee SEV release, and then update the dependency here as well.

Add another commit to this PR to get az-snp-vtpm working again, even if that means using outdated logic. At least everything will be functional for now. Once az-snp-vtpm is updated, we can raise a new PR to update the verifier.

I’ll push a commit implementing option 2 to restore compatibility with az-snp-vtpm, and then we can discuss the next steps.

DGonzalezVillal:
@Xynnn007 @fitzthum @Xynnn007 @ryansavino Pushed the changes to patch the Azure verifier with the bump. Please let me know if there is anything else that needs to be fixed.

deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -188,6 +206,57 @@ fn verify_snp_report(
...
+    if common_name == "VCEK"
+        && get_oid_octets::<64>(&parsed_endorsement_key, HW_ID_OID)? != report.chip_id
+    {
+        bail!("Chip ID mismatch");
+    }
+
+    // tcb version
+    // these integer extensions are 3 bytes with the last byte as the data
+    if get_oid_int(&parsed_endorsement_key, UCODE_SPL_OID)? != report.reported_tcb.microcode {
+        return Err(anyhow!("Microcode version mismatch"));
```
Xynnn007, line 232:
```suggestion
        bail!("Microcode version mismatch");
```

deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -188,6 +206,57 @@ fn verify_snp_report(
...
+    }
+
+    if get_oid_int(&parsed_endorsement_key, SNP_SPL_OID)? != report.reported_tcb.snp {
+        return Err(anyhow!("SNP version mismatch"));
```
Xynnn007, line 236:
```suggestion
        bail!("SNP version mismatch");
```

deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -188,6 +206,57 @@ fn verify_snp_report(
...
+    }
+
+    if get_oid_int(&parsed_endorsement_key, TEE_SPL_OID)? != report.reported_tcb.tee {
+        return Err(anyhow!("TEE version mismatch"));
```
Xynnn007, line 240:
```suggestion
        bail!("TEE version mismatch");
```

deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -188,6 +206,57 @@ fn verify_snp_report(
...
+    }
+
+    if get_oid_int(&parsed_endorsement_key, LOADER_SPL_OID)? != report.reported_tcb.bootloader {
+        return Err(anyhow!("Boot loader version mismatch"));
```
Xynnn007, line 244:
```suggestion
        bail!("Boot loader version mismatch");
```

deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -188,6 +206,57 @@ fn verify_snp_report(
...
+    }
+
+    // verify report signature
+    let sig = ecdsa::EcdsaSig::try_from(&report.signature)?;
+    let data = &bincode::serialize(&report)?[..=0x29f];
+
+    let pub_key = EcKey::try_from(vcek.0.public_key()?)?;
+    let signed = sig.verify(&sha384(data), &pub_key)?;
+    if !signed {
+        return Err(anyhow!("Signature validation failed."));
```
Xynnn007, line 254:
```suggestion
        bail!("Signature validation failed.");
```

deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -188,6 +206,57 @@ fn verify_snp_report(
...
+        return Err(anyhow!("TEE version mismatch"));
+    }
+
+    if get_oid_int(&parsed_endorsement_key, LOADER_SPL_OID)? != report.reported_tcb.bootloader {
+        return Err(anyhow!("Boot loader version mismatch"));
+    }
+
+    // verify report signature
+    let sig = ecdsa::EcdsaSig::try_from(&report.signature)?;
+    let data = &bincode::serialize(&report)?[..=0x29f];
```
Xynnn007, line 249: Here we have a hardcoded `0x29f`, probably it is good to add a note in code it means the [AttestationReport](https://docs.rs/sev/6.1.0/sev/firmware/guest/struct.AttestationReport.html) struct except signature field, or use [memoffset](https://crates.io/crates/memoffset) crate to generate the value.

Xynnn007 (approved):
Overall lgtm. Only last comments and I think it's safe to give an approval.

fitzthum (commented):
Looks good. I suggested some pedantic things. @mkulke ptal at the Azure bits. We should be close to merging.

deps/verifier/src/snp/mod.rs
```diff
@@ -111,32 +131,147 @@ impl Verifier for Snp {
         let SnpEvidence {
             attestation_report: report,
             cert_chain,
-        } = serde_json::from_slice(evidence).context("Deserialize Quote failed.")?;
+        } = serde_json::from_slice(evidence).context(format!(
+            "Deserialize Quote failed. Evidence {}",
+            std::str::from_utf8(evidence)?
```
mkulke, line 136: do we want to add the full evidence to the error string, that could be pretty noisy.

DGonzalezVillal: Good Catch I had add this for debugging and forgot to remove it

DGonzalezVillal:
@fitzthum Comments addressed!

fitzthum (approved):
LGTM

DGonzalezVillal:
Hey guys, sorry about all the pushes yesterday—I made some Git mistakes and had to fix those. Now I’m running into a separate issue and wanted to get your thoughts.

All the e2e tests are passing after integrating the changes in guest-components, but there are two tests that are still failing:

The test using the evidence.json attestation report (CoCo-AS e2e with [TEE=snp Generate Evidence Dynamically=false]). That JSON file was in an old format, so I updated it and added a cargo test to confirm that serialization and deserialization still work if you're using the correct version of the library. That one should pass now.

The failing test in the CoCo-AS Rust Suites during the cargo test run is related to the VLEK report, which is now in an unsupported format.

The problem is that the library needs to determine the processor generation to correctly parse the TCB fields in the report. If the report is version 2, it can’t use the CPUID family/model fields since they aren’t present. In that case, it falls back to checking the CHIP ID field. However, reports from Turin have a different length than pre-Turin (Genoa and Milan), which allows us to distinguish between them. But if the MASK CHIP ID setting is enabled (as it is in these reports), the field is set to all zeros, and the library has no way of identifying the processor generation—so it errors out. That’s what’s happening in this test.

I tried modifying the CHIP ID field to resemble a pre-Turin report, which allows the report to be parsed, but the signature verification then fails due to the modification.

@wobito I heard you provided the VLEK test reports before—could you help me get an updated VLEK report, or should we consider disabling the VLEK tests for now until VLEK support is upstream?

On my end, I’m also working on getting a VLEK report.

Apologies for not catching this earlier—the tests were passing before the recent changes, and I forgot to re-run them afterward.

wobito:
> @wobito I heard you provided the VLEK test reports before—could you help me get an updated VLEK report, or should we consider disabling the VLEK tests for now until VLEK support is upstream?

Hey @DGonzalezVillal - If my memory serves me correctly... I generated the report file use the snpguest tool to generate.. I spun up an AWS EC2 SNP instance and installed https://github.com/virtee/snpguest to get the report.

PR #788, opened 2025-05-14
docker-compose: add env var for RUST_LOG

fitzthum:
This makes it easier to turn on debug printing for Trustee. To enable, you can do something like

`docker-compose --env-file debug.env up`

Where debug.env has

`RUST_LOG=debug`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,8 @@ services:
       context: .
       dockerfile: kbs/docker/coco-as-grpc/Dockerfile
     image: ghcr.io/confidential-containers/staged-images/kbs-grpc-as:latest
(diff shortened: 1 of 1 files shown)


Xynnn007:
the change is directly on the docker-compose file, not consistent with PT message?

fitzthum:
> the change is directly on the docker-compose file, not consistent with PR message?

I think you need both. The changes in the docker compose file tell docker compose which env vars to propagate to the containers. Then you actually set the values of the env vars with an env file.

Xynnn007 (approved):

mythi (approved):
this is useful, thanks!

PR #790, opened 2025-05-15
Fix a tiny typo in kbs document

IsaacYangSLA:
The typo is inside document and not code-related.  However, during setting up the KBS with HTTPS by following the document, it caused small issues.

This PR won't affect any executable files.


fitzthum (approved):
Good catch. Thanks @IsaacYangSLA

PR #793, opened 2025-05-16
docker-compose: automatically generate admin keys

fitzthum:
We expect the user to generate the admin keypair for the KBS (and keyprovider), but this really isn't necessary. This commit does maintain the existing workflow, so users can continue to setup their keys if they would like to and our existing guides are still valid.

We do have to make a few small tweaks to setup the keys from inside the KBS container. First, we have to install openssl inside the container. The staged container images don't have openssl in them yet (until after this PR), so let's hold off on updating the docs for now. I will make a second PR that tests out the staged images and updates the docs. For now, this will only work if you build the images yourself.

Next, we have to change how we mount the keys into the container. Currently, we mount the two key files individually. Docker compose will complain if they don't exist. To get around this we can mount the entire config directory. This does copy a few files into the container that we don't need, but if we want to keep the paths of the keys the same on the host (to not break the current flow), I don't think there's any way around this.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ services:
     command: [
         "/usr/local/bin/kbs",
         "--config-file",
-        "/etc/kbs-config.toml",
+        "/opt/confidential-containers/kbs/user-keys/docker-compose/kbs-config.toml",
       ]
     restart: always # keep the server running
     ports:
@@ -17,10 +17,12 @@ services:
     volumes:
       - ./kbs/data/kbs-storage:/opt/confidential-containers/kbs/repository:rw
       - ./kbs/data/nebula-ca:/opt/confidential-containers/kbs/nebula-ca:rw
-      - ./kbs/config/public.pub:/opt/confidential-containers/kbs/user-keys/public.pub
-      - ./kbs/config/docker-compose/kbs-config.toml:/etc/kbs-config.toml
+      - ./kbs/config:/opt/confidential-containers/kbs/user-keys
     depends_on:
-    - as
+      as:
+        condition: service_started
+      setup:
+        condition: service_completed_successfully
 
   as:
     build:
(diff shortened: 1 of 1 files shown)


Xynnn007 (commented):
I like the aim to help users who forget/do not know to set the key pair. However the way in this PR will add new dependency to KBS image and have some extra commands to KBS. What about handling this via some logic like an init-container, e.g.
```
services:
  kbs:
    image: <kbs>
    command: ...
    restart: always # keep the server running
    volumes:
      - ...
    depends_on:
      - prepare-privatekey

  prepare-privatekey:
    image: debian
    volumes:
      - ./kbs/config:/opt/confidential-containers/kbs/user-keys/
    working_dir: /opt/confidential-containers/kbs/user-keys/
    command: sh -c " if [ ! -f private.key ]; then apt update -y && apt install -y openssl && /usr/bin/openssl genpkey -algorithm ed25519 > private.key && /usr/bin/openssl pkey -in private.key -pubout -out public.pub; fi "

```

mythi:
> We expect the user to generate the admin keypair for the KBS (and keyprovider), but this really isn't necessary. This commit does maintain the existing workflow, so users can continue to setup their keys if they would like to and our existing guides are still valid.

The minor caveat in this PR is that users who prefer to keep the existing flow get exposed to this new flow. I was wondering if it was better to have an opt-in compose file for this new functionality `docker compose up -d -f docker-compose-autokeys.yml` which pulls the original using `include` and sets the "initcontainer" on top.

fitzthum:
> The minor caveat in this PR is that users who prefer to keep the existing flow get exposed to this new flow.

The keys are only created if they don't already exist so I don't think it changes the current flow.

fitzthum:
Ok, I updated this PR to use an additional setup container. It might be overkill for making a few keys, but it can come in handy if we have more complex setup later.

Xynnn007 (approved):
Good. CI passes

Issue #787, opened 2025-05-14
TDX CCEL updates


mythi:
As of today, `tdx-verifier` parses and adds the following CC eventlog claims:

```json
         "ccel": {
            "cmdline": ""
            "kernel": ""
            "initrd": ""
          },
```

whenever found. The search is aware of OVMF and TDshim direct boot flows but fails to find some entries with grub2/systemd-boot/UKI. Moreover, the list is only a subset of the full log which may contain other relevant measurement events to have policy checks written.

The proposal is to change `"ccel: {}" claims to include parsed events, roughly as follow (just one event as an example):

```json
     {
        "details": {
          "tagged_event_data": "LOADED_IMAGE::LoadOptions",
          "tagged_event_data_size": 26,
          "tagged_event_id": 2403017453
        },
        "digests": [
          {
            "alg": "SHA-384",
            "digest": ""
          }
        ],
        "event": "bas64encoded event",
        "index": 3,
        "type": 6,
        "type_name": "EV_EVENT_TAG"
      },
```

fitzthum:
This seems like a good idea. The interface to the policy will be a little more complex with people having to parse this thing, but it's more scalable. Let's just be sure to update the default policy to consume this.

Xynnn007:
Overall I +1 for this.

Some furthur thoughts about policy, which might not be related to this issue.

We now flatten claims, including CCEL, primarily to be able to use it as a single-layer key-value pair in the default policy to compare with the value in RVPS.

I like the nesting method mentioned in this issue, which may imply the need to allow regorus to support some custom writing, correcting the previous use of all RVPS reference values ​​as input - let the statement in the policy prompt to find the RVPS query reference value. In this way, we can completely abandon flattening claims and freely allow nested formats

like the following code
```rego
kernel_ok = if { 
       input["tdx"]["ccel"][0]["details"]["LOADED_IMAGE::LoadOptions"]
       input["tdx"]["ccel"][0]["digests"][0]["digest"] = query_rvps("reference-value-key-of-kernel")
```

PR #802, opened 2025-05-23
deps/eventlog: added TCG2 events eventlog parsing library + policy usage

pawelpros:
This PR removes usage of eventlog-rs and introduce internal ccel library which adds new uefi_event_logs claims accessible through policies for TDX

Fixes: #787

### DIFF
--- a/attestation-service/docs/parsed_claims.md
+++ b/attestation-service/docs/parsed_claims.md
@@ -20,8 +20,23 @@ All platforms will by default have two fixed claims:
 ## Intel TDX
 
 The following fields are optional. Whether they appear depends on whether there is CCEL.
-- `tdx.ccel.kernel`: the measurement of the kernel memory slice in hex
-- `tdx.ccel.kernel_parameters.*`: different kernel parameter items. For example `console=hvc0` will be parsed into a claim `"tdx.ccel.kernel_parameters.console": "hvc0"`. `rw` will be parsed into a claim `"tdx.ccel.kernel_parameters.rw": null`.
+- `tdx.uefi_event_logs` list of objects parsed from ccel log file. Each event can be accessed using fields described below.
+
+Uefi event log entry contains below fields:
+S`tdx.uefi_event_logs[0].index`: Measurement registry index.
+`tdx.uefi_event_logs[0].event_type`: Name of the measurement event from [TCG PC Client Platform Firmware Profile Specification Section 10.4.1](https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v22_02dec2020.pdf).
+`tdx.uefi_event_logs[0].digests[0].alg`: Hash algorithm (`RSA`, `TDES`, `SHA-1`, `SHA-256`, `SHA-384`, `SHA-512`).
+`tdx.uefi_event_logs[0].digests[0].digest`: Digest value calculated for hash defined in previous field.
+`tdx.uefi_event_logs[0].event`: Base64 encoded raw event data.
+`tdx.uefi_event_logs[0].details`: List of attributes parsed from event data. All details properties are optional.
+`tdx.uefi_event_logs[0].details.string`: Parsed UTF-8 value.
+`tdx.uefi_event_logs[0].details.unicode_name`: Parsed Unicode name of the measurement event.
+`tdx.uefi_event_logs[0].details.unicode_name_length`: Unicode name length of the measurement event.
+`tdx.uefi_event_logs[0].details.variable_data`: Base64 encoded event variable data.
+`tdx.uefi_event_logs[0].details.variable_data_length`: Length of the variable data.
+`tdx.uefi_event_logs[0].details.variable_name`: Variable name.
+`tdx.uefi_event_logs[0].details.device_paths`: List of parsed device paths.
+`tdx.uefi_event_logs[0].details.data`: Additional information processed from the event.
 
 The following fields always exist.
 - `tdx.quote.header.version`: The quote format version. Now supports 4 and 5.
--- a/attestation-service/src/token/ear_default_policy_cpu.rego
+++ b/attestation-service/src/token/ear_default_policy_cpu.rego
@@ -89,7 +89,15 @@ else := 3 if {
 ##### TDX
 executables := 3 if {
 	# Check the kernel, initrd, and cmdline (including dmverity parameters) measurements
-	# TODO: add individual CCEL measurements from input.tdx.ccel instead
+	input.tdx.quote.body.rtmr_1 in data.reference.rtmr_1
+	input.tdx.quote.body.rtmr_2 in data.reference.rtmr_2
+	tdx_uefi_event_tdvfkernel_ok
+	tdx_uefi_event_tdvfkernelparams_ok
+}
+
+# Support for Grub boot used by GKE
+else := 4 if {
+	# Check the kernel, initrd, and cmdline (including dmverity parameters) measurements
 	input.tdx.quote.body.rtmr_1 in data.reference.rtmr_1
 	input.tdx.quote.body.rtmr_2 in data.reference.rtmr_2
 }
@@ -103,12 +111,12 @@ hardware := 2 if {
 	input.tdx.quote.body.mr_seam in data.reference.mr_seam
 	input.tdx.quote.body.tcb_svn in data.reference.tcb_svn
 	input.tdx.quote.body.mr_td in data.reference.mr_td
+
 	# Check TCB status
-	# input.tdx.tcb_status == "OK"
+	input.tdx.tcb_status == "OK"
 
 	# Check collateral expiration status
-	# input.tdx.collateral_expiration_status == "0"
-
+	input.tdx.collateral_expiration_status == "0"
 	# Check against allowed advisory ids
 	# allowed_advisory_ids := {"INTEL-SA-00837"}
 	# attester_advisory_ids := {id | id := input.attester_advisory_ids[_]}
@@ -125,6 +133,24 @@ configuration := 2 if {
 	# Check the TD has the expected attributes (e.g., debug not enabled) and features.
 	input.tdx.td_attributes.debug == false
 	input.tdx.quote.body.xfam in data.reference.xfam
+}
+
+tdx_uefi_event_tdvfkernel_ok if {
+	event := input.tdx.uefi_event_logs[_]
+	event.type_name == "EV_EFI_BOOT_SERVICES_APPLICATION"
+	"File(kernel)" in event.details.device_paths
+
+	digest := event.digests[_]
+	digest.digest == data.reference.tdvfkernel
+}
+
+tdx_uefi_event_tdvfkernelparams_ok if {
+	event := input.tdx.uefi_event_logs[_]
+	event.type_name == "EV_EVENT_TAG"
+	event.details.string == "LOADED_IMAGE::LoadOptions"
+
+	digest := event.digests[_]
+	digest.digest == data.reference.tdvfkernelparams
 }
 
 ##### Azure vTPM SNP


attestation-service/docs/parsed_claims.md
```diff
@@ -49,6 +64,14 @@ The following fields always exist.
 - `tdx.quote.size`: Quote body length. Only quote format V5 contains this field.
 - `tdx.quote.body.tee_tcb_svn2`: Array of TEE TCB SVNs (for TD preserving).
 - `tdx.quote.body.mr_servicetd`: If there is one or more bound or pre-bound service TDs, this field is the SHA384 hash of the `TDINFO`s of those service TDs bound. Else, this field is 0.
+- `tdx.tcb_status`: TCB status which can be one of: OK, Min, OutOfDate, OutOfDateConfigurationNeeded, InvalidSignature, Revoked, Unspecified, SoftwareHardeningNeeded, ConfigurationAndSoftwareHardeningNeeded, TdRelaunchAdvised, TdRelaunchAdvisedConfigurationNeeded, Max
```
mythi: let's move doc changes that are not eventlog related to a separate PR.

deps/verifier/src/eventlog/ccel/parser/parsers/boot_services_app.rs
```diff
@@ -0,0 +1,108 @@
...
+    // https://uefi.org/specs/UEFI/2.10/10_Protocols_Device_Path_Protocol.html#generic-device-path-node-structure
+    if device_path.is_empty() {
+        return Ok(result);
+    }
+
+    get_nested_data(device_path, result)
+}
+
+fn print_path(efi_type: u8, efi_sub_type: u8, vendor_data: String) -> String {
+    if efi_type == 1 && efi_sub_type == 1 {
```
mythi: can we get rid of these magic numbers? In general, better doc pointers to UEFI_IMAGE_LOAD_EVENTs in this file would be good.

deps/verifier/src/eventlog/ccel/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+            } else if index == 0 {
+                break;
+            }
+        }
+
+        Ok(CcEventLog { log: event_log })
+    }
+}
+
+fn parse_initial_entry(
```
mythi: add a reference to the TCG spec to explain what these initial entries are

deps/verifier/src/eventlog/ccel/utils.rs
```diff
@@ -0,0 +1,76 @@
...
+
+pub(crate) fn read_u32_le(data: &[u8], index: &mut usize) -> Result<u32, anyhow::Error> {
+    let bytes = get_next_bytes(data, index, size_of::<u32>())?;
+    Ok(LittleEndian::read_u32(bytes))
+}
+
+pub(crate) fn read_u64_le(data: &[u8], index: &mut usize) -> Result<u64, anyhow::Error> {
+    let bytes = get_next_bytes(data, index, size_of::<u64>())?;
+    Ok(LittleEndian::read_u64(bytes))
+}
```
mythi: could we use something like scroll Pread instead of implementing custom ones

deps/verifier/src/eventlog/ccel/rtmr.rs
```diff
@@ -0,0 +1,168 @@
+use crate::eventlog::ccel::tcg_enum::TcgAlgorithm;
+use crate::eventlog::ccel::CcEventLog;
+use anyhow::bail;
+use anyhow::*;
```
mythi: we had issues with the wildcard imports earlier (see #726) so let's not repeat the errors

deps/verifier/src/eventlog/ccel/parser/parsers/ipl.rs
```diff
@@ -0,0 +1,25 @@
...
+
+impl DescriptionParser for EvIplParser {
+    fn parse_description(&self, data: Vec<u8>) -> Result<EventDetails, Error> {
+        if data.len() < 2 {
+            bail!(
+                "Data is too short: expected at least 2 bytes, got {}",
+                data.len()
+            );
+        }
+        let gce_ipl_prefix = [0x2e, 0x00];
```
mythi: what's the meaning of this prefix?

deps/verifier/src/eventlog/ccel/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    let digests;
+    (digests, index) = parse_digests(data, index, digest_size_map)?;
+
+    let event_desc_size = utils::read_u32_le(data, &mut index)?;
+    let event_desc_raw = data[index..(index + event_desc_size as usize)].to_vec();
+    index += event_desc_size as usize;
+
+    let event = STANDARD.encode(&event_desc_raw);
+    let event_result = event_type.get_parser().parse_description(event_desc_raw)?;
```
mythi: Are we actually parsing event "descriptions"? The TCG spec has descriptions for various events but as far as the parsers are concerned, I believe we are parsing "event data"

mythi (commented):
some initial comments

attestation-service/docs/parsed_claims.md
pawelpros: Moved to https://github.com/confidential-containers/trustee/pull/809

deps/verifier/src/eventlog/ccel/mod.rs
pawelpros: Added description for this initial  event

deps/verifier/src/eventlog/ccel/rtmr.rs
pawelpros: Fixed

deps/verifier/src/eventlog/ccel/mod.rs
pawelpros: Renamed parsers and usage to parsing event_data

deps/verifier/src/eventlog/ccel/parser/parsers/ipl.rs
pawelpros: IPL data payload based on samples and actual GKE result were different -> in case of GKE there is structure in this field (as it's vendor defined) and in for example "CCEL_data_grub_out" it's UTF-8 string as content so to provide backward compatibility I added detecting if IPL is plain utf-8 or structure in UTF-16 prefixed with two bytes

deps/verifier/src/eventlog/ccel/parser/parsers/boot_services_app.rs
pawelpros: Added comment + TODO to this function

deps/verifier/src/eventlog/ccel/utils.rs
pawelpros: Good idea - that helped to remove utils class entirely

deps/verifier/src/eventlog/ccel/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+            }
+        }
+
+        Ok(CcEventLog { log: event_log })
+    }
+}
+
+/// Parse the first event in the event log which is the informational event
+/// TCG_EfiSpecIdEvent (See Section 10.4.5.1 Specification ID Version Event)
+/// Docs: https://trustedcomputinggroup.org/wp-content/uploads/TCG-PC-Client-Platform-Firmware-Profile-Version-1.06-Revision-52_pub-3.pdf
```
Xynnn007: This will hep to turn it to a link

```suggestion
/// Docs: <https://trustedcomputinggroup.org/wp-content/uploads/TCG-PC-Client-Platform-Firmware-Profile-Version-1.06-Revision-52_pub-3.pdf>
```

deps/verifier/src/eventlog/ccel/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+            .ok_or_else(|| anyhow!("Overflow in calculating digest_data range"))?;
+        let digest_data = data
+            .get(start..end)
+            .ok_or_else(|| anyhow!("Out of bounds while reading digest_data"))?;
+        *index = end;
+        let actual_size = parse_digest_sizes(digest_data, digest_size_map)?;
+        if actual_size != event_data_size as usize {
+            return Err(anyhow!(
+                "Unexpected data size consumed for detecting digests"
+            ));
```
Xynnn007:
```suggestion
            bail!("Unexpected data size consumed for detecting digests");
```

deps/verifier/src/eventlog/ccel/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+}
+
+fn parse_eventlog_entry(
+    data: &[u8],
+    mut index: usize,
+    digest_size_map: &mut HashMap<TcgAlgorithm, u16>,
+) -> Result<(Option<EventlogEntry>, usize)> {
+    let stop_flag = data
+        .gread_with::<u64>(&mut index, LE)
+        .map_err(|err| anyhow!(err.to_string()))?;
```
Xynnn007: `anyhow` crate could help make it easier for error handling. If we want to directly return the error, we could use
```rust
.gread_with::<u64>(&mut index, LE)?;
```

Or, if we want to add some more context about the error, we could use
```rust
.gread_with::<u64>(&mut index, LE)
.context("failed to parse stop flag")?;
```

deps/verifier/src/eventlog/ccel/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+/// TCG_EfiSpecIdEvent (See Section 10.4.5.1 Specification ID Version Event)
+/// Docs: https://trustedcomputinggroup.org/wp-content/uploads/TCG-PC-Client-Platform-Firmware-Profile-Version-1.06-Revision-52_pub-3.pdf
+fn parse_initial_entry(
+    data: &[u8],
+    index: &mut usize,
+    digest_size_map: &mut HashMap<TcgAlgorithm, u16>,
+) -> Result<()> {
+    let _pcr_index = data
+        .gread_with::<u32>(index, LE)
+        .map_err(|_| anyhow!("Cannot read PCR index"))?;
```
Xynnn007: nit: it would help if you do not like `map_err`
```
.gread_with::<u32>(index, LE)
.context("Cannot read PCR index")?;
```

deps/verifier/src/eventlog/ccel/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+    let event_type = TcgEventType::try_from(event_type_num)
+        .map_err(|_| anyhow!("Unknown event type detected: {:#x}", event_type_num))?;
+
+    let digests;
+    (digests, index) = parse_digests(data, index, digest_size_map)?;
+
+    let event_data_size = data
+        .gread_with::<u32>(&mut index, LE)
+        .map_err(|_| anyhow!("Cannot read event data size"))?;
+    let event_data_raw = data[index..(index + event_data_size as usize)].to_vec();
```
Xynnn007: This is somehow tricky. Let's get the length of data and check if it is `>= index + event_data_size` before we index it. There might be corner cases where the data does not aligns with the length thus causing panic

deps/verifier/src/eventlog/ccel/parser/parsers/boot_services_app.rs
```diff
@@ -0,0 +1,82 @@
...
+use anyhow::{Error, Result};
+use byteorder::{ByteOrder, LittleEndian};
+use scroll::Pread;
+
+pub struct EvBootServicesAppParser;
+
+impl EventDataParser for EvBootServicesAppParser {
+    fn parse(&self, data: Vec<u8>) -> Result<EventDetails, Error> {
+        // ImageLocationInMemory + ImageLengthInMemory + ImageLinkTimeAddress (24) and 8 for length
+        let length_of_device_path = u64::from_le_bytes(data[24..32].try_into()?);
```
Xynnn007: ditto. Let's make a condition judgement of the data length before indexing it.

deps/verifier/src/eventlog/ccel/parser/parsers/boot_services_app.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+fn get_nested_data(
+    device_path_bytes: &[u8],
+    mut result: EventDetails,
+) -> Result<EventDetails, Error> {
+    let efi_type = device_path_bytes[0];
+    let efi_sub_type = device_path_bytes[1];
+    let efi_length = u16::from_le_bytes(device_path_bytes[2..4].try_into()?);
```
Xynnn007: ditto. Let's make a condition judgement of the data length before indexing it.

deps/verifier/src/eventlog/ccel/parser/parsers/boot_services_app.rs
```diff
@@ -0,0 +1,82 @@
+use crate::eventlog::ccel::parser::EventDataParser;
+use crate::eventlog::ccel::EventDetails;
+use anyhow::{Error, Result};
+use byteorder::{ByteOrder, LittleEndian};
+use scroll::Pread;
+
+pub struct EvBootServicesAppParser;
```
Xynnn007: Could we give eaech Parser an inline doc that points to the definition of the document?

deps/verifier/src/eventlog/ccel/parser/parsers/platform_config_flags.rs
```diff
@@ -0,0 +1,124 @@
...
+        let cleaned = raw.chars().filter(|c| !c.is_control()).collect::<String>();
+        result_data.string = Some(cleaned);
+        result_data.data = Some(map);
+        Ok(result_data)
+    }
+}
+
+fn parse_kernel_parameters(kernel_parameters: String) -> Vec<String> {
+    let parameters = kernel_parameters
+        .split(&[' ', '\n', '\r', '\0', '\u{10}'])
```
Xynnn007: Is there any reason that we do not use `split_whitespace()` here?

deps/verifier/src/eventlog/ccel/parser/parsers/platform_config_flags.rs
```diff
@@ -0,0 +1,124 @@
...
+
+            let it = item.split_once('=');
+
+            match it {
+                Some((k, v)) => {
+                    let detail = EventDetail {
+                        name: k.into(),
+                        value: Some(v.into()),
+                    };
+                    let json = serde_json::to_string(&detail).expect("Failed to encode json");
```
Xynnn007: The `expect` here would cause panic. If we want to handle error in the Iterator, there is an official document https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect

deps/verifier/src/eventlog/ccel/parser/parsers/platform_config_flags.rs
```diff
@@ -0,0 +1,124 @@
...
+    fn test_parse_kernel_parameters(#[case] params: String, #[case] result: Vec<String>) {
+        let msg = format!("test: params: {:?}, result: {result:?}", &params);
+
+        let actual_result = parse_kernel_parameters(params);
+
+        let msg = format!("{msg}: actual result: {actual_result:?}");
+
+        if std::env::var("DEBUG").is_ok() {
+            println!("DEBUG: {msg}");
+        }
```
Xynnn007: We usually do not do this. If we want debug logs, `debug!("...")` macro is preferred. And this macro is controlled by an built-in env `RUST_LOG` which could be `debug/info/warn`, etc.

Xynnn007 (commented):
@pawelpros Thanks for this work! I have just revewed part of the logic, some suggestions before diving into it

1. Could we move this submodule into a separate crate under `deps`? 
Because this logic has the potential to become a separate library to use later
2. Please add some more inline documents about the `*Parser` definitions about their definitions to the UEFI spec (if any)
3. Let's not add `test_data/*_out.json` to the PR as they are generated by unit tests.

deps/verifier/src/eventlog/ccel/mod.rs
pawelpros: Done

deps/verifier/src/eventlog/ccel/mod.rs
pawelpros: Done (two findings)

deps/verifier/src/eventlog/ccel/parser/parsers/platform_config_flags.rs
pawelpros: I moved old test here -> removed this few lines not needed for tests

deps/verifier/src/eventlog/ccel/mod.rs
pawelpros: The problem is that the scroll lib returns scroll::Error not anyhow::error and without mapping it cannot be casted (from my understanding)

deps/verifier/src/eventlog/ccel/parser/parsers/platform_config_flags.rs
pawelpros: To avoid this scenario:

```
 lhs:
        "td_payload_info"
rhs:
        "td_payload_info\u0000\u0000"
```

And also with suggested split_whitespace one of test fails `a=b\rc=d\nname_only\0e=foo`

deps/verifier/src/eventlog/ccel/parser/parsers/boot_services_app.rs
pawelpros: Done

deps/verifier/src/eventlog/ccel/parser/parsers/platform_config_flags.rs
pawelpros: Done

deps/verifier/src/eventlog/ccel/mod.rs
pawelpros: As from other comment -> scroll is not automatically mapped to anyhow Error

deps/verifier/src/eventlog/ccel/mod.rs
pawelpros: Done

deps/eventlog/src/ccel/parser/parsers/platform_config_flags.rs
```diff
@@ -0,0 +1,135 @@
...
+pub struct EvPlatformConfigFlagsParser;
+
+#[derive(Debug, Clone, Serialize)]
+pub struct EventDetail {
+    pub name: String,
+    pub value: Option<String>,
+}
+
+/// Parser for EV_PLATFORM_CONFIG_FLAGS.
+/// This event field contents are manufacturer implementation-specific.
```
mythi: is there a way to check what is "manufacturer" is? the implementation below assumes there's always the kernel cmdline.

deps/verifier/src/eventlog/ccel/parser/parsers/ipl.rs
pawelpros: @mythi after some refactor this comment can be closed as sovled?

deps/eventlog/src/ccel/parser/parsers/platform_config_flags.rs
pawelpros: As far as i know it was previously treated as happy structure which was split against white space and other separators -> I added couple tests to verify output which is packed as array of key-value. 

We can assume that data will be UTF-8 concatenated field with various separators. 
Logic between old implementation (existing in trustee) and new one is quite the same in this aspect so no regression

deps/eventlog/src/ccel/rtmr.rs
```diff
@@ -0,0 +1,168 @@
...
+            let hash = accumulate_hash(
+                entry.digests[0].alg,
+                mr_value.clone().to_vec(),
+                digest.as_slice(),
+            )?;
+
+            mr_value.copy_from_slice(&hash);
+        }
+
+        let mut data: [Vec<u8>; 4] = [
```
mythi: we cannot assume it's 4 anymore. See the latest UEFI spec for other vendor details. with that, would it make sense to move the replay logic to the caller side which knows the measurement registry count and the registry width

pawelpros: I refactored RTMR to not hardcode "4" it's now vector of vectors and depends on rtmrs from quote -> if there will be 10 rtmrs -> it will compare them with 10 rtmrs from ccel. If there will be less than 4 rmtrs it will use placeholder for missing ones to compare

deps/eventlog/src/ccel/rtmr.rs
```diff
@@ -0,0 +1,286 @@
...
+
+        Ok(())
+    }
+}
+
+impl TryFrom<CcEventLog> for Rtmr {
+    type Error = anyhow::Error;
+
+    fn try_from(data: CcEventLog) -> anyhow::Result<Self> {
+        let mut result: HashMap<u32, [u8; RTMR_LENGTH_BY_BYTES]> = HashMap::new();
```
mythi: my suggestion is that `CcEventLog` would have something like `replay(index: u32, alg: TcgAlgorithm) -> Result<Vec<u8>>`. This would allow the caller to choose which register and which hash algo is desired. We don't need sha384 specific `RTMR_LENGTH_BY_BYTES`. Errors would happen if no requested index/digest is found.

pawelpros: Refactored RTMR - please take a look

deps/eventlog/src/ccel/tcg_enum.rs
```diff
@@ -0,0 +1,182 @@
...
+            Self::EvEfiAction => Box::new(EvSimpleParser),
+            Self::EvEfiPlatformFirmwareBlob2 => Box::new(SimpleStringParser),
+            Self::EvEfiHandoffTables2 => Box::new(SimpleStringParser),
+            Self::EvEfiVariableBoot2 => Box::new(EvEfiVariableParser),
+            Self::EvEfiVariableAuthority => Box::new(EvEfiVariableParser),
+            _ => Box::new(EvBlankParser),
+        }
+    }
+
+    pub(crate) fn format_name(&self) -> String {
```
mythi: could use a docstring here.

deps/eventlog/src/ccel/parser/parsers/boot_services_app.rs
```diff
@@ -0,0 +1,128 @@
...
+    if device_path.is_empty() {
+        return Ok(result);
+    }
+
+    get_nested_paths(device_path, result)
+}
+
+/// Initial approach for parsing text device nodes described in specification:
+/// https://uefi.org/specs/UEFI/2.10/10_Protocols_Device_Path_Protocol.html#text-device-node-reference
+/// TODO Implement full support for data paths available for device nodes
```
mythi: add comments for type/subtype 1/1 and 4/4 below.

deps/eventlog/src/ccel/parser/mod.rs
```diff
@@ -0,0 +1,9 @@
+use crate::EventDetails;
+use anyhow::Result;
+
+pub mod parsers;
```
mythi: do we need all these layers?

mythi (commented):
It's starting to look good to me.

I'd propose to rephrase the commit messages a bit. The purpose of this work is to improve CoCo attestation tokens to include detailed eventlog events information so

ff4fc92 message body would be

"Add support for confidential computing (TCG2 events) eventlog processing with event type specific parsers and JSON serializable events. The new functionality also includes a replay+integrity check method which is vendor/verifier agnostic."

Also, I think 'kbs: ' prefix in the commits is slightly misleading. 'deps/eventlog:', 'tdx-verifier: ', and 'as:'

deps/eventlog/src/ccel/parser/mod.rs
pawelpros: Hmm it can be flattened - I'll do so

deps/eventlog/src/ccel/parser/parsers/boot_services_app.rs
pawelpros: Done

deps/eventlog/src/ccel/tcg_enum.rs
pawelpros: Done

deps/eventlog/src/ccel/tcg_enum.rs
```diff
@@ -0,0 +1,199 @@
...
+            Self::EvEfiPlatformFirmwareBlob2 => Box::new(SimpleStringParser),
+            Self::EvEfiHandoffTables2 => Box::new(SimpleStringParser),
+            Self::EvEfiVariableBoot2 => Box::new(EvEfiVariableParser),
+            Self::EvEfiVariableAuthority => Box::new(EvEfiVariableParser),
+            _ => Box::new(EvBlankParser),
+        }
+    }
+
+    /// Formatter for TcgEventType which converts enum name to uppercased snake cased name
+    /// Example: EvEventTag -> EV_EVENT_TAG
```
mythi: try `strum` crate here?

attestation-service/docs/parsed_claims.md
```diff
@@ -20,8 +20,23 @@ All platforms will by default have two fixed claims:
 ## Intel TDX
 
 The following fields are optional. Whether they appear depends on whether there is CCEL.
-- `tdx.ccel.kernel`: the measurement of the kernel memory slice in hex
-- `tdx.ccel.kernel_parameters.*`: different kernel parameter items. For example `console=hvc0` will be parsed into a claim `"tdx.ccel.kernel_parameters.console": "hvc0"`. `rw` will be parsed into a claim `"tdx.ccel.kernel_parameters.rw": null`.
+- `tdx.uefi_event_logs` list of objects parsed from ccel log file. Each event can be accessed using fields described below.
+
+Uefi event log entry contains below fields:
+`tdx.uefi_event_logs[0].index`: PCR index.
```
Xynnn007: Do we need to note that the PCR index here can also be [CC Event Log Measurement Register Index](https://uefi.org/specs/UEFI/2.11/38_Confidential_Computing.html)

attestation-service/docs/parsed_claims.md
```diff
@@ -20,8 +20,23 @@ All platforms will by default have two fixed claims:
...
+`tdx.uefi_event_logs[0].event_type`: Name of the measurement event from TCG PC Client Platform Firmware Profile Specification.
```
Xynnn007: Could we add a link to 
```suggestion
`tdx.uefi_event_logs[0].event_type`: Name of the measurement event from [TCG PC Client Platform Firmware Profile Specification Section 10.4.1](https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v22_02dec2020.pdf).
```

attestation-service/docs/parsed_claims.md
```diff
@@ -20,8 +20,23 @@ All platforms will by default have two fixed claims:
...
+`tdx.uefi_event_logs[0].digests[0].alg`: Hash algorithm (RSA, TDES, SHA-1, SHA-256, SHA-384, SHA-512).
```
Xynnn007:
```suggestion
`tdx.uefi_event_logs[0].digests[0].alg`: Hash algorithm (`RSA`, `TDES`, `SHA-1`, `SHA-256`, `SHA-384`, `SHA-512`).
```

deps/eventlog/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
```
Xynnn007, line 1: Hi, we prefer to make the commit message shorter for each line (up to 72 chars), like [this doc](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#git-commit-message-structure) suggests.

deps/eventlog/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+name = "eventlog"
+description = "Parser for TCG Eventlog"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+log.workspace = true
+byteorder = "1.5.0"
```
Xynnn007: We have imported `byteorder` in [verifier](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/Cargo.toml#L27) crate. Could you help to move it to the root Cargo.toml and change both verifier and here `byteorder.workspace = true`? Thanks!

deps/eventlog/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+hex.workspace = true
+sha2.workspace = true
+anyhow.workspace = true
+base64.workspace = true
```
Xynnn007, line 9: one nit, we can reorder the crates in alphabet order for better maintaince.

deps/eventlog/src/ccel/boot_services_app.rs
```diff
@@ -0,0 +1,130 @@
...
+pub struct EvBootServicesAppParser;
+
+/// Parser for EV_IPL event EV_EFI_BOOT_SERVICES_APPLICATION
+/// UEFI_IMAGE_LOAD_EVENT {
+///     EFI_PHYSICAL_ADDRESS ImageLocationInMemory; // PE/COFF image
+///     UINT64 ImageLengthInMemory;
+///     UINT64 ImageLinkTimeAddress;
+///     UINT64 LengthOfDevicePath;
+///     BYTE DevicePath[LengthOfDevicePath]; // See UEFI spec Section EFI Device Path Protocol
+/// };
```
Xynnn007: We can add ``` in scenarios such as structure declarations in inline comments to ensure the format of document rendering. The render is like the content inside [the grey box](https://docs.rs/prometheus/0.14.0/prometheus/#basic-example)

/// \`\`\`
/// UEFI_IMAGE_LOAD_EVENT {
/// ...
/// }
/// \`\`\`
///

deps/eventlog/src/ccel/boot_services_app.rs
```diff
@@ -0,0 +1,130 @@
...
+/// UEFI spec: <https://uefi.org/specs/UEFI/2.10/10_Protocols_Device_Path_Protocol.html#generic-device-path-node-structure>
```
Xynnn007: Let's add the original definition of `UEFI_IMAGE_LOAD_EVENT` in this link, i.e. section 10.2.3 of https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v22_02dec2020.pdf

deps/eventlog/src/ccel/boot_services_app.rs
```diff
@@ -0,0 +1,130 @@
+use super::{EventDataParser, EventDetails};
+use anyhow::{bail, Result};
+use byteorder::{ByteOrder, LittleEndian};
+use scroll::Pread;
+
+pub struct EvBootServicesAppParser;
+
+/// Parser for EV_IPL event EV_EFI_BOOT_SERVICES_APPLICATION
+/// UEFI_IMAGE_LOAD_EVENT {
+///     EFI_PHYSICAL_ADDRESS ImageLocationInMemory; // PE/COFF image
```
Xynnn007:
```suggestion
///     UEFI_PHYSICAL_ADDRESS ImageLocationInMemory; // PE/COFF image
```

deps/eventlog/src/ccel/boot_services_app.rs
```diff
@@ -0,0 +1,130 @@
...
+///     UINT64 ImageLengthInMemory;
+///     UINT64 ImageLinkTimeAddress;
+///     UINT64 LengthOfDevicePath;
+///     BYTE DevicePath[LengthOfDevicePath]; // See UEFI spec Section EFI Device Path Protocol
+/// };
+/// UEFI spec: <https://uefi.org/specs/UEFI/2.10/10_Protocols_Device_Path_Protocol.html#generic-device-path-node-structure>
+impl EventDataParser for EvBootServicesAppParser {
+    fn parse(&self, data: Vec<u8>) -> Result<EventDetails> {
+        let dev_path_offset_start = 24;
```
Xynnn007: Probably we coud use `size_of::<u64>() * 3` here

deps/eventlog/src/ccel/boot_services_app.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        // Calculate the start of the device path and ensure data length
+        let dev_path_start = dev_path_offset_end;
+        let dev_path_end = dev_path_start + (dev_path_len as usize - size_of::<u32>());
+
+        if data.len() <= dev_path_end {
+            return Ok(EventDetails::empty());
+        }
+
```
Xynnn007, line 45: Let's add a condition judgement like (some better language if needed)
```
if data.len() < dev_path_end {
    bail!("data is shorter than claimed by dev_path_offset")
}
```

deps/eventlog/src/ccel/boot_services_app.rs
```diff
@@ -0,0 +1,130 @@
...
+        let device_path_bytes = &data[dev_path_start..dev_path_end];
+        let result_data = EventDetails::empty();
+        get_nested_paths(device_path_bytes, result_data)
+    }
+}
+
+fn get_nested_paths(device_path_bytes: &[u8], mut result: EventDetails) -> Result<EventDetails> {
```
Xynnn007, line 58: There is a corner case where `dev_path_len` is not `0` but is `1`. The functon `get_nested_paths` implies that `dev_path_len` is not shorter than 4. We need to explicitly include the condition judgement before call this function. More better  is to add some notes ahead of the function.

deps/eventlog/src/ccel/boot_services_app.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    get_nested_paths(device_path, result)
+}
+
+/// Initial approach for parsing text device nodes described in specification:
+/// <https://uefi.org/specs/UEFI/2.10/10_Protocols_Device_Path_Protocol.html#text-device-node-reference>
+/// TODO Implement full support for data paths available for device nodes
+fn print_path(efi_type: u8, efi_sub_type: u8, vendor_data: &[u8]) -> String {
+    // Type: 1 (Hardware Device Path) / SubType: 1 (PCI)
+    if efi_type == 1 && efi_sub_type == 1 {
```
Xynnn007: Could we have an enum here for `EfiType` and `EfiSubType`? It will do great help for maintainance and future work

deps/eventlog/src/ccel/boot_services_app.rs
```diff
@@ -0,0 +1,130 @@
...
+        let func_num = vendor_data.pread::<u8>(0).unwrap();
```
Xynnn007: Let's avoid `unwrap()` here by using
```rust
let func_num = vendor_data.pread::<u8>(0)?;
```
or
```
let func_num = vendor_data.pread::<u8>(0).context("failed to read func num")?;
```

deps/eventlog/src/ccel/boot_services_app.rs
```diff
@@ -0,0 +1,130 @@
...
+        let device_num = vendor_data.pread::<u8>(1).unwrap();
```
Xynnn007: Here we also implies that the `vendor_data` is longer than 2 bytes. It might be good to explicitly add the condition judgement

deps/eventlog/src/ccel/efi_variable.rs
```diff
@@ -0,0 +1,136 @@
...
+
+/// Parser for EV_EFI_VARIABLE_AUTHORITY, EV_EFI_VARIABLE_BOOT2, EV_EFI_VARIABLE_BOOT, EV_EFI_VARIABLE_DRIVER_CONFIG
+/// All defined above structures share below structure
+/// UEFI_VARIABLE_DATA {
+///     UEFI_GUID VariableName;
+///     UINT64 UnicodeNameLength;
+///     UINT64 VariableDataLength;
+///     CHAR16 UnicodeName[];
+///     INT8 VariableData[];
+/// }
```
Xynnn007: We can add ``` in scenarios such as structure declarations in inline comments to ensure the format of document rendering. The render is like the content inside [the grey box](https://docs.rs/prometheus/0.14.0/prometheus/#basic-example).

Btw, let's add the note
```
/// Define in section 10.2.6 of <https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v22_02dec2020.pdf>
```

deps/eventlog/src/ccel/efi_variable.rs
```diff
@@ -0,0 +1,136 @@
...
+impl EventDataParser for EvEfiVariableParser {
+    fn parse(&self, data: Vec<u8>) -> Result<EventDetails> {
+        let mut index = 0;
+
+        let guid_size = 16;
```
Xynnn007: This can be a const, probably `pub const GUID_SIZE: usize = 16` in `lib.rs`?

deps/eventlog/src/ccel/efi_variable.rs
```diff
@@ -0,0 +1,136 @@
...
+        let guid = data
```
Xynnn007, line 27: Let's add a condition judgement that `data.len() > guid_size + size_of::<u64>() * 2` before indexing it, as it would cause potential panic.

deps/eventlog/src/ccel/efi_variable.rs
```diff
@@ -0,0 +1,136 @@
...
+        let uname_length: u64 = data
+            .gread_with(&mut index, LE)
+            .map_err(|e| anyhow::anyhow!("Failed to read unicode name length: {:?}", e))?;
+
+        let var_data_length: u64 = data
+            .gread_with(&mut index, LE)
+            .map_err(|e| anyhow::anyhow!("Failed to read variable data length: {:?}", e))?;
+
+        let desc_byte_len = uname_length
+            .checked_mul(2)
```
Xynnn007, line 41: I like the `checked_mul` usage here.

deps/eventlog/src/ccel/efi_variable.rs
```diff
@@ -0,0 +1,136 @@
...
+            .ok_or_else(|| anyhow!("Out of bounds while reading description length"))?
+            as usize;
+
+        let description_bytes = data
+            .get(index..index + desc_byte_len)
```
Xynnn007, line 46: Coo. I noticed the `get` usage here. The [official document](https://doc.rust-lang.org/std/primitive.slice.html#method.get) says 

"""
If given a range, returns the subslice corresponding to that range, or None if out of bounds.
"""


Previously, I have left many reviews that mentioned the judgment of slice length. It seems that it can be replaced by this way of writing, which is simpler.

deps/eventlog/src/ccel/event_tag.rs
```diff
@@ -0,0 +1,65 @@
...
+use scroll::{Pread, LE};
+
+pub struct EvEventTagParser;
+
+/// Parser for EV_EVENT_TAG
+/// TCG_PCClientTaggedEvent{
+///     UINT32 taggedEventID; // SKIPPED
+///     UINT32 taggedEventDataSize;
+///     BYTE taggedEventData[taggedEventDataSize];
+/// }
```
Xynnn007: Let's add ``` here for inline documents as previous commented

deps/eventlog/src/ccel/event_tag.rs
```diff
@@ -0,0 +1,65 @@
...
+impl EventDataParser for EvEventTagParser {
+    fn parse(&self, data: Vec<u8>) -> Result<EventDetails> {
+        let mut index = size_of::<u32>();
+        let data_size = data
```
Xynnn007, line 19: I am not sure if `data.len() < size_of::<u32>()`, whether `gread_with` will panic. If so, we'd better still judge the length.

deps/eventlog/src/ccel/ipl.rs
```diff
@@ -0,0 +1,85 @@
...
+
+/// Parser for EV_IPL event
+/// This event is vendor defined.
+/// For Google Cloud it's prefixed with 2e00
+impl EventDataParser for EvIplParser {
+    fn parse(&self, data: Vec<u8>) -> Result<EventDetails> {
+        if data.len() < 2 {
+            bail!("Data is too short: expected at least 2 bytes");
+        }
+        let gce_ipl_prefix = [0x2e, 0x00];
```
Xynnn007: Could we make this as a `const`?

deps/eventlog/src/ccel/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+pub(crate) use event_tag::EvEventTagParser;
+pub(crate) use ipl::EvIplParser;
+pub(crate) use platform_config_flags::EvPlatformConfigFlagsParser;
+pub(crate) use simple::EvSimpleParser;
+pub(crate) use simple_string::SimpleStringParser;
+
+/// All parser implementations follow structures defined in <https://trustedcomputinggroup.org/wp-content/uploads/TCG-PC-Client-Platform-Firmware-Profile-Version-1.06-Revision-52_pub-3.pdf>
+pub trait EventDataParser: Sync + Send {
+    fn parse(&self, data: Vec<u8>) -> anyhow::Result<EventDetails>;
+}
```
Xynnn007: I see that we do not have another `\n` in this file. Please run `cargo fmt -p eventlog` to format the code.

deps/eventlog/src/ccel/platform_config_flags.rs
```diff
@@ -0,0 +1,134 @@
...
+
+/// Parser for EV_PLATFORM_CONFIG_FLAGS.
+/// This event field contents are manufacturer implementation-specific.
+impl EventDataParser for EvPlatformConfigFlagsParser {
+    fn parse(&self, data: Vec<u8>) -> Result<EventDetails> {
+        let mut result_data = EventDetails::empty();
+        let raw = String::from_utf8(data)
+            .unwrap_or_default()
+            .replace('\u{10}', " ");
+        let map = parse_kernel_parameters(raw.clone())?;
```
Xynnn007: This event is vendor specific. Any reasons/links/docs to reflect that this event is for kernel para?

deps/eventlog/src/ccel/simple_string.rs
```diff
@@ -0,0 +1,81 @@
...
+///     EFI_PHYSICAL_ADDRESS BlobBase;
+///     UINT64 BlobLength;
+/// }
+///
+/// EV_EFI_HANDOFF_TABLES2 contains UEFI_HANDOFF_TABLE_POINTERS2 {
+///     UINT8 TableDescriptionSize; // PARSED
+///     BYTE[TableDescriptionSize] TableDescription; // PARSED
+///     UINT64 NumberOfTables;
+///     UEFI_CONFIGURATION_TABLE TableEntry[NumberOfTables];
+/// }
```
Xynnn007: Let's add ```  around the struct definitions.

deps/eventlog/src/ccel/tcg_enum.rs
```diff
@@ -0,0 +1,202 @@
...
+            _ => Err(()),
+        }
+    }
+}
+
+#[repr(u32)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, EnumString, Display)]
+pub enum TcgEventType {
+    #[strum(serialize = "EV_PREBOOT_CERT")]
+    EvPrebootCert = 0x0,
```
Xynnn007, line 66: nit: let's add an empty line between each enums.

deps/eventlog/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+#[derive(Debug, Clone, Serialize)]
+pub struct ElDigest {
+    pub alg: TcgAlgorithm,
+    #[serde(serialize_with = "serialize_digest_as_hex")]
+    pub digest: Vec<u8>,
+}
+
+#[derive(Debug, Clone, Serialize, PartialEq)]
+pub struct EventDetails {
+    #[serde(skip_serializing_if = "Option::is_none")]
```
Xynnn007, line 41: nit: let's add an empty line between each fields.

deps/eventlog/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+        let mut state = serializer.serialize_struct("EventlogEntry", 6)?;
+        state.serialize_field("details", &self.details)?;
+        state.serialize_field("digests", &self.digests)?;
+        state.serialize_field("event", &self.event)?;
+        state.serialize_field("index", &self.index)?;
+        state.serialize_field("type", &format!("0x{:08X}", self.event_type as u32))?;
+        state.serialize_field("type_name", &self.event_type.to_string())?;
+        state.end()
+    }
+}
```
Xynnn007: Looks like this implementation is mostly like the derived one, except for some fields are not serialized. What about add `#[serde(skip_serializing)]` macro upon those fields not needed to be serialized and get rid of this function?

deps/eventlog/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+    let digest_size = alg.get_digest_size(alg)?;
+    let mut materials = vec![0u8; digest_size];
+
+    for entry in ccel.iter().filter(|e| e.index == index) {
+        for digest in entry.digests.iter().filter(|d| d.alg == alg) {
+            materials = accumulate_hash(alg, materials, &digest.digest)?;
+        }
+    }
+    Ok(materials)
+}
```
Xynnn007: We could optimize the algorithm here. Currently each `ReferenceMeasurement` (assuming that it's of size `m`) will cause a going-through upon the eventlog (of length `m`). Thus the time complexity is `O(mn)`;

We could use a `HashMap<u32, Vec<u8>>` and go through just one time of the CCEL to shorten the time complexity to `O(m)`.

deps/eventlog/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+
+/// Parse the first event in the event log which is the informational event
+/// TCG_EfiSpecIdEvent (See Section 10.4.5.1 Specification ID Version Event)
+/// Docs: <https://trustedcomputinggroup.org/wp-content/uploads/TCG-PC-Client-Platform-Firmware-Profile-Version-1.06-Revision-52_pub-3.pdf>
+fn parse_initial_entry(
+    data: &[u8],
+    index: &mut usize,
+    digest_size_map: &mut HashMap<TcgAlgorithm, u16>,
+) -> Result<()> {
+    let _pcr_index = data
```
Xynnn007, line 228: Similar question about whether `gread_with` would panic if `data` has not enough length. If so, we need a condition judgement here.

Xynnn007 (commented):
Thanks @pawelpros ! I like the fine grained unit test cases. Some nits and also

1. We need to let the CI to cover this crate. Because I found some code snippets that may turn cargo fmt or cargo clippy detection into problems. You only need to add `-p eventlog` to [the three items](https://github.com/confidential-containers/trustee/blob/main/.github/workflows/as-rust.yml#L70-L80).
2. I found that we have an overall struct named `EventDetails` for all different events. For some eventlog formats, we can expect what its content will be like, for example, `UEFI_IMAGE_LOAD_EVENT` contains information such as the address and offset of the image. These contents can be more structured to form something like 
```json
{"offset": 123, "length": 234}
```
So I wonder if (it can be left for future implementation) `EventDetails` can be extended in some way to support more fine-grained parse results (such as changing the `data` field of `EventDetails` to `serde_json::Value`)

attestation-service/docs/parsed_claims.md
pawelpros: included

attestation-service/docs/parsed_claims.md
pawelpros: applied

deps/eventlog/Cargo.toml
pawelpros: Sure - no problem -> applied

deps/eventlog/src/ccel/tcg_enum.rs
pawelpros: applied

deps/eventlog/src/lib.rs
pawelpros: applied

attestation-service/docs/parsed_claims.md
pawelpros: Yes as it's vendor specific -> I'll use proposed name rather than PCR index which is not generic here

deps/eventlog/src/ccel/boot_services_app.rs
pawelpros: I'm working on this on separate branch - quite early stage but should do the trick (I'll send You diff between this PR and that changes after i clean it a little bit :)

deps/eventlog/src/ccel/boot_services_app.rs
pawelpros: nice ! - applied

deps/eventlog/src/ccel/boot_services_app.rs
pawelpros: Added in several places

deps/eventlog/src/ccel/ipl.rs
pawelpros: sure - done

deps/eventlog/src/ccel/efi_variable.rs
pawelpros: moved as const

deps/eventlog/src/ccel/efi_variable.rs
pawelpros: No need - I have tested it (below) with malformed data and won't panic -> in case of guid it will get into else and for greed with I'm mapping potential errors to anyhow so won't panic

deps/eventlog/src/ccel/event_tag.rs
pawelpros: Gread with is checked in tests -> won't panic - anyhow error will be thrown in all cases

deps/eventlog/src/ccel/boot_services_app.rs
pawelpros: Replaced with above 
`if data.len() <= dev_path_end {
            return Ok(EventDetails::empty());
        }`

deps/eventlog/src/ccel/boot_services_app.rs
pawelpros: Added condition and comment

deps/eventlog/src/ccel/boot_services_app.rs
pawelpros: I'll apply it to my separate PR to cover that

deps/eventlog/src/ccel/boot_services_app.rs
pawelpros: same here -> I'll push it to separate PR

deps/eventlog/src/ccel/platform_config_flags.rs
pawelpros: Hmm don't have idea -> this was moved part from existing implementation to follow what currently is on main branch

deps/eventlog/src/lib.rs
pawelpros: It's covered by mapping error from scroll output -> won't panic (tested)

deps/eventlog/src/lib.rs
pawelpros: Ok - refactored

deps/eventlog/src/ccel/boot_services_app.rs
pawelpros: After changes from other PR it will look like:
`fn pci(data: &[u8]) -> Result<String> {
    if data.len() < 2 {
        bail!("PCI data is too short");
    }
    let func_num = data[0];
    let device_num = data[1];
    Ok(format!("Pci({},{})", func_num, device_num))
}`

deps/eventlog/src/lib.rs
pawelpros: Simplified

deps/eventlog/src/ccel/boot_services_app.rs
pawelpros: https://github.com/pawelpros/trustee/pull/2 <- please take a look for structured approach for this element

deps/eventlog/src/ccel/platform_config_flags.rs
mythi: [td-shim spec](https://github.com/confidential-containers/td-shim/blob/main/doc/tdshim_spec.md#guideline) documents the config flags it uses and `parse_kernel_parameters()` focuses on just a subset of the config flags. perhaps we don't need the details here and simply make the string output available if it's possible to decode it. AFAUI, without `parse_kernel_parameters()`, we'd already see something like ` "string": "td_payload_info root=/dev/vda1 console=hvc0 rw"`

deps/eventlog/src/ccel/boot_services_app.rs
Xynnn007: Oh, can you move the unwrap part of the code to this PR? Otherwise, the code that causes the panic version will be merged

deps/eventlog/src/ccel/boot_services_app.rs
Xynnn007: Can you move the unwrap part of the code to this PR? Otherwise, the code that causes the panic version will be merged

Xynnn007:
@pawelpros Thanks. Only the places of unwrap which I think is a blocker. Other things look good to me and we can perfect them later.

deps/eventlog/src/ccel/boot_services_app.rs
pawelpros: @Xynnn007 updated PR -> as I mentioned, I have follow up PR to this but it's too large to merge it with this one

deps/eventlog/src/ccel/platform_config_flags.rs
pawelpros: I'm closing this thread -> If something will be needed we can adjust it later on in my opinion

pawelpros:
@Xynnn007 Let me check `UEFI_HANDOFF_TABLE_POINTERS2` i checked structure with it and it's not supported correctly -> I'll adjust fix in this PR

deps/eventlog/src/ccel/platform_config_flags.rs
Xynnn007: Due to the [document](https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v23_pub.pdf), this is vendor specific and we cannot assume that they are in form `k=v` or `k` splitted by ` `. If we do not have a flag to distinguish the concrete structure, we might want just a string but without this map. We can extend it later, e.g. there is a prefix or something in the string to distinguish which vendor format it is.

pawelpros: Ok - I removed custom implementation in separate commit to have refactored version for history purposes

Xynnn007:
Btw, I want to propose @pawelpros to be a maintainer ofthe repo for the great contribution of this eventlog crate, as well as future maintainance, once after this PR gets merged. 

cc @confidential-containers/trustee-maintainers

Xynnn007:
> @Xynnn007 can you look at https://github.com/pawelpros/trustee/pull/2/files - it's already tested - common device paths added (i tried to find existing lib for that but non of existing implements entire UEFI spec in this aspect)


It is best to make static logic such as DevicePathDispatcher type to processing function into constant code, without the need for runtime new and insert. But we can discuss more about that when a new PR comes.

Issue #801, opened 2025-05-22
Value of configuration for SNP in the default EAR policy


niteeshkd:
### Describe the bug

It seems the following rego statements in [default EAR policy](https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/token/ear_default_policy.rego) are not true and therefore the value of configutaion for SNP is never equal to 2. 

```
input.snp.policy_debug_allowed == 0
input.snp.policy_migrate_ma == 0 
```

Changing the above statements as follows works as expected.
```
input.snp.policy_debug_allowed == "0"
input.snp.policy_migrate_ma == "0" 
```

### How to reproduce

1. Create a SNP pod with the annotation `io.katacontainers.config.hypervisor.kernel_params: "agent.guest_components_rest_api=all"`.
2.  Get attestation token by running `AA_TOKEN=$(oc exec -it ocp-cc-pod -- curl http://127.0.0.1:8006/aa/token?token_type=kbs)` .
3. Parse the token (i.e. `echo $AA_TOKEN | jq ".token" | tr "." "\n" | tr -d \" | base64 -d  | jq` ) and observe the value of  `configuration` under `submods.cpu.ear.trustworthiness-vector`.

### CoCo version information

Trustee v0.12.0

### What TEE are you seeing the problem on

None

### Failing command and relevant log output

```shell

```

fitzthum:
Good find. I would have thought that the evidences supports numeric types. We should figure out where that goes wrong. In the short term let's update the policy tho.

PR #807, opened 2025-05-29
policy: change expected SNP configuration values to Strings

fitzthum:
The SNP verifier returns all of its TCB claims as strings, we could adjust this to return numeric types for certain claims, but the RVPS currently stores all reference values as strings.

In the future we should update the RVPS to work with numeric types as well, but for now let's change the policy to use strings.

Fix https://github.com/confidential-containers/trustee/issues/801


Xynnn007 (approved):
lgtm

PR #816, opened 2025-06-06
KBS: more prometheus metrics

pmores:
Add Prometheus metrics for authentication, attestation and policy evaluation as per discussion at issue #543.

The endpoint output after some successful resource accesses will be similar to
```
$ curl http://127.0.0.1:8080/metrics
# HELP attestation_errors_total Total count of errors during attestation processing
# TYPE attestation_errors_total counter
attestation_errors_total 0
# HELP attestation_requests_total Total count of attestation requests
# TYPE attestation_requests_total counter
attestation_requests_total 3
# HELP attestation_successes_total Total count of attestation successes
# TYPE attestation_successes_total counter
attestation_successes_total{tee_type="sample"} 3
# HELP auth_errors_total Total count of errors during auth processing
# TYPE auth_errors_total counter
auth_errors_total 0
# HELP auth_requests_total Total count of auth requests
# TYPE auth_requests_total counter
auth_requests_total 3
# HELP auth_successes_total Total count of successfully authenticated requests
# TYPE auth_successes_total counter
auth_successes_total 3
# HELP http_request_duration_seconds Distribution of request handling duration
# TYPE http_request_duration_seconds histogram
http_request_duration_seconds_bucket{le="0.0005"} 6
http_request_duration_seconds_bucket{le="0.001"} 7
http_request_duration_seconds_bucket{le="0.005"} 13
http_request_duration_seconds_bucket{le="0.01"} 13
http_request_duration_seconds_bucket{le="0.05"} 13
http_request_duration_seconds_bucket{le="0.5"} 13
http_request_duration_seconds_bucket{le="1"} 13
http_request_duration_seconds_bucket{le="+Inf"} 13
http_request_duration_seconds_sum 0.014657365
http_request_duration_seconds_count 13
# HELP http_request_size_bytes Distribution of request body sizes
# TYPE http_request_size_bytes histogram
http_request_size_bytes_bucket{le="32"} 8
http_request_size_bytes_bucket{le="128"} 17
http_request_size_bytes_bucket{le="512"} 23
http_request_size_bytes_bucket{le="2048"} 26
http_request_size_bytes_bucket{le="8192"} 26
http_request_size_bytes_bucket{le="+Inf"} 26
http_request_size_bytes_sum 7255
http_request_size_bytes_count 26
# HELP http_requests_total Total HTTP requests count
# TYPE http_requests_total counter
http_requests_total 13
# HELP http_response_size_bytes Distribution of response body sizes
# TYPE http_response_size_bytes histogram
http_response_size_bytes_bucket{le="32"} 0
http_response_size_bytes_bucket{le="128"} 0
http_response_size_bytes_bucket{le="512"} 0
http_response_size_bytes_bucket{le="2048"} 0
http_response_size_bytes_bucket{le="8192"} 0
http_response_size_bytes_bucket{le="+Inf"} 0
http_response_size_bytes_sum 0
http_response_size_bytes_count 0
# HELP kbs_policy_approvals_total Total count of requests approved by KBS policy
# TYPE kbs_policy_approvals_total counter
kbs_policy_approvals_total 3
# HELP kbs_policy_errors_total Total count of errors during KBS evaluation
# TYPE kbs_policy_errors_total counter
kbs_policy_errors_total 0
# HELP kbs_policy_evaluations_total Total count of KBS policy evaluations
# TYPE kbs_policy_evaluations_total counter
kbs_policy_evaluations_total 3
# HELP kbs_policy_violations_total Total count of requests denied by KBS policy
# TYPE kbs_policy_violations_total counter
kbs_policy_violations_total 0
# HELP resource_reads_total KBS resource read count
# TYPE resource_reads_total counter
resource_reads_total{resource_path="pvl/test/smazat"} 3
# HELP resource_writes_total KBS resource write count
# TYPE resource_writes_total counter
resource_writes_total{resource_path="pvl/test/smazat"} 1
```

### DIFF
--- a/kbs/src/api_server.rs
+++ b/kbs/src/api_server.rs
@@ -16,7 +16,10 @@ use crate::{
     jwe::jwe,
     plugins::PluginManager,
     policy_engine::PolicyEngine,
-    prometheus::{REQUEST_DURATION, REQUEST_SIZES, REQUEST_TOTAL},
+    prometheus::{
+        KBS_POLICY_APPROVALS, KBS_POLICY_ERRORS, KBS_POLICY_EVALS, KBS_POLICY_VIOLATIONS,
+        REQUEST_DURATION, REQUEST_SIZES, REQUEST_TOTAL,
+    },
     token::TokenVerifier,
     Error, Result,
 };
@@ -280,10 +283,18 @@ pub(crate) async fn api(
 
                 let claim_str = serde_json::to_string(&claims)?;
 
+                KBS_POLICY_EVALS.inc();
                 // TODO: add policy filter support for other plugins
-                if !core.policy_engine.evaluate(&endpoint, &claim_str).await? {
+                if !core
+                    .policy_engine
+                    .evaluate(&endpoint, &claim_str)
+                    .await
+                    .inspect_err(|_| KBS_POLICY_ERRORS.inc())?
+                {
+                    KBS_POLICY_VIOLATIONS.inc();
                     return Err(Error::PolicyDeny);
                 }
+                KBS_POLICY_APPROVALS.inc();
 
                 let response = plugin
                     .handle(&body, query, additional_path, request.method())
--- a/kbs/src/attestation/backend.rs
+++ b/kbs/src/attestation/backend.rs
@@ -20,6 +20,7 @@ use std::collections::HashMap;
 use crate::attestation::session::KBS_SESSION_ID;
 use crate::prometheus::{
     ATTESTATION_ERRORS, ATTESTATION_FAILURES, ATTESTATION_REQUESTS, ATTESTATION_SUCCESSES,
+    AUTH_ERRORS, AUTH_REQUESTS, AUTH_SUCCESSES,
 };
 
 use super::{
@@ -232,9 +233,16 @@ impl AttestationService {
     }
 
     async fn __auth(&self, request: &[u8]) -> anyhow::Result<HttpResponse> {
-        let request: Request = serde_json::from_slice(request).context("deserialize Request")?;
-        let version = Version::parse(&request.version).context("failed to parse KBS version")?;
+        AUTH_REQUESTS.inc();
+
+        let request: Request = serde_json::from_slice(request)
+            .inspect_err(|_| AUTH_ERRORS.inc())
+            .context("deserialize Request")?;
+        let version = Version::parse(&request.version)
+            .inspect_err(|_| AUTH_ERRORS.inc())
+            .context("failed to parse KBS version")?;
         if !VERSION_REQ.matches(&version) {
+            AUTH_ERRORS.inc();
             bail!(
                 "KBS Client Protocol Version Mismatch: expect {} while the request is {}",
                 *VERSION_REQ,
@@ -246,9 +254,12 @@ impl AttestationService {
             .inner
             .generate_challenge(request.tee, request.extra_params.clone())
             .await
+            .inspect_err(|_| AUTH_ERRORS.inc())
             .context("Attestation Service generate challenge failed")?;
 
         let session = SessionStatus::auth(request, self.timeout, challenge);
+
+        AUTH_SUCCESSES.inc();
 
         let response = HttpResponse::Ok()
             .cookie(session.cookie())
--- a/kbs/src/attestation/backend.rs
+++ b/kbs/src/attestation/backend.rs
@@ -18,6 +18,9 @@ use serde_json::json;
 use std::collections::HashMap;
 
 use crate::attestation::session::KBS_SESSION_ID;
+use crate::prometheus::{
+    ATTESTATION_ERRORS, ATTESTATION_FAILURES, ATTESTATION_REQUESTS, ATTESTATION_SUCCESSES,
+};
 
 use super::{
     config::{AttestationConfig, AttestationServiceConfig},
@@ -267,19 +270,26 @@ impl AttestationService {
         attestation: &[u8],
         request: HttpRequest,
     ) -> anyhow::Result<HttpResponse> {
-        let cookie = request.cookie(KBS_SESSION_ID).context("cookie not found")?;
+        ATTESTATION_REQUESTS.inc();
+
+        let cookie = request
+            .cookie(KBS_SESSION_ID)
+            .context("cookie not found")
+            .inspect_err(|_| ATTESTATION_ERRORS.inc())?;
 
         let session_id = cookie.value();
 
-        let attestation: Attestation =
-            serde_json::from_slice(attestation).context("deserialize Attestation")?;
+        let attestation: Attestation = serde_json::from_slice(attestation)
+            .inspect_err(|_| ATTESTATION_ERRORS.inc())
+            .context("deserialize Attestation")?;
         let (tee, nonce) = {
             let session = self
                 .session_map
                 .sessions
                 .get_async(session_id)
                 .await
-                .ok_or(anyhow!("No cookie found"))?;
+                .ok_or(anyhow!("No cookie found"))
+                .inspect_err(|_| ATTESTATION_ERRORS.inc())?;
             let session = session.get();
 
             debug!("Session ID {}", session.id());
@@ -296,6 +306,7 @@ impl AttestationService {
                 let body = serde_json::to_string(&json!({
                     "token": token,
                 }))
+                .inspect_err(|_| ATTESTATION_ERRORS.inc())
                 .context("Serialize token failed")?;
 
                 return Ok(HttpResponse::Ok()
@@ -305,6 +316,7 @@ impl AttestationService {
             }
 
             let attestation_str = serde_json::to_string_pretty(&attestation)
+                .inspect_err(|_| ATTESTATION_ERRORS.inc())
                 .context("Failed to serialize Attestation")?;
             debug!("Attestation: {attestation_str}");
 
@@ -314,6 +326,7 @@ impl AttestationService {
         // deserialize evidence
         let composite_evidence: CompositeEvidence =
             serde_json::from_value(attestation.tee_evidence)
+                .inspect_err(|_| ATTESTATION_ERRORS.inc())
                 .context("failed to parse composite evidence")?;
         let mut evidence_to_verify: Vec<IndependentEvidence> = vec![];
 
@@ -354,7 +367,8 @@ impl AttestationService {
         // additional evidence
         if !composite_evidence.additional_evidence.is_empty() {
             let additional_evidence: HashMap<Tee, TeeEvidence> =
-                serde_json::from_str(&composite_evidence.additional_evidence)?;
+                serde_json::from_str(&composite_evidence.additional_evidence)
+                    .inspect_err(|_| ATTESTATION_ERRORS.inc())?;
             for (tee, tee_evidence) in additional_evidence {
                 evidence_to_verify.push(IndependentEvidence {
                     tee,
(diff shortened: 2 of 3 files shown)


kbs/src/attestation/backend.rs
```diff
@@ -276,11 +279,13 @@ impl AttestationService {
...
-            .verify(tee, &nonce, &attestation_str)
-            .await
-            .context("verify TEE evidence failed")?;
+        let token = match self.inner.verify(tee, &nonce, &attestation_str).await {
+            Ok(token) => token,
+            Err(err) => {
+                ATTESTATION_FAILURES.inc();
+                return Err(err).context("verify TEE evidence failed");
+            }
+        };
```
Xynnn007: There is a way for logging but I am not sure if the compiler will cry for this
```rust
self.inner
    .verify(tee, &nonce, &attestation_str)
    .await
    .inspect(|_| ATTESTATION_FAILURES.inc())
    .context("verify TEE evidence failed")?;
```

kbs/src/attestation/backend.rs
```diff
@@ -230,6 +231,8 @@ impl AttestationService {
         attestation: &[u8],
         request: HttpRequest,
     ) -> anyhow::Result<HttpResponse> {
+        ATTESTATION_REQUESTS.inc();
```
Xynnn007, line 284: nit: Remote attestation (RCAR protocol) consists of two rounds of API calls, `auth` is the first round and `attest` is the second round. I'm not sure if this is appropriate here, maybe we can add some context like [here](https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/implementations/resource/backend.rs#L129)?

kbs/src/policy_engine/opa/mod.rs
```diff
@@ -33,6 +35,8 @@ impl PolicyEngineInterface for Opa {
         resource_path: &str,
         input_claims: &str,
     ) -> Result<bool, KbsPolicyEngineError> {
+        RESOURCE_POLICY_EVALS.inc();
```
Xynnn007: We could move the two resource policy evals to a more [upper place](https://github.com/confidential-containers/trustee/blob/main/kbs/src/api_server.rs#L284-L286), as now policy engine is designed as a plugin interface. If a new plugin is introduced, it could help us save some code lines.

Xynnn007 (commented):
Thanks @pmores for this! some nits

kbs/src/attestation/backend.rs
pmores: That's a great point, the compiler doesn't mind it :-) and it actually works, too (provided I use `inspect_err()`, not `inspect()` which is only called if value is `Ok()`) - thanks!

kbs/src/policy_engine/opa/mod.rs
pmores: Good point, thanks - I put the counters update where I did because I didn't realise this is just a specific plugin implementation (not a great place for this, I agree) and also because this is the function that has natural access to all possible policy evaluation outcomes and thus allows to count them with the highest possible resolution.

I agree this should be moved up the call stack, and especially if we can utilise the `error_response()` function as discussed on issue #543 this should also be possible and easy.

kbs/src/attestation/backend.rs
pmores: I'm not sure if I'm understanding correctly - could you please be more specific as to why you suspect this might not be appropriate here?  My understanding is that `auth` is a separate stage in RCAR, handled [elsewere](https://github.com/confidential-containers/trustee/blob/b5d76310032b9fa298a18294a7735a0ecee13db6/kbs/src/attestation/backend.rs#L194), only invoked when client is not authenticated and not called on subsequent attestations anymore (at least until a session expires) - am I misinterpreting the protocol semantics?

Also, from early tests that I ran, putting the counter increment here leads to results that seem expected and correct, at least to me.

As for adding more context, I'm deliberately omitting all labels at this stage as there doesn't seem to be consensus (@mkulke) as to how to go about labelling.  I'll be very happy to add them once it's clear what makes everybody happy. :-)

Xynnn007: RCAR is based on a Challenge-Response protocol. The `auth` interface call is to obtain a Challenge value from KBS to generate Evidence, and `attest` is to actually submit Evidence to verify its binding with the challenge value. Usually, the number of calls to these two interfaces is equivalent, but it is not ruled out that sometimes due to incompatible network or KBS Client protocol versions, auth is called but attest is not called.


I saw only `attest` is covered but not `auth`. This makes me suspect that this matter is not very clear so I hope that this context could help.

pmores: That makes it a lot clearer, thanks!  Wouldn't `auth` be best tracked then if we add a separate set of counters for it? If I understand correctly `auth` as such can't really fail (unless there is a request deserialisation or nonce generation error, or a version mismatch, it will always respond with a session cookie and thus succeed), `auth_requests_total` and `auth_requests_errors` would seem enough.

kbs/src/prometheus/mod.rs
```diff
@@ -57,6 +57,42 @@ lazy_static! {
         Histogram::with_opts(response_sizes_opts).unwrap()
     };
 
+    /// KBS Policy Evaluations Total
+    pub(crate) static ref RESOURCE_POLICY_EVALS: Counter = {
```
fitzthum: At one point we were talking about moving from the term `resource_policy` to `kbs_policy` since the resource backend is just one of the plugins we support in the KBS, but I think we stopped caring about that.

Xynnn007: the [context](https://github.com/confidential-containers/trustee/blob/main/kbs/src/api_server.rs#L283)

pmores: So should I rename the counters?  @fitzthum does what you say mean that there's ever only a single policy applied in the KBS?

kbs/src/api_server.rs
```diff
@@ -280,8 +283,15 @@ pub(crate) async fn api(
...
+                RESOURCE_POLICY_EVALS.inc();
                 // TODO: add policy filter support for other plugins
-                if !core.policy_engine.evaluate(&endpoint, &claim_str).await? {
+                if !core
+                    .policy_engine
+                    .evaluate(&endpoint, &claim_str)
+                    .await
+                    .inspect_err(|_| RESOURCE_POLICY_ERRORS.inc())?
+                {
+                    RESOURCE_POLICY_VIOLATIONS.inc();
```
Xynnn007: The format of variable `endpoint` looks like `{plugin-name}/{other-parts}`. If you focus on resource related, you can count `endpoint` starts with `resource/` where `{plugin-name}` is `resource`. 

Btw, currently, this policy can indeed cover all plugin calls. The data input to the Policy Engine is a complete `endpoint`.  Thus `KBS policy` would be better? @fitzthum

kbs/src/prometheus/mod.rs
```diff
@@ -57,6 +57,78 @@ lazy_static! {
...
+    };
+
+    /// KBS Auth Errors Total
+    pub(crate) static ref AUTH_ERRORS: Counter = {
+        let opts = Opts::new(
+            "auth_errors_total",
+            "Total count of errors during auth processing",
+        );
+        Counter::with_opts(opts).unwrap()
+    };
```
Xynnn007, line 157: Oh, the same pattern of code breaks out here. It seems that we can consider using macros to simplify it. We can add it in the subsequent PR

Xynnn007 (commented):
Some comments. btw `auth` is good to carry an overall `inspect_err` as you suggested imo.

kbs/src/api_server.rs
pmores: Okay, thanks for clearing this up!  I think I can just go ahead and do the rename since @fitzthum has already expressed his preference gently but firmly. :-)

kbs/src/prometheus/mod.rs
pmores: Actually, this hasn't occurred to me yet but you're totally right.

pmores:
Pushed the final version as far as I'm concerned, please do your final review.  Changes:
- renamed resource policy to KBS policy as per [review request](https://github.com/confidential-containers/trustee/pull/816#discussion_r2135970620)
- added success counters as per [review request](https://github.com/confidential-containers/trustee/pull/816#pullrequestreview-2910485317)
- added `tee_type` labelling to attestation success and failure counters as per [result of the pertinent discussion](https://github.com/confidential-containers/trustee/issues/543#issuecomment-2975898460)

Personally I'm not entirely happy with how attestation counters are handled but after several tests and attempts I believe any improvement is only possible with a fairly substantial refactor of `__attest()`/`attest()` which would not be appropriate at this time in this PR.

kbs/src/attestation/backend.rs
```diff
@@ -268,31 +290,52 @@ impl AttestationService {
...
+            .context("serialize attestation failed")?;
+        let token = self
+            .inner
+            .verify(tee, &nonce, &attestation_str)
+            .await
+            .inspect(|_| {
+                ATTESTATION_SUCCESSES
+                    .with_label_values(&[&tee_type_label])
+                    .inc()
+            })
```
Xynnn007: We can move this out of the nested `inspect` to keep the chain short.

kbs/src/api_server.rs
```diff
@@ -280,9 +283,18 @@ pub(crate) async fn api(
...
+                    .policy_engine
+                    .evaluate(&endpoint, &claim_str)
+                    .await
+                    .inspect_err(|_| KBS_POLICY_ERRORS.inc())?
+                {
+                    KBS_POLICY_VIOLATIONS.inc();
                     return Err(Error::PolicyDeny);
+                } else {
+                    KBS_POLICY_APPROVALS.inc();
                 }
```
Xynnn007: Some personal taste - we even do not need the `else` here

kbs/src/attestation/backend.rs
pmores: done

kbs/src/api_server.rs
pmores: done

kbs/src/attestation/backend.rs
```diff
@@ -324,23 +349,41 @@ impl AttestationService {
...
-            .ok_or(anyhow!("session not found"))?;
+            .ok_or(anyhow!("session not found"))
+            .inspect_err(|_| ATTESTATION_ERRORS.inc())?;
         let session = session.get_mut();
 
+        let tee_type_label = {
+            // it seems impossible to prevent serde from putting double-quotes
+            // around the tee name, get rid of them in a subsequent step
+            let quoted = serde_json::to_string(&session.request().tee)?;
+            quoted[1..quoted.len() - 1].to_owned()
```
fitzthum: We use `trim_start_matches` and `trim_end_match` elsewhere in the codebase. Might be slightly safer than this.

kbs/src/attestation/backend.rs
```diff
@@ -324,23 +349,41 @@ impl AttestationService {
...
+        };
+
+        let token = self
+            .inner
+            .verify(evidence_to_verify)
+            .await
+            .inspect_err(|_| {
+                ATTESTATION_FAILURES
+                    .with_label_values(&[&tee_type_label])
```
fitzthum, line 415: At this point we aren't guaranteed to have a valid tee type so there is a chance of dos.

fitzthum: actually maybe it's fine since we've deserialized the Tee enum

pmores: Yes, exactly my thinking - we take the TEE type from Request, and Request parses `"tee"` into an enum during auth.  A Request with an invalid/made up tee won't even parse (being reasonably paranoid I've actually verified this) and won't make it to the attest phase.

bpradipt:
I think having a readme for the metrics feature will be a good addition.

kbs/src/attestation/backend.rs
pmores: True, thanks, should be fixed now.

pmores:
please take a look

fitzthum:
We should add some info to the website about this.

PR #818, opened 2025-06-09
bump rust toolchain to 1.85.1

mythi:
pin 1.85.1 toolchain version for the repository. Also, drop RUSTC_VERSION definition from CI workflow files as they are mostly redundant.

The actions README states:
"If a toolchain file (i.e., rust-toolchain or rust-toolchain.toml) is found in the root of the repository and no toolchain value is provided, all items specified in the toolchain file will be installed."

Trustee CI checks Rust builds against one version only so we can control CI through rust-toolchain.toml.

NB: Rust MSRV is unchanged: 1.80.x.

### DIFF
--- a/.github/workflows/as-e2e.yml
+++ b/.github/workflows/as-e2e.yml
@@ -28,16 +28,14 @@ jobs:
     name: TEE=${{ matrix.restful_tee_enum }} Generate Evidence Dynamically=${{ matrix.generate_evidence }}
     runs-on: ${{ matrix.runner }}
     env:
-      RUSTC_VERSION: 1.80.0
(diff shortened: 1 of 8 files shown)


mythi:
2. I checked when I saw everything passing. I think it's simply 3.) because we've just fixed things along the way (at least I'm typically doing some sanity checking when a new rust version is released). The toolchain pinning makes this harder because of the need to override the version. 

I can double check 1. tomorrow to be sure.

fitzthum:
> because we've just fixed things along the way 

yeah good point we do have the latest toolchain in the CI

PR #820, opened 2025-06-11
Fix Rust toolchain 1.85.1 to Dockerfiles

Xynnn007:
This PR fixes the rust version of the dockerfile to 1.85.1. The original version used to be the latest one. Please see the specific commit message

### DIFF
--- a/.github/workflows/build-as-image.yml
+++ b/.github/workflows/build-as-image.yml
@@ -49,6 +49,8 @@ jobs:
           instance: ubuntu-24.04-arm
           verifier: cca-verifier
     runs-on: ${{ matrix.instance }}
+    env:
+      RUST_TOOLCHAIN_VERSION: 1.85.1
 
     steps:
     - name: Checkout code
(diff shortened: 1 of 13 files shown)


Xynnn007:
> [cbb9573](https://github.com/confidential-containers/trustee/commit/cbb9573957c5de4b1dbb9d1e0a925bdfcd7c3728) does not look reasonable because in this repo we can now control the version using toolchain pinning.


Yes, I know that. What I'm not sure if the config file is picked up during the docker image build phase. Let me check the CI logs and if so I'll cancel the commit.

Xynnn007:
Good. It picks https://github.com/confidential-containers/trustee/actions/runs/15548417359/job/43774241534?pr=819#step:5:184 (This Action run is an old one). Thanks for the notice, let me get rid of that.

Xynnn007:
Oh, and if that's the case, latest rust makes sense, since every build of it will automatically choose the version based on rust-toolchain.toml, and thus the whole PR isn't needed.

mythi:
> Oh, and if that's the case, latest rust makes sense, since every build of it will automatically choose the version based on rust-toolchain.toml, and thus the whole PR isn't needed.

The only thing that is a bit sub-optimal is that we pull latest but always downgrade for each build which might also bust the container layer build cache. I guess it's OK for Github where caching must be optimized but not sure what developers think.

Issue #768, opened 2025-04-16
Ditch HashValuePair struct in RVPS


fitzthum:
Related to https://github.com/confidential-containers/trustee/issues/637 but more specific. 

The RVPS keeps track of values in hash, value pairs via a struct called HashValuePair.

I think we should get rid of this abstraction and just use a single value instead. We can still keep track of metadata in the parent reference value struct, but it seems wrong to assume that reference values will come in hash, value pairs. The policy itself should handle any transformations that need to be made to a reference value.

wdyt @Xynnn007

Xynnn007:
Agreed. hash algorithm usually do no help to the context

fitzthum:
I'm starting to work on this. Two questions for you @Xynnn007 

1) What about using `serde_json::Value` in place of HashValue. I am thinking we want to support different types of reference values (see the recent issue we had with the integer Values in the SNP policy). We could introduce our own enum for this, but maybe it's simpler to use `serde_json::Value` especially given that the policy engine consumes this.

2) Do we want to continue to support multiple reference values per reference value object? Basically today each reference value is actually a list of reference values. In theory this makes sense because users will often have multiple allowed measurements, but in practice I don't think we've used this feature. For one thing most reference value manifests that I have seen just represent one set of valid measurements. We may want to have multiple sets of reference values rather than each reference value having multiple values. wdyt?

Anyway, I'm thinking the code would either look like this 
```
#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
pub struct ReferenceValue {
    #[serde(default = "default_version")]
    pub version: String,
    pub name: String,
    #[serde(deserialize_with = "primitive_date_time_from_str")]
    pub expiration: DateTime<Utc>,
    pub value: serde_json::Value,
}
```
or like 
```
pub struct ReferenceValue {
    #[serde(default = "default_version")]
    pub version: String,
    pub name: String,
    #[serde(deserialize_with = "primitive_date_time_from_str")]
    pub expiration: DateTime<Utc>,
    pub value: serde_json::Value,
}
```

I have some other fields in mind too, but I will save those for the PR.

Xynnn007:
@fitzthum 
1. `Value` would makes more sense. As you said, we need to handle data that is not just strings.
[...]

PR #821, opened 2025-06-13
rvps: drop HashValue abstraction

fitzthum:
When we first made the RVPS, we assumed that most reference values would be hashes and we created a struct to keep track of this. It turns out that lots of reference values have other formats (integers, strings, etc) and even for value that are hashes, the RVPS only reports the value itself to the AS. As such, let's drop the HashValue struct and instead use a serde_json::Value to store the reference value.

One cool advantage of this is that reference values can take almost any format. For instance, one reference value can actually be a map of a bunch of values. As long as the AS policy knows how to use the value, everything should work out.

I also added a little hint about how we will support a pull model by supporting local and remote values. Support for remote values will be added in future PRs.

See https://github.com/confidential-containers/trustee/issues/768

### DIFF
--- a/attestation-service/src/lib.rs
+++ b/attestation-service/src/lib.rs
@@ -10,12 +10,13 @@ pub mod token;
 
 use crate::token::AttestationTokenBroker;
 
+pub use kbs_types::{Attestation, Tee};
+pub use serde_json::Value;
+
 use anyhow::{anyhow, bail, Context, Result};
 use config::Config;
-pub use kbs_types::{Attestation, Tee};
 use log::{debug, info};
 use rvps::{RvpsApi, RvpsError};
-use serde_json::Value;
 use sha2::{Digest, Sha256, Sha384, Sha512};
 use std::collections::HashMap;
 use strum::{AsRefStr, Display, EnumString};
@@ -257,7 +258,7 @@ impl AttestationService {
     }
 
     /// Query Reference Values
-    pub async fn query_reference_values(&self) -> Result<HashMap<String, Vec<String>>> {
+    pub async fn query_reference_values(&self) -> Result<HashMap<String, Value>> {
         self.rvps
             .get_digests()
             .await
--- a/rvps/src/extractors/extractor_modules/swid/mod.rs
+++ b/rvps/src/extractors/extractor_modules/swid/mod.rs
@@ -5,24 +5,14 @@
 
 use anyhow::*;
 use base64::Engine;
-use chrono::{Months, Timelike, Utc};
 use log::{debug, info};
 
-use crate::{
-    reference_value::{HashValuePair, REFERENCE_VALUE_VERSION},
-    ReferenceValue,
-};
+use crate::ReferenceValue;
 
 use super::Extractor;
 
 #[derive(Default)]
 pub struct SwidExtractor;
-
-/// Default reference value hash algorithm
-const DEFAULT_ALG: &str = "sha384";
-
-/// The reference value will be expired in the default time (months)
-const MONTHS_BEFORE_EXPIRATION: u32 = 12;
 
 const SWID_NS: &str = "http://standards.iso.org/iso/19770/-2/2015/schema.xsd";
 const RIMIM_NS: &str = "https://trustedcomputinggroup.org/resource/tcg-reference-integrity-manifest-rim-information-model/";
@@ -80,28 +70,19 @@ impl Extractor for SwidExtractor {
             // Resource may have multiple hash attributes
             let mut hash_index = 0;
             loop {
-                let hash = resource.attribute((HASH_NS, format!("Hash{hash_index}").as_str()));
-                if hash.is_none() {
+                let Some(hash) =
+                    resource.attribute((HASH_NS, format!("Hash{hash_index}").as_str()))
+                else {
                     break;
-                }
+                };
 
-                let rv_pair = HashValuePair::new(DEFAULT_ALG.into(), hash.unwrap().to_string());
+                let hash_value = serde_json::Value::String(hash.to_string());
                 let name = format!("{rv_name_prefix}.{measurement_name}.hash{hash_index}");
 
                 // Rego does not like dashes
                 let name = name.replace("-", "_");
 
-                let expiration = Utc::now()
-                    .with_nanosecond(0)
-                    .and_then(|t| t.checked_add_months(Months::new(MONTHS_BEFORE_EXPIRATION)))
-                    .unwrap(); // The expiration time is fixed.
-
-                let rv = ReferenceValue {
-                    version: REFERENCE_VALUE_VERSION.into(),
-                    name: name.to_string(),
-                    expiration,
-                    hash_value: vec![rv_pair],
-                };
+                let rv = ReferenceValue::new()?.set_name(&name).set_value(hash_value);
                 rvs.push(rv);
 
                 hash_index += 1;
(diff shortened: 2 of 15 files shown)


rvps/src/extractors/extractor_modules/swid/mod.rs
```diff
@@ -85,23 +75,15 @@ impl Extractor for SwidExtractor {
                     break;
                 }
 
-                let rv_pair = HashValuePair::new(DEFAULT_ALG.into(), hash.unwrap().to_string());
+                let hash_value = serde_json::Value::String(hash.unwrap().to_string());
```
Xynnn007: we can use the following structure to avoid `unwrap`
```rust
let Some(hash) = resource.attribute((HASH_NS, format!("Hash{hash_index}").as_str())) else {
    break;
}
// ... use `hash` variable
```

rvps/src/reference_value.rs
```diff
@@ -49,26 +27,37 @@ fn primitive_date_time_from_str<'de, D: Deserializer<'de>>(
...
-/// * `hash_value`: A set of key-value pairs, each indicates a hash
-///   algorithm and its relative hash value for the artifact.
-///   The actual struct deliver from RVPS to AS is
-///   [`TrustedDigest`], whose simple structure is easy
-///   for AS to handle.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+pub enum ValueType {
+    Local(serde_json::Value),
+    Remote,
+}
```
Xynnn007: Why do we need "remote" here? At my first glance, the idea is that the internal metadata may store a URL pointing to the remote end. When the client initiates a request, RVPS will help to obtain it. 
I suggest using a simple ReferenceValue structure and abandon `ValueType` here. Because ReferenceValue or ValueType is the logical result after the Extractor processes, but if it is a content pointed to by a URI, it means that it also needs to take care of the type information and gets processed by Extractor again. In RVPS, the pull model requires us to add Remote-related logic in the logic before the Extractor, and at least specify its 1. Extractor type 2. URI address. It could be a choice to introduce in a separate PR

attestation-service/src/lib.rs
```diff
@@ -219,7 +219,7 @@ impl AttestationService {
     }
 
     /// Query Reference Values
-    pub async fn query_reference_values(&self) -> Result<HashMap<String, Vec<String>>> {
+    pub async fn query_reference_values(&self) -> Result<HashMap<String, serde_json::Value>> {
```
Xynnn007: Note that we are returning a struct `Value` not provided by `attestation_service` crate now. The user might meet situations where `serde_json` is not included in the code base and `serde_json::Value` is not found.

A workaround is to add `pub use serde_json::Value` to all the files that uses this struct.

A better solution is to wrap `serde_json::Value` into a new struct defined in rvps crate and implement some basic conversion logic from/into `serde_json::Value`.

It's ok to me to perfect it later with way 2 and just use way 1 at first.

Xynnn007 (commented):
Nice @fitzthum . I think we are getting closer to the right destination

rvps/src/reference_value.rs
fitzthum: I see your point. Here's my thinking about the flow. To setup a remote reference value, the admin could provide a manifest that points to remote values. This would go through an extractor. This would allow the configuration for the remote RV to be signed and distributed like an RV itself.

The remote and local RVs would share the same struct because they share the same top-level metadata, but when retrieving a remote value there would be additional logic. The remote value configuration (inside the ValueType enum) could point to different pull modules. If we use coserv to pull a value, we don't need to use the extractors again (I think). If we use a simple curl-based method, we would import the extractors and run them. I think we would also probably implement some caching in this config struct.

Anyway I'm not too attached to this flow and it is maybe a little convoluted. My question for you is if we don't use the existing extractors to configure the remote values, how would an admin setup those values? We don't currently have any other API for that. I think configuring remote and local values in the same way might be nice.

Xynnn007: I want to confirm one more question:
1. Is the URL of the remote reference value here carried in a source information format? (For example, let's assume that CoRIM can carry a URI instead of a real reference value)
2. Or do we just mean that the user manually enters the source of a reference value and then retrieves it when it is actually accessed?

Because the two questions will lead to different design ideas; or can you give a more specific example of an application scenario?

fitzthum: Not sure I totally get the question. There are some reference value manifest formats that support remote values. For instance the TCG RIM-IM SWID thing that is used by nvidia can support links to external documents. Our extractor currently does not support this, but we could extend it so that these RIMs could be parsed into both local and remote reference values as needed. I'm not sure about how widely used this feature is or if other manifest formats have something similar.

In the short term I think we would just extend our sample manifest format to support remote values. Now we also have a front-end for reference values in the KBS client that will create a sample manifest automatically. We could also extend that to support remote values.

That said, there is a problem with the design I am proposing. Most RIMs don't just have one value. They have a bunch of different values. Using one reference value struct to point to a RIM that has many values might not be ideal. I'm not quite sure the best way to handle this yet. To support a pull model, we probably do need to know which remote values can be found at which location, but that information might not even be known ahead of time if the remote RIMs are dynamic.

Xynnn007: Do you mean that the reference value pointed to in the AS's Policy may be a remote URI or a key that indexes the reference value?

The reason that I raise this question is that in my mind, the AS policy even does not need to know the URI, but only need to know the key to index.

> That said, there is a problem with the design I am proposing. Most RIMs don't just have one value. They have a bunch of different values. Using one reference value struct to point to a RIM that has many values might not be ideal. I'm not quite sure the best way to handle this yet. To support a pull model, we probably do need to know which remote values can be found at which location, but that information might not even be known ahead of time if the remote RIMs are dynamic.

Probably we can discuss the remote reference values more when we actually try to integrate it. Before that, as what we are using now, they are all local. Does it make sense to just integrate local form in this PR?

fitzthum: Yeah let me drop the remote thing from this PR and we can discuss more.

I agree that the policy shouldn't need to know a URI. The policy should request an RV by ID. For remote values we will need some kind of mapping from this id to a remote manifest. The problem is that these manifests usually don't have just one value and technically we might not know exactly which values they do contain until they are actually pulled. So I suspect we will need another abstraction to keep track of a group of RVs or a class of RVs that are associated with a remote manifest.

Xynnn007: A URI based id would do help. Parent path could mean a manifest and child path could mean the subpath in it.

There is also another problem, s.t. TE(target environment). In some cases,  some reference values of different fields must be matched altogether, and this kind of matching is named TE.

One manifest can have multiple TEs, and one TE can have multiple RVs. There are two questions for this
1. How to define an id of both TE and RV
2. How to design TE rule in AS policy.

But we can disscuss and resolve TE case once we need it later.

fitzthum: Yeah I agree. I think the URI could be useful here. We could define a remote value fetching mechanism that could match on a URI (either a specific id or a part of an id) and pull in values. We'd just need to think a little bit about where this goes in the flow and how to handle caching, but we can figure that out in the implementation.

attestation-service/src/lib.rs
fitzthum: OK I added `pub use serde_json::Value` to the RVPS and AS libs.

Xynnn007: Nice. I think we can now use `Result<HashMap<String, Value>>`?

fitzthum: Ok. I kind of like using the full path when the enum name is so generic, but I will update the signatures.

Xynnn007: Oh yes, it would be better to treat the type as "provided by our crate" after we have pub use it.

Xynnn007 (approved):

PR #824, opened 2025-06-18
AS: fix test cases and documents

Xynnn007:
This patch fixes the tests and documents of the CoCoAS. cc @fitzthum


attestation-service/src/bin/restful/mod.rs
```diff
@@ -36,7 +36,7 @@ type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct AttestationRequest {
-    individual_attestation_request: Vec<IndividualAttestationRequest>,
+    verification_requests: Vec<IndividualAttestationRequest>,
```
fitzthum, line 39: Good, but you gotta change the for loop below.

fitzthum (approved):
LGTM just fix the thing it's complaining about.

Maybe we should add the different AA front-ends to the integration tests at some point although currently the KBS only supports builtlin or gRPC connections.

attestation-service/src/bin/restful/mod.rs
Xynnn007: Oh yes. What a premature celebration

mythi (approved):

Issue #542, opened 2024-10-21
RFC: Run Trustee inside of an enclave


fitzthum:
Since Trustee needs to run in a trusted environment, it could make sense to run it inside of an enclave. This is easier said and done. 

Here is a sketch of how we could run Trustee inside of not just any enclave, but CoCo.

One one level, it should be easy to run Trustee in CoCo. We can already deployee Trustee in k8s. If we add some runtime classes, it will work for CoCo. On top of that, we probably want to use sealed secrets to provide certain parts of the configuration, such as the token signing keys.

The big question is how we attest this environment. It might seem like we have a tautological situation here, but actually I think it's reasonable to have a basic attestation tool that can attest Trustee only.

So the question is really how we create a simple, static attestation tool (likely part of trustee-cli) that will listen for a resource request, carry out the KBS protocol, and provide secrets. Hopefully we can reuse logic from Trustee, but the implementation is still a little fuzzy to me.

PR #825, opened 2025-06-18
Add support for Hashicorp vault as a resource backend

bpradipt:


kbs/Cargo.toml
```diff
@@ -105,6 +109,20 @@ tempfile.workspace = true
 rstest.workspace = true
 reference-value-provider-service.path = "../rvps"
 serial_test = "3.0"
+toml = "0.8"
+
+[target.'cfg(not(any(target_arch = "s390x", target_arch = "aarch64")))'.dev-dependencies]
+kbs-client = { path = "../tools/kbs-client" }
```
fitzthum: These changes seem unrelated to this PR?

kbs/docs/vault_kv.md
```diff
@@ -0,0 +1,256 @@
+# HashiCorp Vault KV secrets engine resource backend
+
+[HashiCorp Vault](https://developer.hashicorp.com/vault) is a secrets
+management tool that securely stores and tightly controls access to tokens,
+passwords, certificates, encryption keys, and other secrets. This plugin
```
fitzthum: Technically this is a resource backend rather than a plugin so you may want to avoid using the term plugin here.

kbs/src/plugins/implementations/resource/vault_kv.rs
```diff
@@ -0,0 +1,245 @@
+// Copyright (c) 2024 Confidential Containers Contributors.
```
fitzthum: nit: 2025

kbs/src/plugins/implementations/resource/vault_kv.rs
```diff
@@ -0,0 +1,245 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use super::backend::{ResourceDesc, StorageBackend};
+use anyhow::{Context, Result};
+use derivative::Derivative;
+use log::info;
+use serde::Deserialize;
+use std::collections::HashMap;
+
+#[cfg(feature = "vault")]
```
fitzthum: Since you are conditionally importing this module in `mod.rs`, I don't think you need to have all these feature flags in this file. At least there should be a way to set things up so that they aren't necessary.

kbs/src/plugins/implementations/resource/vault_kv.rs
```diff
@@ -0,0 +1,245 @@
...
+        // For KV v1, we expect the secret to contain a "data" field with the actual secret
+        // If "data" field exists, use it; otherwise, serialize the entire HashMap
+        let secret_bytes = if let Some(data) = secret_data.get("data") {
+            data.as_bytes().to_vec()
+        } else if let Some(value) = secret_data.get("value") {
+            value.as_bytes().to_vec()
+        } else {
+            // If no specific field, serialize the entire secret as JSON
+            serde_json::to_vec(&secret_data).context("Failed to serialize secret data")?
+        };
```
fitzthum: This big conditional seems slightly un-idiomatic. I might just use two `if let` statements that return `secret_bytes` directly. 

Also, this doesn't quite mach the comment above, which doesn't mention the `value` field.

kbs/src/plugins/implementations/resource/vault_kv.rs
bpradipt: Hmm, let me explore this further. Without the feature flag here, it resulted in build failures for me. May be I was doing some mistake in the env

bpradipt: Issue on my side.. works fine with feature flag only in the mod.rs

kbs/src/plugins/implementations/resource/vault_kv.rs
bpradipt: Updated to align with other backend providers. Just returning an error when the secret data is not available in vault

kbs/config/kbs-config.toml
```diff
@@ -26,3 +26,24 @@ insecure_api = true
...
+#
+# HTTPS Configuration Options:
+# - verify_ssl: Set to true to enable SSL certificate verification
+# - ca_certs: Array of paths to custom CA certificate files for self-signed or enterprise CA certificates
+#
+# Note: Token must have read/write permissions for the specified mount path
+# Example Vault policy:
+# path "secret/*" {
+#   capabilities = ["create", "read", "update", "delete", "list"]
+# }
```
Xynnn007, line 49: Probably we can move the `verify_ssl` and `ca_certs` parts to the comment after the corresponding item.

kbs/src/plugins/implementations/resource/vault_kv.rs
```diff
@@ -0,0 +1,212 @@
...
+        token = "hvs.test-token"
+        "#;
+
+        let config: VaultKvBackendConfig = toml::from_str(config_str).unwrap();
+        assert_eq!(config.mount_path, "secret");
+        assert_eq!(config.verify_ssl, false); // Default value
+        assert_eq!(config.ca_certs, None); // Default value
+    }
+
+    #[tokio::test]
```
Xynnn007: Looks like this test case does not need to be async, thus`#[test]` is enough

kbs/src/plugins/implementations/resource/vault_kv.rs
```diff
@@ -0,0 +1,212 @@
...
+        assert_eq!(expected_path, "test/secret/test-key");
+    }
+
+    #[test]
+    fn test_vault_config_https_options() {
+        // Test minimal HTTPS config
+        let config_str = r#"
+        vault_url = "https://vault.example.com:8200"
+        token = "hvs.test-token"
+        verify_ssl = false
```
Xynnn007, line 218: For deseralization, `rstest` can be a choice to make it dry if you like.

Examples: 
- https://github.com/confidential-containers/trustee/blob/main/kbs/src/config.rs#L143-L477

kbs/src/plugins/implementations/resource/vault_kv_test.rs
```diff
@@ -0,0 +1,561 @@
+#[cfg(test)]
```
Xynnn007: For rust we often do not have a separate `*_test.rs` file like golang. We can merge the whole content to `vault_kv.rs` file.

kbs/src/plugins/implementations/resource/vault_kv_test.rs
```diff
@@ -0,0 +1,561 @@
...
+        match backend.read_secret_resource(resource_desc).await {
+            Ok(data) => {
+                println!(
+                    "Successfully read secret: {:?}",
+                    String::from_utf8_lossy(&data)
+                );
+                assert!(!data.is_empty());
+            }
+            Err(e) => {
+                // Test might fail if secret doesn't exist - that's expected
```
Xynnn007: How to distinguish a failed reading and a reading that does not exist? If we run `cargo test`, they both will make the test green.

kbs/src/plugins/implementations/resource/vault_kv_test.rs
```diff
@@ -0,0 +1,561 @@
...
+                    }
+                }
+            }
+            Err(e) => {
+                println!(
+                    "Write failed (may be expected if Vault server isn't configured): {}",
+                    e
+                );
+            }
+        }
```
Xynnn007: The same question as before. We cannot distinguish the real error of reading/writing and errors caused by no secret. btw we have a hidden habit of "trying to avoid nested conditions", which can make the logic and readability of the code better.
```rust
let _ = backend
        .write_secret_resource(resource_desc.clone(), test_data)
        .await
        .inspect_err(|e| {
            println!(
                "Write failed (may be expected if Vault server isn't configured): {}",
                e
            )
        })
        .expect("Failed to write secret resource");

    let data = backend
        .read_secret_resource(resource_desc)
        .await
        .expect("Failed to read secret resource");
    println!(
        "Successfully read back secret: {:?}",
        String::from_utf8_lossy(&read_data)
    );
```

kbs/src/plugins/implementations/resource/vault_kv_test.rs
```diff
@@ -0,0 +1,561 @@
...
+    }
+
+    #[tokio::test]
+    #[ignore]
+    async fn vault_ssl_with_ca_verification() {
```
Xynnn007: I noticed that the tests share a common structure, s.t. 
1. Given different config file
2. write secret
3. read secret

Probably we can merge them into just two function, s.t. `vault_ssl_test` and `vault_nossl_test`, and use the config as the rstest `case`?

kbs/test/config/resource-kbs-vault-ssl-insecure.toml
```diff
@@ -0,0 +1,21 @@
+[http_server]
```
Xynnn007: Are the three `toml`s mentioned somewhere else?

Xynnn007 (commented):
Thanks @bpradipt ! Nice code, nice tests. Only some code optimization ideas, but the logic looks good to me

kbs/src/plugins/implementations/resource/vault_kv_test.rs
bpradipt: I see.. Since these tests were integration tests requiring a running vault instance, I kept it separate. I'll move it to vault_kv.rs

kbs/src/plugins/implementations/resource/vault_kv_test.rs
bpradipt: Vault either returns key doesn't exist or an API error. I'll distinguish these error types in the next iteration.

kbs/test/Makefile
```diff
@@ -227,10 +227,162 @@ stop: stop-kbs stop-resource-kbs
...
+VAULT_TOKEN_FILE := $(WORK_DIR)/vault-token
+VAULT_KBS_CONFIG_PATH := ./config
+VAULT_CA_KEY := $(WORK_DIR)/vault-ca.key
+VAULT_CA_CERT := $(WORK_DIR)/vault-ca.pem
+VAULT_SERVER_KEY := $(WORK_DIR)/vault-server.key
+VAULT_SERVER_CERT := $(WORK_DIR)/vault-server.pem
+
+# Download and setup Vault binary
+$(VAULT_BIN):
+	@if [ ! -f $(VAULT_BIN) ]; then \
```
mkulke, line 246: isn't this guaranteed by the makefile? if the VAULT_BIN file is present the target will be fulfilled?

kbs/src/plugins/implementations/resource/vault_kv_test.rs
mkulke: by convention integration tests are in `$ROOT/test`, for some reason we have `$ROOT/integration-test`, so it should probably go there.

kbs/src/plugins/implementations/resource/vault_kv_test.rs
mkulke: I agree, if the tests need some special setup to succeed, they should be ignored, but we shouldn't swallow errors. the println! console output will not be visible in the shell on that runs `cargo test`, unless you specify --no-capture

kbs/src/plugins/implementations/resource/vault_kv_test.rs
```diff
@@ -0,0 +1,561 @@
...
+        let resource_desc = ResourceDesc {
+            repository_name: "test-repo".to_string(),
+            resource_type: "test-type".to_string(),
+            resource_tag: "write-test".to_string(),
+        };
+
+        let test_data = b"test-secret-data-for-write";
+
+        // Test write operation
+        match backend
```
mkulke: to avoid this nesting we can early exit:

```rust
let Err(e) = backend.write_secret_reseource.await {
   println!("Write failed (may be expected if Vault server isn't configured): {}", e);
   return;
}
```

kbs/test/Makefile
bpradipt: Rerun of the test case will download the files again. This is to avoid it.

Xynnn007 (approved):
LGTM. Thanks @bpradipt !

oliver-steinbrecher:
Thanks a lot for proposing this important change. it will help clients with their setups a lot. Next step would be to bring trustee into a TEE / CC-protected setup - see https://github.com/confidential-containers/trustee/issues/542

kbs/src/plugins/implementations/resource/vault_kv_test.rs
fitzthum: I would argue that these aren't really end-to-end integration tests like we have in the rust integration tests or the Makefile tests. Those use the API of the KBS to attest, get a resource, etc. This is more like a unit test that requires a vault instance to be running. I guess the line is a little blurry, but I am fine with this living here, especially since we do have an end-to-end test in the Makefile. I would maybe have preferred that to go in the Rust integration tests but no big deal.

fitzthum:
Ok let's merge this unless there is any more review from @mkulke

mkulke (approved):

PR #828, opened 2025-06-20
Add initdata toml plaintext support

Xynnn007:
This PR mainly adds the KBS protocol and the AS verification function for initdata toml. The corresponding PR is https://github.com/confidential-containers/guest-components/pull/1031


deps/verifier/src/sample/mod.rs
```diff
@@ -80,7 +70,7 @@ fn parse_tee_evidence(quote: &SampleTeeEvidence) -> Result<TeeEvidenceParsedClai
     let claims_map = json!({
         "svn": quote.svn,
         "report_data": quote.report_data,
-        "init_data": quote.init_data,
+        "init_data": "",
```
fitzthum: If we don't check the init-data, we shouldn't set this claim at all due to [this logic](https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/token/ear_broker.rs#L429). The EAR broker will add the raw init-data to the token as long as the init-data claim is set.

kbs/src/attestation/backend.rs
```diff
@@ -45,6 +45,15 @@ lazy_static! {
...
 pub type TeeEvidence = serde_json::Value;
 
+#[derive(Deserialize)]
+pub struct KbsEvidence {
+    /// The TEE evidence.
+    evidence: String,
+
+    /// The init data is optional. Note that it is initdata plaintext.
+    init_data_toml: Option<String>,
+}
```
fitzthum: Not sure this struct is really necessary, especially after our discussion about putting the init-data in the protocol itself rather than in the evidence.

fitzthum (commented):
Looks good. Two comments.

kbs/src/attestation/backend.rs
Xynnn007: Yup. This is the old version of code. I just updated with changes upon kbs-types.

deps/verifier/src/sample/mod.rs
Xynnn007: Thanks, fixed this.

mkulke (approved):

Xynnn007:
blocked by a dependency conflict error. will be fix via https://github.com/veraison/rust-apiclient/pull/24

Issue #834, opened 2025-06-26
Update the sample policy to reflect the new EAR submod naming conventions


thomas-fossati:
In a [recent commit](https://github.com/confidential-containers/trustee/commit/2278ffabcb5eeaf4eb5f4eb868fbf0f99cd01ff6), the EAR submod naming convention has changed slightly to add an explicit index alongside the submod’s “basename” (i.e., from `cpu` to `cpu0`).

This new expectation should be reflected in KBS's sample policies, where it is still using the previous naming convention:

https://github.com/confidential-containers/trustee/blob/b37ad43e0a50938df36b00e8851560626a505f97/kbs/sample_policies/affirming.rego#L7

PR #835, opened 2025-06-26
sample-polcy: fixup affirming policy

fitzthum:
With multi-device attestation, the EAR submods are accessed with the Tee class and the index of the device among all devices of the same class.

In other words "cpu" -> "cpu0"

We introduced the affirming policy while the multi-device PR was being developed and we missed updating the submod index.

Fixes: https://github.com/confidential-containers/trustee/issues/834


Xynnn007 (approved):

bpradipt (approved):
/lgtm

PR #839, opened 2025-06-27
kbs-client: show file name when cannot read it

liangxiao1:
It's not clear to know which file is not found while passing more file path parameters like tee-key-file, cert-file, attestation-token.

With this patch:
$ kbs-client --url https://trusteeserver:8080 --cert-file trustee_keys/srv_host.crt get-resource --tee-key-file trustee_keys/tee_key.pem --attestation-token /tmp/attestation_token_x --path default/test/model_encrypt_key 
Failed to read:/tmp/attestation_token_x
Error: No such file or directory (os error 2)

Without this patch:
$ kbs-client --url https://trusteeserver:8080 --cert-file trustee_keys/srv_host.crt get-resource --tee-key-file trustee_keys/tee_key.pem --attestation-token /tmp/attestation_token_x --path default/test/model_encrypt_key 
Error: No such file or directory (os error 2)

### DIFF
--- a/tools/kbs-client/src/main.rs
+++ b/tools/kbs-client/src/main.rs
@@ -153,14 +153,14 @@ async fn main() -> Result<()> {
     let cli = Cli::parse();
 
     let kbs_cert = match cli.cert_file {
-        Some(p) => vec![std::fs::read_to_string(p)?],
+        Some(ref p) => vec![std::fs::read_to_string(p).inspect_err(|_|eprintln!("Failed to read:{}",p.display()))?],
         None => vec![],
     };
 
     match cli.command {
         Commands::Attest { tee_key_file } => {
             let tee_key = match tee_key_file {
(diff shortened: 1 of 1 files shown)


tools/kbs-client/src/main.rs
```diff
@@ -205,7 +205,8 @@ async fn main() -> Result<()> {
             }
         }
         Commands::Config(config) => {
-            let auth_key = std::fs::read_to_string(config.auth_private_key)?;
+            let auth_private_key = &config.auth_private_key;
```
fitzthum: nit: `autn_private_key` -> `auth_private_key_path`

liangxiao1: Thanks, @fitzthum updated.

Xynnn007 (approved):
Thanks @liangxiao1 ! Welcome to the Confidential Containers

PR #840, opened 2025-06-27
Use complex reference value types

fitzthum:
Now that we have support for complex reference value types, let's use them.

Specifically, let's update the `kbs-client` so that we can specify integer and boolean reference values, as well as RVs that are single values (rather than a list of values).

Then, let's add some integer tcb claims to the sample verifier to show how this works.

Then, update the SNP verifier to return boolean and integer claims where it makes sense. @bpradipt @DGonzalezVillal @ryansavino might want to check this part. This will change what users need to do to provision reference values.

We may want to do something similar for other platforms. 

See commit messages for more information.

### DIFF
--- a/deps/verifier/src/sample/mod.rs
+++ b/deps/verifier/src/sample/mod.rs
@@ -84,6 +84,12 @@ fn parse_tee_evidence(quote: &SampleTeeEvidence) -> Result<TeeEvidenceParsedClai
 
         // Generally TCB claims should originate from the attester.
         "launch_digest": "abcde",
+
+        // TCB Claims can be any type supported by serde_json
+        "platform_version": {
+            "major": 1,
+            "minor": 4,
+        },
     });
 
     Ok(claims_map as TeeEvidenceParsedClaim)
--- a/deps/verifier/src/snp/mod.rs
+++ b/deps/verifier/src/snp/mod.rs
@@ -381,27 +381,27 @@ pub(crate) fn verify_report_tcb(
 pub(crate) fn parse_tee_evidence(report: &AttestationReport) -> TeeEvidenceParsedClaim {
     let claims_map = json!({
         // policy fields
-        "policy_abi_major": format!("{}",report.policy.abi_major()),
-        "policy_abi_minor": format!("{}", report.policy.abi_minor()),
-        "policy_smt_allowed": format!("{}", report.policy.smt_allowed()),
-        "policy_migrate_ma": format!("{}", report.policy.migrate_ma_allowed()),
-        "policy_debug_allowed": format!("{}", report.policy.debug_allowed()),
-        "policy_single_socket": format!("{}", report.policy.single_socket_required()),
+        "policy_abi_major": report.policy.abi_major(),
+        "policy_abi_minor": report.policy.abi_minor(),
+        "policy_smt_allowed": report.policy.smt_allowed(),
+        "policy_migrate_ma": report.policy.migrate_ma_allowed(),
+        "policy_debug_allowed": report.policy.debug_allowed(),
+        "policy_single_socket": report.policy.single_socket_required(),
 
         // versioning info
-        "reported_tcb_bootloader": format!("{}", report.reported_tcb.bootloader),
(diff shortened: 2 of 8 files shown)


deps/verifier/src/sample/mod.rs
```diff
@@ -84,6 +84,12 @@ fn parse_tee_evidence(quote: &SampleTeeEvidence) -> Result<TeeEvidenceParsedClai
 
         // Generally TCB claims should originate from the attester.
         "launch_digest": "abcde",
+
+    // TCB Claims can be any type supported by serde_json
+    "platform_version": {
+        "major": 1,
+        "minor": 4,
+    },
```
Xynnn007: Let's add indents here for alignment.

integration-tests/tests/get_resource.rs
```diff
@@ -124,10 +125,16 @@ async fn get_secret_not_contraindicated(harness: &TestHarness) -> Result<()> {
 
     // so this test will fail if run in an enclave.
     harness
-        .set_reference_value("svn".to_string(), "1".to_string())
+        .set_reference_value("svn".to_string(), json!(["1".to_string()]))
         .await?;
     harness
-        .set_reference_value("launch_digest".to_string(), "abcde".to_string())
+        .set_reference_value("launch_digest".to_string(), json!(["abcde".to_string()]))
```
Xynnn007: I have noticed that we have
```
json!(["1"])
```

below this line, so we might not need `.to_string()`s in `json!` macro?

tools/kbs-client/src/main.rs
```diff
@@ -139,11 +140,28 @@ enum ConfigCommands {
...
+        /// If set, the value will be parsed as a bool.
+        #[clap(long, action, group = "resource_type")]
+        as_bool: bool,
+        /// By default the reference value will be a single
+        /// member in a list.
+        /// If this argument is set, the reference value
+        /// will be a single value.
+        #[clap(long, action)]
+        as_single_value: bool,
+    },
```
Xynnn007, line 164: I wonder if we could simply `let value: Value = serde_json::from_str()` to deserialize the givene reference value than define complex types here.

The benefit is that we could support to inject nested structrued reference values. We could add some unit tests for this.

deps/verifier/src/snp/mod.rs
```diff
@@ -381,27 +381,27 @@ pub(crate) fn verify_report_tcb(
...
-        "platform_tsme_enabled": format!("{}", report.plat_info.tsme_enabled()),
-        "platform_smt_enabled": format!("{}", report.plat_info.smt_enabled()),
+        "platform_tsme_enabled": report.plat_info.tsme_enabled(),
+        "platform_smt_enabled": report.plat_info.smt_enabled(),
 
         // measurements
-        "measurement": format!("{}", STANDARD.encode(report.measurement)),
-        "report_data": format!("{}", STANDARD.encode(report.report_data)),
-        "init_data": format!("{}", STANDARD.encode(report.host_data)),
+        "measurement": serde_json::Value::String(STANDARD.encode(report.measurement)),
```
Xynnn007: We might not need `serde_json::Value::String(` here for `json!` will automatically convert it into a string enum?

tools/kbs-client/src/main.rs
fitzthum: The rvps-tool can be used to specify arbitrary JSON reference values. We could also add a JSON option to the kbs-client, but I think we should keep the option to specify a single value without having to do any formatting.

Xynnn007: Ok, I think this might be complicating the api than just json content/path. But I guess I can live with it. btw, `as_single_value` is actually a combination of both `as_string` and `as_list` options? woud we split them?

fitzthum: `as_single_value` is unrelated to `as_string`. Numeric values will also be in a list by default unless you use `as_single_value`. This is for backwards compatibility since most platforms use `in` for every check in the policy.

Xynnn007 (approved):

PR #841, opened 2025-07-01
KBS: add initdata and runtimedata for KBS protocol

Xynnn007:
This patch changes the Attestation message format to include two parts of claims, s.t. `init-data` and `runtime-data`. This provides flexibility and extensibility for future use and also easy to include current tee-keys and nonce claims.

### DIFF
--- a/kbs/docs/kbs_attestation_protocol.md
+++ b/kbs/docs/kbs_attestation_protocol.md
@@ -143,7 +143,13 @@ evidence from the HW-TEE platform and organizes it into the following payload:
      * It is valid until the next time an attestation is required. Its hash must
      * be included in the HW-TEE evidence and signed by the HW-TEE hardware.
      */
-    "tee-pubkey": $pubkey
+    /* Initdata plaintext */
+    "init-data": {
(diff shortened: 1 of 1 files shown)


Xynnn007:
The lint error is not related to this PR.

kbs/docs/kbs_attestation_protocol.md
```diff
@@ -155,7 +161,24 @@ The KBS matches the attestation evidence to an attestation challenge with the
 HTTP Cookie that the KBC includes in the HTTP request that contains the
 `Attestation` payload.
 
-- `tee-pubkey`
+- `init-data`
+   Init data is an **OPTIONAL** structured document associated with a TEE instance, typically
+   containing mutable configuration settings.
```
mkulke: "mutable" might be a misleading term here, indicating that something is able to change during its lifetime (initdata itself should be immutable. no).
```suggestion
   containing measured configuration settings.
```

Xynnn007: Oh yes. Thanks for pointing out!

PR #843, opened 2025-07-02
lint: fix code format with cargo fmt

Xynnn007:


tools/kbs-client/src/main.rs
```diff
@@ -153,14 +153,18 @@ async fn main() -> Result<()> {
     let cli = Cli::parse();
 
     let kbs_cert = match cli.cert_file {
-        Some(ref p) => vec![std::fs::read_to_string(p).inspect_err(|_|eprintln!("Failed to read:{}",p.display()))?],
+        Some(ref p) => vec![std::fs::read_to_string(p)
+            .inspect_err(|_| eprintln!("Failed to read:{}", p.display()))?],
```
pawelpros: Could you improve a little bit error msg (all in this file):
```suggestion
            .inspect_err(|_| eprintln!("Failed to read: {}", p.display()))?],
```

pawelpros (approved):

tools/kbs-client/src/main.rs
Xynnn007: Thanks, fixed.

PR #849, opened 2025-07-09
Make the integration tests more generic.

fitzthum:
Previously we were using a somewhat clunky switch statement and a bunch of individual functions to capture the logic of different tests. With a bit of refactoring, we can just use one function to handle all of the different cases. With the new layout, we can easily add tests that use different policies, different reference values, and have different results. 

I plan to extend the test coverage with more policies once this PR is merged.


integration-tests/src/common.rs
```diff
@@ -54,14 +54,41 @@ const DENY_ALL_POLICY: &str = "
...
-    Custom(String),
+    Custom(&'static str),
 }
 
 pub enum RvpsType {
     Builtin,
     Remote,
 }
 
+// An enum that selects between TestParameter configurations
```
Xynnn007: Let's use `///` to convert it into an inline document.

integration-tests/tests/get_resource.rs
```diff
@@ -2,83 +2,130 @@
...
-#[case::ear_allow_all(TestParameters{attestation_token_type: "Ear".to_string(), rvps_type: RvpsType::Builtin }, "allow_all".to_string())]
-#[case::simple_allow_all(TestParameters{attestation_token_type: "Simple".to_string(), rvps_type: RvpsType::Builtin }, "allow_all".to_string())]
-#[case::ear_deny_all(TestParameters{attestation_token_type: "Ear".to_string(), rvps_type: RvpsType::Builtin }, "deny_all".to_string())]
-#[case::simple_deny_all(TestParameters{attestation_token_type: "Simple".to_string(), rvps_type: RvpsType::Builtin }, "deny_all".to_string())]
-#[case::contraindicated(TestParameters{attestation_token_type: "Ear".to_string(), rvps_type: RvpsType::Builtin }, "contraindicated".to_string())]
-#[case::not_contraindicated(TestParameters{attestation_token_type: "Ear".to_string(), rvps_type: RvpsType::Remote }, "not_contraindicated".to_string())]
-#[case::not_contraindicated_device(TestParameters{attestation_token_type: "Ear".to_string(), rvps_type: RvpsType::Remote }, "not_contraindicated_device".to_string())]
-#[case::contraindicated_device(TestParameters{attestation_token_type: "Ear".to_string(), rvps_type: RvpsType::Remote }, "contraindicated_device".to_string())]
+//
```
Xynnn007, line 18: Noticed that you are trying to use the comments to distinguish the different types of the unit tests. Could we use the test name to distinguish? Then it can be reflected via terminal.

integration-tests/tests/get_resource.rs
```diff
@@ -2,83 +2,130 @@
...
+    else {
+        // If the test does not pass, we have a problem.
+        if secret.is_err() {
+            bail!("Failed to get secret when test is expected to pass.");
+        }
+        // If we get a secret, make sure it's the right one.
+        else {
+            assert_eq!(secret?, SECRET_BYTES);
+        }
+    }
```
Xynnn007: Oh. I noticed that the `Result`s and `unwrap()`s are mixing in the integration test crate. 
I suggest that we either use error handling for everything (which will cause the top level to return an error), or use `unwrap` for everything (which will panic in place).

For (integration) tests, we usually use the latter as it can give more context when panic.

integration-tests/tests/get_resource.rs
fitzthum: The names are already fairly descriptive, but I can make them more specific if yo have any suggestions. The comments are partly there to keep it from being one huge block of cases, which is hard to navigate.

integration-tests/tests/get_resource.rs
fitzthum: Ok, I changed these asserts to bails. Ik it's a little more common to panic in tests, but we already have error handling throughout the integration test crate. The tests print a ton of logs so it's usually pretty easy to figure out what went wrong.

integration-tests/tests/get_resource.rs
Xynnn007: Ok. A choice is to add a long common prefix to the same group of tests. but it is not a big deal. I am ok if you do not want to do any change.

integration-tests/tests/get_resource.rs
```diff
@@ -2,83 +2,133 @@
...
+        if secret.is_err() {
+            bail!("Failed to get secret when test is expected to pass.");
+        }
+        // If we get a secret, make sure it's the right one.
+        else {
+            if secret? != SECRET_BYTES {
+                bail!("Secret retrieved, but secret has wrong value");
+            }
+        }
+    }
```
Xynnn007, line 130: What we want is to compare the expected Result to the got one, right? The nested structure would be somehow straightforward. We can mark the `expected_result` as `Result<String>` and take a simple condition
```rust
if format!("{expected_result:?}") != format!("{secret:?}") {
    bail!("result assertion failed.");
}
```

integration-tests/tests/get_resource.rs
fitzthum: I reworked the names a little bit.

integration-tests/tests/get_resource.rs
fitzthum: Not sure if that's really idiomatic, but I guess it should work in this case.

fitzthum: Actually looks like that doesn't work since the formatted version of the error includes the entire backtrace. (at least on some platforms). I went back to the old way although it is a little clunky.

PR #850, opened 2025-07-10
Enable CORS support in trustee for browser scenario

RodgerZhu:
[Background]:
We are working on a Confidential AI project (https://github.com/intel/confidential-computing-zoo/tree/main/cczoo/confidential_ai),which is built based-on TDX + Open-webui. We will leverage attestation service (restful api) in trustee to do the attestation related work. Open-webui(https://github.com/open-webui/open-webui) is an open-sourced project which provides a webui for AI services.

[Description]
When Open-webui tries to access as via http, it will show below error message:
Access to XMLHttpRequest at 'http://xxxx:8080/attestation' from origin 'http://xxxx:18080' has been blocked by CORS policy: Resp                  onse to preflight request doesn't pass access control check: It does not have HTTP ok status.

[New]
Add CORS policy to permit access from open-webui request.

Test scripts:
1. To use cors mode for browser, please add below cmd in as-restful Dockerfile:
```
ENV ALLOWED_ORIGIN="http://localhost"
CMD exec restful-as --socket 0.0.0.0:8080 --config-file /etc/config.json --allowed_origin ${ALLOWED_ORIGIN}
```
2. Build restful-as docker images
3. run restful-as container with new parameter ALLOWED_ORIGIN for specific region access to as-restful service.
```
   docker run -d --network=host -e ALLOWED_ORIGIN=http://origin_address:port/  -v /etc/sgx_default_qcnl.conf:/etc/sgx_default_qcnl.conf -p 8080:8080 docker_img_id
```
4. Run attest request in open-webui to check attestation status

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,6 +45,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "actix-cors"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "daa239b93927be1ff123eebada5a3ff23e89f0124ccb8609234e5103d5a5ae6d"
+dependencies = [
+ "actix-utils",
+ "actix-web",
+ "derive_more",
+ "futures-util",
+ "log",
+ "once_cell",
+ "smallvec",
+]
+
+[[package]]
 name = "actix-http"
 version = "3.11.0"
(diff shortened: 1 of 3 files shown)


RodgerZhu:
> @RodgerZhu btw, the CI is not passed.

Updated the format.

PR #851, opened 2025-07-10
Add TPM verifier

bpradipt:


attestation-service/src/token/ear_default_policy_cpu.rego
```diff
@@ -202,4 +202,18 @@ configuration := 2 if {
...
+	input.tpm.svn in data.reference.tpm_svn
+}
+
+executables := 3 if {
+	input.tpm.pcrs[11] in data.reference.tpm_pcr11
+}
+
+configuration := 2 if {
+	input.tpm.pcrs[11] in data.reference.tpm_pcr11
+}
```
fitzthum, line 227: Not totally sure it makes sense to have two claims that depend on the same condition.

The claims do all get default values at the top, but if you don't want to use one of the claims, you can just set it to zero if the TPM platform is in use. 

Something like this
```
configuration := 0 if {
	input.tpm
}
```

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+pub fn verify_tpm_quote_signature(_tpm_quote: &TpmQuote) -> Result<()> {
+    // TODO: Implement actual TPM quote signature verification using AK pubkey 
+    Ok(())
+}
+
+pub fn parse_tpm_evidence(ev: &Evidence) -> Result<TeeEvidenceParsedClaim> {
+    let claims_map = json!({
+        "svn": ev.svn,
+        "report_data": ev.report_data,
+        "signature": ev.tpm_quote.signature,
```
fitzthum: Not sure that we need/want the signature as a claim. Do we expect the policies to do something with this?

deps/verifier/src/sample/mod.rs
```diff
@@ -14,6 +14,20 @@ struct SampleTeeEvidence {
     report_data: String,
 }
 
+#[derive(Deserialize, Debug)]
+pub struct Evidence {
+    pub svn: String,
+    pub report_data: String,
+    pub tpm_quote: TpmQuote,
```
fitzthum: I would probably keep the sample verifier and the TPM verifier totally separate, but I'm not quite sure what you have in mind.

fitzthum (commented):
This is still a draft, so I won't bug you about it, but I left a couple basic comments that might be useful.

attestation-service/src/token/ear_default_policy_cpu.rego
bpradipt: I see.. I was not exactly sure what to have it here. thanks

deps/verifier/src/sample/mod.rs
bpradipt: My thinking was using tpm quote in sample attester for testing. But re-thinking about it, it's not really needed as we can use swtpm. I'll remove it

bpradipt: This will also avoid changing the GHA which are failing due to missing TPM deps

deps/verifier/src/tpm/mod.rs
bpradipt: Correct.

alicefr:
The kbs-client now depends on the tpm libraries, so it is difficult to statically build it. Would it possible to dymanically build it in the container image kbs-client-image?

mkulke:
is there an option to re-use some parts of that within the az-*-vtpm verifiers? or is too specifc and not worth it?

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,142 @@
...
+use serde_json::json;
+use sha2::Digest;
+use sha2::Sha256;
+use tss_esapi::structures::{Attest, Public,Signature as TpmSignature};
+use tss_esapi::traits::UnMarshall;
+
+use super::*;
+
+#[derive(Deserialize, Debug)]
+pub struct Evidence {
```
mythi, line 32: would BIOS eventlog be available on these systems you are targeting?

bpradipt:
> is there an option to re-use some parts of that within the az-*-vtpm verifiers? or is too specifc and not worth it?

I'll check and revert

fitzthum:
@mkulke 

> is there an option to re-use some parts of that within the az-*-vtpm verifiers? or is too specifc and not worth it?

At some point we should try to develop a more flexible, composable TPM interfac where the TPM and the underlying platform are handled separately with the binding logic imlpemented on top. One option  for this is to treat the vTPM as a second device and check the binding in the policy. We don't need to implement this yet, but worth thinking about.

mkulke:
> @mkulke
> 
> > is there an option to re-use some parts of that within the az-*-vtpm verifiers? or is too specifc and not worth it?
> 
> At some point we should try to develop a more flexible, composable TPM interfac where the TPM and the underlying platform are handled separately with the binding logic imlpemented on top. One option for this is to treat the vTPM as a second device and check the binding in the policy. We don't need to implement this yet, but worth thinking about.

yes, in theory I can imagine this could be reworked into a composite evidence. not sure how this will look like in detail: e.g. the init-data binding would look different than on a SNP/TDX vm, so that would be part of the TPM device. also, the binding from a CPU to TPM would not be determined by guest-components, so that would have to be abstracted for az-cvm-vtpm TEEs.

bpradipt:
@Xynnn007 @mkulke @fitzthum this is ready for review. I have added an e2e test that should help to test the flow using sw TPM. The commit message has the details. The verifier uses pre-created AK public keys to verify the quote.

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+
+        // Decode the attestation message
+        let message = general_purpose::STANDARD
+            .decode(&self.message)
+            .context("Failed to decode message from base64")?;
+
+        // Parse PCR values from hex strings to byte arrays
+        let pcrs = self.parse_pcr_values()?;
+
+        // Construct Quote with the corrected signature format
```
mkulke, line 60: it's a bit unfortunate that we are creating a json just to be able to parse it into the quote type, but yeah there is no public constructor sadly :/

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+        Self::new(config).unwrap_or_else(|_| Self {
+            trusted_ak_hashes: HashSet::new(),
+        })
+    }
+}
+
+impl TpmVerifier {
+    pub fn new(config: config::TpmVerifierConfig) -> Result<Self> {
+        let mut trusted_ak_hashes = HashSet::new();
+        if let Some(keys_dir) = config.trusted_ak_keys_dir {
```
mkulke: maybe early return here to safe a level of indent?

```
let Some(keys_dir) = config.trusted_ak_keys_dir else {
   return Ok(Self { trusted_ak_hashes });
}
```

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+                    "Number of trusted AK keys ({}) exceeds the limit ({}). Only the first {} keys will be loaded.",
+                    entries.len(),
+                    config.max_trusted_ak_keys,
+                    config.max_trusted_ak_keys
+                );
+            }
+
+            for entry in entries.into_iter().take(config.max_trusted_ak_keys) {
+                let path = entry.path();
+                if path.is_file() {
```
mkulke: we can reduce a level of indent, if we continue here

```rust
if !path.is_file() {
  continue;
}
```

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+                        continue;
+                    }
+
+                    // Try to read as text first to avoid UTF-8 errors on binary files
+                    let key_content = match fs::read_to_string(&path) {
+                        Ok(content) => content,
+                        Err(_) => {
+                            // Skip files that can't be read as UTF-8 text
+                            debug!("Skipping non-UTF8 file: {:?}", path);
+                            continue;
```
mkulke: personally, I think we are too forgiving here. If a user specifies a folder w/ public keys and there is a `.pub` file that isn't readable, we should bail, IMO.

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+
+    debug!("Expected report_data ({} bytes): {}", expected_report_data.len(), hex::encode(expected_report_data));
+    debug!("Quote nonce ({} bytes): {}", nonce.len(), hex::encode(&nonce));
+
+    // TPM attester pads report_data to 64 bytes, so we need to pad expected_report_data the same way
+    const TPM_REPORT_DATA_SIZE: usize = 64;
+    let mut padded_expected = expected_report_data.to_vec();
+    padded_expected.resize(TPM_REPORT_DATA_SIZE, 0);
+
+
```
mkulke: newline

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+        true => {
+            debug!("TPM report_data verification completed successfully");
+            Ok(())
+        }
+        false => {
+            debug!("Nonce and padded expected are different");
+            Err(TpmVerifierError::NonceMismatch.into())
+        }
+    }
+}
```
mkulke: true/false should be covered with if/else statements (i think there is some clippy rule enforcing this)

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+        };
+
+        // 1. Check if the provided AK public key is trusted
+        let ak_public_bytes = general_purpose::STANDARD.decode(&ev.ak_public)?;
+        let ak_public_hash = Sha256::digest(&ak_public_bytes).to_vec();
+
+        match self.trusted_ak_hashes.contains(&ak_public_hash) {
+            true => {}
+            false => return Err(TpmVerifierError::UntrustedAkKey.into()),
+        }
```
mkulke: maybe a bit clearer:

```rust
       if !self.trusted_ak_hashes.contains(&ak_public_hash) {
         return Err(TpmVerifierError::UntrustedAkKey.into());
       }
```

deps/verifier/src/lib.rs
```diff
@@ -137,7 +140,26 @@ pub fn to_verifier(tee: &Tee) -> Result<Box<dyn Verifier + Send + Sync>> {
             }
         }
 
-        Tee::Tpm => todo!(),
+        Tee::Tpm => {
+            cfg_if::cfg_if! {
+                if #[cfg(feature = "tpm-verifier")] {
+                    let config_path = std::env::var("TPM_CONFIG_FILE")
+                        .unwrap_or_else(|_| "/etc/tpm_verifier.json".to_string());
```
Xynnn007: Can we add a log here like `info!` or `debug!` to show the config actually used.

deps/verifier/src/lib.rs
```diff
@@ -137,7 +140,26 @@ pub fn to_verifier(tee: &Tee) -> Result<Box<dyn Verifier + Send + Sync>> {
...
+                    let config = match tpm::config::Config::try_from(std::path::Path::new(&config_path)) {
+                        std::result::Result::Ok(c) => c.tpm_verifier,
+                        std::result::Result::Err(e) => {
+                            log::warn!("Failed to load TPM config file: {}. Using default.", e);
+                            tpm::config::TpmVerifierConfig::default()
```
Xynnn007: No suggestions here, just to ensure that "if failed to read the specfied config file, it will use the default one than raise an error" is expected.

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+const MAX_TRUSTED_AK_KEYS: usize = 100;
+const INITDATA_PCR: usize = 8;
+
+// Simple hex decode function since hex crate isn't available
+fn decode_hex(s: &str) -> Result<Vec<u8>, std::num::ParseIntError> {
+    (0..s.len())
+        .step_by(2)
+        .map(|i| u8::from_str_radix(&s[i..i + 2], 16))
+        .collect()
+}
```
Xynnn007: I see `verifier` imports `hex` [here](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/Cargo.toml#L49). Is there anything wrong that causes "isn't available"?

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+
+
+    match nonce == padded_expected {
+        true => {
+            debug!("TPM report_data verification completed successfully");
+            Ok(())
+        }
+        false => {
+            debug!("Nonce and padded expected are different");
+            Err(TpmVerifierError::NonceMismatch.into())
```
Xynnn007: Looks like we are mixing `anyhow` and `thiserror` in the code. The `Err(TpmVerifierError::NonceMismatch.into())` here is equal to 
`bail!("TPM quote nonce doesn't match expected report_data")`.

Usually, if we want to use thiserror, we can define a `Result<T> = std::result::Result<T, TpmVerifierError>` and return the new `Result` in all `pub` functions. If we want to return the `anyhow::Result`, we do not need to define the consts but directly write error message in-flight.

I suggest to get rid of `thiserror`and use anyhow for all places to make it simpler

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+            trusted_ak_hashes: HashSet::new(),
+        })
+    }
+}
+
+impl TpmVerifier {
+    pub fn new(config: config::TpmVerifierConfig) -> Result<Self> {
+        let mut trusted_ak_hashes = HashSet::new();
+        if let Some(keys_dir) = config.trusted_ak_keys_dir {
+            let entries = fs::read_dir(&keys_dir)?.collect::<Result<Vec<_>, _>>()?;
```
Xynnn007: I am not sure if `.` and `..` will be treated as an entry. If so, we might need to skip them

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+                    config.max_trusted_ak_keys
+                );
+            }
+
+            for entry in entries.into_iter().take(config.max_trusted_ak_keys) {
+                let path = entry.path();
+                if path.is_file() {
+                    // Only process files with .pub extension
+                    if let Some(extension) = path.extension() {
+                        if extension != "pub" {
```
Xynnn007: We should mention this explicitly in the README that only files with name `*.pub` are valid.

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+                    let key_content = match fs::read_to_string(&path) {
+                        Ok(content) => content,
+                        Err(_) => {
+                            // Skip files that can't be read as UTF-8 text
+                            debug!("Skipping non-UTF8 file: {:?}", path);
+                            continue;
+                        }
+                    };
+
+                    let pkey = PKey::public_key_from_pem(key_content.as_bytes())
```
Xynnn007: we might need to change base64 to PEM [there](https://github.com/confidential-containers/trustee/pull/851/commits/5bda66d6de52544a22fe40b37ddfc8c93c3b1d67) ?

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+}
+
+fn verify_nonce(quote: &Quote, expected_report_data: &[u8]) -> Result<()> {
+    let nonce = quote.nonce()?;
+
+    debug!("Expected report_data ({} bytes): {}", expected_report_data.len(), hex::encode(expected_report_data));
+    debug!("Quote nonce ({} bytes): {}", nonce.len(), hex::encode(&nonce));
+
+    // TPM attester pads report_data to 64 bytes, so we need to pad expected_report_data the same way
+    const TPM_REPORT_DATA_SIZE: usize = 64;
```
Xynnn007: Can we move this const to the top of the file?

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+    let mut padded_expected = expected_report_data.to_vec();
+    padded_expected.resize(TPM_REPORT_DATA_SIZE, 0);
```
Xynnn007, line 211: Sorry but I am not sure what is the difference between`0` and`'\0'` here?

deps/verifier/src/tpm/mod.rs
Xynnn007: we can take a `cargo fmt` here, as usually two empty lines will be optimized to one

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+            // sha256(0x00 * 32 || expected_init_data_hash)
+            let mut input = [0u8; 64];
+            input[32..].copy_from_slice(expected_init_data_hash);
+            let digest = openssl::sha::sha256(&input);
+
+            let pcrs: Vec<&[u8; 32]> = quote.pcrs_sha256().collect();
+            let init_data_pcr = pcrs[INITDATA_PCR];
+            match &digest == init_data_pcr {
+                true => Ok(()),
+                false => Err(TpmVerifierError::InitDataMismatch.into()),
```
Xynnn007: same for thiserror vs anyhow

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+    };
+    let pcrs: Vec<&[u8; 32]> = quote.pcrs_sha256().collect();
+    let mut tpm_values = serde_json::Map::new();
+    for (i, pcr) in pcrs.iter().enumerate() {
+        tpm_values.insert(
+            format!("pcr{:02}", i),
+            serde_json::Value::String(hex::encode(pcr)),
+        );
+    }
+    map.insert("tpm".to_string(), serde_json::Value::Object(tpm_values));
```
Xynnn007: looks like we have a nested structure here -- the parsed claims will be `tpm.tpm.xxx`. Is this expected?

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+    map.insert(
+        "init_data".into(),
+        serde_json::Value::String(hex::encode(pcrs[INITDATA_PCR])),
+    );
+    map.insert(
+        "report_data".into(),
+        serde_json::Value::String(hex::encode(quote.nonce()?)),
+    );
```
Xynnn007: we have already inserted the two items here https://github.com/confidential-containers/trustee/pull/851/commits/5bda66d6de52544a22fe40b37ddfc8c93c3b1d67#diff-a600e603880544f2930d0f17c6e89b7602c457edbe62df1a3cc748169d723485R356-R363

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+        }
+
+        // 5. Verify init data hash
+        verify_init_data(expected_init_data_hash, &ev.quote)?;
+
+        // 6. Parse claims
+        let mut claims = parse_tee_evidence(&ev.quote);
+        extend_claim(&mut claims, &ev.quote)?;
+
+        Ok(vec![(claims, "cpu".to_string())])
```
Xynnn007, line 305: cc @fitzthum . I'm still not convinced by this submodule or teetype definition. TPM is actually a bootchain measurement, or rather, measurement of the guest software, not the "CPU." Given that most measurement currently rely on the CPU TEE for guest softwares, I'm fine with this for now in this patch. However, I'll consider replacing teetype with a better descriptive method later—currently, it seems its sole purpose is to define and index different submodule processing logic and results within the EAR module.

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+            "TPM PCR[8] doesn't match expected init_data_hash"
+        );
+    }
+
+    #[rstest]
+    fn test_missing_field_error() {
+        let error = TpmVerifierError::MissingField("test_field".to_string());
+        assert_eq!(error.to_string(), "Evidence field missing: test_field");
+    }
+}
```
Xynnn007: these unit tests fall on the same question of thiserror vs anyhow

kbs/Cargo.toml
```diff
@@ -40,6 +40,8 @@ nebula-ca-plugin = []
 # Use HashiCorp Vault KV v1 as KBS backend
 vault = ["vaultrs"]
 
+#tpm-verifier = ["attestation-service/tpm-verifier"]
+
```
Xynnn007: Is this something left?

Xynnn007 (commented):
Thanks @bpradipt and @alicefr for this! some nits in round 1 : )

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,439 @@
...
+        // 3. Verify PCRs
+        verify_pcrs(&ev.quote)?;
+
+        // 4. Verify nonce/report data
+        match expected_report_data {
+            ReportData::Value(expected_report_data) => {
+                verify_nonce(&ev.quote, expected_report_data)?;
+            }
+            ReportData::NotProvided => {}
+        }
```
mkulke:
```rust
if let ReportData::Value(report_data) {
   verify_nonce(&ev.quote, report_data)?;
}
```

kbs/test/setup-swtpm.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+set -euo pipefail
+
+WORK_DIR="./work"
+TRUSTED_TPM_KEYS_DIR="$WORK_DIR"/tpm_trusted_keys
+TPM_STATE_DIR="/tmp/tpmdir"
+
+setup_swtpm() {
+    sudo modprobe tpm_vtpm_proxy
```
mkulke: can we move sudo to the caller of this script? (in theory swtpm shouldn't require root, no?)

kbs/Cargo.toml
```diff
@@ -40,6 +40,8 @@ nebula-ca-plugin = []
 # Use HashiCorp Vault KV v1 as KBS backend
 vault = ["vaultrs"]
 
+#tpm-verifier = ["attestation-service/tpm-verifier"]
```
mkulke: leftover?

deps/verifier/src/tpm/mod.rs
bpradipt: As I see the generic format is  "$tee".tpm...  and that's why it looks odd here as tee is "tpm" :-(

kbs/test/setup-swtpm.sh
bpradipt: Unfortunately swtpm needs root access. I have updated the commit message with the ownership details and why the test must be run as root.

bpradipt:
@mkulke @Xynnn007 thanks a lot for your review comments. I have created separated commits to address the comments which I'll squash it later. The reason to keep it separate for now is to help myself not to repeat the same mistakes again :-) 
Apologies if I have missed something or overlooked.

deps/verifier/src/tpm/mod.rs
bpradipt: I didn't understand this comment :-( .. Where is `\0` ?

deps/verifier/src/tpm/mod.rs
bpradipt: I'm using the crate now.

deps/verifier/src/lib.rs
bpradipt: Yeah, it's expected. Config file is not mandatory

deps/verifier/src/tpm/mod.rs
Xynnn007: Oh. I mean why do we use `padded_expected.resize(TPM_REPORT_DATA_SIZE, 0);` than `padded_expected.resize(TPM_REPORT_DATA_SIZE, '\0');`. Are the two equal? I am not sure for this, just a question

deps/verifier/src/tpm/mod.rs
mkulke: yeah, I think conceptually a tpm attester is challenging some assumptions that we have, eventually in the new multi-device scheme, I imagine tpm-only evidences would map to a composite evidence with `CPU: [], device: [TPM]`. If we don't want do this yet, we can maybe unnest the PCRs and put them in into the top level, afaiu the evidence is just serde_json types w/o schema.

deps/verifier/src/tpm/mod.rs
bpradipt: Honestly I don't know, so I asked Gemini and this is what it says
```
padded_expected.resize(TPM_REPORT_DATA_SIZE, 0)
This is the most common and idiomatic way to pad a byte vector (Vec<u8>). The second argument is a u8 value, and 0 is a clear and direct way to specify the byte with the value zero.

padded_expected.resize(TPM_REPORT_DATA_SIZE, '\0')
This uses the null character literal, '\0'. When used in a context requiring a u8, Rust converts this character to its underlying numeric value, which for the null character is 0. So, you are still padding with the byte 0.
```

mkulke: yes, it will be coerced into the same. but since we operate on bytes the using a char is a bit confusing

deps/verifier/src/tpm/mod.rs
mkulke: I think this is "hardware" evidence in some form, so it's not just measurements, but also a signature from the HW's private key. but I agree, it's not CPU, the hierarchy is confusing. Would we be able to have a sort of "null" cpu and TPM as a device?

deps/verifier/src/tpm/mod.rs
bpradipt: My mistake. I interpreted the map wrongly. The claim should be now just tpm.pcrXX

deps/verifier/src/tpm/mod.rs
bpradipt: You are right. This will not be filtered. Also there seems to be another issue with this approach. It'll read all the files irrespective of the limit. I'll update the implementation in the next iteration.

kbs/test/setup-swtpm.sh
```diff
@@ -0,0 +1,73 @@
...
+                --pcr-banks - \
+                --display
+
+    swtpm chardev --tpm2 \
+                --tpmstate dir="$TPM_STATE_DIR" \
+                --vtpm-proxy \
+                --daemon \
+                --log file="$TPM_STATE_DIR"/tpm.log,level=20 \
+                --flags not-need-init \
+		--pid file="$TPM_STATE_DIR"/swtpm.pid
```
mkulke, line 30: indent

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+
+#[async_trait]
+impl Verifier for TpmVerifier {
+    async fn evaluate(
+        &self,
+        evidence: TeeEvidence,
+        expected_report_data: &ReportData,
+        expected_init_data_hash: &InitDataHash,
+    ) -> Result<Vec<(TeeEvidenceParsedClaim, TeeClass)>> {
+        // Try to deserialize in guest-components format first, then fall back to original format
```
mkulke: Why do we expect different `TeeEvidence` (actually just a an alias to a String afaik)?

deps/verifier/src/lib.rs
```diff
@@ -139,14 +142,34 @@ pub fn to_verifier(tee: &Tee) -> Result<Box<dyn Verifier + Send + Sync>> {
...
                 } else {
                     bail!("feature `nvidia-verifier` is not enabled for `verifier` crate.")
+
+                }
+            }
+        }
+        Tee::Tpm => {
+            cfg_if::cfg_if! {
+                if #[cfg(feature = "tpm-verifier")] {
+                    let config_path = std::env::var("TPM_CONFIG_FILE")
```
mkulke: Instead of having a config file per verifier (which can get unwieldy quickly) can we have a verifier-specitic section in the attestion-server config?

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+            let mut input = [0u8; 64];
+            input[32..].copy_from_slice(expected_init_data_hash);
+            let digest = openssl::sha::sha256(&input);
+
+            let pcrs: Vec<&[u8; 32]> = quote.pcrs_sha256().collect();
+            let init_data_pcr = pcrs[INITDATA_PCR];
+            if &digest == init_data_pcr {
+                Ok(())
+            } else {
+                bail!("TPM PCR[8] doesn't match expected init_data_hash")
```
mkulke:
```suggestion
                bail!(format!("TPM PCR[{INITDATA_PCR}] doesn't match expected initdata hash"))
```

deps/verifier/src/lib.rs
bpradipt: I'm not sure. But I'll check and revert

deps/verifier/src/tpm/mod.rs
fitzthum: I think it probably makes sense to have the `vTPM` Tee class here. This will either mean adding a possible public key path in the KBS JWT code (because there will be no `cpu0`) or we could amend the AS to treat the vTPM as an additional device rather than a primary device. In the second case, when the vTPM is used without Tee hardware, the sample attester would be used for the primary evidence. I think this lines up reasonably well with the use case for this PR.

deps/verifier/src/lib.rs
bpradipt: @mkulke would it make sense to have a verifier config file with sections for different verifiers instead of coupling it with attestation-service config ?
Also I ran into circular dep issues when adding the tpm config file into attestation-service handling. May be I'm doing something wrong, but before spending more time, I wanted to understand if this is really what you are looking at? or is there a simpler approach ?

bpradipt: Can I add a verifiers section in the attestation-service config and just process this section in the tpm-verifier? Would that be acceptable ?

mkulke: afaik the verifier dep is only used by attestation-service, so there shouldn't be a circular dep in theory. If we desire to keep all the TEE specifics in the verifier module then we need to export a `verifier::Config` from the verifier and use that in the attestation-service Config struct.

mkulke: > Can I add a verifiers section in the attestation-service config and just process this section in the tpm-verifier? Would that be acceptable ?

yes, this is what I meant

deps/verifier/src/lib.rs
```diff
@@ -149,15 +155,28 @@ pub fn to_verifier(
...
-        Tee::Tpm => todo!(),
-
         Tee::Nvidia => {
             cfg_if::cfg_if! {
                 if #[cfg(feature = "nvidia-verifier")] {
                     let nvidia_config = _config.map(|c| c.nvidia_verifier).unwrap_or(None);
                     Ok(Box::<nvidia::Nvidia>::new(nvidia::Nvidia::new(nvidia_config)) as Box<dyn Verifier + Send + Sync>)
                 } else {
                     bail!("feature `nvidia-verifier` is not enabled for `verifier` crate.")
+
```
Xynnn007: This empty line should be moved above the `Tee::Tpm => {` line, right?

deps/verifier/src/tpm/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+        let hash = Sha256::digest(&key_bytes).to_vec();
+        Ok(hash)
+    }
+
+    pub fn new(config: config::TpmVerifierConfig) -> Result<Self> {
+        let mut trusted_ak_hashes = HashSet::new();
+
+        let Some(keys_dir) = config.trusted_ak_keys_dir else {
+            return Ok(Self { trusted_ak_hashes });
+        };
```
Xynnn007: A question: according to https://github.com/confidential-containers/trustee/pull/851/files#diff-6dd737e62e929168d6420c106ea492f271aa3810f22fa02c5e79ea5800741b9fR18-R19, `trusted_ak_keys_dir` will be given a default value `DEFAULT_TRUSTED_AK_KEYS_DIR` if user does not specify it in the config. That might mean that anyway the `trusted_ak_keys_dir` is `Some(_)`. I am not totally sure - If so, I wonder if `trusted_ak_keys_dir` can be defined without the type `Option`, and given a default value. In this way we can avoid extra condition judgement.

deps/verifier/src/lib.rs
bpradipt: yeah ;)

deps/verifier/src/tpm/mod.rs
bpradipt: Thanks for the suggestion. I'll make the necessary changes

deps/verifier/src/tpm/mod.rs
bpradipt: I have not seen it in the test systems. However I'll take a relook and figure out addition in a future PR.

PR #852, opened 2025-07-14
kbs/config/kubernetes: update image tags for the release

mythi:
The new images must be updated in main before the release so that they are pulled when installing the from the release branch.

### DIFF
--- a/kbs/config/kubernetes/base/kustomization.yaml
+++ b/kbs/config/kubernetes/base/kustomization.yaml
@@ -5,7 +5,7 @@ namespace: coco-tenant
 images:
 - name: kbs-container-image
   newName: ghcr.io/confidential-containers/key-broker-service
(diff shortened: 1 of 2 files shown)


fitzthum (commented):
Ah good thinking. Better to do this before the release for once.

The only question is what we should be setting the tag to. The v0.13.0 image is actually from the last release. In theory we should be releasing Trustee before Kata. We would first run the release helper script to tag the latest image as the release image. Then we would bump the kustomization files. Then we would update kata to point to the released images. Doing all those steps isn't always going to be practical tho.

Btw, I think we had to respin the CI due to a little race condition in starting up the services.

mythi:
> > The only question is what we should be setting the tag to. The v0.13.0 image is actually from the last release.
> 
> ah! I did a blind +1 bump and forgot about this. I really meant v0.14.0.

updated

PR #853, opened 2025-07-15
Use Rego extensions to get reference values

fitzthum:
Today the RVPS provides all known reference values to the AS to evaluate the AS policy.

This has several downsides. For instance, the set of all reference values may be much larger than the set of reference values that are needed to evaluate the AS policy on a particular platform. Also, this doesn't allow us to support a pull model, because all the RVs have to be known ahead of time. Finally, this doesn't allow us to keep track of which reference values a policy actually uses, which is information that we may want to report in the attestation token at some point.

Enter rego extensions, Rust functions (well, closures really) that can be called from the policy. This PR adds the extension `get_reference_value`, which takes a reference value name and fetches it from the RVPS. This was deceptively difficult to implement. See commit messages for more info. 

This PR isn't quite done, but it's ready to review. I may need to fix up a few configs. I didn't fix up the simple policy, which in its current state won't really work. We could a) update the policy to use `get_reference_value` b) allow the simple policy to continue getting all the RVs in a JSON map. wdyt @Xynnn007?


attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -10,19 +10,24 @@ use sha2::{Digest, Sha384};
...
 
 use super::{EvaluationResult, PolicyDigest, PolicyEngine, PolicyError};
 
-#[derive(Debug, Clone)]
+use crate::RvpsApi;
+
+#[derive(Clone)]
 pub struct OPA {
     policy_dir_path: PathBuf,
+    rvps: Arc<Mutex<dyn RvpsApi + Send + Sync>>,
```
Xynnn007, line 22: We can move the parameter to `evaluate` API than a member of `OPA` struct, which can help to avoid deliver the parameter in too many levels.

attestation-service/src/rvps/builtin.rs
```diff
@@ -17,14 +17,20 @@ impl BuiltinRvps {
...
-    async fn verify_and_extract(&mut self, message: &str) -> Result<()> {
-        self.rvps.verify_and_extract(message).await?;
+    fn verify_and_extract(&mut self, message: &str) -> Result<()> {
+        self.rvps.verify_and_extract(message)?;
         Ok(())
     }
 
-    async fn get_digests(&self) -> Result<HashMap<String, serde_json::Value>> {
-        let hashes = self.rvps.get_digests().await?;
+    fn get_digest(&self, id: String) -> Result<serde_json::Value> {
```
Xynnn007, line 25: We can rename all `get_digest` to `get_reference_value` imo

attestation-service/src/token/ear_default_policy_cpu.rego
```diff
@@ -90,18 +96,20 @@ else := 3 if {
 
 ##### TDX
 executables := 3 if {
+    input.tdx
 	# Check the kernel, initrd, and cmdline (including dmverity parameters) measurements
-	input.tdx.quote.body.rtmr_1 in data.reference.rtmr_1
-	input.tdx.quote.body.rtmr_2 in data.reference.rtmr_2
+	input.tdx.quote.body.rtmr_1 in get_reference_value("rtmr_1")
```
Xynnn007, line 105: I think we can comment out all the usage of `get_reference_value`, just leave the determined `A=B` rules.

attestation-service/src/token/ear_broker.rs
```diff
@@ -248,7 +243,7 @@ impl AttestationTokenBroker for EarAttestationTokenBroker {
             let policy_id = format!("{}_{}", policy_ids[0], tee_claims.tee_class);
```
Xynnn007, line 243: Probably not related to this PR. I have tried locally with ear policy naming rules, it is somehow troublesome. I suggest to get rid of coupling the tee_class name with policy name, that means

The `policy_id` given to attestation-service's API is actually the `policy_id` that was just registered. The current behavior is that if we want to verify with policy id `some-id`, we need to register a policy with id `some-id_cpu`, and the backend storage will also show `some-id_cpu.rego`.

attestation-service/src/token/ear_broker.rs
```diff
@@ -248,7 +243,7 @@ impl AttestationTokenBroker for EarAttestationTokenBroker {
...
             let policy_results = self
                 .policy_engine
-                .evaluate(&reference_data, &tcb_claims_json, &policy_id, rules)
+                .evaluate("{}", &tcb_claims_json, &policy_id, rules)
```
Xynnn007, line 246: We can change the OPA API's data to `Option` and use `None` here.

rvps/src/storage/mod.rs
```diff
@@ -50,11 +50,11 @@ impl ReferenceValueStorageConfig {
 pub trait ReferenceValueStorage {
     /// Store a reference value. If the given `name` exists,
     /// return the previous `Some<ReferenceValue>`, otherwise return `None`
-    async fn set(&self, name: String, rv: ReferenceValue) -> Result<Option<ReferenceValue>>;
+    fn set(&self, name: String, rv: ReferenceValue) -> Result<Option<ReferenceValue>>;
 
-    // Retrieve reference value by name
-    async fn get(&self, name: &str) -> Result<Option<ReferenceValue>>;
+    /// Retrieve reference value by name
+    fn get(&self, name: String) -> Result<Option<ReferenceValue>>;
```
Xynnn007, line 56: we can unify all `name`, `id` things to a same term, probably `id` in this PR.

Xynnn007 (commented):
The changes are more than I once imagined, but the [actix-web framework](https://docs.rs/actix-rt/latest/actix_rt/) claims to be
```
Tokio-based single-threaded async runtime for the Actix ecosystem.
```

Thus it is hard for us to block a single thread to run async jobs. So I am ok with the way this PR shows, although it might cause performance decrease.

attestation-service/src/token/ear_broker.rs
fitzthum: Are you suggesting that we share one policy between all the different classes? This is possible but it would make the policy really big.

Xynnn007: Not even wrong. What I mean is that the choice of which policy to use is made by the caller of the AS, and the caller can choose different policy combinations to verify different devices according to the needs. 

The potential challenge here is the need to support multiple policies. But we can reuse the `appraisal.policy_id` with a `,` to separate.

Also, the KBS does not support to select the policy id now. It's KBS' responsibility to make a change, but not AS.

fitzthum: Yeah I am open to allowing the caller to specify the policy for each class individually. I didn't do this at first because specifying a policy per class will make the API more complex, but I can see it is a little confusing to append things to the policy. id  If we can come up with a nice way to change the API, I wouldn't be against it.

Xynnn007: Yes, I have composed some different ways in the issue https://github.com/confidential-containers/trustee/issues/863

fitzthum:
Superseded by #887

PR #870, opened 2025-07-23
KBS: refactor in prometheus, active connections metric

pmores:
More gradual progress on KBS's Prometheus endpoint:
- capture Prometheus metric objects' repeated/boilerplate initialisation code in macros [as suggested earlier](https://github.com/confidential-containers/trustee/pull/816#discussion_r2139146420)
-  add the active connections metric [as proposed](https://github.com/confidential-containers/trustee/issues/543#issuecomment-2916097383).

### DIFF
--- a/kbs/src/api_server.rs
+++ b/kbs/src/api_server.rs
@@ -17,8 +17,8 @@ use crate::{
     plugins::PluginManager,
     policy_engine::PolicyEngine,
     prometheus::{
-        KBS_POLICY_APPROVALS, KBS_POLICY_ERRORS, KBS_POLICY_EVALS, KBS_POLICY_VIOLATIONS,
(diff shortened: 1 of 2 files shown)


Xynnn007 (approved):
Thanks @pmores . I love the macro definition!

lmilleri:
Would it make sense to use a naming convention for the metrics? For example:
"Application Prefix (Namespace): Metric names should start with a single-word prefix relevant to the application or domain the metric belongs to. This acts as a namespace, preventing collisions and making it clear which application a metric originates from."
I think this could help to identify trustee specific metrics

pmores:
I think that's a good idea.  Which prefix or naming convention would you guys consider useful?

We should merge this one first though.

fitzthum (approved):
LGTM

PR #871, opened 2025-07-23
Eventlog | Support to parse AAEL

Xynnn007:
This patch gets rid of separate AAEL and merge it into CCEL. Relative parsing logic is in this PR.

I have test this within TDX guest on alibaba cloud.

```shell
./ttrpc-aa-client extend-runtime-measurement --domain coco.com --operation some-operation --content some-content
./ttrpc-aa-client get-evidence --runtime-data AAAA
```

The returned evidence can be verified by AS and get the parsed claims as
```json
{
    "uefi_event_logs": [
        {
            "details": {
                "string": "TdxTable"
            },
            // Original CCEL
            "digests": [
                {
                    "alg": "SHA-384",
                    "digest": "951b045980c718ed0f6909c99e159831ea81abd818f8b761412ca2725b5803363e7990365390419dc58dfb789968a3b2"
                }
            ],
            "event": "CVRkeFRhYmxlAAEAAAAAAAAAr5a7k/K5uE6UYuC6dFZCNgCQgAAAAAAA",
            "index": 1,
            "type_name": "EV_EFI_HANDOFF_TABLES2"
        },
                        ...
           // AAEL inside EV_EVENT_TAGv
        {
            "details": {
                "data": {
                    "content": "some-content",
                    "domain": "coco.com",
                    "operation": "some-operation"
                },
                "string": "coco.com some-operation some-content",
                "unicode_name": "AAEL"
            },
            "digests": [
                {
                    "alg": "SHA-384",
                    "digest": "2330183d89750c6a374ac9827d33a33636e7876733544584c8ccc88df0df87eada37e5d71d0773502234c2b5fbc7448e"
                }
            ],
            "event": "TEVBQSQAAABjb2NvLmNvbSBzb21lLW9wZXJhdGlvbiBzb21lLWNvbnRlbnQ=",
            "index": 4,
            "type_name": "EV_EVENT_TAG"
        },
        // Nested AAEL structure
                                {
                            "details": {
                                "data": {
                                    "content": {
                                        "key": "value",
                                        "nested-key": {
                                            "sub-key": "value"
                                        }
                                    },
                                    "domain": "coco.com",
                                    "operation": "some-operation"
                                },
                                "string": "coco.com some-operation {\"key\": \"value\", \"nested-key\": {\"sub-key\": \"value\"}}",
                                "unicode_name": "AAEL"
                            },
                            "digests": [
                                {
                                    "alg": "SHA-384",
[...]

PR #873, opened 2025-07-24
deps/eventlog: improve EV_IPL data parser

pawelpros:
Added removing \0 from event output


pawelpros:
@Xynnn007 It's small improvement which should clean sample from https://github.com/confidential-containers/trustee/pull/871

PR #877, opened 2025-07-25
ci: remove pull_request_target trigger from e2e wf

mkulke:
pull_request_target is being used to run e2e test on real TEE hardware, since this hardware is only available as private runners. PR authors run their tests in the context of a forked repository where the required self-hosted runners do not exist (usually).

GHs pull_request_target workflow opens up various exploit vectors, since it'll execute foreign code in the context of the repository.

The require-authorize-label technique that we employ is not acknowledged by security linters and a bandaid at best (once the label is applied a malicious PR author can push whatever they want).

Since we have rarely cought problems on e2e runs in PRs recently we can probably live with after-the-fact testing on pushes to main.


mkulke:
> Can we remove the CI helper script?

which script?

fitzthum:
> which script?

[This guy](https://github.com/confidential-containers/trustee/blob/main/kbs/hack/ci-helper.sh). Looks like it is only called from some code you are removing here.

fitzthum (approved):
LGTM

PR #881, opened 2025-07-29
eventlog: add sm3 hash algorithm

Xynnn007:
sm3 is a hash algorithm defined in ISO/IEC 10118-
3:2018. The Tcg Algorithm enum can be found in

table 3 of

https://trustedcomputinggroup.org/wp-content/uploads/TCG-Algorithm-Registry-Family-2.0-Level-00-Revision-1.35_pub.pdf


mythi (approved):
Looks simple :+1:

PR #886, opened 2025-07-30
workflows: fixup token permissions

fitzthum:
Appease the scorecard.

Let's see if these are the right permissions.

### DIFF
--- a/.github/workflows/build-kbs-client-image.yml
+++ b/.github/workflows/build-kbs-client-image.yml
@@ -8,6 +8,10 @@ on:
         type: string
         required: false
 
+permissions:
+  contents: read
+  packages: write
+
 jobs:
   build_kbs_client_image:
     strategy:
--- a/.github/workflows/kbs-e2e-azure-vtpm.yml
+++ b/.github/workflows/kbs-e2e-azure-vtpm.yml
@@ -5,9 +5,15 @@ on:
     branches:
     - main
 
+permissions:
+  contents: read
+
 jobs:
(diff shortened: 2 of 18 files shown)


.github/workflows/build-kbs-client-image.yml
```diff
@@ -8,6 +8,10 @@ on:
         type: string
         required: false
 
+premissions:
+  contents: read
+  packages: right
```
mythi: `right` does not look right :-)

.github/workflows/kbs-e2e-azure-vtpm.yml
```diff
@@ -5,6 +5,10 @@ on:
     branches:
     - main
 
+permissions:
+  contents: read
+  packages: write
```
mythi: AFAUI, top-level permissions don't need `write` if you move it down to `checkout`. similar case in `kbs-e2e.yml`

.github/workflows/build-kbs-client-image.yml
fitzthum: 😧

fitzthum:
Good catch @mkulke. Looks like Actionlint dropped the `fail_on_error` option, which is interesting. I added a commit to fix things up. This causes a failure which I will fix after I take some typing classes.

mythi (approved):
LGTM

mythi:
FYI: there seems to be some post-merge permission errors with workflows.

PR #887, opened 2025-07-31
Add query_reference_value for AS policy

Xynnn007:
This is a workaround for https://github.com/confidential-containers/trustee/pull/853. In this version we will use a separate thread to do querying things and no changes upon other parts.

cc @fitzthum


fitzthum:
Ah yes. The integration tests use the default policy for the CPU tests and a custom one for the GPU tests. Let me try your PR with my changes to those policies and see what happens.

fitzthum:
I've got the commit to update the policy in a branch [here](https://github.com/fitzthum/kbs/tree/extension-rvps). On my PR you mentioned something about having the extensions commented out at first. What did you mean by that? 

I think we will also need a fixup to keep the OPA linter from complaining about the extension. I have something like that for the other PR but it doesn't work yet. I will move that over and get it fixed up.

attestation-service/src/policy_engine/opa/mod.rs
```diff
@@ -89,19 +94,70 @@ impl PolicyEngine for OPA {
...
+                        });
+
+                        // We wait for 10 seconds at most.
+                        match rx.recv_timeout(Duration::from_secs(10)) {
+                            Ok(Ok(v)) => {
+                                let json_value = serde_json::to_value(v)?;
+                                let value: regorus::Value = serde_json::from_value(json_value)?;
+                                Ok(value)
+                            }
+                            Ok(Err(e)) => bail!("Failed to get Reference Value from RVPS: {e:?}"),
```
Xynnn007: @fitzthum btw now if the rv-key is not found in rvps, it will raise an error, while the original behavior will do nothing.

fitzthum: That's good. Having the extension return nothing can lead to unclear behavior.

fitzthum:
Looks like you may need to add `--rego-v1` to the command in my second commit. I don't need this locally for whatever reason.

attestation-service/src/policy_engine/opa/mod.rs
Xynnn007: Yup, for this, I have pushed a fix-up commit to get rid of all query statement in default policy to avoid errors when no rv is inserted to rvps but AS runs.

Xynnn007: Please take a look if it is good and I will rebase it to the previous commit.

Xynnn007: This [commit](https://github.com/confidential-containers/trustee/pull/887/commits/503b171873d69f211745f4a19609b0659eaf6176). And also this commit will influence the `integration-test` if sample device is not enabled. The default policy will not do query thus `policy_contraindicated` case will fail.

So, we might need to do a little change upon integration-test to set attestation policy to for each case.

Xynnn007: done. I did not rebase the last fix-up commits, as it brings some extra changes.

If we raise an error when reference value cannot be found from RVPS, some integration tests will break. So I add some more logic to specify attestation policies.

please take another look if it is good to take a rebase and fixup into the [commit](https://github.com/confidential-containers/trustee/pull/887/commits/bbda4af0c9e11e72dc2c3c5462fea75c9b17c762).

fitzthum: > Yup, for this, I have pushed a fix-up commit to get rid of all query statement in default policy to avoid errors when no rv is inserted to rvps but AS runs.

This shouldn't be necessary. The extensions should only be executed when the line that they are on is executed. Notice how in my PR I added a new statement to the top of each block that just tests if the paltform is detected. If that statement fails it should short-circuit the rest of the block and skip all the extensions.

Xynnn007: Good. What if a platform is chosen, but RV of a single field cannot be found in RVPS, i.e. failed to execute `get_reference_value`? I think this should be raised as an error. So I changed the default policies to just include the single claim names like
```rego
	input.azsnptvpm
	input.azsnpvtpm.measurement
	input.azsnpvtpm.tpm.pcr11
```

than querying them by default 
```rego
	input.azsnptvpm
	input.azsnpvtpm.measurement == get_reference_value("azsnpvtpm.measurement")
	input.azsnpvtpm.tpm.pcr11 == get_reference_value("azsnpvtpm.tpm.pcr11")
```

And tell people if they want to use RV, they should manually change the policy. Note that previously if no RV is found in RVPS, always the policy enforcement will return true, which should not be the expected behavior.

Xynnn007:
cc @fitzthum Updated and test locally. (let me check if CI runs well)

if a rule like the following but "reserved2" is not an existing key in RVPS
```
configuration := 2 if {
	input.sgx.reserved2 == query_reference_value("reserved2")
}
```

It will raise an error like this:
```
0: Failed to evaluate policy `default_cpu`
1: 
--> default_cpu:311:49
    |
311 |   input.sgx.reserved2 == query_reference_value("reserved2")
    |                                                 ^
error: Get Reference Value from RVPS failed: Reference value with key `reserved2` not found
```

Thus I deletes all the parts that uses `query_reference_value` in default policy.

attestation-service/src/policy_engine/opa/mod.rs
fitzthum: Expecting users to change the default policy is a non-starter for me. It should work out of the box and users should never need to change it.

So I think we should return to the original behavior where the function returns null if the reference value isn't found. We just need to double check that if the LHS and RHS are both null, the argument won't evaluate to True. I don't think it will.

If a clause depends on reference values that aren't set, it is reasonable for to not be evaluate (and the claim to have the default value) than for there to be an explicit error.

Xynnn007: Oh, you're right. I confused the situation where the execution was stopped midway due to an error with the situation where the evaluation level was low.

A Warning should be reported here to indicate that there is no reference value, but null is returned.

fitzthum: Ok, looks good.

fitzthum: But make sure to add the extensions back into the default policies.

Xynnn007: One tradeoff: this will cause a lot of query warnings of `no rv` when doing evaluation and query requests. 

two ways

1. Tell people how to configure the AS policy, and use `query_reference_value` API, like a guide or something. The default policy does not interacts with RVPS, but only do some hardcode value check, thus no warnings and no queries.
2.  By default define RV keys on all Right-hand value inside the default policy, and enable all the query statement. thus expectedly a lot of queries will happen and a lot of warnings.

fitzthum: I think it's ok to have warnings. Users should define reference values. It's good to warn them if they don't. As I mentioned above the extension should only be called for the platform they are using, so it will actually be fairly informative for them to be warned about all the missing values.

I don't want users to have to change the attestation service policy itself.

Xynnn007: Good. short-circuit shows that only the first `warn!` will occur, and others will all be skipped.

Xynnn007:
This PR can get merged as we still are not using new policy crate. I will do adoption work in later PR

attestation-service/src/ear_token/broker.rs
```diff
@@ -15,18 +15,124 @@ use openssl::ec::{EcGroup, EcKey};
...
+    #[serde(default = "Option::default")]
+    pub cert_url: Option<String>,
+
+    // PEM format certificate chain.
+    #[serde(default = "Option::default")]
+    pub cert_path: Option<String>,
+}
+
+#[derive(Deserialize, Debug, Clone, PartialEq)]
+pub struct Configuration {
```
fitzthum: Are these changes related to this commit?

attestation-service/docs/policy.md
```diff
@@ -30,13 +19,25 @@ These dimensions can be used to return a detailed, but generic description of th
...
+AS policies will be provide `input`.
 
 `input` are the TCB claims generated by the verifier.
 
 See the [default policy](../src/ear_token/ear_default_policy_cpu.rego) for an example.
 
+> [!NOTE]
+> We now provide a new extension in AS policy. If any field or value should be fetched from RVPS,
+> we can use function `query_reference_value(key_name)` in the rego to get reference value with key
+> id `key_name` from RVPS. If no key is found in RVPS, the policy enforcement will fail.
```
fitzthum: To be precise, the clause where the extension is used will evaluate to false/null.

Xynnn007: Oh. Let me fix this

attestation-service/src/ear_token/broker.rs
Xynnn007: Oh. The commits look messy. Let me do a tidy.

PR #890, opened 2025-08-01
verifier: Add support for nvidia-verifier

cclaudio:
Adds a nvidia-verifier for the nvidia-attester proposed in https://github.com/confidential-containers/guest-components/pull/1016

This was tested only the test cases defined.

More details in the commit messages.

TODO:
- [x] Update the documentation
- [x] Test with actual hardware and the nvidia-attester


deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -0,0 +1,164 @@
...
+        .get_extension_unique(&fwid_oid)?
+        .ok_or_else(|| anyhow!("FwId oid not found in the NVIDIA signing certificate"))?
+        .value;
+
+    // The FwId data is the last 48 bytes
+    let fwid = value
+        .windows(48)
+        .rev()
+        .next()
+        .ok_or(anyhow!("Unexpected oid data size "))?;
```
mkulke:
```suggestion
        let fwid: [u8; 48] = value[value.len() - 48..].try_into().context("Failed to extract fwid")?;
```

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+            let device_claims = NvDeviceReportAndCertClaim::try_build_from(&device.arch, device.uuid.as_str(), &report)?;
+
+            // Verify certificate chain
+            let signing_cert = cert_chain
+                .verify(&device_claims.opaque_data.OPAQUE_FIELD_ID_FWID)?;
+
+            // Verify attestation report freshness
+            report.validate_freshness(expected_report_data)?;
+
+            // TODO! Check if the singing certificate has been revoked
```
mkulke:
```suggestion
            // TODO! Check if the signing certificate has been revoked
```

deps/verifier/src/nvidia/report.rs
```diff
@@ -0,0 +1,71 @@
...
+            .public_key()?
+            .public_key_to_der()?;
+
+        let verifying_key = VerifyingKey::from_public_key_der(public_key.as_slice())?;
+        verifying_key.verify(signed_data, &signature)
+            .map_err(|e| anyhow!(e.to_string()))
+    }
+
+    pub fn validate_freshness(self, report_data: &ReportData) -> Result<()> {
+        if let ReportData::Value(expected_report_data) = report_data {
```
mkulke: maybe push this to the caller site, so we do not validate the nonce if ReportData is not used?

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+            // Verify the attestation report signature using the report_bytes rather than the parsed report.
+            // Otherwise, we would need to ensure that they have the same representation in memory.
+            NvidiaAttestationReport::verify_signature(report_vec.as_slice(), &HOPPER_SIGNATURE_LENGTH, signing_cert)?;
+
+            all_claims.devices_claims.push(device_claims);
+        }
+      
+        let value = serde_json::to_value(all_claims).context("serializing NVIDIA evidence claims into JSON")?;
+
+        Ok((value as TeeEvidenceParsedClaim, "nvidia".to_string()))
```
fitzthum: This Tee class should be something like `gpu` or `switch` rather than `nvidia` which is actually the Tee Type.

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+
+#[derive(Debug, Default, Serialize)]
+struct MeasurementBlockClaim {
+    index: u8,
+    measurement_type: u8,
+    measurement_value: String,
+}
+
+#[derive(Debug, Default, Serialize, Deserialize)]
+#[allow(non_snake_case)]
```
fitzthum: Any reason to not use snake case here?

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -0,0 +1,164 @@
...
+            .ok_or(anyhow!("Leaf certificate not found in the NVIDIA certificate chain"))
+    }
+
+    /// Use OpenSSL to cryptographically verify the NVIDIA certificate chain.
+    /// 
+    /// Returns the signing certificate if its chain of trust can be verified all the
+    /// way up to the trusted NVIDIA root CA.
+    pub fn verify(&self, expected_fwid: &str) -> Result<&X509> {
+        if self.certs.len() < 2 {
+            bail!("Certificate chain must be >= 2");
```
fitzthum: This can be slightly clearer. We mean that there need to be two or more certs, right?

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+    // Case1: Valid Hopper certificate chain
+    #[case(1, FWID, true)]
+    // Case2: Bad Hopper certificate chain. Only the root CA
+    #[case(2, FWID, false)]
+    // Case3: Bad Hopper certificate chain. Missing intermediate CA
+    #[case(3, FWID, false)]
+    // Case4: Bad Hopper certificate chain. Missing actual signing certificate
+    #[case(4, FWID, false)]
+    fn test_verify_certificate_chain_for_hopper(
+        #[case] case_number: usize,
```
fitzthum: You can just use a string here and specify the file path in the case macro above. That will eliminate the case statement.

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        self.check_root_ca_is_trusted()?;
+
+        // Signing certificate to be verified its chain of trust
```
fitzthum: typo here. `its` -> `in the` or something like that

deps/verifier/src/nvidia/report.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// 0        - SPDM_GET_MEASUREMENT_REQUEST   - 37
+    /// 37       - SPDM_GET_MEASUREMENT_RESPONSE  - vary
+    pub fn decode(report_bytes: &[u8], signature_length: &usize) -> Result<Self> {
+        let (request_bytes, response_bytes) = report_bytes
+            .split_at_checked(SPDM_GET_MEASUREMENT_REQUEST_SIZE)
+            .ok_or(anyhow!("Attestation report data is too small"))?;
+
+        Ok( Self {
+            request: SpdmGetMeasurementRequest::decode(request_bytes)?,
+            response: SpdmGetMeasurementsResponse::decode(response_bytes, signature_length)?,
```
mythi: did you check if `spdm-rs` spdmlib measurements from CCC could be used here?

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        // Trusted root certificate
+        let trusted_certs = {
+            let mut builder = X509StoreBuilder::new().unwrap();
+            builder.add_cert(NVIDIA_ROOT_CA.cert.clone())?;
+            builder.build()
+        };
+
+        // Untrusted certificate chain
+        // OpenSSL 1.1.0+ considers the root certificate to not be part of the chain, while 1.0.2 and LibreSSL do
```
mythi: how this translates to SSL library versions people *should be using*? would it be better to highlight the part that is specific to LibSSL (if we want to support it)

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+        ).unwrap();
+
+        let expected_claim = include_str!("../../test_data/nvidia/hopperAttestationReport-claims.txt");
+
+        assert_eq!(expected_claim.to_string(), json);
+    }
+
+    #[test]
+    fn test_verify_hopper_report_signature() {
+        // Certificate chain is in PERM format
```
```suggestion
        // Certificate chain is in PEM format
```

mythi (commented):
nice PR

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -0,0 +1,164 @@
...
+        let mut intermediate_certs = Stack::<X509>::new().unwrap();
+        for index in 1..self.certs.len()-1 {
+            intermediate_certs.push(self.certs[index].clone())?;
+        }
```
fitzthum, line 123: Might be slightly cleaner to just pop the last value off the vec and then iterate through with `for cert in self.certs` rather than messing around with the indices.

deps/verifier/src/nvidia/cert_chain.rs
cclaudio: Improved that part. Thanks.

deps/verifier/src/nvidia/report.rs
cclaudio: we could push that to the attester, but then the attester would need to be able to decode and verify the attestation report as well. Not sure we want to do that

deps/verifier/src/nvidia/mod.rs
cclaudio: Thanks, changed to "gpu"

deps/verifier/src/nvidia/mod.rs
cclaudio: Hmmm ... I just wanted the fields to look like the ones in nvtrust.
I changed that to snake case and renamed the field only at serde serialization/deserialization time.

deps/verifier/src/nvidia/cert_chain.rs
cclaudio: Changed to `Certificate chain must have at least the root CA and signing certificates`

deps/verifier/src/nvidia/mod.rs
cclaudio: I am happy to get rid of the inner case statement, but I did not find a way to do that yet.

Passing a path (either string or &str) as argument to the test case is easy, the problem is that AFAIK it cannot be converted to string literal for the `include_bytes` / `include_str` :-(

deps/verifier/src/nvidia/cert_chain.rs
cclaudio: oops ... fixed, thanks!

deps/verifier/src/nvidia/report.rs
cclaudio: Yes, I did that some time ago. I don't remember all the details, but I remember that spdm-rs is not published in the `crates.tio` and also its documentation says that some manual steps (patching) are required in the setup. At the end of the day, I decided to have a simple spdm implementation, we don't need much.

deps/verifier/src/nvidia/cert_chain.rs
cclaudio: I think we only care about openssl, so I changed the comment to `OpenSSL 1.1.0+ considers the root certificate to not be part of the chain`

For this code to work properly (and possibly the token code https://github.com/confidential-containers/trustee/blob/main/kbs/src/token/jwk.rs#L167), we may need to ensure that we are linking to the right openssl version, for example, we could add somethink like https://docs.rs/openssl/latest/openssl/#feature-detection to the build.rs. That is probably out of scope for this PR ;-)

deps/verifier/src/nvidia/mod.rs
cclaudio: uhhhh ... fixed. Actually, I had to refactor that code. Thanks

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+    measurements: Vec<MeasurementBlockClaim>,
+    opaque_data: OpaqueDataClaim,
+}
+
+#[derive(Debug, Default, Serialize)]
+struct MeasurementBlockClaim {
+    index: u8,
+    measurement_type: u8,
+    measurement_value: String,
+}
```
fitzthum: I think you can simplify this. We can probably just put these measurement blocks in one map of index -> measurement_value. This will be easier for the policy to consume.

cclaudio: Ok, I will do that.

deps/verifier/src/nvidia/mod.rs
fitzthum: Can you put the `include_str!` macro inside the case or define a const for each file.

cclaudio: that's worth trying

deps/verifier/src/nvidia/cert_chain.rs
cclaudio: Yeah, I agree. I am using `split_last()` now

```rust
        let mut intermediate_certs = Stack::<X509>::new().unwrap();
        let (_, certs_without_root) = self.certs.split_last().unwrap();
        for cert in certs_without_root {
            intermediate_certs.push(cert.clone())?;
        }
```

deps/verifier/src/nvidia/report.rs
mkulke: ah, I meant pushing it up the fn call stack (i.e. not calling validate_freshness() if there's no report data).

deps/verifier/src/nvidia/report.rs
mythi: OK, sounds like it makes sense to go with what's in this PR.

deps/verifier/Cargo.toml
```diff
@@ -73,6 +75,8 @@ ear = { version = "0.3.0", optional = true }
 x509-parser = { version = "0.17.0", optional = true }
 reqwest.workspace = true
 bitflags = { version = "2.8.0", features = ["serde"] }
+nvml-wrapper = { version = "0.11.0", optional = true, default-features = false, features = ["serde"]}
+p384 = { version = "0.13.1" }
```
Xynnn007: Let's mark this as `optional` too and enable it with feature `nvidia-verifier`

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -0,0 +1,164 @@
...
+///
+/// subject=CN = NVIDIA Device Identity CA, O = NVIDIA
+/// issuer=CN = NVIDIA Device Identity CA, O = NVIDIA
+pub struct NvidiaCertificateChain {
+    certs: Vec<X509>,
+}
+
+impl NvidiaCertificateChain {
+    pub fn decode(bytes: &[u8]) -> Result<Self> {
+            let certs = X509::stack_from_pem(bytes)?;
```
Xynnn007: looks like the idents are not same - will you run `cargo fmt` to fix this?

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        // Check if leaf certificate FwId matches the expected one
+        let cert_fwid = get_fwid_from_cert(leaf_cert)?;
+        if cert_fwid != expected_fwid {
+            bail!("Fwid mismatch: certificate {}, evidence {}", cert_fwid, expected_fwid);
+        }
+
+        // Trusted root certificate
+        let trusted_certs = {
+            let mut builder = X509StoreBuilder::new().unwrap();
```
Xynnn007: Can we replace `unwrap()` here to `map_err(..)?` or `?`? Unwraps will cause panic

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -0,0 +1,164 @@
...
+            builder.add_cert(NVIDIA_ROOT_CA.cert.clone())?;
+            builder.build()
+        };
+
+        // Untrusted certificate chain
+        // OpenSSL 1.1.0+ considers the root certificate to not be part of the chain, while 1.0.2 and LibreSSL do
+        let mut intermediate_certs = Stack::<X509>::new().unwrap();
```
Xynnn007: The same for this `unwrap`

deps/verifier/src/nvidia/cert_chain.rs
Xynnn007: Probably we can use the following to avoid annoying indexes
```suggestion
        for cert in &self.certs {
            intermediate_certs.push(cert.clone())?;
        }
```

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -0,0 +1,164 @@
...
+        for index in 1..self.certs.len()-1 {
+            intermediate_certs.push(self.certs[index].clone())?;
+        }
+
+        let mut context = X509StoreContext::new().unwrap();
```
Xynnn007: Let's get rid of this `unwrap` too

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -0,0 +1,164 @@
...
+        let verified = context
+            .init(
+                &trusted_certs,
+                leaf_cert,
+                &intermediate_certs,
+                |c| c.verify_cert()
+            )
+            .map_err(|e| anyhow!(e.to_string()))?;
```
Xynnn007, line 130: can we add more context here? like
```suggestion
            .map_err(|e| anyhow!("failed to init the X509 store: {e}"))?;
```

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -0,0 +1,164 @@
...
+        if !verified {
+            bail!("Leaf certificate failed to verify the certificate");
+        }
+        
+        Ok(leaf_cert)
+    }
+}
+
+pub fn get_fwid_from_cert(cert: &X509) -> Result<String>{
+    let fwid_oid = oid!(2.23.133.5.4.1);
```
Xynnn007: If this is a const value, I recomment to claim it as a `const` on top of the file with some comments

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+
+impl TryFrom<&MeasurementBlock> for MeasurementBlockClaim {
+    type Error = Error;
+
+    fn try_from(value: &MeasurementBlock) -> Result<Self> {
+        let claim = MeasurementBlockClaim {
+                index: value.index,
+                measurement_type: value.measurement.value_type,
+                measurement_value: hex::encode(value.measurement.value.clone()),
```
Xynnn007: I think we do not need `.clone()` here to avoid memory copy. But i am not sure - let's take a `cargo clippy` to get some guides

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+        Ok(claim)
+    }
+}
+
+impl TryFrom<&SpdmGetMeasurementsResponse> for SpdmGetMeasurementsResponseClaim {
+    type Error = Error;
+
+    fn try_from(response: &SpdmGetMeasurementsResponse) -> Result<Self> {
+        // Measurements
+        let mut measurements: Vec<MeasurementBlockClaim> = Vec::with_capacity(response.measurement_record.len());
```
Xynnn007: we can just claim it with new function and push to the end
```suggestion
        let mut measurements: Vec<MeasurementBlockClaim> = Vec::new();
```

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+        for m in &response.measurement_record {
+            measurements.push(MeasurementBlockClaim::try_from(m)?);
+        }
+
+        // Opaque data        
+        let opaque_data_string = serde_json::to_string(&response.opaque_data)?;
+        let opaque_data: OpaqueDataClaim = serde_json::from_str(opaque_data_string.as_str())?;
```
Xynnn007: Rust will help to do the `String` to `&str` conversion with `&`
```suggestion
        let opaque_data: OpaqueDataClaim = serde_json::from_str(&opaque_data_string)?;
```

deps/verifier/src/nvidia/report.rs
Xynnn007: Agree with @mkulke . ReportData is an enum for both a value or None. If it is None, we can skip verification/validation

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+struct NvDeviceEvidence {
+    device_evidence_list: Vec<NvDeviceReportAndCert>,
+}
+
+#[derive(Debug, Deserialize)]
+struct NvDeviceReportAndCert {
+    arch: DeviceArchitecture,
+    uuid: String,
+    evidence: String,
+    certificate_chain: String,
```
Xynnn007: This field has different name with [certificate](https://github.com/confidential-containers/guest-components/pull/1016/files#diff-bc2897e411e13ef2db535e2278f433f7b710f3fc1dffcb62596557f5a4d81e84R34)

Also there is a question: does this mean that the attester side should carry the whole cert chain from root to the chip along with the evidence? I did not see the logic on the attester side - also, for other platforms, we often get the intermediate certs on the verifier side

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+        let mut all_claims = NvDeviceEvidenceClaim::default();
+        all_claims.devices_claims.push(device_claims);
+        
+        let value = serde_json::to_value(all_claims).unwrap();
+        let json = serde_json::to_string(&value).unwrap();
+
+        let _ = fs::write(
+            "hopperAttestationReport-claims.txt",
+            &json,
+        ).unwrap();
```
Xynnn007: This will write a file to cwd. Is this expected?

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,323 @@
...
+        assert!(status.is_ok())
+    }
+
+    #[test]
+    // The FwId is last 48 bytes of the "2.23.133.5.4.1" extension
+    // Command: openssl asn1parse -i -in hopper_singing_cert.pem
+    fn test_parse_fwid_from_certificate() {
+        let signing_cert = X509::from_pem(include_bytes!("../../test_data/nvidia/hopper_signing_cert.pem"))
+            .map_err(|_| anyhow!("hopper_singing_cert.pem failed to read"))
+            .unwrap();
```
Xynnn007: If we are in a unit test, feel free to panic
```suggestion
            .expect("hopper_singing_cert.pem failed to read");
```

deps/verifier/src/nvidia/report.rs
```diff
@@ -0,0 +1,71 @@
...
+        let signature = Signature::from_slice(signature_bytes)?;
+
+        // Extract EC public key from certificate
+        let public_key = signing_cert
+            .public_key()?
+            .public_key_to_der()?;
+
+        let verifying_key = VerifyingKey::from_public_key_der(public_key.as_slice())?;
+        verifying_key.verify(signed_data, &signature)
+            .map_err(|e| anyhow!(e.to_string()))
```
Xynnn007: Let's add more context to the error
```suggestion
            .map_err(|e| anyhow!("failed to verify the signature: {e}"))
```

deps/verifier/src/nvidia/spdm_request.rs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        let mut request = SpdmGetMeasurementRequest::default();
+
+        request.spdm_version = bytes[0];
+        request.request_response_code = bytes[1];
+        request.param1 = bytes[2];
+        request.param2 = bytes[3];
+        request.nonce.copy_from_slice(&bytes[4..36]);
+        request.slot_id_param = bytes[36];
```
Xynnn007: This looks like a logic trap. Usually `SPDM_GET_MEASUREMENT_REQUEST_SIZE` is 37. But how we get 37 is by `size_of`. The corner case is `size_of:<SPDM_GET_MEASUREMENT_REQUEST_SIZE>` is less than 36 on some platforms thus indexings will cause panic.

A potential way is to directly set `SPDM_GET_MEASUREMENT_REQUEST_SIZE` to 37

deps/verifier/src/nvidia/spdm_response.rs
```diff
@@ -0,0 +1,429 @@
...
+        let mut response = SpdmGetMeasurementsResponse::default();
+
+        response.spdm_version = bytes[0];
+        response.request_response_code = bytes[1];
+        response.param1 = bytes[2];
+        response.param2 = bytes[3];
+        response.number_of_blocks = usize::from(bytes[4]);
+
+        // MeasurementRecordLength
+        let array: [u8; 3] = bytes[5..8].try_into().unwrap();
```
Xynnn007: let's get rid of this `unwrap` with `?` or `map_err(..)?`

deps/verifier/src/nvidia/spdm_response.rs
```diff
@@ -0,0 +1,429 @@
...
+            .ok_or(anyhow!("Nonce overflow"))?;
+        response.nonce.copy_from_slice(nonce);
+        offset += 32;
+
+        // OpaqueDataLength
+        let array: [u8; 2] = bytes
+            .get(offset..offset+2)
+            .ok_or(anyhow!("OpaqueDataLength overlow"))?
+            .try_into()
+            .unwrap();
```
Xynnn007: let's get rid of this `unwrap` with `?` or `map_err(..)?`

deps/verifier/src/nvidia/spdm_response.rs
```diff
@@ -0,0 +1,429 @@
...
+        response.signature.extend_from_slice(signature);
+
+        Ok(response)
+    }
+}
+
+#[derive(Debug, Default)]
+struct OpaqueData {}
+
+impl OpaqueData {
```
Xynnn007: I did not see any member functions that manipulates/uses the `OpaqueData` struct itself. So we might not need to claim this struct - but only define functions like 

`fn decode_opaque_data(bytes: &[u8]) -> Result<HashMap<String, Value>> `
?

deps/verifier/src/nvidia/spdm_response.rs
```diff
@@ -0,0 +1,429 @@
...
+    /// 3      - DMTFSpecMeasurementValue     - DMTFSpecMeasurementValueSize
+    pub fn decode(bytes: &[u8]) -> Result<Self> {
+        if bytes.len() < 3 {
+            bail!(anyhow!("DMTF measurement must be bigger than 3 bytes"));
+        }
+        let mut measurement = Self::default();
+
+        measurement.value_type = bytes[0];
+
+        let array: [u8; 2] = bytes[1..3].try_into().unwrap();
```
Xynnn007: let's get rid of this `unwrap` with `?` or `map_err(..)?`

deps/verifier/src/nvidia/spdm_response.rs
```diff
@@ -0,0 +1,429 @@
...
+            // Measurement specification
+            block.measurement_specification = bytes[offset];
+            offset += 1;
+
+            if block.measurement_specification != DMTF_MEASUREMENT_SPECIFICATION_VALUE {
+                bail!("Only measurements encoded in the DMTF layout are supported");
+            }
+            
+            // Measurement size
+            let array: [u8; 2] = bytes[offset..offset+2].try_into().unwrap();
```
Xynnn007: let's get rid of this `unwrap` with `?` or `map_err(..)?`

deps/verifier/src/nvidia/spdm_response.rs
```diff
@@ -0,0 +1,429 @@
...
+            .get(3..3+measurement.value_size)
+            .ok_or(anyhow!("DMTF measurement overflow"))?;
+        measurement.value.extend_from_slice(value);
+
+        Ok(measurement)
+    }
+}
+
+
+struct MeasurementRecord {}
```
Xynnn007: This is the same. If this struct object is not used in the member functions, we do not need to define this struct. we can move 
`pub fn decode(number_of_blocks: usize, bytes: &[u8]) -> Result<Vec<MeasurementBlock>>` under `impl MeasurementBlock` and rename to `pub fn decode(number_of_blocks: usize, bytes: &[u8]) -> Result<Vec<Self>>`

Xynnn007 (commented):
Thanks @cclaudio ! What a huge work! Some nits at first glance; and you can take a `cargo fmt` to format the code

deps/verifier/src/nvidia/cert_chain.rs
cclaudio: yes, I forgot to do that.
Done. Thanks

deps/verifier/src/nvidia/cert_chain.rs
cclaudio:
```suggestion
        let (_, certs_without_root) = self
            .certs
            .split_last()
            .ok_or(anyhow!("Failed to split NVIDIA certificates"))?;
        for cert in certs_without_root {
            intermediate_certs.push(cert.clone())?;
        }
```
I am slightly changing that

deps/verifier/src/nvidia/spdm_response.rs
cclaudio: I have multiple structs in the `spdm_response.rs` that have no field, mainly because they are intermediate representations of the original data and they don't need to be stored. However, each of these structs have their own semantic in the spdm.

```suggestion
#[derive(Debug, Default)]
struct OpaqueData;

impl OpaqueData {
```
What if we turn the struct into a zero-size type (ZST), does that help? By definition ZST consumes no memory at runtime.

I think there multiple advantages of keeping `impl OpaqueData {` (e.g. semantic, readability, maintainability and more idiomatic), but if you think otherwise I am happy to rename and move the functions.

deps/verifier/src/nvidia/spdm_response.rs
cclaudio: Hmm ... I guess I focused too much on the other parts of the code and got relaxed about details such as using unwrap outside of test cases.
I just reviewed all my unwraps. Thanks

deps/verifier/src/nvidia/spdm_response.rs
cclaudio:
```suggestion
struct MeasurementRecord;
```
Same idea. Is it better if we turn it into a zero-sized type?

deps/verifier/src/nvidia/spdm_response.rs
Xynnn007: Yea, readability and maintainability are good properties. There are also some observations, that it seems that `OpaqueData` should be one item but the function `decode` returns a `OpaqueDataType` -> `data` mapping.

One question
1. Can we assert that in a `SpdmGetMeasurementsResponse`, no same `OpaqueDataType`s exist? If two same `OpaqueDataType`s exist, when building the hashmap the former one will be overwritten.

Also, for readability and maintainability, I suggest to have a more complete definition, something like this

```rust
pub struct OpaqueData {
    r#type: OpaqueDataType,
    size: usize,
    data: Value,
}

and implementations

```
impl OpaqueData {
    pub fn from_bytes(bytes: &[u8]) -> Result<Self>
}
```

This can be a clearer imo.

Comment (cclaudio) [reply deps/verifier/src/nvidia/spdm_response.rs]: Nice. Applying the suggestions and pushing an updated branch today.

File: deps/verifier/src/nvidia/cert_chain.rs (hunk 19 above)
Comment (mythi) [reply deps/verifier/src/nvidia/cert_chain.rs:130]: how about `.context("failed to init the X509 store")?;`

**Review by fitzthum — APPROVED** [review]
LGTM

File: deps/verifier/src/nvidia/mod.rs [hunk 35]
```diff
@@ -0,0 +1,191 @@
...
+            let b64_engine = base64::engine::general_purpose::STANDARD;
+
+            let cert_chain_vec: Vec<u8> = b64_engine.decode(device.certificate)?;
+            let report_vec: Vec<u8> = hex::decode(device.evidence)?;
+
+            let report_filename = format!("/tmp/{}_report.txt", &device.uuid);
+            let cert_filename = format!("/tmp/{}_certs.txt", &device.uuid);
+
+            let _ = std::fs::write(report_filename, hex::encode(&report_vec)).unwrap();
+            let _ = std::fs::write(cert_filename, hex::encode(&cert_chain_vec)).unwrap();
```
Xynnn007: looks like we do not use the two files anymore, and do not delete them. Is this expected?

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+            )?;
+            trace!("{}", &report);
+
+            // Build the device claims
+            let device_claims =
+                NvDeviceReportAndCertClaim::new(&device.arch, device.uuid.as_str(), &report);
+            let value = serde_json::to_value(device_claims)
+                .context("serializing NVIDIA evidence claims into JSON")?;
+
+            let json = serde_json::to_string(&value).unwrap();
```
Xynnn007: Let's try to get rid of the `unwrap()`s

deps/verifier/src/nvidia/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+            let claims_filename = format!("/tmp/{}_claims.txt", &device.uuid);
+
+            let _ = std::fs::write(claims_filename, &json).unwrap();
+
```
Xynnn007: The same for this. Is this something left for degbugging?

cclaudio:
All feedbacks applied. Thank you
I tested the patches with an actual H100 GPU device

deps/verifier/src/nvidia/spdm_response.rs
Xynnn007: Oh. I did not see the reply for my question and the feedback for my comment. Did I miss it somewhere?

deps/verifier/src/nvidia/mod.rs
cclaudio: Yes, I used that for debugging.  I can just remove it

deps/verifier/src/nvidia/mod.rs
cclaudio: you're right, we don't. Good catch, thanks.

deps/verifier/src/nvidia/mod.rs
cclaudio: Oh no. Removing and double checking

cclaudio:
> > I tested the patches with an actual H100 GPU device
> 
> Tested on my side too and I get the report verified OK. One observation which I think needs thinking still before merging is how to get the claims better aligned with NRAS based tokens. I think interoperability is key to help policy authors.
> 
> ```
> $ sudo RUST_LOG=none ./target/release/kbs-client attest|jq -R 'split(".") | .[1] | @base64d | fromjson | .submods.gpu0."ear.veraison.annotated-evidence".nvidia.config | keys' 
> [
>   "board_id",
>   "chip_sku",
>   "chip_sku_mod",
>   "cpr_info",
>   "driver_version",
>   "fwid",
>   "gpu_info",
>   "measurement_count",
>   "nvdec0_status",
>   "project",
>   "project_sku",
>   "project_sku_mod",
>   "protected_pcie_status",
>   "vbios_version"
> ]
> ```

Are you suggesting to have all of them upercase and with the `OPAQUE_FIELD_ID_` prefix, just like nvtrust?

OPAQUE_FIELD_ID_CERT_ISSUER_NAME
OPAQUE_FIELD_ID_CERT_AUTHORITY_KEY_IDENTIFIER
OPAQUE_FIELD_ID_DRIVER_VERSION
OPAQUE_FIELD_ID_GPU_INFO
OPAQUE_FIELD_ID_SKU
OPAQUE_FIELD_ID_VBIOS_VERSION
OPAQUE_FIELD_ID_MANUFACTURER_ID
OPAQUE_FIELD_ID_TAMPER_DETECTION
OPAQUE_FIELD_ID_SMC
OPAQUE_FIELD_ID_VPR
OPAQUE_FIELD_ID_NVDEC0_STATUS
OPAQUE_FIELD_ID_MSRSCNT
OPAQUE_FIELD_ID_CPRINFO
OPAQUE_FIELD_ID_BOARD_ID
OPAQUE_FIELD_ID_CHIP_SKU
OPAQUE_FIELD_ID_CHIP_SKU_MOD
OPAQUE_FIELD_ID_PROJECT
OPAQUE_FIELD_ID_PROJECT_SKU
OPAQUE_FIELD_ID_PROJECT_SKU_MOD
OPAQUE_FIELD_ID_FWID
OPAQUE_FIELD_ID_PROTECTED_PCIE_STATUS
OPAQUE_FIELD_ID_SWITCH_PDI
OPAQUE_FIELD_ID_FLOORSWEPT_PORTS
OPAQUE_FIELD_ID_POSITION_ID
OPAQUE_FIELD_ID_LOCK_SWITCH_STATUS
OPAQUE_FIELD_ID_GPU_LINK_CONN
OPAQUE_FIELD_ID_SYS_ENABLE_STATUS
OPAQUE_FIELD_ID_OPAQUE_DATA_VERSION
OPAQUE_FIELD_ID_CHIP_INFO
OPAQUE_FIELD_ID_INVALID

If so, I am fine with that. I preferred to go without that because I think it is cleaner and `OPAQUE_FIELD_ID_` is something that ressembles SPDM terminology, not sure if policy creators would need to know about that.

By the way, with NRAS based tokens, would the policy have `board_id` or any of the other fields you listed? Last time I ran nvtrust to verify the evidences, it just printed a bunch of high level information such as PolicyA passed, PolicyB passed, PolicyC failed, etc.

Xynnn007:
@cclaudio one thing left https://github.com/confidential-containers/trustee/pull/890#discussion_r2350522992

mythi:
> We may have some more changes to apply depending on @mythi 's answer

Perhaps we go with what we have now but add that note that the claims might still change. Checking NRAS output and sharing it here is not possible due to their EULA it seems.

mythi:
> @cclaudio one thing left [#890 (comment)](https://github.com/confidential-containers/trustee/pull/890#discussion_r2350522992)

it looks there are other comments too that are not fully addressed, https://github.com/confidential-containers/trustee/pull/890#discussion_r2331762990

deps/verifier/src/nvidia/spdm_response.rs
cclaudio: > 1. Can we assert that in a SpdmGetMeasurementsResponse, no same OpaqueDataTypes exist? If two same OpaqueDataTypes exist, when building the hashmap the former one will be overwritten.

Yes. I added this to the end of `OpaqueData::decode()`
```rust
            // Check if the key already exists
            if opaque_data.contains_key(&key) {
                bail!(anyhow!("Duplicated OpaqueDataType: {}", &key));
            }
```

> Also, for readability and maintainability, I suggest to have a more complete definition
> pub struct OpaqueData {
    r#type: OpaqueDataType,
    size: usize,
    data: Value,
}

In SPDM, the OpaqueData is a *free-form* field, if present. Well, NVIDIA is using OpaqueData, but we would not be able to use the OpaqueData definition you described as-is because NVIDIA is using the OpaqueData to carry a list of items (as many as available for the device), not just one item.

So, I added the `NvidiaOpaqueDataItem`  with the fields you described @Xynnn007 and updated the description of the structs and functions, as below to convey that. Inside the decode() function, we are creating and populating a `NvidiaOpaqueDataItem` and then inserting part of it to the hashmap.

``` rust
/// OpaqueData field in the SPDM GET_MEASUREMENT response message, which is a *free-form* field if present
#[derive(Debug, Default)]
struct OpaqueData;

/// NVIDIA is using the OpaqueData to carry OpaqueDataType values, as many as available for the device.
#[derive(Debug, Default)]
struct NvidiaOpaqueDataItem {
    r#type: OpaqueDataType,
    size: usize,
    data: Value,
}

impl OpaqueData {
    /// Decode the OpaqueData for Nvidia. We store the OpaqueDataItems
    /// in a hashmap, which is more convenient for generating device claims
    ///
    /// Layout for each OpaqueDataItem
    ///
    /// OFFSET   - FIELD                        - SIZE(in bytes)
    /// 0        - DataType                     - 2
    /// 2        - DataSize                     - 2
    /// 4        - Data                         - DataSize
    pub fn decode(bytes: &[u8]) -> Result<HashMap<String, Value>> {
```
With that, I think that the code would be relatively easy to read and maintain, but let me know if you would like to have OpaqueData more descriptive.

cclaudio:
Rebased to latest.
Fixed clippy warnings.

Xynnn007 (approved):

PR #893, opened 2025-08-04
kbs: prometheus build info metric

pmores:
This simple change should be the last addition necessary to deal with all of metrics proposed in issue #543.


kbs/Cargo.toml
```diff
@@ -113,3 +113,4 @@ toml = "0.9"
 
 [build-dependencies]
 tonic-build = { workspace = true, optional = true }
+chrono = "*"
```
Xynnn007: we can use `chrono.workspace = true` here

Xynnn007 (commented):
lgtm. Only one comment

kbs/Cargo.toml
pmores: Ah yes, sure, sorry, I overlooked that.

pmores: Should be fixed now.

kbs/build.rs
```diff
@@ -5,9 +5,23 @@
 
 #![allow(missing_docs)]
 #[allow(unused_imports)]
+use std::env;
```
fitzthum: I don't see where you are using this.

kbs/src/prometheus/mod.rs
```diff
@@ -144,6 +144,20 @@ lazy_static! {
...
+    /// KBS Build Info
+    pub(crate) static ref BUILD_INFO: Gauge = {
+        let opts = Opts::new(
+                "build_info",
+                "KBS binary build info",
+            )
+            .const_labels(std::collections::HashMap::from([
+                ("version".to_owned(), env!("CARGO_PKG_VERSION").to_owned()),
+                ("git_hash".to_owned(), env!("GIT_HASH").to_owned()),
+                ("build_date".to_owned(), env!("BUILD_DATE").to_owned())
```
fitzthum: I wonder if we should have these env vars be scoped to the KBS in case we implement something similar for the AS. Probably not necessary for now but worth remembering.

kbs/build.rs
pmores: That's because I'm not ;-D , good catch, thanks!  I don't remember adding this either.  It must have been vscode, I use it for some tasks and this is one of its more annoying features.

Should be fixed now.

kbs/src/prometheus/mod.rs
pmores: Just to make sure I understand correctly, do you mean renaming GIT_HASH to KBS_GIT_HASH and BUILD_DATE to KBS_BUILD_DATE?

I'm thinking of adding one more commit to this PR to prepend "kbs_" to all metrics names because [we seem to agree on that](https://github.com/confidential-containers/trustee/issues/543#issuecomment-3149175866).  I could scope the env vars as well while at it.

pmores:
Added both the env vars scoping (merged into the initial commit) and the metric names scoping in an additional separate commit.

Xynnn007 (approved):

pmores:
Just a small change to fix a clippy warning that was breaking CI.

PR #894, opened 2025-08-04
ci: fix permissions for nested workflows

mkulke:
Some necessary permissions weren't passed down to nested workflows when they are being invoked.

### DIFF
--- a/.github/workflows/as-docker-build.yml
+++ b/.github/workflows/as-docker-build.yml
@@ -13,4 +13,7 @@ permissions: {}
 jobs:
   check_as_image_build:
     uses: ./.github/workflows/build-as-image.yml
+    permissions:
+      contents: read
+      packages: write
     secrets: inherit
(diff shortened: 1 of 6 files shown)


mythi:
it looks there's one more with:
```
The workflow is not valid. .github/workflows/kbs-e2e-azure-vtpm.yml (Line: 28, Col: 3): Error calling workflow 'confidential-containers/trustee/.github/workflows/kbs-e2e.yml@7273a44bd4b30e4f7e6bbaf964fc70f3c420f0f3'. The workflow is requesting 'contents: read', but is only allowed 'contents: none'. .github/workflows/kbs-e2e-azure-vtpm.yml (Line: 28, Col: 3): Error calling workflow 'confidential-containers/trustee/.github/workflows/kbs-e2e.yml@7273a44bd4b30e4f7e6bbaf964fc70f3c420f0f3'. The nested job 'build-binaries' is requesting 'contents: read', but is only allowed 'contents: none'.
```

PR #900, opened 2025-08-05
Improvements to Multi-Device Attestation

fitzthum:
This PR does 3 things. 

1) Adds a default GPU policy. Even if we change how we handle the names of policies in the future, it's still a good idea to ship a default policy.

2) Make the affirming policy more restrictive. Specifically, make the affirming policy check the trust tier of all submods.

3) Allow a verifier to return multiple sets of tee evidence each with its own tee class as suggested by @mythi 

See commit messages for more info.


attestation-service/src/token/ear_broker.rs
```diff
@@ -144,14 +144,21 @@ pub struct EarAttestationTokenBroker {
...
         )?;
         info!("Loading default AS policy \"default_cpu.rego\"");
 
+        let gpu_policy_b64 = base64::engine::general_purpose::URL_SAFE_NO_PAD
+            .encode(include_str!("ear_default_policy_gpu.rego"));
+        policy_engine
+            .set_policy("default_gpu".to_string(), gpu_policy_b64)
+            .await?;
+        info!("Loading default AS policy \"default_gpu.rego\"");
+
```
Xynnn007: can we change the way to set the default policy here? Probably we could do a little refactoring upon the logic of  `PolicyEngineType::OPA.to_policy_engine` API to only receive the `policy_dir`. Putting (default) policy logic can be all moved here by writing files.

Or the current PR would explicitly have two ways to set policies: both initialization and call `set_policy`. This is duplicated while they are all default/initial policies.

deps/verifier/src/sample/mod.rs
```diff
@@ -79,6 +79,10 @@ fn parse_tee_evidence(quote: &SampleTeeEvidence) -> Result<TeeEvidenceParsedClai
             "major": 1,
             "minor": 4,
         },
+
+        // An example of a claim representing the platform configuration.
+        // The sample platform is in a sense only for debugging.
+        "debug": false,
```
Xynnn007, line 86: Note that this is hardcode and will always be `false` no matter what attester side sends. We might need to highlight this via notes.

.github/workflows/as-rust.yml
```diff
@@ -40,7 +40,10 @@ jobs:
       - name: OPA policy.rego fmt and check
         run: |
           opa fmt -d attestation-service/src/token/*.rego | awk '{ print } END { if (NR!=0) { print "run `opa fmt -w <path_to_rego>` to fix this"; exit 1 } }'
-          opa check attestation-service/src/token/*.rego
+          set -e
+          for file in attestation-service/src/token/*.rego ; do
+            opa check "$file"
+          done
```
Xynnn007: Do you think it a good idea to have a Makefile target to cover policy check?

attestation-service/src/lib.rs
```diff
@@ -209,22 +209,26 @@ impl AttestationService {
                 None => InitDataHash::NotProvided,
             };
 
-            let (claims_from_tee_evidence, tee_class) = verifier
+            //let (claims_from_tee_evidence, tee_class) = verifier
```
Xynnn007: This is for debugging?

deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -153,7 +153,7 @@ impl Verifier for AzSnpVtpm {
         let mut claim = parse_tee_evidence_az(&snp_report);
         extend_claim(&mut claim, &evidence.quote)?;
 
-        Ok((claim, "cpu".to_string()))
+        Ok(vec![(claim, "cpu".to_string())])
```
Xynnn007, line 156: When I review this code, I have a design question. Why not we move the TeeClass abstraction into the structure of the parsed claims? that means use
```rust
async fn evaluate(
        &self,
        evidence: TeeEvidence,
        expected_report_data: &ReportData,
    ) -> Result<TeeEvidenceParsedClaim>
```
and
```json
{
  "nvidia": {
    "teeClass": "gpu",
    ...
  }
}

// or
{
  "nvidia": [
    {
      "teeClass": "gpu",
    ...
    },
    {
      "teeClass": "switch",
    ...
    },
   ...
  ]
}

```

than
```rust
async fn evaluate(
        &self,
        evidence: TeeEvidence,
        expected_report_data: &ReportData,
    ) -> Result<Vec<(TeeEvidenceParsedClaim, TeeClass)>> {
```

it will bring two similar structure claims, s.t. a gpu claim
```json
{
  "nvidia": {
    ...
  }
}
```

and a switch claim
```json
{
  "nvidia": {
    ...
  }
}
```

Xynnn007 (commented):
Some nits

deps/verifier/src/az_snp_vtpm/mod.rs
mythi: > Why not we move the TeeClass abstraction into the structure of the parsed claims?

How `TeeClass` is passed is independent of this change. The change in this PR won't change the end result but only allows "one verifier to represent multiple devices.".

attestation-service/src/lib.rs
fitzthum: whoops. yeah

deps/verifier/src/az_snp_vtpm/mod.rs
fitzthum: Yeah this is slightly orthogonal to the PR, but I think it's a reasonable proposal. To me, it seems better to keep them separate since the tee class isn't part of the claims that end up in the token. Since we have a different policy for each tee class, there is no need to check the tee class as part of the policy. In the token the tee class is used as the submod name, so again it isn't needed in the claims themselves.

deps/verifier/src/sample/mod.rs
fitzthum: I made the comments a little more explicit.

.github/workflows/as-rust.yml
fitzthum: added something basic. there might be a slicker way to do it. this will be quite useful for my development flow actually.

attestation-service/src/token/ear_broker.rs
fitzthum: ok. updated. there is some special logic for the default policies. we don't want the default policy to overwrite an existing policy. i made a new wrapper method called `set_default_policy` to handle this.

attestation-service/Makefile
```diff
@@ -37,5 +37,15 @@ install:
 		install -D -m0755 $(TARGET_DIR)/$$bin_name $(DESTDIR); \
 	done
 
+opa-check:
+	for rego_file in src/token/*.rego ; do \
+		opa fmt -d $${rego_file} && \
+		opa check $${rego_file} ; \
+	done
+
+opa-check-2:
```
Xynnn007: Why we have a `2` here?

attestation-service/src/token/ear_broker.rs
```diff
@@ -144,13 +144,17 @@ pub struct EarAttestationTokenBroker {
...
+        let policy_engine =
+            PolicyEngineType::OPA.to_policy_engine(Path::new(&config.policy_dir))?;
+
+        policy_engine
+            .set_default_policy("default_cpu", include_str!("ear_default_policy_cpu.rego"))
+            .await?;
+
+        policy_engine
+            .set_default_policy("default_gpu", include_str!("ear_default_policy_gpu.rego"))
+            .await?;
```
Xynnn007: Oh.  Sorry for bothering, but the `set_default_policy` and `set_policy` are somehow overlapped for the only difference is whether override or not. Imo we can simply write file here
```rust
if !Path::new(<policy_dir>/default_cpu).exists() {
  fs::write(include_str!("ear_default_policy_gpu.rego"), <policy_dir>/default_cpu)?;
}
```

This could help avoid `async`-ify call stack, and also avoid a much similar member function.

deps/verifier/src/az_snp_vtpm/mod.rs
Xynnn007: Ok. let's keep this topic away from the PR, but I still think the way of tee class name to index policy can be optimized.

attestation-service/Makefile
fitzthum: whoops that was for testing

attestation-service/src/token/ear_broker.rs
fitzthum: I don't really like checking whether the file exists in the caller. That seems redundant especially as we support more TeeTypes. We could make the `set_policy` and `set_default_policy` methods sync, but that's sort of unrelated to this PR.

Xynnn007: TBO, the `set_default_policy` interface is only called at creation time throughout the policy engine's lifecycle. Methods are usually a runtime conception, not initialization. Here, we want to determine the policy engine's default policy only during initialization. A good approach would be to create a builder form like

```rust
let policy_engine = PolicyEngineBuilder::new()
    .with_default_policy(<name>, <content>)
    .with_default_policy(<name>, <content>)
    .build();
```

This seems a bit heavy in some way, so I recommend simply determining the path directly above. Anyway, introducing a new function especially overlapped with an existing one is not ideal fwiw.

mythi: how about going back to the original setup but changing things to pass a map of id/path defaults.

fitzthum: @Xynnn007 The builder pattern seems like overkill for now. 

Remember that the OPA policy engine is just one implementation of the policy engine trait. The trait provides a `set_policy` method because the details of how the policies are stored are up to the implementation of the trait. Writing the policy directly to the filesystem in the caller breaks the interface of the policy engine. Setting a policy and setting a default policy seem to be two distinct operations that the policy engine trait should support so I like the idea of having `set_default_policy` be a simple wrapper.

Xynnn007: still. default policy is an initialization conception. I am not in favor of a trait function. As a trait function hints that we can call it anytime of the lifecycle of the policy agent object but it is not expected. This would be the same as set policy.

If we do not want to do grey-box operation like filesystem writing, builder would be a good choice. Or, any other proper ways avoid the new function would be fine.

fitzthum: First, `set_default_policy` can be called at any point. I can rename it to `set_policy_if_not_present` if that would be more clear. It is not strictly an initialization mechanism.

Second, 
> a trait function hints that we can call it anytime of the lifecycle of the policy agent 

I don't think that is correct. It's true that some people don't like to add `new()` to a trait because it can cause problems if one struct implements multiple traits, but it is not the case that every method needs to be available for the entire life of a struct. There are tons of well-known crates that do not follow this pattern. Think about any trait that has an `open` and `close` method. There are plenty of examples inside Trustee as well. For instance, the `session` trait has methods that can only be called at certain times.

fitzthum: Ok, I started working on a builder approach, but there are a couple snags. First, we instantiate the policy backend [via an enum](https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/policy_engine/mod.rs#L52-L73). This doesn't fit very well with the builder pattern. We could use the enum to generate the builder, but this would mean that we would need a trait for the builder as well. There are some workarounds here, but they seem worse than what we currently have.

It's also a bit inconvenient to set a policy from the builder because we need [this method](https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/policy_engine/opa/mod.rs#L50) to check the policy id. I guess we could make this a function instead. More generally tho, setting the policy from both the builder and the OPA struct itself leads to duplication. It seems a little fishy to be doing the same thing in both places. For instance, the builder struct needs to keep track of the `policy_dir_path` just like the OPA struct does. Setting the policy isn't really building the policy engine. 

Anyway, I can continue with this approach but it doesn't look very promising so this is your chance to say "let's just keep this and adjust it later." cc @Xynnn007 @mythi

Xynnn007: Ah.

Let's go back to my original comment: what we are trying to figure out is to avoid duplication apis and logic; I must admit that builder pattern might not be the best choice and sorry for leading to the wrong way. we can achieve the same goal like this [commit](https://github.com/Xynnn007/kbs/commit/6f72a4743536630da164d8213ae29a0db8cc3a4a) by multiplex the existing API.

PR #905, opened 2025-08-07
e2e-tests: run tests unprivileged

mkulke:
The e2e tests can run in an unprivileged mode, causing less pollution on self-hosted runners.


mythi (approved):
nice improvement

PR #913, opened 2025-08-13
Rework Extractor Module Interface

fitzthum:
Currently we have an Extractor interface and an ExtractorModules interface. I think we can simplify things a lot here. See commit message for more details. 

I also tweaked how the swid extractor generates reference value names. Currently the names are too specific and make writing a policy tricky. See commit message for more details.


rvps/src/extractors/extractor_modules/swid/mod.rs
```diff
@@ -107,7 +102,7 @@ mod tests {
 
         let mut found = false;
         for rv in rvs {
-            if rv.name == "NVIDIA_Corporation.GH100.96_00_74_00_1C.Measurement_12.hash1" {
+            if rv.name == "NVIDIA_Corporation.GPU.Measurement_12.hash1" {
```
Xynnn007: Is `NVIDIA_Corporation.GPU.Measurement_12.hash1` a typical rv id inside rvps? It does not include any version information. Two design ways
1. include the version inside the key itself. 
2. include the version inside the value. keys may collide.

we should consider a typical scenario where a user writes an AS policy which refers to a reference value. Do you have some ideas?

rvps/src/extractors/swid/mod.rs
```diff
@@ -41,7 +51,8 @@ impl Extractor for SwidExtractor {
             .attribute((RIMIM_NS, "PlatformManufacturerStr"))
             .ok_or(anyhow!("Could not find manufacturer information."))?
             .replace(" ", "_");
-        let edition = meta.attribute("edition")
+        let edition = meta
+            .attribute("edition")
             .ok_or(anyhow!("Could not find edition information."))?;
```
Xynnn007, line 56: This change should be included in the previous commit

rvps/src/extractors/mod.rs
```diff
@@ -5,38 +5,59 @@
...
+
+pub struct Extractors {
     /// A map of provenance types to Extractor instances
-    extractors_instance_map: HashMap<String, ExtractorInstance>,
+    extractor_map: HashMap<String, ExtractorInstance>,
+}
+
+#[derive(Deserialize, Clone, Debug, PartialEq, Default)]
+pub struct ExtractorsConfig {
+    swid_extractor: swid::SwidExtractorConfig,
```
Xynnn007: I think we can follow the plugin design in kbs like [this](https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/plugin_manager.rs#L62-L76), s.t. using an enum for `ExtractorsConfig`

rvps/src/extractors/extractor_modules/swid/mod.rs
fitzthum: > we should consider a typical scenario where a user writes an AS policy which refers to a reference value. Do you have some ideas?

That's exactly why I made this change. The version information should not be part of the reference value id because then the policy would have to change whenever the device was updated to a new version. Instead, the device will return a claim that says what the version is. The policy will check this version claim against a reference value. So the user just needs to load the RIM for the device and set a claim for the version. They don't need to touch the policy.

Xynnn007: Good. What about a collision? e.g. two different RVs with a same key `NVIDIA_Corporation.GPU.Measurement_12.hash1`, how to store and use?

Using a list to store different version objects?

fitzthum: Something like this would best be handled by adding a context selector to the RVPS. Until we have that users will just load the reference values for one version of the device.

rvps/src/extractors/mod.rs
fitzthum: I tweaked this slightly to make the swid config optional. Note that the plugin config and the extractor config aren't quite doing the same thing. For plugins, we use the config partly to enable the plugins. For extractors, the extractors are always enabled (unless the feature is disabled) and we have the option to pass in a config.

Anyway, I think this approach (a struct with optional fields) is cleaner for this and might actually be better for the plugins too. It's a lot less code and basically the same process for adding a new sub-config. I think using enums makes sense when we can only have one config enabled at a time, but it's not necessary here. A list of enums is fine but it's actually a little looser semantically because you could have multiple entries with the same value. A struct with optional fields seems like exactly what we want. wdyt?

rvps/src/extractors/extractor_modules/swid/mod.rs
Xynnn007: In my previous assumption, RVPS only had one input for the index reference value. This is to make it easier to write AS policies. If we introduce context, where should it be specified?

rvps/src/extractors/mod.rs
Xynnn007: Agreed. Only one point is to raise an info or warn when no config of specific extractor is given and a default one is used.

Xynnn007 (approved):
lgtm. Only one part for a visable default config notification

rvps/src/extractors/extractor_modules/swid/mod.rs
fitzthum: A reference value URI could be useful for this. These reference values could be something like `default/nvidia/GPU.Measurement_12.hash1`. Then if you want to load in multiple sets of reference values you can replace `default` with something else. The query for the reference value could be constructed with the first field coming from the init-data.

rvps/src/extractors/mod.rs
fitzthum: ok. warning added

PR #914, opened 2025-08-14
ear: Dereference value before as_i8

Jakob-Naucke:
When running `as_i8` on a reference to a single-digit `regorus::Value::Number`, it works fine. For a multiple-digit number, you'll get

```
Attestation evaluation failed: Policy claim value not i8
```

Dereference instead.

This probably went unnoticed because in AR4SI, "good" values are single digit and "bad" values are double digit, and attestation still failed iff we wanted it. **Very technically**, this fix could inadvertently allow access to resources that had an overly permissive resource policy, but were still effectively guarded by double digit attestation policy values. I'll leave it to the maintainers whether this requires a disclaimer on the next release.

### DIFF
--- a/attestation-service/src/token/ear_broker.rs
+++ b/attestation-service/src/token/ear_broker.rs
@@ -252,7 +252,7 @@ impl AttestationTokenBroker for EarAttestationTokenBroker {
                 .await?;
 
             for (k, v) in &policy_results.rules_result {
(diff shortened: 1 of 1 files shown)


anakrish:
I am puzzled by this. @Jakob-Naucke Can you share a tiny snippet that reproduces the issue? Please do create an issue in the regorus repo.

PR #918, opened 2025-08-18
docs: add debug flag for CoCo initdata

Xynnn007:
For CoCo, initdata will have a new field to control debugging mode. This will not influence the original overall initdata spec for it will only appear in `[data]` field.

### DIFF
--- a/.github/workflows/kbs-rust.yml
+++ b/.github/workflows/kbs-rust.yml
@@ -6,11 +6,17 @@ on:
       - "main"
     paths:
       - 'kbs/**'
+      - 'attestation-service/**'
+      - 'deps/**'
+      - 'rvps/**'
       - '.github/workflows/kbs-rust.yml'
       - 'Cargo.toml'
   pull_request:
     paths:
       - 'kbs/**'
+      - 'attestation-service/**'
+      - 'deps/**'
+      - 'rvps/**'
       - '.github/workflows/kbs-rust.yml'
       - 'Cargo.toml'
 


kbs/docs/initdata.md
```diff
@@ -239,7 +239,7 @@ producer side and consumer side use the same encoding.
 
 Confidential Containers (CoCo) leverages Initdata to inject configurations like
 [kata-agent's policy](https://github.com/kata-containers/kata-containers/blob/main/docs/how-to/how-to-use-the-kata-agent-policy.md),
-configurations for [guest components](https://github.com/confidential-containers/guest-components).
+configurations for [guest components](https://github.com/confidential-containers/guest-components) and a `debug` flag to enable debug mode for [guest components](https://github.com/confidential-containers/guest-components).
```
mythi: debug flag is a configuration too so perhaps simplify
```suggestion
configurations (including a flag to enable debugging) for [guest components](https://github.com/confidential-containers/guest-components). 
```

yafu-1:
> The new version of az-vtpm uses an underlying sev build (SEV) version 6.2.1, which contains the [arch-specific](https://doc.rust-lang.org/core/arch/x86_64/index.html) statements shown in the error message. Previously, the SEV 6.2.1 crate was located in the `verifier` crate, and its build was controlled by a feature, which did not trigger CCA build tests.
> 
> Now, az-vtpm is referenced by the ITA feature. Therefore, when the ITA feature is enabled, both az-vtpm and SEV 6.2.1 will be built, resulting in an error on the Arm platform.
> 
> For CI coverage, Let me add a commit to update KBS Rust CI to be triggered whenever changes introduced in `deps/*`, `attestation-service/*` or `rvps/*` first.
> 
> For this issue, probably introducing a `x86-64` feature in `az-vtpm` crate would help?

Sorry for that. How about disabling the ITA feature on aarch64 since it's not meaningful for aarch64 platforms?

fitzthum (approved):
I'm fine with this.

> This will not influence the original overall initdata spec for it will only appear in [data] field.

This is what I was saying in https://github.com/confidential-containers/confidential-containers/issues/303. The init-data spec allows for anything to be written into the data section, but we do not follow this in kata.

Xynnn007:
@yafu-1 It can be a good workaround. At the same time we can highlight the issue inside code, use something like a `TODO: ` comment, as @mythi mentioned to explicitly record it. Would you like to put a PR for that?

mythi:
> @yafu-1 It can be a good workaround. At the same time we can highlight the issue inside code, use something like a `TODO: ` comment, as @mythi mentioned to explicitly record it. Would you like to put a PR for that?

Is there a reason to keep the commit that added the failure? Perhaps just revert that.

yafu-1:
> @yafu-1 It can be a good workaround. At the same time we can highlight the issue inside code, use something like a `TODO: ` comment, as @mythi mentioned to explicitly record it. Would you like to put a PR for that?

Sure. I can handle that.

mythi:
> > > @yafu-1 It can be a good workaround. At the same time we can highlight the issue inside code, use something like a `TODO: ` comment, as @mythi mentioned to explicitly record it. Would you like to put a PR for that?
> > 
> > 
> > Is there a reason to keep the commit that added the failure? Perhaps just revert that.
> 
> In my understanding, to support SNP attestation reports newer than v2 on azure CVM, that commit is required.

OK, sounds like we cannot support that until the sev crate is fixed.

yafu-1:
> > In my understanding, to support SNP attestation reports newer than v2 on azure CVM, that commit is required.
> 
> OK, sounds like we cannot support that until the sev crate is fixed.
Just to confirm, do you mean building intel-trust-authority-as on aarch64 failed here?

mkulke:
sth like:

```rust
        // Determine SEV-SNP CPU generation in order to parse platform status accordingly.
        // On non-x86_64 targets, bail out with an Unknown error so the crate still compiles.
        #[cfg(target_arch = "x86_64")]
        let raw_cpuid = unsafe { std::arch::x86_64::__cpuid(0x8000_0001) }
            .eax
            .to_le_bytes();
        #[cfg(not(target_arch = "x86_64"))]
        {
            return Err(crate::error::UserApiError::Unknown);
        }
```

yafu-1:
@mythi @mkulke Thanks for the suggestion, and sorry for the late reply — I just got back from holiday. A fix has already been merged in the sev crate by @fitzthum:
https://github.com/virtee/sev/pull/329

fitzthum:
@yafu-1 can you rebase this PR

PR #920, opened 2025-08-20
KBS token parsing modifications

tylerfanelli:
This PR introduces two new changes for find and parsing the TEE public key within attestation tokens. Changes include:

- Adding a new token path to find the TEE public key, `/tee-pubkey`.
- Requiring the TEE public key JSON value to be represented as a valid JWK string.


tylerfanelli:
It seems the e2e tests are failing due to token TEE pubkey string not being in JWK format. @Xynnn007 what format is the TEE pubkey represented by in `guest_components`? Could ensure its in JWK form for uniformity?

kbs/src/token/mod.rs
```diff
@@ -96,7 +98,12 @@ impl TokenVerifier {
         for path in &self.extra_teekey_paths {
             if let Some(pkey_value) = claim.pointer(path) {
                 debug!("Extract tee public key from {path}");
-                return TeePubKey::deserialize(pkey_value).map_err(|_| Error::TeePubKeyParseFailed);
+                match pkey_value {
+                    Value::String(s) => {
+                        serde_json::from_str(s).map_err(Error::TeePubKeyParseFailed)?
+                    }
+                    _ => return Err(Error::TeePubKeyInvalidJwkString),
+                }
```
Xynnn007: Note that your new logic will assume the JWK is a string, while the original logic assumes it is a JSON object. The difference (I did not follow the JWK format but to show the difference):

Original one
```json
{
  "tee-pubkey": {
    "a": "123"
  }
}
```

change in this PR
```json
{
  "tee-pubkey": "{\"a\":\"123\"}"
}
```

tylerfanelli: Thank you, this was a mistake in the way I was serializing my key, not in the way it was being deserialized. There may still need to be a modification there but it can be addressed in another PR. I've dropped that commit.

tylerfanelli: Thanks, I modified the commit to allow the public key to be represented as a string, but not requiring it to be.

Xynnn007: Would you like to change the producer logic to be a JSON object, rather than change this place? I believe it is more reasonable to serialize JWK as an object in json than a string

tylerfanelli: I'm trying to work around a bug in another crate that is resulting in objects to be represented as strings in `serde_json`. You're correct that this is a workaround. I'll mark this as draft for now and try to fix the problem in the producing crate.

mythi:
> what format is the TEE pubkey represented by in `guest_components`?

not sure what the error is but to answer your question: https://github.com/confidential-containers/trustee/blob/main/kbs/docs/kbs_attestation_protocol.md#key-format

kbs/src/token/mod.rs
tylerfanelli: The serialization issue was fixed in `rust-ear` version `0.4.0`. Commit removed.

tylerfanelli:
@Xynnn007 #959 is a placeholder for the producer issue. This PR can move forward now.

Xynnn007 (approved):

Issue #875, opened 2025-07-25
Eventlog | Check integrity binding to the event data


Xynnn007:
Currently, when verifying the binding relationship between eventlog and rtmr/ccmr/pcr, we only extract the el_digest field in each event and then replay the calculation. However, we do not verify whether the value in the el_digest field corresponds to the original event data.

According to the TCG2 spec definition, different event types may have different bindings ways to el_digest. We need to provide verification of this binding relationship to prevent a correct digest from being followed by an incorrect eventdata. Our policy can be written based on eventdata.

pawelpros:
I run basic SHA calculation on all events from tests in Trustee and received:

List of raw event_data to hash comparison which are correct:
1. EvEfiVariableAuthority
2. EvEfiGptEvent
3. EvSeparator
4. EvEfiVariableDriverConfig
5. EvEfiAction

List of custom hash process (invalid comparison):
1. EvPlatformConfigFlags
2. EvEfiPlatformFirmwareBlob2
3. EvEfiHandoffTables2
4. EvEfiBootServicesApplication
5. EvEfiVariableBoot
6. EvEventTag
7. EvIpl
8. EvNoAction

PR #922, opened 2025-08-20
deps/eventlog: add digest match event data

pawelpros:
Added event digest comparison with event data as part of solution for #875


deps/eventlog/src/lib.rs
```diff
@@ -28,6 +28,7 @@ pub struct CcEventLog {
 #[derive(Debug, Clone, Serialize)]
 pub struct EventlogEntry {
     pub details: EventDetails,
+    pub digest_matches_event: bool,
```
Xynnn007, line 31: Any reason that we did not directly return failure when the digest does not match the event? - or, do you know why some entries does not match the digest?

mythi: there are two types of measurements. in some cases the description is just a hint what the measurement is about

Xynnn007: Can we only extract the contents that are cryptographically binded to the contents to the parsed claims, and abondon those whose integrity is not protected? The digest binding algorithm is decided by the event type, so this can be practical?

pawelpros: I checked against available samples what are stats for this checking:
[1] EV_EFI_HANDOFF_TABLES2: valid=0, invalid=4
[1] EV_EFI_PLATFORM_FIRMWARE_BLOB2: valid=0, invalid=4
**[1] EV_EFI_VARIABLE_DRIVER_CONFIG: VALID
[1] EV_SEPARATOR: VALID**
[1] EV_PLATFORM_CONFIG_FLAGS: valid=0, invalid=12
[1] EV_EFI_VARIABLE_BOOT: valid=0, invalid=18
**[2] EV_EFI_ACTION: VALID**
[1] EV_EFI_VARIABLE_AUTHORITY: VALID
**[2] EV_EFI_GPT_EVENT: VALID**
[2] EV_EFI_BOOT_SERVICES_APPLICATION: valid=0, invalid=12
[3] EV_IPL: valid=0, invalid=132
[3] EV_EVENT_TAG: valid=0, invalid=4
[1] EV_POST_CODE: valid=0, invalid=1
[2] EV_EFI_VARIABLE_BOOT: valid=0, invalid=7
**[2] EV_SEPARATOR: VALID
[4] EV_EVENT_TAG: VALID**
[0] EV_NO_ACTION: valid=0, invalid=2

As far as I checked:
- Not all events can be checked by design (either by custom way for calculating hash)
- Only subset of them are really signed event data (see list above)
- EV_EFI_PLATFORM_FIRMWARE_BLOB2 & EV_EFI_VARIABLE_BOOT cannot be checked at all 
- Rest of missing should be potentially brought from platform firmware implementation but it may vary.

From that point of view -> that's the reason why it's flag rather hard stop for processing further

pawelpros: Further more I tried to find proper way for calculating some of the tags using replying PCR extensions for particular events (some entries have for example different hash for same content) so there is logic behind it but I cannot find what exactly

Xynnn007: Thanks for the details. Looks like we cannot reach a perfect solution in short time. I have an idea, do you think this is appropriate? :

Let's add two different fields to `EventDetail` to distinguish between integrity-protected data fields and integrity-unprotected data fields. We'll also mark event types whose hashing algorithms are currently difficult to identify with "TODO" comments in the code.

Xynnn007: This would be helpful to policy writers to let them explicitly know some fields can be tampered and some cannot.

pawelpros: Hmm what about changing "bool" value to enum: PROTECTED / UNPROTECTED / NOT_APPLICABLE - one field and would be explicit ?

Xynnn007: that would be fine also

mythi: What additional value a non-boolean result would give and what would be the meaning/use-case of 'not applicable'? only `true` let's policies to check against non-binary reference values (e.g., expected AAEL log entries)  instead of expected checksums. The baseline for all events is that only checksums can be trusted.

pawelpros: I have updated PR to support currently valid events (usage for enum flag).
baseline is correct but real scenario - only subset of events can be successfully checked with hash(eventdata). For some probably it can be added and some cannot be trusted at all in this case

deps/eventlog/src/ccel/tcg_enum.rs
```diff
@@ -249,4 +249,23 @@ impl TcgEventType {
...
+                | Self::EvEfiVariableDriverConfig
+        )
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize)]
+pub enum DigestMatch {
+    Match,
+    Mismatch,
+    Unsupported,
```
mythi: My preference is to keep the boolean value as I proposed originally since `Unsupported` has no use and is also not as simple as `!can_verify_digest()`:  for example we have known `EvEventTag` events that are known not to give a digest match.

pawelpros: Intention behind approach was separating something which cannot be checked from that which can be checked but failed because mismatch

pawelpros: @Xynnn007 @mythi to not going back and forth with this -> I can update it to bool or leave it as is but needs some confirmation.

**Bool** option:
- Pros:
  - simplier for later rego usage (check against bool)
  - less implementation (no `enum` and no `can_verify_digest` method)
  - in basic scenario user probably mostly of the time will use bool true or skip this field usage if it will be always false
  - if there will be any new implementation around checking event digest appropriate for particular event it will be transparent for implementation (new overloaded method for checking and that's all)
- Cons:
  - There will be no track for events that for now will be always treated as "not supported" / "not protected" (but maybe documentation will be enough for it)
  
**Enum** option:
- Pros
  - easier to track if this field can be used as requirement in rego
  - in defined can_verify_digest there will be always 
  - [OPEN] - maybe there will be usage for it better than read the docs ?
- Cons
  - Enum + more complexity in solution
  - additional method `can_verify_digest` which will behave as additional place to check if new implementation will be added and probably will be some kind of documentation
  - if new custom digest check will be implemented for particular event it will need 2 changes (method itself + updating can_verify_digest

Xynnn007: Ok. Bool sounds good to me. We can perfect them later and now just record them as TODO like you said

pawelpros: Docs updated -> bool approach applied

mythi (approved):
LGTM

PR #923, opened 2025-08-20
attestation-service: drop apt-key usage in Dockerfiles and tests

mythi:
apt-key has been deprecated for some time. With Debian 13 / trixie it's gone. rust:latest is a moving target and jumped from bookworm to trixie and builds started failing due to missing apt-key.

Remove all occurences of apt-key to gpg based tooling.


BbolroC:
I was wondering how we deal with the failing CI jobs for arm. Just ignoring them and moving forward?

BbolroC (approved):
LGTM, thanks @mythi !

Issue #863, opened 2025-07-19
Promote EAR Token issurance


Xynnn007:
### Background

Currently, the Attestation Service (AS) has some areas for improvement in supporting EAR, including the following issues:
1. EAR only supports a single Policy ID.
2. There is an inconsistency between the Policy ID provided during the verification and the one automatically read in the backend. For example, if the Policy ID provided for verification is `policy-id-test`, the backend Policy ID automatically read is `policy-id-test_<tee-class>`. This means the uploaded policy name does not match the one actually being used. Users need to consider different TEE classes and place different policies accordingly.

---

### Modification Goals

We aim to enable EAR to support multiple policies, while minimizing the user's effort. The goal is to make it intuitive such that "the policy ID I want to use is exactly the one I input."

---

### Modification Suggestions

Here are several different solutions, each with its own pros and cons. I have listed them here for selection.

#### Existing Solution

The input parameters for the AS are as follows:
```json
{
    "verification_requests": [
        {
            "evidence": <evidence>,
            "tee": <used-to-select-tee-handler>,
         },
         ...
    ],
    "policy-ids": <policy-id-slice>
}
```

In this solution, when verifying each individual evidence, the Verifier plugin will return an additional `tee-class` value. The AS then verifies that the Policy ID used for this evidence is `policy-id_<tee-class>`. When organizing the returned EAR, only the first Policy ID from the "policy-ids" field will be included.

#### Improvement Solution 1
The input parameters for the AS are as follows:

```json
{
    "verification_requests": [
        {
            "evidence": <evidence>,
            "tee": <used-to-select-tee-handler>,
            "policy-ids": <policy-id-slice>
         },
         ...
    ]
}
```

Each verification request is translated into an individual EAR submodule. The name of the submodule corresponds to the tee field name (or derived rom the verifier plugin as current. Another choice is to define more finegrained tee class like `NvidiaGpu` and `NvidiaSwitch` as tee enum, but this is out of topic of this issue), and the Trustworthiness Vector is computed based on the policies’ evaluations. The minimum value of the same Trustworthiness Vector type is selected, and then the final evaluation is made.

Moreover, the appraisal-policy-id is a concatenated string, using a separator (e.g., comma), to represent all the Policy IDs.

A typical result EAR (the AS response is a JWT, and the claims are as follows):

```json
{
  "eat_profile": "tag:github.com,2024:confidential-containers/Trustee",
  "iat": 123, 
  "ear.verifier-id": {
[...]

PR #928, opened 2025-08-26
plugins: expose init-data to plugins

fitzthum:
The init-data contains measured guest configuration. Exposing this to the plugin backend can allow for plugins to implement sophisticated logic based on measured guest data.

For example, a plugin could parse the guest policy and issue a certificate based on the properties of the guest (i.e. which container images are allowed to run).

Some of this logic can already be implemented today via a KBS policy, but exposing the init-data to the plugin itself allows for more flexibility.

The init-data is not always present. If a guest does not set the init-data or if the attestation agent does not validate the init-data and put the init_data_claims into the token, then this feature can't be used. Thus, the init-data argument is an Option.

### DIFF
--- a/kbs/src/api_server.rs
+++ b/kbs/src/api_server.rs
@@ -269,7 +269,7 @@ pub(crate) async fn api(
                 // Plugin calls need to be authorized by the admin auth
                 core.admin_auth.validate_auth(&request)?;
                 let response = plugin
-                    .handle(&body, query, additional_path, request.method())
+                    .handle(&body, query, additional_path, request.method(), None)
                     .await
                     .map_err(|e| Error::PluginInternalError { source: e })?;
 
@@ -282,6 +282,8 @@ pub(crate) async fn api(
                     .map_err(|_| Error::TokenNotFound)?;
 
                 let claims = core.token_verifier.verify(token).await?;
+                let init_data = claims
+                    .pointer("/submods/cpu0/ear.veraison.annotated-evidence/init_data_claims");
 
                 let claim_str = serde_json::to_string(&claims)?;
 
@@ -299,7 +301,7 @@ pub(crate) async fn api(
                 KBS_POLICY_APPROVALS.inc();
 
                 let response = plugin
-                    .handle(&body, query, additional_path, request.method())
+                    .handle(&body, query, additional_path, request.method(), init_data)
                     .await
                     .map_err(|e| Error::PluginInternalError { source: e })?;
                 if plugin
--- a/kbs/src/plugins/implementations/nebula_ca.rs
+++ b/kbs/src/plugins/implementations/nebula_ca.rs
@@ -399,6 +399,7 @@ impl ClientPlugin for NebulaCaPlugin {
         query: &str,
         path: &str,
         method: &Method,
+        _init_data: Option<&serde_json::Value>,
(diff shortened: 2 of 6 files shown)


kbs/src/api_server.rs
```diff
@@ -282,6 +282,8 @@ pub(crate) async fn api(
                     .map_err(|_| Error::TokenNotFound)?;
 
                 let claims = core.token_verifier.verify(token).await?;
+                let init_data = claims
+                    .pointer("/submods/cpu0/ear.veraison.annotated-evidence/init_data_claims");
```
mythi, line 286: since it's just a subset of the token claims, any reason why we shouldn't make the full token available instead?

fitzthum: Yeah good point. There's a little bit of a tradeoff here. Giving plugin authors the full plugin could make the interface more powerful but it makes things more complicated for plugin authors. I couldn't think of a use case for a plugin using the whole token so I thought this would be a little more convenient. wdyt?

Xynnn007:
I think bring semantics related to KBS/kbs-protocol from token to KBS is a good point. But if we bring all initdata from the token it would be somehow coupled between AS and KBS. My suggestion is to define a good spec that records fields/data that KBS protocol / KBS uses. The fields can be set by runtime data and/or initdata. And the Attestation Service plugin of KBS can extract them and deliver to upper layer handlers, like plugins. This will help to cover the gap of different AS backend and abstraction of AS and KBS. For example, a `client_id` can be set via initdata. Thus then we can extract exactly `client_id` from the initdata. In this way what delivered to KBS logic from token is exactly a struct with field `client_id` and other defined fields.

The point two of https://github.com/confidential-containers/trustee/issues/863#issuecomment-3212657019

fitzthum:
@Xynnn007 I feel like this is best left to the plugins themselves. We might be able to standardize some parts of the init-data, but we don't really know what plugins will need and it's likely that different plugins will have different requirements. It's probably simplest to have the plugins parse the init-data themselves.

It's true that this does create some sort of requirement for the init-data to exist and for the AS to return it in the token, but note that we are passing `Option<Value>` here so really all we're saying is that some plugins might require init-data, which seems reasonable. Not every plugin will necessarily be compatible with every deployment.

Xynnn007:
> I feel like this is best left to the plugins themselves. We might be able to standardize some parts of the init-data, but we don't really know what plugins will need and it's likely that different plugins will have different requirements. It's probably simplest to have the plugins parse the init-data themselves.

Initdata represents metadata for the entire instance, while the portion used by KBS is defined only within the KBS remote attestation protocol. Consider that the `policy.rego` and `cdh` configuration files are not relevant to KBS, as they are workload-independent and only concern the specific PodVM instance.

Therefore, we need to define the content required by an extensible plug-in within the scope of the KBS protocol, such as extending the AA configuration file and e.g. `cc_kbc` related parts, allowing it to accept formatted configuration items such as `[[plugin.XXX]]` type than define it in the whole initdata scope.


> It's true that this does create some sort of requirement for the init-data to exist and for the AS to return it in the token, but note that we are passing `Option<Value>` here so really all we're saying is that some plugins might require init-data, which seems reasonable. Not every plugin will necessarily be compatible with every deployment.

Yes, `Option<Value>` makes sense and I am in favor of it.

PR #929, opened 2025-08-27
deps/verifier: Add DCAP supplemental data claims for SGX/TDX

pawelpros:
List of added claims:
- earliest_expiration_date,
- earliest_issue_date,
- is_cached_keys,
- is_dynamic_platform,
- is_smt_enabled,
- latest_issue_date,
- pck_crl_num,
- platform_provider_id,
- root_ca_crl_num,
- root_key_id,
- sgx_type,
- tcb_date,
- tcb_eval_num.

Applicable to both SGX and TDX, more details in their respective "Quote Generation Library and Quote Verification Library" docs.

### DIFF
--- a/attestation-service/docs/parsed_claims.md
+++ b/attestation-service/docs/parsed_claims.md
@@ -40,22 +40,22 @@ UEFI event log entry contains below fields:
 
 The following fields always exist.
 - `tdx.quote.header.version`: The quote format version. Now supports 4 and 5.
-- `tdx.quote.header.att_key_type`: enum of the algorithm used in signature.
+- `tdx.quote.header.att_key_type`: Enum of the algorithm used in signature.
 - `tdx.quote.header.tee_type`: TDX is always 0x81.
-- `tdx.quote.header.reserved`: reserved.
+- `tdx.quote.header.reserved`: Reserved.
 - `tdx.quote.header.vendor_id`: UID of QE Vendor. QE is a signed software component inside TEE to help to generate tdx quote.
 - `tdx.quote.header.user_data`: Custom attestation key owner data.
 - `tdx.quote.body.mr_config_id`: Software-defined ID for non-owner-defined configuration of the guest TD – e.g., run-time or OS configuration.
 - `tdx.quote.body.mr_owner`: Software-defined ID for the guest TD’s owner.
 - `tdx.quote.body.mr_owner_config`: Software-defined ID for owner-defined configuration of the guest TD – e.g., specific to the workload rather than the run-time or OS.
 - `tdx.quote.body.mr_td`: Measurement of the initial contents of the TD.
-- `tdx.quote.body.mrsigner_seam`: measurement of a 3rd party tdx-module's signer (SHA384 hash). If it is 0, the tdx-module is from Intel.
-- `tdx.quote.body.report_data`: software defined ID for non-owner-defined configuration on the guest TD.
-- `tdx.quote.body.seam_attributes`: for tdx1.0, must be 0.
+- `tdx.quote.body.mrsigner_seam`: Measurement of a 3rd party tdx-module's signer (SHA384 hash). If it is 0, the tdx-module is from Intel.
+- `tdx.quote.body.report_data`: Software defined ID for non-owner-defined configuration on the guest TD.
+- `tdx.quote.body.seam_attributes`: For tdx1.0, must be 0.
 - `tdx.quote.body.td_attributes`: TD's attributes.
-- `tdx.quote.body.mr_seam`: Measurement of the SEAM module
+- `tdx.quote.body.mr_seam`: Measurement of the SEAM module.
 - `tdx.quote.body.tcb_svn`: TEE hardware tcb version, defined and meaningful to Intel. everytime firmware updates this field will change.
-- `tdx.quote.body.xfam`: TD's XFAM
+- `tdx.quote.body.xfam`: TD's XFAM.
 - `tdx.quote.body.rtmr_0`: Runtime extendable measurement register 0.
 - `tdx.quote.body.rtmr_1`: Runtime extendable measurement register 1.
 - `tdx.quote.body.rtmr_2`: Runtime extendable measurement register 2.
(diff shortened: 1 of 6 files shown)


attestation-service/docs/parsed_claims.md
```diff
@@ -64,14 +64,27 @@ The following fields always exist.
...
+- `tdx.latest_issue_date`: Date time value in RFC3339 format - The latest issueDate value among all collaterals.
+- `tdx.pck_crl_num`: Indication of the freshness of the PCK cert used.
+- `tdx.platform_provider_id`: The Platform Provisioning ID.
+- `tdx.root_ca_crl_num`: Indication of the freshness of the Root CA cert used.
+- `tdx.root_key_id`: ID of the collateral’s root signer (hash of Root CA’s public key SHA-384).
+- `tdx.sgx_type`: The type of memory used in SGX. Can be one of (`Standard`, `Scalable`, `Scalable with Integrity`).
+- `tdx.tcb_date`: Date time value in RFC3339 format - Earliest date between tcbInfo and qeIdentity.
+- `tdx.tcb_eval_num`: Indication of the freshness of the reference values used.
+- `tdx.tcb_status`: TCB status which can be one of: OK, Min, OutOfDate, OutOfDateConfigurationNeeded, InvalidSignature, Revoked, Unspecified, SoftwareHardeningNeeded, ConfigurationAndSoftwareHardeningNeeded, TdRelaunchAdvised, TdRelaunchAdvisedConfigurationNeeded, Max
```
Xynnn007:
```suggestion
- `tdx.tcb_status`: TCB status which can be one of: `OK`, `Min`, `OutOfDate`, `OutOfDateConfigurationNeeded`, `InvalidSignature`, `Revoked`, `Unspecified`, `SoftwareHardeningNeeded`, `ConfigurationAndSoftwareHardeningNeeded`, `TdRelaunchAdvised`, `TdRelaunchAdvisedConfigurationNeeded`, `Max`
```

deps/verifier/src/sgx/claims.rs
```diff
@@ -89,14 +84,11 @@ pub fn generate_parsed_claims(quote: sgx_quote3_t) -> Result<TeeEvidenceParsedCl
...
     );
     parse_claim!(quote_body, "mr_enclave", quote.report_body.mr_enclave);
-    parse_claim!(quote_body, "reserved2", quote.report_body.reserved2);
     parse_claim!(quote_body, "mr_signer", quote.report_body.mr_signer);
-    parse_claim!(quote_body, "reserved3", quote.report_body.reserved3);
     parse_claim!(quote_body, "config_id", quote.report_body.config_id);
     parse_claim!(quote_body, "isv_prod_id", quote.report_body.isv_prod_id);
     parse_claim!(quote_body, "isv_svn", quote.report_body.isv_svn);
     parse_claim!(quote_body, "config_svn", quote.report_body.config_svn);
-    parse_claim!(quote_body, "reserved4", quote.report_body.reserved4);
```
Xynnn007: Is the field `reserved4` deleted from the quote? The same question for other `reserved*`. I think if it is still there, we can keep it as-is.

attestation-service/docs/parsed_claims.md
pawelpros: Applied

deps/verifier/src/sgx/claims.rs
pawelpros: Ok reverted changes to reserved fields

mythi: > I think if it is still there, we can keep it as-is.

the thinking was to drop claims from the token that have no use-case. can you think of any?

attestation-service/docs/parsed_claims.md
mythi: tcb_status docs could also stay unchanged in this changeset

deps/verifier/src/sgx/claims.rs
Xynnn007: @mythi I have no idea. But to sincerely try to given more detail about the evidence is our aim. It is there, so we parse it out.

mythi: parsing is fine but adding zero fillers to the claims is not necessary. but this PR is for other purpose so we can get back to that reserved topic in a separate PR

pawelpros: The only usage of reserved field is keep up with full list of possible values so if reserved won't be reserved any more probably it will be renamed so code won't compile and dev will know that field previously reserved is now in use -> but from claims perspective it's useless in my opinion.

But as Mythi said -> it can be aligned later on in separate PR

Xynnn007: Yes. We can rename/split the field when it's defined in future. For now users might not use the claims in policy, but we can still leave them there

attestation-service/docs/parsed_claims.md
```diff
@@ -64,14 +64,27 @@ The following fields always exist.
 - `tdx.quote.size`: Quote body length. Only quote format V5 contains this field.
 - `tdx.quote.body.tee_tcb_svn2`: Array of TEE TCB SVNs (for TD preserving).
 - `tdx.quote.body.mr_servicetd`: If there is one or more bound or pre-bound service TDs, this field is the SHA384 hash of the `TDINFO`s of those service TDs bound. Else, this field is 0.
-- `tdx.tcb_status`: TCB status which can be one of: OK, Min, OutOfDate, OutOfDateConfigurationNeeded, InvalidSignature, Revoked, Unspecified, SoftwareHardeningNeeded, ConfigurationAndSoftwareHardeningNeeded, TdRelaunchAdvised, TdRelaunchAdvisedConfigurationNeeded, Max
```
mythi: as suggested, this change can move to #930

pawelpros: Reverted

pawelpros:
@mythi commit msg re-phrased - should be detailed right now :)

Xynnn007 (approved):
LGTM. Thanks @pawelpros !

PR #930, opened 2025-08-27
deps/verifier: Change mapping values for TCB_STATUS

pawelpros:


deps/verifier/src/intel_dcap/claims.rs
```diff
@@ -10,27 +10,36 @@ struct SgxQlQvResultWrapper(pub sgx_ql_qv_result_t);
 impl SgxQlQvResultWrapper {
     pub fn as_str_arr(&self) -> Value {
         let mapping = match self.0 {
-            sgx_ql_qv_result_t::SGX_QL_QV_RESULT_OK => "OK",
-            sgx_ql_qv_result_t::SGX_QL_QV_RESULT_MIN => "Min",
```
Xynnn007, line 14: Hi, how do we handle `SGX_QL_QV_RESULT_MIN`?

pawelpros: From recent experience `SGX_QL_QV_RESULT_CONFIG_NEEDED` is alias for MIN that's why i have removed `SGX_QL_QV_RESULT_MIN` as duplicated entry

Xynnn007 (approved):
LGTM. Thanks @pawelpros !

Xynnn007:
@mythi Oh. I did not see any comments against this. Does it make sense to put a revert and discuss more here?

pawelpros:
@mythi as part for further policy hardening this change is useful and more descriptive than previous one in my opinion. Policy validation against this field is next step after this PR. By keeping tcb_status value aligned with knowledge what should be expected  from policy point of view it will be easier to create specific policy requirements based on available options. 

But yes -> it's breaking change in terms of policy usage.

PR #931, opened 2025-08-27
Bump SEV Crate to fix ARM build issue

fitzthum:
Test https://github.com/virtee/sev/pull/329

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -716,7 +716,7 @@ version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b3d0900c6757c9674b05b0479236458297026e25fb505186dc8d7735091a21c"
 dependencies = [
- "bincode",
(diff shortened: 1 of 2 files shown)


fitzthum:
Compilation error seems to be good https://github.com/confidential-containers/trustee/actions/runs/17279027197/job/49043047711?pr=931

Xynnn007:
Good. https://github.com/virtee/sev/pull/329 has been merged. Let's move it forward

fitzthum:
ok, we know have a `6.3.1` release so I've updated the Cargo file and rebuilt the lock and things work. (I dropped the test commit from this PR, but I also ran it through the CI with that commit). Technically this PR might not be necessary since Cargo will probably pick up the point release on its own. Still, it's good to rebuild the lockfile. Fortunately the az crates are happy with the point release as well. 

cc @Xynnn007 @mythi

mythi (approved):
Thanks!

PR #938, opened 2025-09-03
Revert "deps/verifier: Change mapping values for TCB_STATUS"

Xynnn007:
Reverts confidential-containers/trustee#930

See comment https://github.com/confidential-containers/trustee/pull/930#issuecomment-3248391728

cc @mythi @pawelpros


pawelpros:
@Xynnn007 let's close this revert -> I'll prepare separate PR to move from vec! to string enum - we will have in history second approach so we can treat it as alternative if we need to change it back (if needed).

PR #944, opened 2025-09-08
AS: fix unit test with extra field for rvps

Xynnn007:
To fix error

error[E0063]: missing field `extractors` in initializer of `reference_value_provider_service::Config`
Error:   --> attestation-service/src/config.rs:94:42
   |
94 |         rvps_config: RvpsConfig::BuiltIn(RvpsCrateConfig {
   |                                          ^^^^^^^^^^^^^^^ missing `extractors`

PR #945, opened 2025-09-08
AS: allow underscore trustworthiness claim names in ear policy

Xynnn007:
The definition of Trustworthiness Claims in AR4SI uses hyphens, but in rego a variable can only be concatted with underscores.

Thus we explicitly convert underscores to hyphens in the ear processor to allow underscores in the rego policy.


Xynnn007:
Let me get this rebased after https://github.com/confidential-containers/trustee/pull/944 is merged.

fitzthum (approved):
Good catch. We fixed this previously in https://github.com/confidential-containers/trustee/pull/742 but it looks like I dropped it in the multi-device PR. The fix is still there in the tests and we don't typically use the vectors with underscores so nobody noticed.

PR #954, opened 2025-09-11
deps/verifier: Revert vector result approach for TCB_STATUS

pawelpros:
- updated description for tcb_status claim field
- reverted tcb_status result format from array to simple string
- Updated default ear policy rego

### DIFF
--- a/attestation-service/docs/parsed_claims.md
+++ b/attestation-service/docs/parsed_claims.md
@@ -85,7 +85,15 @@ The following fields always exist.
 - `tdx.sgx_type`: The type of memory used in SGX. Can be one of (`Standard`, `Scalable`, `Scalable with Integrity`).
 - `tdx.tcb_date`: Date time value in RFC3339 format - Earliest date between tcbInfo and qeIdentity.
 - `tdx.tcb_eval_num`: Indication of the freshness of the reference values used.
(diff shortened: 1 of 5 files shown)


attestation-service/docs/parsed_claims.md
```diff
@@ -85,7 +85,16 @@ The following fields always exist.
...
+  - `ConfigurationAndSWHardeningNeeded` - This status is combination of `SWHardeningNeeded` and `ConfigurationNeeded`.
+  - `OutOfDate` - The attesting platform software and/or firmware is not patched in accordance with the latest TCB Recovery (TCB-R).
+  - `OutOfDateConfigurationNeeded` - The attesting platform is not patched in accordance with the latest TCB-R. Hardware configuration is needed.
+  - `TDRelaunchAdvised` - The platform firmware and software are at the latest security patching level but the TD was launched prior to the application of new TDX TCB components using a TD Preserving update. Re-launching the TD will change the attestation result.
+  - `TDRelaunchAdvisedConfigurationNeeded` - The platform firmware and software are at the latest security patching level but there are platform hardware configurations that may expose the TD to vulnerabilities. Re-launching the TD will change the attestation result.
+  - `TerminalStatus` - This is helper status which informs that received status shouldn't be processed further. In standard scenario this status will never be actual claim value.
```
mythi:
```suggestion
  - `TerminalStatus` - This is helper status which informs that received status shouldn't be processed further. Trustee will never return this `tcb_status` claim but a quote verification error instead.
```

pawelpros: TerminalStatus was moved as comment to code -> no needs for introducing it in docs

PR #955, opened 2025-09-11
docs: Reformat + aligned TEE available options

pawelpros:
- Reformatted config.md
- Reformatted kbs_attestation_protocol.md
- Added actual TEE values

### DIFF
--- a/kbs/docs/kbs_attestation_protocol.md
+++ b/kbs/docs/kbs_attestation_protocol.md
@@ -74,7 +74,7 @@ The payload format of the request is as follows:
   "version": "0.1.1",
   /*
    * Type of HW-TEE platforms where KBC is located,
-   * e.g. "intel-tdx", "amd-sev-snp", etc.
+   * e.g. "tdx", "sev", "snp", etc.
    */
   "tee": "$tee",
   /* Reserved fields to support some special requests sent by HW-TEE. */


kbs/docs/kbs_attestation_protocol.md
```diff
@@ -594,34 +619,40 @@ The payload of the POST request should like:
...
- The payload of the POST request should like:
+
+The payload of the POST request should like:
 
  ```json
  {
-    "policy": <base64encoded policy>
- }
+  "policy": <base64encoded
+  policy>
```
Xynnn007: We might not need the new line here?

pawelpros: Thanks, fixed that already - auto formatter was not so good.

kbs/docs/kbs_attestation_protocol.md
```diff
@@ -91,8 +91,26 @@ KBC can communicate normally according to this field.
 
 - `tee`
 
-Used to declare the type of HW-TEE platform where KBC is located, the valid
-values are `intel-tdx`, `intel-sgx` and `amd-sev-snp`.
+Used to declare the type of HW-TEE platform where KBC is located, the valid values are:
```
mythi: The table below is misleading because we would only use the "primary tee" here and not any of the devices. Also, from the protocol perspective there are not "valid" or "invalid" values as such. 

How  about:
`Used to declare the type of HW-TEE platform where KBC is located. Currently known HW-TEE platforms are:`

(with the devices removed)

pawelpros: Done + changed commit msg to include info about tool used to reformat

Issue #904, opened 2025-08-07
[RFC] Helm Chart vs Operator for Trustee Deployment


zvonkok:
### Describe the bug

**Background**

Trustee can already be installed in Kubernetes via the dedicated trustee-operator CRD.

Yet many users run GitOps pipelines (ArgoCD, Flux, etc.) that treat “a Helm chart in an OCI repo” as the lingua franca for deploying third-party components. For these users, an operator is an extra controller to secure, upgrade and RBAC.

**Problem Statement**

We need to decide whether the project:
- Officially ships a Helm chart (in-tree or in a sibling repo);
- Continues to recommend only the operator;
- Maintains both, with clear guidance on when to pick which.

Without a clear stance we risk:
- Diverging install paths → fractured bug reports & docs;
- Duplicate effort keeping two deployment artifacts in sync;
- Confusion about versioning (chart vs operator bundle vs Trustee image tag

**Helm Chart – Pros / Cons**

Aspect | Helm Chart | Operator
-- | -- | --
Installation UX | 1-line helm install, fits GitOps | Requires OLM/OperatorHub or kustomize build && kubectl apply
Lifecycle | Stateless; upgrades are “replace” semantics | Reconciliation & drift-correction built-in
Custom Resources | None (simpler RBAC) | Introduces KbsConfig CRD
Configurability | values.yaml is familiar; can gate features behind --set | CR fields + configmaps; richer validation
Complex orchestrations (micro-services split) | Template logic gets hairy | Operator handles multi-deployment wiring
Security surface | No additional controller to trust | Extra controller, watch permissions

**Proposed Next Steps**

Gather feedback on questions from maintainers & users.

If consensus emerges to add a chart:
- Draft a minimal POC chart aligning with current operator defaults.
- Automate linting (Chart Testing CI) and OCI artifact publishing.
- Document the new Helm chart flow

mythi:
> I think the decision of one vs the other decides what target you hope/expect will pick up trustee.

I also think a unified install approach (either Helm or the operator) won't work because "vanilla k8s" users tend to prefer Helm and the operator based flows are commonly used in Openshift environments where the pre-installed OLM nicely integrates with "community operators " from operatorhub.io and Redhat operators.

fitzthum:
What are the key features of the config CR? What would we be missing if we just used a Helm chart vs operator plus Helm chart.

lmilleri:
[...]

PR #956, opened 2025-09-12
kbs-config.toml: introduce admin_api_ro flag to forbid set_* operations

esposem:
In operator-based deployments, kbs-client (and in general the admin API)  should not be allowed to try and modify the resources, for two reasons:
    
    1. trustee runs in a container so the fs is r/o
    2. management of such resources is by design delegated to kubernetes configmaps

Therefore add admin_api_ro flags defaulting to false under the [admin]  section of kbs-config.toml.

### DIFF
--- a/kbs/docs/config.md
+++ b/kbs/docs/config.md
@@ -204,10 +204,11 @@ Detailed [documentation](https://docs.trustauthority.intel.com).
 
 The following properties can be set under the `[admin]` section.
 
-| Property          | Type    | Description                                                       | Required | Default |
-|-------------------|---------|-------------------------------------------------------------------|----------|---------|
-| `auth_public_key` | String  | Path to the public key used to authenticate the admin APIs        | No       | None    |
-| `insecure_api`    | Boolean | Whether KBS will not verify the public key when called admin APIs | No       | `false` |
+| Property               | Type    | Description                                                       | Required | Default |
+|------------------------|---------|-------------------------------------------------------------------|----------|---------|
+| `auth_public_key`      | String  | Path to the public key used to authenticate the admin APIs        | No       | None    |
+| `insecure_api`         | Boolean | Whether KBS will not verify the public key when called admin APIs | No       | `false` |
+| `admin_api_read_only`  | Boolean | Limit the admin APIs to only get resources.                       | No       | `false` |
 
 ### Policy Engine Configuration
 
--- a/kbs/src/admin/mod.rs
+++ b/kbs/src/admin/mod.rs
@@ -2,7 +2,10 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-use actix_web::{http::header::Header, HttpRequest};
+use actix_web::{
+    http::{header::Header, Method},
+    HttpRequest,
+};
 use actix_web_httpauth::headers::authorization::{Authorization, Bearer};
 use config::AdminConfig;
 use jwt_simple::{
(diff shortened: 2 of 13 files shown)


kbs/docs/config.md
```diff
@@ -207,6 +207,7 @@ The following properties can be set under the `[admin]` section.
 |--------------------------|--------------|------------------------------------------------------------------------------------------------------------|----------|----------------------|
 | `auth_public_key`        | String       | Path to the public key used to authenticate the admin APIs                                                 | No       | None                 |
 | `insecure_api`           | Boolean      | Whether KBS will not verify the public key when called admin APIs                                          | No       | `false`              |
+| `admin_api_ro`           | Boolean      | Limit the admin APIs to only get resources. Useful for operator-based deployments                      | No       | `false`              |
```
mkulke: missing an ident?

mkulke (commented):
thansk! I think this flag is necessary feature atm. a couple of questions on cosmetics and semantics:

- should we call the toggle and variable/field in the code `admin_api_read_only`? `ro` is not very grepable, and it's unidiomatic in rust to sacrifice characters for clarity.
- Is it only the admin_api or the whole state of the service that is read_only? because only the admin can change state, we can drop the `admin_api_...` prefix.
- it looks like we apply the restriction to POST methods. are we using PUT/DELETE/PATCH to alter the state? maybe we can turn it around and assert GET | HEAD when the read_only toggle is set? (and hope that no one in a plugin corner performs state changes with GET 🤞)
- AFAIU, auth_public_key=Some(...) would be overridden by read_only=true. Can we change the internal representation (not the user toggles) of AdminConfig to reflect that by using an auth enum: 

```rust
enum AdminAuth {
   None,
   PubKey(PathBuf),
   ReadOnly,
}

pub struct AdminConfig {
   pub insecure_api: bool,
   pub auth: AdminAuth,
}
```

kbs/src/admin/mod.rs
```diff
@@ -35,12 +44,17 @@ impl TryFrom<AdminConfig> for Admin {
...
         Ok(Self {
             public_key: Some(key),
+            admin_api_read_only: false,
         })
     }
 }
 
 impl Admin {
     pub(crate) fn validate_auth(&self, request: &HttpRequest) -> Result<()> {
+        if request.method() != Method::GET && self.admin_api_read_only {
```
mkulke: maybe allow HEAD too?

mkulke (approved):
looks good, maybe run the formatter to satisfy the ci

mkulke:
> basically the storage way to upload a rsvp via kbs-config.toml bypasses all security checks

I don't think this is an argument about security. it's rather about processing-functionality available in the rvps-tool that is absent when using the trustee operator.

(the overarching question, if I understood it correctly, would be whether it's desirable at all to use the trustee operator for deployment and management. see [here](https://github.com/confidential-containers/trustee/issues/904). my take: afaik the operator is currently the closest to anything resembling a production deployment, since authn/z, state and other non-functional aspects are still WIP in trustee itsself)

esposem:
FWIW I added another flag for the rvps standalone server too, but didn't document it in kbs as it doesn't make sense (rvps running in kbs is only accessible via admin apis as far as I understood). This should cover also the rvps-tool case.

rvps/src/config.rs
```diff
@@ -15,6 +15,9 @@ pub struct Config {
 
     #[serde(default)]
     pub extractors: Option<ExtractorsConfig>,
+
+    #[serde(default)]
+    pub rvps_read_only: bool,
```
Xynnn007: This is in scope of RVPS now, so we might not need the prfix `rvps_`?

fitzthum:
Btw @esposem let me clarify something. Is it all operator-based deployments that don't support writing to the filesystem or is that just on OCP?

lmilleri:
> Btw @esposem let me clarify something. Is it all operator-based deployments that don't support writing to the filesystem or is that just on OCP?

@fitzthum I think the rationale of having this PR is to make sure that in operator-based deployments, the trustee instance is enforced to be (almost) stateless. This should help the trustee HA use case (every trustee instance must have the same file system content - policies, ref-values, etc). This works when using k8s configmaps and secrets for modifications.
Btw maybe this is just the first iteration, considering the future evolution of trustee, e.g. fs -> db, etc

fitzthum:
Ok, now we have more sophisticated admin roles that allow fine-grained control over which endpoints are enabled/

@esposem please re-open if still needed.

PR #958, opened 2025-09-15
Update rust-ear dependency

tylerfanelli:
Update the rust-ear dependency to latest version.


fitzthum:
I should also return to https://github.com/confidential-containers/trustee/pull/664 at some point now that we have a release.

PR #960, opened 2025-09-16
kbs: Update kbs-types, remove duplicate definitions

tylerfanelli:
CompositeEvidence, RuntimeData, and InitData were all being deserialized with each KBS attestation request. Therefore, they were further integrated into the KBS protocol and added to kbs-types. Update kbs-types and use its definitions of these structs.

### DIFF
--- a/.github/workflows/kbs-e2e-azure-vtpm.yml
+++ b/.github/workflows/kbs-e2e-azure-vtpm.yml
@@ -37,7 +37,7 @@ jobs:
   #     contents: read
   #   with:
   #     runs-on-test: '["self-hosted","azure-cvm-tdx"]'
-  #     tee: aztdxvtpm
+  #     tee: az-tdx-vtpm
   #     tarball: kbs.tar.gz
 
   snp-e2e-test:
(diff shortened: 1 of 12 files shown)


tylerfanelli:
Marked as draft, waiting on https://github.com/confidential-containers/kbs-types/pull/74

tylerfanelli:
> LGTM

Thanks. @mkulke I had to modify the rego policy for the Azure vTPM strings. As it turns out hyphenated variable names are converted to snake case. In the rego policy, az-{snp, tdx}-vtpm actually needs to be represented as az_{snp, tdx}_vtpm. I've made this change.

tylerfanelli:
Seems the two failing tests are unrelated to this change.

mkulke:
> > LGTM
> 
> Thanks. @mkulke I had to modify the rego policy for the Azure vTPM strings. As it turns out hyphenated variable names are converted to snake case. In the rego policy, `az-{snp, tdx}-vtpm` actually needs to be represented as `az_{snp, tdx}_vtpm`. I've made this change
> 
> This only applies to the variable name **within the rego policy**.

thank you. I wasn't aware that we use values as keys in the claims hierarchy. but maybe we can fix that in general and turn this into a discriminated union type (not in this PR, though):

```
input.az_snp_vtpm.reported_tcb_bootloader == "1.2.3"
```

```
input.tee == "az-snp-vtpm"
input.reported_tcb_bootloader == "1.2.3"
```

mkulke:
@tylerfanelli I opened #972 to discuss this

PR #966, opened 2025-09-23
kbs: update ITA v2 token claim paths

mythi:
With the move to ITA v2 appraisal API that supports composite evidence, the token claim paths got changed. The claims we used previously are now under 'tdx'.


mythi:
tests are now passing too

Issue #971, opened 2025-09-25
[pkcs11 plugin] different errors


nmwael:
I am in the midst of connecting the plugin to our HSM,

First error I got was a compile error:

```
error[E0308]: mismatched types
   --> kbs/src/plugins/implementations/pkcs11.rs:141:68
    |
141 |         let attribute_map = session.get_attribute_info_map(object, value_attribute.clone())?;
    |                                     ----------------------         ^^^^^^^^^^^^^^^^^^^^^^^ expected `&[AttributeType]`, found `Vec<AttributeType>`
    |                                     |
    |                                     arguments to this method are incorrect
    |
    = note: expected reference `&[AttributeType]`
                  found struct `Vec<AttributeType>`
note: method defined here
   --> /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cryptoki-0.10.0/src/session/object_management.rs:486:12
```

I fixed that error here:
https://github.com/nmwael/trustee/commit/1438160aaf410369ac4f45276209dc18853ebe31

Next error I got (which might be specific to our HSM):
```
Error: PluginManager initialization failed

Caused by:
    0: Initialize 'pkcs11' plugin failed
    1: unable to generate RSA wrap key pair
    2: Function::GenerateKeyPair: PKCS11 error: The template specified for creating an object is incomplete, and lacks some necessary attributes.  See Section 4.1 for more information.
```
Adding some of the missing metadata helped me progress to next stage:
https://github.com/nmwael/trustee/commit/6386a0fe025d52d3187212bb0e6d3585ccf9cadf

Im not done and are working further on this, so more things could appear. When done, I can wrap it up in a PR.

fitzthum:
cc @tylerfanelli

tylerfanelli:
@nmwael thanks for reporting this.

I opened #973 to fix the initial compile errors. I used `softhsm2` to test the module with. Can you verify you're running into the same issue with softhsm2? I'll try to re-create as well to narrow down where the issue lies.

nmwael:
@tylerfanelli ,  I will check with softhsm2 and see.  I wonder if it's some kind of indirect requirement as we had todo the same with the Thales HSM Java package (thinking of the meta data)..

nmwael:
continued in #1017

Issue #970, opened 2025-09-25
[pkcs11 plugin] not included in default image


nmwael:
The pkcs11 plugin is not included in the default kbs image:
https://github.com/confidential-containers/trustee/blob/1fdbe2a5e5a2ae80ffc53a07dbf3bcc24476490b/kbs/docker/Dockerfile#L44

There's no switch to enable it like the ALIYUN feature.

I could make a PR if wanted. As I am working on it anyhow.

tylerfanelli:
Are you referring to the `cryptoki` rust dependency? I don't think there's any package dependency needed. If a user intends to use softhsm2, then that would need to be installed, but I don't think we should make the assumption and install it for them.

PR #973, opened 2025-09-25
PKCS11 fixes

tylerfanelli:


kbs/Cargo.toml
```diff
@@ -7,7 +7,7 @@ documentation.workspace = true
 edition.workspace = true
 
 [features]
-default = ["coco-as-builtin", "coco-as-grpc", "intel-trust-authority-as"]
+default = ["coco-as-builtin", "coco-as-grpc", "intel-trust-authority-as", "pkcs11"]
```
Xynnn007: We can include this in the `--features` flag whenever it is needed, but we do not need to mark it as `default`

tylerfanelli: Thanks, I've added it to the `Dockerfile` instead.

Xynnn007 (approved):
LGTM. Thanks @tylerfanelli !

mythi:
#970 and #971 can be closed now?

PR #977, opened 2025-09-29
Add remote verifier for NVIDIA devices

fitzthum:
Allow NVIDIA devices to be verified via [NRAS](https://docs.attestation.nvidia.com/api-docs/nras.html).

Why do we want to use an external service for GPU verification when Trustee can already attest Hopper GPUs locally?

First, when we use NRAS, we don't have to worry about the reference values or revocation. NVIDIA attestation reports are relatively complex and require several reference values to validate. These reference values aren't related to the workload and ideally will not be something that the user/admin has to worry about. NRAS will automatically find the correct reference values and check if they are still valid. It also handles revocation of the attestation report itself, which the local verifier currently does not.

Second, NRAS supports more devices than the local verifier does. This PR only supports Hopper GPUs with the remote verifier, but this is a temporary limitation. NRAS itself can also attest switches and this functionality will be added to Trustee soon. As more devices are added to NRAS, we automatically pickup support for them in Trustee, unlike with the local verifier.

The limitation of NRAS is that users are expected to enter in a [licensing agreement](https://docs.nvidia.com/attestation/technical-docs-nras/latest/nras_license.html) with NVIDIA to use it. As such, the default is the local verifier.

There are a few limitations to this PR that will be fixed in future PRs
- No support for switches.
- The cert chain of the JWKs fetched from NRAS (via HTTPS) is not checked.
- Requests to NRAS are made per device when they could be batched.

I added a unit test that evaluates evidence from an H100 both locally and remotely. It is ignored in the remote case to avoid spamming NRAS, but it does work. This PR also adds some support for allowing verifiers to take configuration files.

### DIFF
--- a/deps/verifier/src/nvidia/README.md
+++ b/deps/verifier/src/nvidia/README.md
@@ -1,0 +1,24 @@
+# Verifying NVIDIA devices with the Trustee Attestation Service
+
+This verifier provides two ways to verify NVIDIA devices.
+The `local` verifier will parse the hardware evidence (SPDM messages)
+and extract the measurements.
+The policy can then compare these measurements with reference values.
+
+The `remote` verifier uses the NVIDIA NRAS service to validate the evidence.
+To use this, the user should first enter into a licensing agreement with NVIDIA.
+The agreement is described [here](https://docs.nvidia.com/attestation/technical-docs-nras/latest/nras_license.html)
+and has provisions for research and development.
+When the `remote` verifier is enabled, NRAS handles evaluating the evidence
+against reference values.
+Rather than providing the raw HW measurements as TCB Claims, the `remote` verifier
+exports claims relating to each step of the verification process.
+The policy checks these claims to make sure that attestation has been completed
+successfully.
+
+The remote verifier can be enabled with the following entry in the attestation service
+config file.
+```
+[verifier_config.nvidia_verifier.verifier]
+    type = "Remote"
+```
--- a/attestation-service/docs/parsed_claims.md
+++ b/attestation-service/docs/parsed_claims.md
@@ -148,7 +148,7 @@ Note: The TD Report and TD Quote are fetched during early boot in this TEE. Kern
 
 ## NVIDIA
 
-Hopper GPU H100
+The local verifier only supports Hopper and returns the following claims.
 
 - `arch`: Device architecture. Only `Hopper` is supported
 - `measurements`: List of measurements and its respective index
@@ -167,6 +167,10 @@ Hopper GPU H100
 - `config.project_sku_mod`: Project SKU mod
 - `config.protected_pcie_status`: Protected PCIe status
 - `config.vbios_version`: Device VBIOS version
+
+The remote verifier exports the claims that come from NRAS, which are listed [here](https://github.com/NVIDIA/nvtrust/blob/main/guest_tools/attestation_troubleshooting_guide.md#version-30).
+Claims version 3 is used. The `x-nvidia-overall-att-result` from the overall claims is included
+along with the full set of detached claims.
 
 ## AMD SEV-SNP
 


attestation-service/src/config.rs
```diff
@@ -23,6 +25,10 @@ pub struct Config {
     /// The Attestation Result Token Broker Config
     #[serde(default)]
     pub attestation_token_broker: AttestationTokenConfig,
+
+    /// Optional configuration for verifier modules
+    #[serde(default)]
+    pub verifier_config: Option<VerifierConfig>,
```
Xynnn007, line 31: Maybe we can move the verifier-specific config looking-up logic to the concrete verifier logic, like via env

https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/cca/mod.rs#L79-L80

And we can show this via README.md of nvidia verifier. This way can help us to avoid a mixing configuration file, part of whose fields are usually enabled by a feature flag.

attestation-service/src/token/ear_default_policy_gpu.rego
```diff
@@ -4,6 +4,51 @@ import rego.v1
...
 
+default executables := 33
+
+default configuration := 36
+
 hardware := 2 if {
 	input.sampledevice.svn in data.reference.device_svn
 }
+
+# GPUs verified by NRAS
```
mythi, line 15: I think we first need to discuss can we align the claims between the two verifier configurations. I raised the same in #890. In fact at the moment (this PR included) we have three sets of nvidia claims for users to learn. 

Perhaps that is not fully possible due to RIM info missing from the local verifier. Is there anything that is common between the sets?

Ideally, the default GPU policy matches with the default verifier config (=local) claims.

deps/verifier/src/nvidia/mod.rs
```diff
@@ -78,8 +103,120 @@ impl NvDeviceReportAndCertClaim {
...
+            )
+        };
+
+        // We could store this in the struct to avoid multiple calls
+        // to the JWKs endpoint, but for now do it here for simplicity.
+        // The struct doesn't live very long anyway.
+        let jwks = NrasJwks::new().await?;
+
+        let response = NrasResponse::from_str(&res.text().await?)?;
+        response.validate(&jwks)?;
```
mythi, line 171: we have `token::jwk::JwkAttestationTokenVerifier` on the KBS side that can do roughly the same. Perhaps see how that can be reused here.

deps/verifier/src/nvidia/nras_jwks.rs
```diff
@@ -0,0 +1,57 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use anyhow::{anyhow, bail, Result};
+use std::collections::HashMap;
+
+/// Accessing NRAS requires entering into a licensing agreement with NVIDIA.
+/// Using Trustee with the NRAS remote verifier assumes that you have done this.
+pub const NRAS_JWKS_URL: &str = "https://nras.attestation.nvidia.com/.well-known/jwks.json";
```
mythi, line 11: if they offered a `/.well-known/openid-configuration` with `jwks_uri` info, it'd be possible to reuse `JwkAttestationTokenVerifier::new()` pulling also.

deps/verifier/src/nvidia/mod.rs
```diff
@@ -78,8 +103,120 @@ impl NvDeviceReportAndCertClaim {
...
+        config: &NvidiaRemoteVerifierConfig,
+    ) -> Result<(TeeEvidenceParsedClaim, String)> {
+        let b64_engine = base64::engine::general_purpose::STANDARD;
+
+        let (arch, tee_class, endpoint) = match device.arch {
+            DeviceArchitecture::Hopper => ("HOPPER", "gpu", "gpu"),
+            _ => todo!(),
+        };
+
+        let evidence_b64 = b64_engine.encode(hex::decode(device.evidence)?);
```
mythi, line 131: oh, so this got changed on NRAS side rather quickly and the hex encoded evidence is not required anymore :/ Perhaps we need to fix the `nvidia-attester` evidence encoding asap. Going to check ITA plans for this.

deps/verifier/src/nvidia/mod.rs
```diff
@@ -78,8 +103,120 @@ impl NvDeviceReportAndCertClaim {
...
+        let request_json = json!({
+            "nonce": hex::encode(expected_nonce_vec),
+            "arch": arch,
+            "evidence_list": [
+                {
+                    "evidence": evidence_b64,
+                    "certificate": device.certificate,
+                }
+            ],
+            "claims_version": "3.0"
```
mythi, line 150: are there any docs we could fill in `docs/parsed_claims.md`

deps/verifier/src/nvidia/nras_jwks.rs
```diff
@@ -0,0 +1,57 @@
...
+
+#[derive(Clone, Debug)]
+pub struct NrasJwks {
```
mythi, line 14: could just use `jwk::JwkSet` and `find(kid)`?

deps/verifier/src/nvidia/mod.rs
```diff
@@ -177,4 +311,59 @@ mod tests {
...
+    async fn test_evaluation(#[case] local_verifier: bool) {
+        let b64_engine = base64::engine::general_purpose::STANDARD;
+
+        let device_uuid: &str = "1111-2222-33333-444444-555555";
+
+        let expected_nonce: &str =
+            "931d8dd0add203ac3d8b4fbde75e115278eefcdceac5b87671a748f32364dfcb";
+        let expected_nonce_vec: Vec<u8> = hex::decode(expected_nonce).unwrap();
+
+        // The report certificate chain is stored in PERM format
```
```suggestion
        // The report certificate chain is stored in PEM format
```

deps/verifier/src/nvidia/mod.rs
```diff
@@ -74,6 +102,124 @@ impl NvDeviceReportAndCertClaim {
...
+        // but for now, check one device at a time.
+        let request_url = format!(
+            "{}/{}",
+            config.verifier_url.clone().unwrap_or(NRAS_URL.to_string()),
+            endpoint
+        );
+
+        let request_json = json!({
+            "nonce": hex::encode(expected_nonce_vec),
+            "arch": arch,
```
mythi: `DeviceArchitecture` implements `Display` so perhaps the following could be an option:

```suggestion
            "arch": format!("{device.arch}").to_uppercase(),
```

attestation-service/src/config.rs
fitzthum: I think it's best to have all configuration in one place. It's more work for users if they have to set some environment variables and create new config files. I was thinking about changing CCA here, but it's out of scope of this PR.

I can add an example config to the docs.

attestation-service/src/token/ear_default_policy_gpu.rego
fitzthum: I would like to head in this direction, but I don't think we can do it quite yet. I'm trying to figure out if we should change approaches for the local verifier in the future (have it be less dependent on reference values), which would change the claims completely.

deps/verifier/src/nvidia/mod.rs
fitzthum: Yeah I was considering this. I think we may want to move Jwk to deps and share the code, but I left that as future work. The implementation here isn't a heavy lift.

deps/verifier/src/nvidia/nras_jwks.rs
fitzthum: yeah that would be nice. I can ask around. I am planning a future PR to check the cert chain one way or another and this can bring some optimzations.

deps/verifier/src/nvidia/nras_jwks.rs
fitzthum: Oh good point. I think I will still want the `NrasJwks` struct to handle some loading things. I am also planning to add some internal logic there to only load once. The dictionary could be replaced by `jwk::JwkSet`. I'm not sure there is a huge benefit to this. Maybe this is future work?

deps/verifier/src/nvidia/mod.rs
mythi: Future work sounds OK. The flows here are the same than what `token::jwk::JwkAttestationTokenVerifier` users currently need so having a unified would make sense. Also any cert check improvements would benefit both.

deps/verifier/src/nvidia/nras_jwks.rs
mythi: The benefit is if we can have all users to use the same code. The needs are the same.

deps/verifier/src/nvidia/README.md
```diff
@@ -0,0 +1,24 @@
...
+and has provisions for research and development.
+When the `remote` verifier is enabled, NRAS handles evaluating the evidence
+against reference values.
+Rather than providing the raw HW measurements as TCB Claims, the `remote` verifier
+exports claims relating to each step of the verification process.
+The policy checks these claims to make sure that attestation has been completed
+successfully.
+
+The remote verifier can be enabled with the following entry in the attestation service
+config file.
```
mythi, line 20: Is there no authentication/API key needed?

fitzthum: Authentication happens by checking the JWTs from NRAS based on the JWKs. At the moment I am not checking the cert chain of the JWKs, which come from an NRAS endpoint. Basically we are relying on HTTPS to ensure that we get the right JWKs. I will add support for checking these against a CA which can be provided via the config file.

mythi (approved):
Looks OK for merging

attestation-service/src/config.rs
Xynnn007: Good.

attestation-service/src/token/ear_default_policy_gpu.rego
```diff
@@ -4,6 +4,53 @@ import rego.v1
...
+	input.nvidia.x_nvidia_gpu_driver_rim_signature_verified
+	input.nvidia.x_nvidia_gpu_driver_rim_version_match
+
+	input.nvidia.x_nvidia_gpu_vbios_rim_fetched
+	input.nvidia.x_nvidia_gpu_vbios_rim_measurements_available
+	input.nvidia.x_nvidia_gpu_vbios_rim_schema_validated
+	input.nvidia.x_nvidia_gpu_vbios_rim_signature_verified
+	input.nvidia.x_nvidia_gpu_vbios_rim_version_match
+
+	input.nvidia.measres
```
cclaudio, line 55: Is this also default for the local verifier?

fitzthum: This policy will also be executed with the local verifier. Any clauses that depend on values set by the remote verifier (i.e. `nvidia.x_nvidia_gpu_vbios_rim_version_match`) will be false when using the local verifier. The local verifier should probably have its own clauses for its own tcb claims. It shouldn't really be a problem if both verifiers happen to set the same tcb claim. I will look into reconciling the claims more in future work.

cclaudio: Sounds good

cclaudio: Sounds good

PR #980, opened 2025-09-30
Introduce keylime TEE attestation service support

tylerfanelli:
TEEs are most likely found in security-sensitive environments where applications are deployed on untrusted systems such as the cloud or edge. In additions of TEE protections, some users in these environments would also like runtime attestation services to ensure that systems running sensitive applications are not compromised at any point; and if they are, that a trusted party can be notified and take action.

[Keylime](https://keylime.dev/) (whitepaper found [here](https://raw.githubusercontent.com/keylime/keylime/master/docs/old/tci-acm.pdf)) is a CNCF project to measure and verify the boot and runtime environments of systems using TPMs and the Linux IMA subsystem.

For general purpose confidential VMs on AMD SEV-SNP (and eventually Intel TDX), it is recommended to use the [Secure VM Service Module (SVSM)](https://github.com/coconut-svsm/svsm/blob/main/README.md) for privileged operations and device emulations to guest operating systems. One crucial service offered by the SVSM is marshaling a virtual TPM (vTPM) that a guest OS can use in a trusted manner without the respective TPM hardware.

However, since the SVSM runs at an even higher privilege than the guest OS, it must also be attested before any meaningful processing can take place in the confidential VM. As such, there has been work done to add attestation to the SVSM boot process in order to establish trust in:
- The SVSM firmware responsible for bootstrapping the rest of the system.
- The vTPM that is provided by the SVSM.

For SVSM attestation, the Trustee KBS is already supported. However, no one solution provides the needed TEE+firmware attestation of Trustee along with the runtime attestation services provided by Keylime. As such, to set up the required services for this combined use-case, one would need to deploy:

```
            ┌───────────┐                                         
  ┌───────┐ │Trustee    │ ┌────────┐      ┌────────┐ ┌─────────┐  
  │Trustee│ │Attestation│ │Trustee │      │Keylime │ │Keylime  │  
│ │KBS    │ │Service    │ │RVPS    │ │  │ │Verifier│ │Registrar│ │
│ └───────┘ └───────────┘ └────────┘ │  │ └────────┘ └─────────┘ │
│                                    │  │                        │
└─────────────────┬──────────────────┘  └────────────┬───────────┘
                  │                                  │            
                                                                  
                 TEE                             TPM+IMA          
                 boot                            runtime          
              attestation                       attestation       
```

Although complex, this is entirely able to be automated in a cloud native environment. However, without a cloud native environment, this offers considerable complexity to deploy for general purpose VMs. Also, for legacy VM applications (i.e. not running other cloud-native software within them), the TEE attestation policies are relatively simple. Thus, there was proposal in Keylime to [add simple TEE verification](https://github.com/keylime/enhancements/blob/master/123_verifier_evidence_types.md) for the sole purpose of establishing a trusted vTPM in confidential environments. With this, legacy VMs could use SVSM for attestation and use their vTPMs in exactly the same way they always have been.

This PR adds a backend attestation service module for the Keylime TEE verifier with the optional `keylime-as` feature. It also creates and signs EAR tokens on behalf of the Keylime Verifier for successfully-attested clients. With this, the TEE boot+runtime attestation scenario is simplified to:

```
  ┌───────┐┌────────┐┌─────────┐  
  │Trustee││Keylime ││Keylime  │  
│ │KBS    ││Verifier││Registrar│ │
│ └───────┘└────────┘└─────────┘ │
│                                │
└────────────────┬───────────────┘
                 │                
                                  
                TEE               
                boot              
                 +                
              runtime             
             attestation              
```

Legacy clients can also run legacy applications and use their vTPM in a trusted manner with SVSM.

### TODO in later patch series

There still requires some proper integration between Keylime TPM and TEE policy support. As such, the EAR token marshaling still could use some improvement to report the specific verifier policies used to attest TEE evidence. Also, the Keylime TEE verifier only supports SEV-SNP at the moment, with eventual plans for TDX (contingent on SVSM support for the latter).

### DIFF
--- a/kbs/config/kbs-config-keylime-as.toml
+++ b/kbs/config/kbs-config-keylime-as.toml
@@ -1,0 +1,20 @@
+[http_server]
+insecure_http = true
+
+[attestation_token]
+insecure_key = true
+
+[attestation_service]
+type = "keylime-tee"
+base_url = "https://0.0.0.0:8881"
+api_version_major = 2
+api_version_minor = 4
+cv_ca_path = "/var/lib/keylime/cv_ca"
+
+[[plugins]]
+name = "resource"
+type = "LocalFs"
+dir_path = "/opt/confidential-containers/kbs/repository"
+
+[admin]
+insecure_api = true


kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,428 @@
...
+        let evidence_str = serde_json::to_string(&evidence_json)
+            .context("unable to convert Keylime TEE evidence to JSON string")?;
+
+        let resp = self
+            .client
+            .post(format!(
+                "{}/v{}.{}/verify/evidence",
+                self.config.base_url, self.config.api_version_major, self.config.api_version_minor
+            ))
+            .header(CONTENT_TYPE, "application.json")
```
Xynnn007: Maybe it should be `application/json`?

Xynnn007 (commented):
Am I right about the dataflow about the design?

1. keylime verified the evidence, and it only returns Ok or Fail
2. If Ok, kbs-keylime will sign a ear token and return

kbs/src/attestation/keylime.rs
tylerfanelli: Yes, thank you. Fixed.

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,428 @@
...
+            RawValue::Bytes(Bytes::from(report.host_data.as_slice())),
+        ));
+
+        return Ok(RawValue::Map(map));
+    }
+
+    Err(anyhow!("unsupported TEE"))
+}
+
+// TODO: Gather policy (used for specified claims) used to attest the evidecne and add it to the
```
fitzthum: nit: `evidecne`

Cargo.toml
```diff
@@ -51,14 +51,15 @@ regorus = { version = "0.2.6", default-features = false, features = [
...
-    "default-tls",
+    "native-tls",
 ] }
 rstest = "0.18.1"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.143"
 serde_qs = "0.13.0"
 serde_with = { version = "3.14.0", features = ["base64", "hex"] }
 serial_test = { version = "3.2.0", features = ["async"] }
+sev = { version = "6.3.1", default-features = false, features = ["openssl", "snp"] }
```
fitzthum: Including this in the KBS seems a little fishy.

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,428 @@
...
+        };
+
+        log::debug!("keylime response: {:#?}", response);
+
+        match response.status {
+            KeylimeAttestationStatus::Success => {
+                if response.results.valid > 0 {
+                    // TEE attestation successful, get claims from response and marshal an EAR
+                    // token from the claims. Respond to client with signed EAR token.
+                    let claims = claims(&data.tee, evidence_to_verify[0].tee_evidence.clone())?;
```
fitzthum: `data` == `evidence_to_verify[0]`, right? Maybe this could be simplified. 

Also it seems a little odd to get the claims from the input. Keylime doesn't return any kind of verified claims? Like I mentioned above, it doesn't seem ideal to have a bunch of Tee specific logic for parsing the claims. Ideally this would already have been done by keylime.

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,428 @@
...
+            return Err(anyhow!("TEE public key must be elliptic curve key"));
+        };
+
+        let evidence_json = json!({
+            "type": data.tee,
+            "data": json!({
+                "attestation_report": evidence_to_verify[0].tee_evidence,
+                "nonce": runtime.nonce,
+                "tee_pubkey_x_b64": x,
+                "tee_pubkey_y_b64": y,
```
fitzthum: So keylime has its own way of packing the nonce and tee pubkey into the report data? Does this match what we do on the attester side?

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,428 @@
...
+    pub build_name: String,
+    pub cv_ca_path: String,
+}
+
+pub struct KeylimeTeeHandler {
+    config: KeylimeTeeConfig,
+    client: Client,
+    cert: X509,
+    priv_key_pem: Vec<u8>,
+    pub_key: Rsa<Public>,
```
fitzthum, line 50: These keys are all used for making the JWT attestation token, right? Do we do anything to validate the response coming from keylime (besides https)? I guess keylime itself doesn't return any kind of JWT.

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,428 @@
...
+        extensions
+            .register("tee-pubkey", 5, RawValueKind::Map)
+            .context("unable to register token TEE public key")?;
+
+        extensions
+            .set_by_name("tee-pubkey", tee_pubkey.into())
+            .context("unable to set token TEE public key")?;
+
+        let token = Ear {
+            profile: "tag:github.com,2023:veraison/ear".to_string(),
```
fitzthum: Is this the right profile?

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,428 @@
...
+    Err(anyhow!("unsupported TEE"))
+}
+
+// TODO: Gather policy (used for specified claims) used to attest the evidecne and add it to the
+// Appraisal. Build the submods and add them to the EAR token.
+fn submods(_claims: RawValue) -> Result<BTreeMap<String, Appraisal>> {
+    let _map: BTreeMap<String, Appraisal> = BTreeMap::new();
+    let _appraisal = Appraisal::new();
+
+    Ok(BTreeMap::from([("".to_string(), Appraisal::new())]))
```
fitzthum: This doesn't seem right.

Aside from the claims getting lost here, I wonder if you have enough information to construct an Appraisal. For instance do you plan on having an AR4SI vector in here?

kbs/src/attestation/keylime.rs
tylerfanelli: I don't think I understand the question. The KBS generates the nonce, but the Keylime Verifier recreates the freshness hash and compares it with report data. This is similar to what is done on the attester side, yes.

kbs/src/attestation/keylime.rs
tylerfanelli: This group of keys serves two purposes:

- The first being mTLS between the KBS/Keylime so they both are confident they are talking to the right counterpart. Since we have mTLS, we don't need to validate the response.

- The second being giving the KBS the ability to sign JWTs on behalf of the Keylime Verifier. As mentioned before, the Keylime Verifier requires mTLS to communicate with. For a CVM scenario with SVSM, it is not wise to give the CVM access to the mTLS certs, so we restrict communication to the KBS, which acts as a intermediary to the Keylime Verifier.

kbs/src/attestation/keylime.rs
tylerfanelli: Good point, I'll try to produce the claims in Keylime (it's really just building a JSON map after all) and can remove the `sev` dependency at that point.

tylerfanelli:
@fitzthum One broad note on the EAR support in this PR.

Keylime is currently integrating EAR tokens within their own infrastructure, and we'll likely use that at some point. However, for our current attestation use-case with SVSM (unlocking encrypted storage and initializing a persistent vTPM), we really don't have a need for an EAR token other than using it to unwrap a LUKS key (via the KBS PKCS11 module). That is, we don't really care about the claims themselves **at this point**, but moreso "did Keylime sign this token, indicating it is ok. There is no third party investigating the claims that Keylime attests to, like seen when sharing EAR tokens.

However, at some point SVSM will likely write the EAR token into the persistent storage as an artifacts, such that the eventual client that is brought up could use it for whatever purpose, so we do need to support this use case at some point. But for now, it is good enough for a bringup of trusted, persistent vTPMs with SVSM.

tylerfanelli:
@fitzthum After reviewing your comments, I opted to just return a simple JWT (i.e. not EAR yet) from the KBS, rather than a half-baked EAR implementation. This is only temporary, as Keylime itself is working out how to address EAR tokens internally at the moment.

I also added claims support returned from the Keylime Verifier, as I agreed that it made more sense to come from there. I'll add EAR support in another series in the future.

Would you mind re-reviewing?

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,301 @@
...
+#[async_trait]
+impl Attest for KeylimeTeeHandler {
+    async fn verify(&self, evidence_to_verify: Vec<IndependentEvidence>) -> anyhow::Result<String> {
+        if evidence_to_verify.len() != 1 {
+            return Err(anyhow!(
+                "only one TEE evidence type currently supported for keylime attestation"
+            ));
+        }
+
+        let data = &evidence_to_verify[0];
```
mythi: perhaps add a check for keylime supported `Tee`s here?

kbs/src/attestation/keylime.rs
fitzthum: In the attester we populate the runtime data not just with the nonce, but with the nonce and tee public key (per the RuntimeData struct). When we validate the report data we want to make sure both of those things match. So the question is how does keylime handle that? You are separating the report data into three fields here. Will keylime recombine them in the right way to calculate the expected report data? Does it only put the nonce in the report data? Does it fail if things don't match? etc

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,301 @@
...
+                .context("unable to deserialize keylime /verify/evidence results")?
+        };
+
+        log::debug!("keylime response: {:#?}", response);
+
+        match response.status {
+            KeylimeAttestationStatus::Success => {
+                if response.results.valid > 0 {
+                    // TEE attestation successful, get claims from response and marshal an EAR
+                    // token from the claims. Respond to client with signed EAR token.
```
fitzthum: Should probably update this comment

kbs/src/attestation/keylime.rs
fitzthum: Ok, I guess it could be more clear which is which.

fitzthum (commented):
Looking pretty good. Should add some docs, but this seems reasonable to me.

tylerfanelli:
> Looking pretty good. Should add some docs, but this seems reasonable to me.

Great. Yes, I'll add some docs showing purpose and how to set up.

kbs/src/attestation/keylime.rs
tylerfanelli: Yes, keylime will combine them in the same way to calculate the expected report data. Both the nonce and TEE pubkey are included. It fails if the expected report data doesn't match the actual.

tylerfanelli:
@fitzthum @Xynnn007 @mythi I've cleaned up the API to open the possibility of other CPU TEE formats to be supported. Would you mind another review?

Adding documentation in the meantime.

Cargo.toml
```diff
@@ -51,7 +51,7 @@ regorus = { version = "0.2.6", default-features = false, features = [
     "std",
 ] }
 reqwest = { version = "0.12", default-features = false, features = [
-    "default-tls",
+    "native-tls",
```
Xynnn007: any reason to change this to native-tls? `default-tls` will use pure rust underlying crypto crates, while native-tls will use openssl

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,432 @@
...
+    InvalidSignature,
+    #[serde(rename = "tee_attestation.vcek_fetch")]
+    SevSnpVcekFetch,
+    #[serde(rename = "tee_attestation.invalid_public_key")]
+    SevSnpInvalidPublicKey,
+}
+
+#[async_trait]
+impl Attest for KeylimeTeeHandler {
+    async fn verify(&self, evidence_to_verify: Vec<IndependentEvidence>) -> anyhow::Result<String> {
```
Xynnn007: `anyhow::Result` has already been imported. We can directly use `Result<String>` here w/o anyhow

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,432 @@
...
+                            .decode(tee_data.req.nonce)
+                            .context("unable to decode runtime nonce from base64")?,
+                        tee_data.pubkey,
+                    )
+                } else {
+                    // TEE attestation failed, return the reasons for failure to the client.
+                    return Err(anyhow!(
+                        "client attestation failed: {:?}",
+                        response.results.failures
+                    ));
```
Xynnn007:
```suggestion
bail!("client attestation failed: {:?}", response.results.failures);
```

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,432 @@
...
+/// Configuration of the Keylime verifier.
+#[derive(Clone, Debug, Derivative, Deserialize, PartialEq, Default)]
+pub struct KeylimeVerifierConfig {
+    /// Base URL of the verifier.
+    pub base_url: String,
+    /// API version (i.e. verifier version {MAJOR}.{MINOR}).
+    pub api_version_major: u8,
+    pub api_version_minor: u8,
+    /// Path of the verifier CA certificates.
+    pub cv_ca_path: String,
```
Xynnn007, line 38: I noticed that the certs to be looked up are `{cv_ca_path}/cacert.crt`, `{cv_ca_path}/server-cert.crt` and some others.  I recommend writing this down in a user guide documentation, including how to organize the configuration file names in this directory and the meaning of each file.

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,432 @@
...
+            fs::read(&path).context(format!(
+                "unable to read Keylime server public key file from {}",
+                path
+            ))?
+        };
+
+        // Build an HTTP client to communicate with the Keylime verifier. Establish an identity
+        // using the verifier's certificates.
+        let client = ClientBuilder::new()
+            .identity(
```
Xynnn007, line 261: Looks like `server_priv_pem` and `server_x509_pem` is used for KBS side to make mTLS channel with keylime verifier. Some questions:
1. This key and cert are not the ones that keylime verifier server uses to launch TLS server, right?
2. Is it better to rename them to `client_priv_pem` and `client_x509_pem`?
3. Logically this requires the keylime verifier side needs to set a trust root that trustes this client x509 cert - or did I miss anything?

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,432 @@
...
+
+impl TryFrom<Vec<IndependentEvidence>> for TeeData {
+    type Error = anyhow::Error;
+
+    fn try_from(evidence_list: Vec<IndependentEvidence>) -> Result<Self, Self::Error> {
+        // Only one set of TEE evidence can be verified by Keylime at the moment.
+        if evidence_list.len() != 1 {
+            return Err(anyhow!(
+                "only one TEE evidence type currently supported for keylime attestation"
+            ));
```
Xynnn007:
```suggestion
bail!("only one TEE evidence type currently supported for keylime attestation");
```

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,432 @@
...
+
+        // Only EC keys are permitted as the TEE public key.
+        let TeePubKey::EC {
+            crv: _,
+            alg: _,
+            x,
+            y,
+        } = tee_pubkey.clone()
+        else {
+            return Err(anyhow!("TEE public key must be elliptic curve key"));
```
Xynnn007:
```suggestion
bail!("TEE public key must be elliptic curve key");
```

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,432 @@
...
+    /// TEE public key y coordinate.
+    #[serde(rename = "tee-pubkey-y-b64")]
+    y: String,
+}
+
+/// Evidence formatting based on TEE architecture.
+#[derive(Debug, Deserialize, Serialize)]
+struct TeeEvidence {
+    /// TEE architecture.
+    tee: Tee,
```
Xynnn007, line 421: One question I'm curious about is that the `Tee` enum definitions are defined in kbs-types. Can Keylime Verifier recognize these enums?

Cargo.toml
tylerfanelli: [Setting the ClientBuilder's identity (for mTLS)](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.identity) requires either `native-tls` or `rustls-tls`. It seems like you're looking for `rustls-tls`? I can switch.

Side note: I'd likely prefer OpenSSL, is there a reason you want to avoid it?

kbs/src/attestation/keylime.rs
tylerfanelli: Yes, currently working on a setup document that will include this.

kbs/src/attestation/keylime.rs
tylerfanelli: Not directly, but the enum is serialized to a string. For example, `Tee::Snp` serializes to `"snp"`. Keylime reads this string.

Cargo.toml
Xynnn007: Yes, rustls. 

We usually want to build static compilation to ensure that, for example, a binary can be executed on all x86 platforms without having to worry about the OS. Static compilation of openssl is not simple. Although KBS and AS have not yet achieved full static compilation, it is still generally preferred to avoid openssl under the same circumstances.

kbs/src/attestation/keylime.rs
Xynnn007: Good. I think @mythi raise a good question [idea](https://github.com/confidential-containers/trustee/pull/980#discussion_r2397303506) about whether we need to add logic to filter the serializations that can be recognized by keylime. wdyt?

tylerfanelli: Yes, I've done this implicitly by [deserializing the client request to `TeeArchitectureEvidence`](https://github.com/confidential-containers/trustee/pull/980/commits/9469a188777f63ce52966c5891a1a29a8bc93074#diff-23959aa73cef42f2a013c4bb1c06ef456f0dcabdca09f62fd8a142747d02ca9fR341-R343). SNP is the [only enum member](https://github.com/confidential-containers/trustee/pull/980/commits/9469a188777f63ce52966c5891a1a29a8bc93074#diff-23959aa73cef42f2a013c4bb1c06ef456f0dcabdca09f62fd8a142747d02ca9fR414-R424) at the moment, so this implicitly ensures that all evidence is SEV-SNP evidence.

I guess I could still add the explicit TEE check at the beginning, as there's no error message indicating the TEE is invalid.

Cargo.toml
tylerfanelli: Understood. I've made the change.

kbs/src/attestation/keylime.rs
tylerfanelli: I've added the explicit check [here](https://github.com/confidential-containers/trustee/pull/980/commits/931f3dae85e3b982987a024d256593ce9a74edda#diff-23959aa73cef42f2a013c4bb1c06ef456f0dcabdca09f62fd8a142747d02ca9fR356-R359).

kbs/src/attestation/keylime.rs
tylerfanelli: > Looks like `server_priv_pem` and `server_x509_pem` is used for KBS side to make mTLS channel with keylime verifier. Some questions:
> 
>     1. This key and cert are not the ones that keylime verifier server uses to launch TLS server, right?
> 

Yes, it is. In traditional Keylime setups, the Verifier and Registrar use the server certificates, and are considered all part of the server. We're adding onto that and making the KBS also a part of this setup.

Xynnn007: Okay, this is a bit strange, because we're using the KeylimeVerifier certificate as the client certificate to access KeylimeVerifier. When you mention Verifier using the server certificate, are you referring to the Verifier's externally exposed interface using this certificate?

So, I'm a bit skeptical about whether a client certificate is necessary. In other words, do we need to add a TrustRoot for the target KeylimeVerifier root certificate to the client logic?

tylerfanelli: Sorry, I think I misunderstood. KBS is using the verifier's **client** certificates for mTLS. They are not using the same certificates to communicate with each other.

However, since the KBS signs JWTs on behalf of the verifier, it also requires its public/private keypair to do that.

Xynnn007: What is the verifier's client certificate? Do you mean that by default keylime enables mTLS and connections between KeylimeVerifier and Agent use that?



> However, since the KBS signs JWTs on behalf of the verifier, it also requires its public/private keypair to do that.

Do you mean KBS should use KeylimeVerifier's key pair? I am not an expert on keylime; do you mean that KeylimeVerifier have another key pair except for https key pair?

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,456 @@
...
+            bail!("only one TEE evidence type currently supported for keylime attestation");
+        }
+
+        let data = &evidence_list[0];
+
+        match data.tee {
+            Tee::Snp => (),
+            _ => bail!("invalid TEE"),
+        }
+
```
Xynnn007, line 363:
```suggestion
if data.tee != Tee::Snp {
    bail!("invalid TEE");
}
```

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,456 @@
...
+        };
+
+        // Build a request to specify the type of TEE and its respective evidence for the Keylime
+        // /verify/evidence TEE handler.
+        let req = TeeRequest {
+            tee_evidence: TeeEvidence::from(tee_arch_evidence),
+            nonce: runtime.nonce,
+            x,
+            y,
+        };
```
Xynnn007, line 391: I think this is what the KBS delivers to KeylimeVerifier. Now the SNP report data := hash(runtime-data), and runtime-data := pubkey + nonce

How can the KeylimeVerifier check if the runtime data (thus the nonce) is bound to the given evidence? I see [only nonce](https://github.com/confidential-containers/trustee/pull/980/files#diff-23959aa73cef42f2a013c4bb1c06ef456f0dcabdca09f62fd8a142747d02ca9fR121-R124) is given here but not pubkey. If not, there will be risk of replay attack

tylerfanelli: The `x` and `y` parameters are the coordinates of the EC public key that are hashed into the report. We only support EC keys, so can give those parameters over directly. Those represent the public key.

kbs/src/attestation/keylime.rs
tylerfanelli: While this makes complete sense at the moment, I want to be able to just add to this `match` statement as new TEE architectures become supported, like TDX.

kbs/src/attestation/keylime.rs
Xynnn007: Ok  got it

kbs/src/attestation/keylime.rs
Xynnn007: Ok. I once thought this would be detected by cargo clippy as a lint error but it is not. Btw you might need to add feature `keylime-as` to be covered by  Rust lint checker [here](https://github.com/confidential-containers/trustee/blob/main/.github/workflows/kbs-rust.yml#L33)

kbs/src/attestation/keylime.rs
```diff
@@ -0,0 +1,456 @@
...
+
+            Certificate::from_pem(&x509_pem).context(
+                format!(
+                    "bytes in Keylime verifier CA certificate file ({}) not a valid PEM-encoded X509 certificate",
+                    path
+                )
+            )?
+        };
+
+        // Build the TLS identity from client ceritifcate and private key.
```
berrange, line 202: Trivial typo in "ceritifcate"

PR #982, opened 2025-09-30
protocol: support backwards compatibility for v0.2.0

fitzthum:
Many people have run into problems using the older kbs clients with current versions of Trustee. Although the KBS Protocol is not backwards compatible (prior to v1.0.0), in this case it is possible to support this behavior with a relatively small set of changes. This does not imply that we will offer similar support for any other versions. It is unlikely that we will. In fact, this PR might not even be merged, but if not you can at least apply it yourself. If we merge this PR, we might drop support for v0.2.0 in the future.

The changes here are as follows.

* Update the version check to support multiple versions. This is pretty straightforward.
* Allow v0.4.0 attestation messages to be created from v0.2.0 attestation messages. This is also somewhat clean. See comments for details.
* Tweak the runtime data generation to reconcile it with v0.2.0 clients which do not have composite evidence. This is the ugliest part of the PR imo.

Overall I think this is probably clean enough to support, especially given how many people have run into problems here. That said, it may differ a little bit from how we implement backwards compatibility in the future. For instance we may want to consider adding versioned endpoints.

Some of these changes could also go in kbs-types, but because the KBS protocol itself does not say anything about backwards comaptibility, I consider this a feature of Trustee and have implemented it here. This could be adjusted in the future.

It's also worth thinking about whether this creates any security issues (i.e. downgrading the protocol to skip attesting devices). I don't think it does, but you may want to ponder this independently.

### DIFF
--- a/kbs/src/attestation/backend.rs
+++ b/kbs/src/attestation/backend.rs
@@ -29,21 +29,29 @@ use super::{
     Error, Result,
 };
 
-static KBS_MAJOR_VERSION: u64 = 0;
-static KBS_MINOR_VERSION: u64 = 4;
-static KBS_PATCH_VERSION: u64 = 0;
-
 lazy_static! {
-    static ref VERSION_REQ: VersionReq = {
-        let kbs_version = Version {
-            major: KBS_MAJOR_VERSION,
-            minor: KBS_MINOR_VERSION,
-            patch: KBS_PATCH_VERSION,
+    static ref SUPPORTED_PROTOCOL_VERSIONS: Vec<VersionReq> = {
+        let mut versions = Vec::new();
+
+        let v0_4_0 = Version {
+            major: 0,
+            minor: 4,
+            patch: 0,
             pre: Prerelease::EMPTY,
             build: BuildMetadata::EMPTY,
         };
+        versions.push(VersionReq::parse(&format!("={v0_4_0}")).unwrap());
 
-        VersionReq::parse(&format!("={kbs_version}")).unwrap()
+        let v0_2_0 = Version {
+            major: 0,
+            minor: 2,
+            patch: 0,
+            pre: Prerelease::EMPTY,
+            build: BuildMetadata::EMPTY,
+        };
+        versions.push(VersionReq::parse(&format!("={v0_2_0}")).unwrap());
+
+        versions
     };
 }
 
@@ -214,11 +222,14 @@ impl AttestationService {
         let version = Version::parse(&request.version)
             .inspect_err(|_| AUTH_ERRORS.inc())
             .context("failed to parse KBS version")?;
-        if !VERSION_REQ.matches(&version) {
+        if !SUPPORTED_PROTOCOL_VERSIONS
+            .iter()
+            .any(|req| req.matches(&version))
+        {
             AUTH_ERRORS.inc();
             bail!(
-                "KBS Client Protocol Version Mismatch: expect {} while the request is {}",
-                *VERSION_REQ,
+                "Unsupported KBS Protocol Version: supported versions are {:?}, but got {}",
+                *SUPPORTED_PROTOCOL_VERSIONS,
                 request.version
             );
         }
@@ -263,10 +274,7 @@ impl AttestationService {
 
         let session_id = cookie.value();
 
-        let attestation: Attestation = serde_json::from_slice(attestation)
-            .inspect_err(|_| ATTESTATION_ERRORS.inc())
-            .context("deserialize Attestation")?;
-        let (tee, nonce) = {
+        let (tee, nonce, protocol_version) = {
             let session = self
                 .session_map
                 .sessions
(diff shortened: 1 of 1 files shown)


mkulke (commented):
Can we have tests with 0.2 fixtures?

gcoon151:
I love this and hope we offer this up for very selfish reasons.

fitzthum:
> Probably there is more than one thing that can go wrong, not only the kbs_protocol.
I'm running a peer-pod environment in azure that should be compatible with protocol 0.2.0:

Yes, it is somewhat debatable if this is in scope of the KBS Protocol. Even with this PR there are plenty of ways for things to break. In this case, it would be trivial to allow serde to deserialize this struct from either string. I suggested this on https://github.com/confidential-containers/guest-components/pull/1115 and I would still like to see that implemented, but probably not in this PR.

fitzthum:
That said there might be a deeper incompatibility in the Az evidence. If we fixed the deserialization of the enum I'm not sure if attestation would actually work on Az given some of the crate bumps. cc @mkulke

mkulke:
> @fitzthum Probably there is more than one thing that can go wrong, not only the kbs_protocol. I'm running a peer-pod environment in azure that should be compatible with protocol 0.2.0:
> 
> ```
> [2025-10-06T16:45:12Z ERROR kbs::error] AttestationError(RcarAuthFailed { source: deserialize Request
>     
>     Caused by:
>         unknown variant `azsnpvtpm`, expected one of `az-snp-vtpm`, `az-tdx-vtpm`, `nvidia`, `sev`, `sgx`, `snp`, `tdx`, `cca`, `csv`, `se`, `hygondcu`, `tpm`, `sample`, `sampledevice` at line 1 column 36 })
> [2025-10-06T16:45:12Z INFO  actix_web::middleware::logger] 100.64.0.6 "POST /kbs/v0/auth HTTP/1.1" 401 159 "-" "attestation-agent-kbs-client/0.1.0" 0.000143
> [2025-10-06T16:45:13Z ERROR kbs::error] AttestationError(RcarAuthFailed { source: deserialize Request
> ```

yes, this particular issue might be fixable by silently converting the strings for 0.2. evidence, but Tobin's right: there have been various sev crate bumps in between releases that might had an effect on how evidence are being serialized. Also, at some point the vTPM's EKpub was added as part of the evidence, so we had accommodate that in the verifier.

So far there has been a known-to-work permutation of kata+gc+trustee with each CoCo release. If we now want to bolt on backwards compatibility, that's a bit of a chore, but doable, I think.

In any case, to claim backwards compatibility to work (and keep working in the future) we would need unit tests that use v0.2.0 evidence fixtures, no?

mythi:
> Would it make sense to target for backwards compatibility from 0.4.0 onwards ?

What that means needs to be defined first. A couple of corner cases were already discussed: e.g., if an attester changes the evidence format the verifier fails to understand or if the TEE name gets changed. These are opague to the protocol but users will see errors if Trustee/gc versions do not match. IOW, the compatibility goes beyond just the KBS protocol version.

Agrek11:
Another data point from a real-world adopter — hit the same
  RcarAuthFailed on Trustee `main` with a CAA v0.20.0 AA. The
  guest-components SHA pinned in CAA v0.20.0's versions.yaml emits only
  `{"tee":"snp"}` to /kbs/v0/auth; current Trustee rejects with "missing
  field 'version'", matching fitzthum's PR description. Confirming the
  symptom is still live for the peer-pods → Trustee main pairing 7+
  months into this conversation.

  On @mkulke's v0.2 fixtures ask — I can capture wire-fixture JSON from
  that AA's actual `/kbs/v0/auth` body (real bytes from a running
  SEV-SNP PodVM at the pinned SHA, not hand-crafted). Happy to send as a
  follow-up patch against this branch if that's what unblocks the PR.

  That said — if the direction surfaced by @bpradipt + @mkulke (promote
  0.4.0 → 1.0.0 + formal schema + gc/trustee version matrix in CI) is
  the path forward instead, won't spend cycles on fixtures for a PR
  that gets superseded. Defer to your call on which thread to pull.

mkulke:
I don't think there is strong reason against this, other than it giving a false sense of stability. this PR attempts to provide backwards compatibility for changes in the protocol envelope

the tee-specific payload inside the envelope is not versioned. a major bump of a tee crate might introduce changes to the evidence structure and you'd still run into incompatibility issues. Currently in trustee we maintain an openapi schema, but the evidence is an untyped map:

```
      properties:
        primary_evidence:
          description: Primary TEE evidence. Deserialization dependent on underlying attestation service.
          type: object
```

there's a point about this PR being a net-improvement, but I still think trustee needs to solve this problem comprehensively. currently there is some drift between the implementation, the docs and the openapi schema as schema compliance doesn't seem to be enforced in a test or elsewhere. The untyped object for the evidence should probably have a version sibling, so the service can at least react to some changes. Typed evidence fragments would be even better, but that's a bit of toil.

PR #984, opened 2025-10-01
gha: SEV deprecation

arvindskumar99:


fitzthum (approved):
We've never support SEV in Trustee anyway so fine to remove.

deps/verifier/src/lib.rs
```diff
@@ -47,7 +47,6 @@ pub mod intel_dcap;
 
 pub fn to_verifier(tee: &Tee) -> Result<Box<dyn Verifier + Send + Sync>> {
     match tee {
-        Tee::Sev => todo!(),
```
Xynnn007, line 66: Rust will require all possibilities to be exhausted when matching enum, and sev is defined in kbs-types, so there will be a compilation error here. We need to delete this enum from the upstream, [here](https://github.com/confidential-containers/kbs-types/blob/main/src/lib.rs#L34)

ryansavino (approved):
Need to make the change in kbs-types that @Xynnn007 mentioned. But otherwise LGTM, pending that change. @arvindskumar99 is opening that PR now.

deps/verifier/src/lib.rs
arvindskumar99: @Xynnn007 I've opened a PR in kbs-types with this change: https://github.com/confidential-containers/kbs-types/pull/78

Xynnn007: https://github.com/confidential-containers/kbs-types/pull/78 has been merged.

mythi: @arvindskumar99 please update

arvindskumar99: @Xynnn007 if everything else looks good, can you go ahead and approve the PR? Need one more green check before it can get merged.

Xynnn007: There are still CI errors. Please fix them and force push. Also you need to update the [kbs-types](https://github.com/confidential-containers/trustee/blob/main/Cargo.toml#L43) version the the latest which does not include `sev` enum

PR #989, opened 2025-10-07
Add Caching backend to Trustee and use it for caching AMD VCEKs

fitzthum:
Introduce a Cache backend to Trustee. This is added as a new package in `deps` to avoid circular dependencies (since we use the Cache in the AS and the Verifier). We can use the same cache logic for the KBS and RVPS in the future.

Introduce one simple caching backend which is just a HashMap. This doesn't support any kind of eviction or expiration, but that can be implemented in a more complex caching backend.

The AS config file can be populated with initial cache values that will be added to whatever caching backend is used at startup. This allows cache values (such as VCEKs) to be pre-provisioned in offline environments. cc @bpradipt 

Update the SNP verifier to use the cache when fetching the VCEK. Use the VCEK URL as the key for the cache. I haven't tested this with an SNP system yet. cc @ryansavino

### DIFF
--- a/attestation-service/src/config.rs
+++ b/attestation-service/src/config.rs
@@ -1,6 +1,7 @@
 use crate::rvps::RvpsConfig;
 use crate::token::AttestationTokenConfig;
 
+use cache::CacheConfig;
 use serde::Deserialize;
 use std::fs::File;
 use std::path::{Path, PathBuf};
@@ -23,6 +24,10 @@ pub struct Config {
     /// The Attestation Result Token Broker Config
     #[serde(default)]
     pub attestation_token_broker: AttestationTokenConfig,
+
+    /// Configuration for Attestation Service caching
+    #[serde(default)]
+    pub cache: CacheConfig,
 }
 
 fn default_work_dir() -> PathBuf {
@@ -48,6 +53,7 @@ impl Default for Config {
             work_dir: default_work_dir(),
             rvps_config: RvpsConfig::default(),
             attestation_token_broker: AttestationTokenConfig::default(),
+            cache: CacheConfig::default(),
         }
     }
 }
@@ -86,6 +92,8 @@ mod tests {
         rvps::RvpsConfig,
         token::{ear_broker, simple, AttestationTokenConfig},
     };
+
+    use cache::{CacheConfig, CacheType};
     use reference_value_provider_service::storage::{local_fs, ReferenceValueStorageConfig};
 
     #[rstest]
@@ -100,7 +108,11 @@ mod tests {
             issuer_name: "test".into(),
             signer: None,
             policy_dir: "/var/lib/attestation-service/policies".into(),
-        })
+        }),
+        cache: CacheConfig {
+            r#type: CacheType::Simple,
+            initial_values: None,
+        }
     })]
     #[case("./tests/configs/example2.json", Config {
         work_dir: PathBuf::from("/var/lib/attestation-service/"),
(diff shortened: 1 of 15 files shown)


fitzthum:
This PR triggers `clippy::large-enum-variant`. @confidential-containers/trustee-maintainers can we skip this warning?

attestation-service/src/config.rs
```diff
@@ -1,6 +1,7 @@
 use crate::rvps::RvpsConfig;
 use crate::token::AttestationTokenConfig;
 
+use cache::CacheConfig;
```
bpradipt, line 4: Adding a doc for the caching backend will be helpful, including some guidance on implementing other cache types like local directory with pre-populated certs.

Xynnn007 (commented):
Let's consider a scenario where Intel TDX provides a hardware-endorsed certificate (PCK) caching service implementation (PCCS). Typically, cloud vendors maintain such a service to centrally manage the cloud cache (avoiding the single point of bottleneck caused by requesting Intel's official PCS, similar to AMD KDS). The rationale is that cloud machines frequently access these endorsement keys and certificates, eliminating the need to maintain multiple caching services.

In the AMD scenario (including CSV), while the KDS access API is relatively transparent, I still believe that implementing a CEK key caching service is AMD's responsibility. We can have the SNP verifier connect to a designated caching service; this caching service can be maintained by the user or the cloud.

A future I personally am more optimistic about is if the query interface for hardware endorsement can be unified (potentially with https://www.ietf.org/archive/id/draft-howard-rats-coserv-04.html and it will take a very long time and might fail), we can have a unified endorsement cache service implementation, but this implementation does not have to be included in KBS. We can open a new project, such as in CoCo or anywhere else to maintain it. The purpose of this implementation is to unify different architectures and vendors (PCCS, KDS cache, etc.)

mkulke (commented):
there is also https://crates.io/crates/http-cache-reqwest which is a bit less manual. It uses different heuristics, but it worked for me on KDS

Xynnn007:
Ok, if you want to pre-provision keys, both by a new cache crate, or reqwest extension we need some operations or a script to do this. (local cache we need to download ceks, run command to insert them to the storage; reqwest, we need to run command to pre-download the ceks, and copy the cache binary to specific path).

When I tried to refactor all storage things in trustee, I met some overlap work as this. There is a comment [here](https://github.com/confidential-containers/trustee/pull/1000#issuecomment-3404380279).

please take a look if we could also use a common db for this?

fitzthum:
Btw a somewhat related issue is whether the verifiers should be stateful at all. Currently we create a new verifier for each request to the AS. This means that today the only way to have any kind of persistent state with the verifier would be through some kind of cache or storage.

fitzthum:
The conflicts in this PR are probably mostly related to some extra fields being added to the config file. Shouldn't be too difficult to resolve. 

There is a bigger question about the relationship between the cache and the KV Storage. They are similar, but a bit different. For instance, we don't really need the cache to be persistent or shared between Trustee instances. On the other hand it would be nice if there were an easy way to pre-provision the cache, which we don't entirely have with storage. So I think they are two different things and I wouldn't mind having both abstractions. We could also have a caching backend that points to KV Storage.

That said, this does all add complexity so maybe it's fine to just use the second option even though it's not as nice for pre-provisioning. I don't have a strong opinion here.

amd-aliem:
Prototyped & tested a solution following the example from @mkulke https://github.com/confidential-containers/trustee/pull/989#issuecomment-3400679521 - seems to work fine, for my setup I added a nw rule to block traffic to the amd kds and saw it fail w/o prototype enabled and succeed with it enabled.

Likely will change the config format to be more specific (so it's not too big of a deal to keep supporting it alongside new better options in the future), but would appreciate some feedback on this doc to see if there are glaring issues w/ design: https://github.com/amd-aliem/trustee/blob/amd-cert-cache/attestation-service/docs/amd-offline-certificate-cache.md

If not I'll open a PR for review new week hopefully

bpradipt:
@fitzthum is this PR relevant any more?

fitzthum:
Superseded.

PR #991, opened 2025-10-08
kbs/pkcs11: minimal changes needed for working on an actual hsm

nmwael:
Tested on Thales Luna HSM in FIPS-140 level 3 mode

### DIFF
--- a/kbs/src/plugins/implementations/pkcs11.rs
+++ b/kbs/src/plugins/implementations/pkcs11.rs
@@ -2,12 +2,21 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+//! This module provides plugin support for the cryptographic backend.
+//!
+//! For more information about PKCS#11 and the methodologies used, see the following
+//! [PKCS#11 Usage Guide] (https://docs.oasis-open.org/pkcs11/pkcs11-ug/v3.2/pkcs11-ug-v3.2.html)
+//! [PKCS#11 Specification v3.0](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/pkcs11-spec-v3.2.html)
+//! [PKCS#11 Base Specification v3.0](https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/pkcs11-base-v3.0-os.html).
 use crate::plugins::resource::{ResourceDesc, StorageBackend};
 use actix_web::http::Method;
 use anyhow::{anyhow, bail, Context, Result};
 use cryptoki::{
     context::{CInitializeArgs, Pkcs11},
-    mechanism::Mechanism,
+    mechanism::{
+        rsa::{PkcsMgfType, PkcsOaepParams, PkcsOaepSource},
+        Mechanism, MechanismType,
+    },
     object::{Attribute, AttributeInfo, AttributeType, KeyType, ObjectClass},
     session::{Session, UserType},
     types::AuthPin,
@@ -195,19 +204,22 @@ impl Pkcs11Backend {
     fn wrap_key_new(session: &mut Session, label: &Uuid) -> Result<()> {
         let public_template = vec![
             Attribute::Token(true),
-            Attribute::Private(false),
+            Attribute::Private(true),
             Attribute::KeyType(KeyType::RSA),
             Attribute::Class(ObjectClass::PUBLIC_KEY),
             Attribute::ModulusBits(4096.into()),
-            Attribute::Label(format!("{}-public", label).into()),
+            Attribute::PublicExponent(vec![0x01, 0x00, 0x01]), // 65537
+            Attribute::Encrypt(true),
+            Attribute::Label(format!("trustee-{}-public", label).into()),
         ];
 
         let private_template = vec![
             Attribute::Token(true),
             Attribute::Private(true),
             Attribute::KeyType(KeyType::RSA),
+            Attribute::Decrypt(true),
             Attribute::Class(ObjectClass::PRIVATE_KEY),
-            Attribute::Label(format!("{}-private", label).into()),
+            Attribute::Label(format!("trustee-{}-private", label).into()),
         ];
 
         let (_, _) = session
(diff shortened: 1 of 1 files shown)


kbs/src/plugins/implementations/pkcs11.rs
```diff
@@ -233,19 +239,26 @@ impl Pkcs11Backend {
...
             .context("unable to find public wrap key in PKCS11 module")?;
 
+
+
         let encrypted = self
             .session
             .lock()
             .await
-            .encrypt(&Mechanism::RsaPkcs, pubkey.remove(0), body)
+// NOTICE deprecated by NIST
```
fitzthum: Comment alignment. Also what exactly is deprecated here?

kbs/src/plugins/implementations/pkcs11.rs
```diff
@@ -233,19 +239,26 @@ impl Pkcs11Backend {
...
+            .encrypt(&Mechanism::RsaPkcsOaep(PkcsOaepParams::new(MechanismType::SHA1,PkcsMgfType::MGF1_SHA1,PkcsOaepSource::empty())), pubkey.remove(0), body)
             .context("unable to encrypt HTTP body with public wrap key")?;
+//        .encrypt(&Mechanism::RsaPkcsOaep(PkcsOaepParams::new(MechanismType::SHA256,PkcsMgfType::MGF1_SHA256,PkcsOaepSource::empty())), pubkey.remove(0), body)
```
fitzthum: Probably just delete this line.

kbs/src/plugins/implementations/pkcs11.rs
```diff
@@ -259,7 +272,9 @@ impl Pkcs11Backend {
             .session
             .lock()
             .await
-            .decrypt(&Mechanism::RsaPkcs, privkey.remove(0), body)
+// NOTICE deprecated by NIST            
+            .decrypt(&Mechanism::RsaPkcsOaep(PkcsOaepParams::new(MechanismType::SHA1,PkcsMgfType::MGF1_SHA1,PkcsOaepSource::empty())), privkey.remove(0), body)
+//            .decrypt(&Mechanism::RsaPkcsOaep(PkcsOaepParams::new(MechanismType::SHA256,PkcsMgfType::MGF1_SHA256,PkcsOaepSource::empty())), privkey.remove(0), body)
```
fitzthum: Same as above.

kbs/src/plugins/implementations/pkcs11.rs
nmwael: AFAIK its the minimal longterm FIPS-140 valid crypt. However softhsm does not support it. So we have to do additional work to allow the user to configure algorithms, which is probably a good idea any how.

kbs/src/plugins/implementations/pkcs11.rs
```diff
@@ -15,6 +15,8 @@ use cryptoki::{
 use derivative::Derivative;
 use serde::Deserialize;
 use std::{path::PathBuf, sync::Arc};
+use cryptoki::mechanism::MechanismType;
+use cryptoki::mechanism::rsa::{PkcsMgfType, PkcsOaepParams, PkcsOaepSource};
```
tylerfanelli: Please move these up with the other `cryptoki` imports.

kbs/src/plugins/implementations/pkcs11.rs
```diff
@@ -195,19 +197,22 @@ impl Pkcs11Backend {
     fn wrap_key_new(session: &mut Session, label: &Uuid) -> Result<()> {
         let public_template = vec![
             Attribute::Token(true),
-            Attribute::Private(false),
+            Attribute::Private(true),
```
tylerfanelli, line 207: Is the public key "private" in the sense that it cannot be read?

nmwael: yes it revolves around hsm access, this touches the subject on a high level : https://docs.oasis-open.org/pkcs11/pkcs11-ug/v3.2/pkcs11-ug-v3.2.html#_Toc195624249 , it is not related to the specific object itself.

I've added documentation links to the source. To help clear it up a bit. Sometimes things can be counter intuitive, even with the documentation it can be a little hard to deduct something because of the complexity.

kbs/src/plugins/implementations/pkcs11.rs
```diff
@@ -233,19 +248,34 @@ impl Pkcs11Backend {
...
+                &Mechanism::RsaPkcsOaep(PkcsOaepParams::new(
+                    MechanismType::SHA1,
+                    PkcsMgfType::MGF1_SHA1,
+                    PkcsOaepSource::empty(),
+                )),
+                pubkey.remove(0),
+                body,
+            )
             .context("unable to encrypt HTTP body with public wrap key")?;
+        //        .encrypt(&Mechanism::RsaPkcsOaep(PkcsOaepParams::new(MechanismType::SHA256,PkcsMgfType::MGF1_SHA256,PkcsOaepSource::empty())), pubkey.remove(0), body)
```
Xynnn007: is this something left useless?

kbs/src/plugins/implementations/pkcs11.rs
```diff
@@ -259,7 +289,17 @@ impl Pkcs11Backend {
...
+            .decrypt(
+                &Mechanism::RsaPkcsOaep(PkcsOaepParams::new(
+                    MechanismType::SHA1,
+                    PkcsMgfType::MGF1_SHA1,
+                    PkcsOaepSource::empty(),
+                )),
+                privkey.remove(0),
+                body,
+            )
+            //            .decrypt(&Mechanism::RsaPkcsOaep(PkcsOaepParams::new(MechanismType::SHA256,PkcsMgfType::MGF1_SHA256,PkcsOaepSource::empty())), privkey.remove(0), body)
```
Xynnn007: is this something left useless?

nmwael: good point, i'll cleanup those and make a RFC issue first..

tylerfanelli (approved):
LGTM, thanks for the great work.

Should we consider switching the algorithm to one not deprecated by NIST?

EDIT: I see this discussion was started on #1017, thanks.

nmwael:
> Oh, and can you squash the commits into one?

@tylerfanelli  redone from clean branch, as squash wasn't possible.

fitzthum:
> Not a blocker for the PR, but I don't see in CI where this ignored test is run.
Does it make sense in this or a separate PR to add a conditional test to CI? Or was the original intent, that this sort of feature is outside of 'official' support?

We should add SoftHSM support to the CI and unskip these tests. I think @nmwael will look into this in future PRs.

PR #992, opened 2025-10-08
Centralize base path variable

spotlesstofu:
During #776 I found the hardcoded `/opt/confidential-containers` paths to come in the way for running Trustee as a non-root user. Changing all the paths at runtime means traversing the whole KbsConfig struct, not optimal.

Here I propose a possible alternative, to set the default base path at runtime with a shared logic.

This approach can look heavy, in that it introduces a new package just to share a function. Over time the shared package can become more useful to share more project-wide config. For example if we decide to split the api server from the kbs package. I'd be willing to populate the shared config package a bit more but right now I don't know what else could be moved.

Please propose alternative approaches.

If we like this, I'll have to fix the tests, and possibly make the tests run in both the root and non-root scenario.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -614,6 +614,7 @@ dependencies = [
  "toml 0.9.7",
  "tonic",
  "tonic-build",
+ "trustee-config",
  "uuid",
  "verifier",
 ]
@@ -3216,6 +3217,7 @@ dependencies = [
  "toml 0.9.7",
  "tonic",
  "tonic-build",
+ "trustee-config",
  "uuid",
  "vaultrs",
 ]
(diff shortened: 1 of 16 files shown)


mythi:
> Maybe we can consider split data and state thing in future?

Sure.

On the PR itself, I feel `/opt/confidential-containers` vs `home_dir()` has limitations but it of course depends on what we want. For example, all container based deployments can have `/opt/confidential-containers` container path writable by a non-root process very easily.

spotlesstofu:
@mythi do you advise to have a backwards compatibility mechanism? That is, if `/opt/confidential-containers` exists and is writable, use it.

mythi:
Proposing we close this. I believe things have changed with the storage rework @Xynnn007

PR #995, opened 2025-10-10
tests: add test for admin endpoints

fitzthum:
Make sure that the admin endpoints cannot be accessed when using the wrong admin key.

This is not totally rigorous (not all endpoints and not all failure modes) but it's a good start.

### DIFF
--- a/integration-tests/tests/admin.rs
+++ b/integration-tests/tests/admin.rs
@@ -1,0 +1,141 @@
+// Copyright (c) 2025 by NVIDIA.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::{bail, Result};
+use log::info;
+use openssl::pkey::PKey;
+use rstest::rstest;
+use serial_test::serial;
+
+extern crate integration_tests;
+use crate::integration_tests::common::{KbsConfigType, PolicyType, TestHarness};
+
+//
+// Set the kbs policy with the a valid admin private key
+// and with the wrong admin private key.
+//
+#[rstest]
+#[case::set_policy_with_valid_key(true)]
+#[case::set_policy_with_invalid_key(false)]
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+#[serial]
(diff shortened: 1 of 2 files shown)


integration-tests/tests/admin.rs
```diff
@@ -0,0 +1,168 @@
...
+        } else {
+            if res
+                .as_ref()
+                .unwrap_err()
+                .to_string()
+                .contains("Admin Token verification failed")
+            {
+                return Ok(());
+            }
+        }
```
Xynnn007:
```suggestion
```suggestion
match res {
            std::result::Result::Ok(_) => bail!("Admin key is invalid, but admin operation was successful"),
            Err(e) if e.to_string().contains("Admin Token verification failed") => return Ok(()),
            _ => {}
        }
```
```

File: integration-tests/tests/admin.rs [hunk 2]
```diff
@@ -0,0 +1,168 @@
...
```
Xynnn007:
```suggestion
match res {
            std::result::Result::Ok(_) => bail!("Admin key is invalid, but admin operation was successful"),
            Err(e) if e.to_string().contains("Admin Token verification failed") => return Ok(()),
            _ => {}
        }
```

integration-tests/tests/admin.rs
```diff
@@ -0,0 +1,168 @@
...
```
Xynnn007:
```suggestion
match res {
            std::result::Result::Ok(_) => bail!("Admin key is invalid, but admin operation was successful"),
            Err(e) if e.to_string().contains("Admin Token verification failed") => return Ok(()),
            _ =>{},
        }
```

integration-tests/tests/admin.rs
```diff
@@ -0,0 +1,168 @@
...
+    res
+}
+
+/*
+    // Set Secret
+    info!("TEST: setting secret");
+    harness
+        .set_secret(SECRET_PATH.to_string(), SECRET_BYTES.as_ref().to_vec())
+        .await?;
+*/
```
Xynnn007: Is this something left?

fitzthum: whoops

PR #996, opened 2025-10-13
AS: fix ear trust vector claim names

### LINKED ISSUE #972: Revisiting underscores/hyphens in EAR vs Rego
in #945 we introduced a change that will convert input keys from hyphens to underscores. while EAR specifies keys with hyphens. 

https://github.com/confidential-containers/trustee/blob/1fdbe2a5e5a2ae80ffc53a07dbf3bcc24476490b/attestation-service/src/token/ear_broker.rs#L267-L271

https://www.ietf.org/archive/id/draft-ietf-rats-ar4si-09.html#name-supportable-trustworthiness-cl

However, afaik Rego is able to operate on data structures that have keys with hyphens in it. it just needs a bracket accessor syntax:

```rego
hello if input["my-message"].foo == "world"
```

(similar to how python and javascript deal with dicts)

IMO it would be more intuitive to use the original hyphenizes keys in policies. It should not be ambigious, because `input.my-message.foo` isn't a valid syntax in a policy.

There is an additional nuisance by lowercasing the keys. Currently we use the `tee` value as key in the claims hierarchy: e.g. `snp.reported_tcb == "1.2.3"` If the tee has hyphens (az-snp-vtpm) then this gets lowercased (`az_snp_vtpm`) which adds friction when writing policies. 

We can avoid that kind of value-as-key nesting and replace it with a discriminated union layout:

```json
{ "tee": "snp", "reported_tcb": "1.2.3" }
```

or 

```json
{ "tee": "snp", "claims": { "reported_tcb": "1.2.3" } }
```

the latter might be easier to express in serde, but I think both are possible (and currently the claims map is untyped anyway, I guess?)

Xynnn007:
AR4SI trust vector names are using `-`. However, rego policy does not support variable names like `file-system`, thus we can only use `file_system` as variable names.

This patch uses a map "result" to use the `file-system` like member names to align with the ar4si standard.

Note that regorus does not support eval rule like
```
data.policy.result["file-system"]
```

thus we only eval rule `data.policy.result` and process it in the code logic.

Fixes #972

### DIFF
--- a/attestation-service/src/token/simple.rs
+++ b/attestation-service/src/token/simple.rs
@@ -224,23 +224,17 @@ impl AttestationTokenBroker for SimpleAttestationTokenBroker {
         let reference_data = serde_json::to_string(&reference_data)?;
         let tcb_claims = serde_json::to_string(&flattened_claims)?;
 
-        let rules = vec!["allow".to_string()];
-
         let mut policies = HashMap::new();
         for policy_id in policy_ids {
             let policy_results = self
                 .policy_engine
-                .evaluate(&reference_data, &tcb_claims, &policy_id, rules.clone())
+                .evaluate(&reference_data, &tcb_claims, &policy_id)
                 .await?;
 
-            // TODO add policy allowlist
-            let Some(result) = policy_results.rules_result.get("allow") else {
-                bail!("Policy results must contain `allow` claim");
-            };
-
-            let result = result
+            let result = policy_results
+                .rules_result
                 .as_bool()
-                .context("value `allow` must be a bool in policy")?;
+                .context("value `result` must be a bool in policy")?;
             if !result {
                 bail!("Reject by policy {policy_id}");
             }
--- a/attestation-service/src/token/simple_default_policy.rego
+++ b/attestation-service/src/token/simple_default_policy.rego
@@ -30,9 +30,9 @@ package policy
 import future.keywords.every
 import future.keywords.if
 
-default allow := false
+default result := false
 
-allow if {
+result if {
 	every k, v in input {
 		# `judge_field`: Traverse each key value pair in the input and make policy judgments on it.
 		#


attestation-service/src/token/ear_broker.rs
```diff
@@ -247,32 +247,32 @@ impl AttestationTokenBroker for EarAttestationTokenBroker {
...
             // The cpu tee class is loaded as the default.
             let policy_id = format!("{}_{}", policy_ids[0], tee_claims.tee_class);
             let policy_results = self
                 .policy_engine
-                .evaluate(&reference_data, &tcb_claims_json, &policy_id, rules)
+                .evaluate(
+                    &reference_data,
+                    &tcb_claims_json,
+                    &policy_id,
+                    vec!["result".to_string()],
```
mkulke: would it make sense to move the result unwrapping into the rego impl? if we'd implement another policy engine, maybe we don't need the nesting.

Xynnn007: Do you mean to extract concrete fields like `file-system` from the result and define them here?

I tried but regorus does not seem to support eval like `data.policy.result["file-system"]` or `data.policy.result["executables"]`. It can only be eval that `data.policy.result`. That means the next level under `result` cannot be got successfully. Not sure if it is intended, thus I bring the extraction logic out here.

cc @anakrish

mkulke: I think it's fine to move the "rules" to the policy, although we don't need the rules parameter in the `evaluate()` fn anymore, if it's just hardcoded to `["result"]`, no?

similarly, the `EvaluationResult` type would always evaluate to `Value`

anakrish: @Xynnn007,
[Engine::eval_rule](https://docs.rs/regorus/latest/regorus/struct.Engine.html#method.eval_rule) requires the exact path of a rule. The result if successful will be a regorus::Value.

[Engine::eval_query](https://docs.rs/regorus/latest/regorus/struct.Engine.html#method.eval_query) can take any query expression. The result if successful will be QueryResult.

The former knows that it is evaluating a rule and thus can do bit more efficient evaluation by avoiding query processing.

Note that we are working on a Rego Virtual Machine and compiler where regorus can compile Rego into optimized instructions. This will blur the performance difference between a query vs rule - the query can be precompiled and likely be evaluated as efficiently as a rule.
Here is an early prototype playground for Rego Virtual Machine: https://anakrish.github.io/rego-virtual-machine-playground/

Xynnn007: Thanks. I tried in the playground with rego
```rego
package basic.conditions

import rego.v1

default result := {}

result := {
    "is_admin": is_admin,
    "is_active": is_active,
    "access_level": access_level
}

is_admin if input.user.role == "admin"
is_active if input.user.status == "active"

access_level := "high" if is_admin
access_level := "medium" if {
    not is_admin
    is_active
}
access_level := "low" if {
    not is_admin
    not is_active
}
```

and Entry Point
```
data.basic.conditions.result.is_admin
```

it fails with error `not a valid rule path`. So it treats `basic.conditions.result` as rule path, and you suggested to use `eval_query`?

fitzthum: Hey @anakrish feel free to say that this is a terrible idea, but what about adding an optional Rust feature to regorus to allow `-`es. Is there some fundamental reason why OPA doesn't do this?

anakrish: Hey @fitzthum, 

Supporting `-` in identifiers (rule, var names etc) would likely be a localzied parser change. However, doing so would break compatibility with OPA which would still treat - as a substraction/set-diff expression. Ideally, we want policies to be written in a way that all the existing tools in the rego ecosystem will be able to process it. Regal (the Rego linter) will likely fail with - in identifiers. So I'm not inclined towards adding support for it in regorus.

Hey @anderseknert, have you considered supporting `-` within identifiers in the rego language.

anakrish: @Xynnn007 

With Regorus, the recommendation is to use eval_rule whereever possible. If needed, sub fields of the rule's value can be extracted from the returned regorus::Value.

eval_query is best used for complex queries. E.g `t := data.package.rule[_]; t > 5`

anderseknert: Hey, and thanks for the ping, @anakrish! Always happy to help, but I feel like there's some context lost on me here 😅 The example policy @Xynnn007 posted didn't use hyphens in identifiers anywhere, so I didn't quite follow how that was related, or why their query would fail with that `not a valid rule path` message? 

Anyway, OPA allows string terms to contain `-` (or any character really) as long as properly quoted, so `data.policy.result["file-system"]` is certainly a valid pointer to a document. In fact, all of the Regal linter rules follow a naming convention like that one (see e.g. [annotation-without-metadata](https://github.com/open-policy-agent/regal/blob/main/bundle/regal/rules/bugs/annotation-without-metadata/annotation_without_metadata.rego)).  Naming of _var terms_ is more limited though, so while `foo["bar-baz"]` is a valid rule head ref, you can't do:

```rego
package p

["foo-bar"].baz := ...
```

This is mostly for historical reasons I believe, as in the past a rule head ref was always only a var, while now full refs are allowed. With that in mind, a "rule path" isn't really a concept in OPA's [document model](https://www.openpolicyagent.org/docs/philosophy#the-opa-document-model), which describes only _documents_. If we limit the scope to virtual documents for now (i.e. those produced by evaluating Rego) the two examples below build the same document, only in different ways:

```rego
package a.b

c := 1
```
```rego
package a

b.c := 1
```

Would `c` be the "rule" in both examples? Or would `c` be the rule in the first example, while `b.c` is the rule in the 2nd example? And what is then `b` in the examples? This is not made any easier to answer when variables also contribute to a document, like:

```rego
package a

b.c[d].e := f if {
    some d in ...
    # ...
}
```

That's not to say it's pointless to talk about rules! Just that the separation of packages and rules that might make sense when you author a policy doesn't make as much sense when we talk about queries, as they're rather unaware of what they point to :)

anakrish: Sorry for getting back to this a bit late.

@anderseknert Thanks for the detailed response.

When I read @fitzthum below comment:

>Hey @anakrish feel free to say that this is a terrible idea, but what about adding an optional Rust feature to regorus to allow -es. Is there some fundamental reason why OPA doesn't do this?

I understood it to be want variables with "-" in their names.
E.g:
```rego
   file-system-root := "/"
```
@fitzthum, Can you confirm what the ask is?

anderseknert: All good! Personally I wouldn’t mind more freedom in naming, but I doubt it’s something we can easily change at this point. At least not without massive effort for relatively little gain. So… it’s what it is :)

PR #997, opened 2025-10-13
Feat/abondon simple token

### LINKED ISSUE #940: [AS] Get rid of Simple Token
For some time, we maintained both the simple and EAR token formats. The simple format was not removed because downstream users (primarily maintained by us Alibaba Cloud) were still using it. After consultation, our downstream maintainers are planning to switch to the EAR format. This issue proposes that CoCo upstream remove support for the simple token and gathers feedback.

Xynnn007:
Close #940.

This patch gets rid of simple token and related docs, tests.

### DIFF
--- a/attestation-service/src/ear_token/ear_default_policy_gpu.rego
+++ b/attestation-service/src/ear_token/ear_default_policy_gpu.rego
@@ -8,8 +8,29 @@ default executables := 33
 
 default configuration := 36
 
+default file_system := 0
+
+default instance_identity := 0
+
+default runtime_opaque := 0
+
+default storage_opaque := 0
+
+default sourced_data := 0
+
 hardware := 2 if {
 	input.sampledevice.svn in data.reference.device_svn
+}
+
+result := {
+	"executables": executables,
+	"hardware": hardware,
+	"configuration": configuration,
+	"file-system": file_system,
+	"instance-identity": instance_identity,
+	"runtime-opaque": runtime_opaque,
+	"storage-opaque": storage_opaque,
+	"sourced-data": sourced_data,
 }
 
 # GPUs verified by NRAS


attestation-service/README.md
```diff
@@ -148,7 +148,7 @@ The results of every policy that is evaluated are included in the attestation to
 
 **Note**: Please refer to the [Policy Language](https://www.openpolicyagent.org/docs/latest/policy-language/) documentation for more information about Rego.
 
-If the policy is not updated, the AS will use the [default policy](src/token/ear_default_policy_cpu.rego).
+If the policy is not updated, the AS will use the [default policy](src/ear_token/ear_default_policy_cpu.rego).
```
bpradipt, line 128: @Xynnn007 any possibility to add a note here that stating that Simple token is  removed and EAR token is the only available one. And may be some additional text on what the user should do to move away from simple token.

Xynnn007: Yes. Added

attestation-service/README.md
```diff
@@ -144,11 +121,18 @@ Supported Verifier Drivers:
...
-The results of every policy that is evaluated are included in the attestation token.
+The results of the policy evaluation will be reflected in the token.
 
 **Note**: Please refer to the [Policy Language](https://www.openpolicyagent.org/docs/latest/policy-language/) documentation for more information about Rego.
 
-If the policy is not updated, the AS will use the [default policy](src/token/ear_default_policy_cpu.rego).
+If the policy is not updated, the AS will use the [default policy](src/ear_token/ear_default_policy_cpu.rego).
+
+> [!NOTE]
+> Now only EAR token is supported, and Simple token is deprecated.
```
bpradipt:
```Suggestion
> Only EAR token is supported, and Simple token is removed.
```

attestation-service/README.md
```diff
@@ -144,11 +121,18 @@ Supported Verifier Drivers:
...
+> If you want to switch to EAR token, you need to change the `attestation_token_broker`
```
bpradipt:
```Suggestion
> If you were using Simple token earlier, you will need to change the `attestation_token_broker`
```

PR #999, opened 2025-10-14
CI fixes for Trustee CLI

spotlesstofu:
Follow-up to #776.

- Rename package to trustee-cli for clarity.
- Re-init Makefile and Dockerfile from KBS. Fixes missing dependencies and arm64 build variables.
- Add new test - Rust Suites.

### DIFF
--- a/tools/trustee-cli/src/run.rs
+++ b/tools/trustee-cli/src/run.rs
@@ -53,30 +53,33 @@ fn get_config(
     });
 
     // Set home dir for CoCoASBuiltIn attestation service
-    match &mut config.attestation_service.attestation_service {
-        CoCoASBuiltIn(as_config) => {
-            as_config.work_dir = replace_base_dir(as_config.work_dir.as_path(), trustee_home_dir);
+    if let CoCoASBuiltIn(as_config) = &mut config.attestation_service.attestation_service {
+        as_config.work_dir = replace_base_dir(as_config.work_dir.as_path(), trustee_home_dir);
 
-            // Handle RVPS config paths
-            if let attestation_service::rvps::RvpsConfig::BuiltIn(rvps_config) =
-                &mut as_config.rvps_config
+        // Handle RVPS config paths
+        if let attestation_service::rvps::RvpsConfig::BuiltIn(rvps_config) =
+            &mut as_config.rvps_config
+        {
+            if let reference_value_provider_service::storage::ReferenceValueStorageConfig::LocalFs(
+                local_fs_config,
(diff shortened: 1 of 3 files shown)


tools/trustee-cli/Cargo.toml
```diff
@@ -1,11 +1,15 @@
...
-name = "trustee"
+name = "trustee-cli"
 version.workspace = true
 authors.workspace = true
 description.workspace = true
 documentation.workspace = true
 edition.workspace = true
 
+[[bin]]
+name = "trustee"
```
fitzthum, line 10: I think there's an open question of what we want this to be called, but we can leave that for later.

tools/trustee-cli/src/run.rs
```diff
@@ -53,30 +53,33 @@ fn get_config(
...
+                    replace_base_dir(Path::new(&local_fs_config.file_path), trustee_home_dir)
                         .to_string_lossy()
                         .into_owned();
             }
         }
-        _ => {}
+
+        // Handle attestation token broker paths
+        if let attestation_service::token::AttestationTokenConfig::Ear(ear_config) =
+            &mut as_config.attestation_token_broker
```
fitzthum, line 76: This is about to be removed in https://github.com/confidential-containers/trustee/pull/997

.github/workflows/trustee-cli-rust.yml
```diff
@@ -0,0 +1,84 @@
...
+        components: rustfmt, clippy
+
+    - name: Building dependencies installation
+      run: |
+        sudo apt-get update && sudo apt-get install -y protobuf-compiler libprotobuf-dev libtss2-dev
+
+    - name: Install TPM dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y libtss2-dev
```
fitzthum, line 60: There is also an open question about how we want to package the cli. Ideally it should be easy for people to use and install, but we do have a number of dependencies here.

tools/trustee-cli/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    curl \
+    gpg \
+    gnupg-agent \
+    git \
+    sudo
```
Xynnn007: We might not need this?

Xynnn007:
> Put a fix #1008. Let's see if it works


Fixed. https://github.com/confidential-containers/trustee/actions/runs/18530624036

PR #1000, opened 2025-10-14
Add `deps/key-value-storage` crate

Xynnn007:
There will be a series of PRs, mainly to:
1. adding a new crate for key-value-storage abstraction used in AS/KBS/RVPS.
2. adding a new policy-engine crate to share common code in AS/KBS.
3. Let AS/KBS/RVPS use the new crates and update related tests/docs/deployments

This patch will introduce a key-value storage, with a hashmap in memory implementation and a postgres DB implementation, which will facilitate deployment in a Kubernetes cluster and allow admin users to write policies.

### DIFF
--- a/.github/workflows/as-rust.yml
+++ b/.github/workflows/as-rust.yml
@@ -7,6 +7,7 @@ on:
       - 'attestation-service/**'
       - 'rvps'
       - 'deps/verifier'
+      - 'deps/key-value-storage'
       - '.github/workflows/as_rust.yml'
       - 'Cargo.toml'
   pull_request:
@@ -14,6 +15,7 @@ on:
       - 'attestation-service/**'
       - 'rvps'
       - 'deps/verifier'
+      - 'deps/key-value-storage'
       - '.github/workflows/as_rust.yml'
       - 'Cargo.toml'
   create:
@@ -71,12 +73,12 @@ jobs:
 
       - name: Run cargo test
         run: |
-          sudo -E PATH="$PATH" -s cargo test -p attestation-service -p reference-value-provider-service -p verifier -p eventlog
+          sudo -E PATH="$PATH" -s cargo test -p attestation-service -p reference-value-provider-service -p verifier -p eventlog -p key-value-storage
 
       - name: Run cargo fmt check
         run: |
-          sudo -E PATH="$PATH" -s cargo fmt -p attestation-service -p reference-value-provider-service -p eventlog -p verifier --check
+          sudo -E PATH="$PATH" -s cargo fmt -p attestation-service -p reference-value-provider-service -p eventlog -p verifier -p key-value-storage --check
 
       - name: Run rust lint check
         run: |
-          sudo -E PATH="$PATH" -s cargo clippy -p attestation-service -p reference-value-provider-service -p eventlog -p verifier -- -D warnings -A clippy::derive_partial_eq_without_eq
+          sudo -E PATH="$PATH" -s cargo clippy -p attestation-service -p reference-value-provider-service -p eventlog -p verifier -p key-value-storage -- -D warnings -A clippy::derive_partial_eq_without_eq


deps/key-value-storage/src/postgres/README.md
```diff
@@ -0,0 +1,48 @@
...
+- Stores policies in the `policy` table (`policy_id` primary key, `content` as base64-url encoded text)
+- Safe parameter binding via `sqlx`
+- Input validation for `policy_id` (ASCII alphanumeric, `-`, `_`)
+- Digest calculation for listing policies using SHA-384 (returned as hex string)
+
+### Quick start (with Docker)
+Use the provided helper scripts to spin up a local PostgreSQL for development:
+
+```bash
+bash set-up.sh
```
fitzthum, line 16: Can we add this to the docker-compose instead?

deps/key-value-storage/src/postgres/README.md
```diff
@@ -0,0 +1,48 @@
...
+Teardown:
+```bash
+docker stop policy-postgres && docker rm policy-postgres
+```
+
+### Configuration
+`PostgresClient::new` accepts a `Config` with the following fields and defaults:
+- `db` (default: `policy`)
+- `username` (default: `postgres`)
+- `password` (optional)
```
fitzthum, line 35: Thinking ahead to running Trustee in CoCo, we'll want to provision these credentials via attestation. It might be easier to do that if we also had an option to set a path to the credentials so that we can put them in a sealed secret.  wdyt?

deps/policy-engine/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+        policy_id: &str,
+        evaluation_rules: Vec<String>,
+    ) -> Result<EvaluationResult> {
+        let policy = self.get_policy(policy_id).await?;
+        self.engine
+            .evaluate(data, input, policy_id, &policy, evaluation_rules)
+            .await
+    }
+
+    pub async fn set_policy(&self, policy_id: &str, policy: &str, overwrite: bool) -> Result<()> {
```
fitzthum: Today the KBS only supports one policy at a time. Probably in the future we should change this, but for this PR do you plan on keeping that restriction or not?

Also, if we are using one policy backend for both the AS and the KBS, how do we prevent collisions? Will the policy ids have some kind of prefix?

deps/policy-engine/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+pub mod engine;
+pub use engine::*;
+
+pub mod error;
+pub use error::*;
+use serde::Deserialize;
+
+#[derive(Debug, Clone, Deserialize, Default, PartialEq)]
+pub struct PolicyEngineConfig {
+    pub backend: PolicyStorageConfig,
```
fitzthum: It might be more clear to call this storage rather than backend.

deps/key-value-storage/src/postgres/README.md
Xynnn007: Yes, docker-compose and kubernetes settings are left as TODO items as this PR is still under development.

deps/key-value-storage/src/postgres/README.md
Xynnn007: As the credentials used by KBS (like ITA credential) are all included inside the config file, it might be practical to mark the whole KBS config file as the credential?

deps/policy-engine/src/lib.rs
Xynnn007: Good question. The KBS policy number is somehow vertical to this PR, we could resolve it later?

> how do we prevent collisions?

Nice question. For postgres, the configuration can specify different tables. The final config file would have multiple backend/storage sections in different parts (AS policy storage/RVPS RV storage/KBS policy storage)

Xynnn007:
Hi folks, thanks for showing interests for this work.

During the implementation process, I found that there is some overlap with what @fitzthum  is doing at https://github.com/confidential-containers/trustee/pull/989/files, concretely about the implementation of a key-value similar interface to work as storage, so some ideas

1. For 989 the aim is to resolve the cache problem of SNP. Please @fitzthum to confirm, whether to use the cache mechanism of reqwest itself in that PR, or still hope to use an additional KV storage to let the trustee solve it.
2. Nice suggestion from @mkulke and I also had realized this so mark this as draft :P. Let's decouple the storage into a separate crate. But let's wait 1 to be determined and then we can decide who implements the storage crate.

Xynnn007:
OK I think this PR gets ready for review. Although the PR size is big, I believe different commits have relative well-decoupled structure thus code review can be straightforward. Please share any comments if the code is still hard to review/carry on.

Xynnn007:
btw the RVPS change should be based on https://github.com/confidential-containers/trustee/pull/887 which leverages `set` and `get` api than `get_values`

deps/key-value-storage/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+#[cfg(feature = "postgres")]
+pub mod postgres;
+
+#[async_trait]
+pub trait KeyValueStorage: Send + Sync {
+    async fn set_key(&self, key: String, value: String, overwrite: bool) -> Result<()>;
+
+    async fn list_keys(&self) -> Result<Vec<String>>;
+
+    async fn get_key(&self, key: String) -> Result<String>;
```
mkulke: _key is maybe redundant? just `set`, `get`, `list`?

deps/key-value-storage/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+pub use error::{KeyValueStorageError, Result};
+
+pub mod simple;
+
+#[cfg(feature = "postgres")]
+pub mod postgres;
+
+#[async_trait]
+pub trait KeyValueStorage: Send + Sync {
+    async fn set_key(&self, key: String, value: String, overwrite: bool) -> Result<()>;
```
mkulke: maybe add a `ttl: Option<std::time::Duration>` param?

deps/key-value-storage/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+pub mod error;
+pub use error::{KeyValueStorageError, Result};
+
+pub mod simple;
+
+#[cfg(feature = "postgres")]
+pub mod postgres;
+
+#[async_trait]
+pub trait KeyValueStorage: Send + Sync {
```
mkulke, line 27: should we add `delete()` to the trait?

deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+use sqlx::{postgres::PgPoolOptions, query, Row};
+use tracing::{debug, info, instrument};
+
+use crate::{KeyValueStorage, KeyValueStorageError, Result};
+
+#[derive(Deserialize, Derivative, Clone, PartialEq)]
+#[derivative(Debug)]
+pub struct Config {
+    /// The name of the PostgreSQL database.
+    #[serde(default = "default_db")]
```
mkulke: we can implement `Default` for Config, and then the set `#[serde(default)]` on the whole struct, we don't need all the default_ fn'.

deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    #[serde(default = "default_username")]
+    pub username: String,
+
+    /// The password of the PostgreSQL database.
+    #[derivative(Debug = "ignore")]
+    #[serde(default)]
+    pub password: Option<String>,
+
+    /// The port of the PostgreSQL database.
+    #[serde(default = "default_port")]
```
mkulke: should this be a u16?

deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+            .map(|password| format!("{}:{password}@", config.username))
+            .unwrap_or(format!("{}@", config.username));
+
+        let url = format!(
+            "postgres://{}{}:{}/{}",
+            auth_part, config.host, config.port, config.db
+        );
+
+        let pool = PgPoolOptions::new()
+            .max_connections(5)
```
mkulke: make this a const maybe?

deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    /// Check if the key is valid.
+    ///
+    /// The key is valid if it only contains ASCII alphanumeric characters, `-` or `_`.
+    /// No spaces and other special characters are allowed to prevent SQL injection.
+    fn is_valid_key(key: &str) -> bool {
+        key.chars()
+            .all(|c| c.is_ascii_alphanumeric() || c == '-' || c == '_')
+    }
+}
```
mkulke: maybe move this to a static fn on the trait? all storage impls should share the same key requirements, no?

deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    #[instrument(skip_all, name = "PostgresClient::set_key")]
+    async fn set_key(&self, key: String, value: String, overwrite: bool) -> Result<()> {
+        if !Self::is_valid_key(&key) {
+            return Err(KeyValueStorageError::SetKeyFailed {
+                source: anyhow::anyhow!("key contains invalid characters"),
+                key,
+            });
+        }
+
+        // we do the base64 encoding here to avoid SQL injection.
```
mkulke: I don't need think we need to convert the value payload to base64 to avoid injections, since the sqlx queries are parameterized statements `value` is bound. Maybe the column-type for value should be a byte array? (`bytea` in postgres schema)

deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+                key,
+            })?;
+
+        Ok(())
+    }
+
+    #[instrument(skip_all, name = "PostgresClient::list_keys")]
+    async fn list_keys(&self) -> Result<Vec<String>> {
+        let sql = format!("SELECT (key) FROM {}", self.table);
+        let rec = query(&sql)
```
mkulke:
```rust
    let keys= sqlx::query_scalar::<_, String>(&sql)
        .fetch_all(&*self.pool)
        .await
        .map_err(|e| KeyValueStorageError::ListKeysFailed { source: e.into() })?;

    Ok(keys)
```

deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        let b64_value: String = value.get("value");
+        let policy = base64::engine::general_purpose::URL_SAFE_NO_PAD
+            .decode(b64_value)
+            .map_err(|_| KeyValueStorageError::MalformedValue {
+                source: anyhow!("failed to decode value with base64 URL_SAFE_NO_PAD"),
+            })?;
+        let value =
+            String::from_utf8(policy).map_err(|_| KeyValueStorageError::MalformedValue {
+                source: anyhow!("The value is not a valid UTF-8 string"),
+            })?;
```
mkulke: why should the value be a valid utf-8 string? it could be some binary key material, no?

deps/key-value-storage/src/postgres/README.md
```diff
@@ -0,0 +1,52 @@
+# PostgreSQL backend for key-value storage
+
+This module provides a PostgreSQL-backed implementation of `KeyValueStorage` for the key-value storage system. Key-value pairs are stored in a single table with the value content base64-url encoded to prevent SQL injection and ensure safe storage of arbitrary text.
+
+## Features
+
+- Stores key-value pairs in a configurable table (`key` primary key, `value` as base64-url encoded text)
```
mkulke: I would reconsider the base64 url requirement, see above. (and, if we want to keep the b64 encoding we want to document that specific variant of base64-encoding that we use, so external tooling can work with that requirement)

deps/key-value-storage/src/postgres/README.md
```diff
@@ -0,0 +1,52 @@
...
+`PostgresClient::new` accepts a `Config` with the following fields and defaults:
+
+- `db` (default: `postgres`)
+- `username` (default: `postgres`)
+- `password` (optional)
+- `port` (default: `5432`)
+- `host` (default: `localhost`)
+- `table` (default: `key_value`)
+
+Connection string format constructed internally:
```
mkulke, line 40: it would be nice if we would support `POSTGRES_URL` envs? that is sort of a common conventions and there are [crates](https://docs.rs/pg-connection-string/latest/pg_connection_string/) that verify/parse them.

deps/key-value-storage/src/lib.rs
Xynnn007: nice idea. Let me do the change

deps/key-value-storage/src/lib.rs
Xynnn007: How about add this field to the backend storage config? The caller can always use [tokio::timeout](https://docs.rs/tokio/1.48.0/tokio/time/fn.timeout.html) to do this

deps/key-value-storage/src/lib.rs
Xynnn007: Yes, let me add this

deps/key-value-storage/src/lib.rs
mkulke: oh, this is not a timeout, this is time-to-live (e.g. redis `set EXAT 60 ... `). it's a useful property to have, in postgres this can be either not supported or implemented with an `expired_at` column

Xynnn007: Oh I have misunderstood. I am not sure if in future there will be more config items, so let me add a context config as a parameter here. By default the ttl will be none

deps/key-value-storage/src/postgres/README.md
fitzthum: Yeah that could work.

deps/key-value-storage/src/postgres/mod.rs
Xynnn007: Yes it's u16

deps/key-value-storage/src/postgres/mod.rs
Xynnn007: Good. Originally I thought different storage impl will have different behaviors of whether/how to check the keys, but to have a common check is a good idea to have less imcompability and secure issues.

deps/key-value-storage/src/postgres/mod.rs
Xynnn007: I changed all the value to binary, thus do not need this anymore.

deps/key-value-storage/src/postgres/README.md
Xynnn007: Added

deps/key-value-storage/src/lib.rs
Xynnn007: Leave a TODO for ttl support of postgres in code

deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+pub const VALUE_COLUMN: &str = "value";
+
+/// The environment variable name for the PostgreSQL URL.
+pub const POSTGRES_URL_ENV_VAR: &str = "POSTGRES_URL";
+
+#[derive(Deserialize, Derivative, Clone, PartialEq)]
+#[derivative(Debug)]
+pub struct Config {
+    /// The name of the PostgreSQL database.
+    #[serde(default)]
```
mkulke: nit: you can move this to the struct I think.

deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+            "INSERT INTO {} ({KEY_COLUMN}, {VALUE_COLUMN}) VALUES ( $1, $2 )",
+            self.table
+        );
+
+        if context.overwrite {
+            sql = format!("{sql} ON CONFLICT ({KEY_COLUMN}) DO UPDATE SET {VALUE_COLUMN} = $2");
+        }
+
+        // TODO: add TTL support for PostgreSQL
+        // now we just ignore the TTL
```
mkulke: nit: log warning if it's set and we ignore it

deps/key-value-storage/src/postgres/README.md
```diff
@@ -0,0 +1,54 @@
+# PostgreSQL backend for key-value storage
+
+This module provides a PostgreSQL-backed implementation of `KeyValueStorage` for the key-value storage system. Key-value pairs are stored in a single table with the value content base64-url encoded to prevent SQL injection and ensure safe storage of arbitrary text.
```
mkulke: doc needs adjustment?

deps/key-value-storage/src/postgres/README.md
```diff
@@ -0,0 +1,54 @@
...
+## Quick start (with Docker)
+
+Use the provided helper scripts to spin up a local PostgreSQL for development:
+
+```bash
+bash set-up.sh
+```
+
+What it does:
+- Starts a container named `policy-postgres` on host port `6432`
```
mkulke: since the store is supposed to be used for all sort of state, not just policies, it probably makes sense to just call it postgres here.

deps/key-value-storage/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+pub mod error;
+pub use error::{KeyValueStorageError, Result};
+
+pub mod simple;
+
+#[cfg(feature = "postgres")]
+pub mod postgres;
+
+#[derive(Default)]
+pub struct SetContext {
```
mkulke: I think "Context" could be misleading here. I would suggest "SetParams", "Context" usually provides meta data and hooks that are independent of the business logic of a fn (like correlation_id, UserId)

deps/key-value-storage/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+    async fn get(&self, key: String) -> Result<Option<Vec<u8>>>;
+
+    /// Delete a value for a key.
+    /// Return the deleted value if it exists.
+    async fn delete(&self, key: String) -> Result<Option<Vec<u8>>>;
+
+    /// Check if the key is valid.
+    ///
+    /// The key is valid if it only contains ASCII alphanumeric characters, `-` or `_`.
+    /// No spaces and other special characters are allowed to prevent SQL injection.
```
mkulke: I think it's reasonable to restrict the valid key chars (we want them to be representable as url path segments). However, if we want to use this store for kbs secrets, this restriction will be a breaking change probably, since keys like `luks.key` are allowed, no?

In any case this is not about SQL injection, we don't string-template keys into the SQL statements.

mkulke (commented):
thanks! looks good, a couple of nits remaining

deps/key-value-storage/src/lib.rs
fitzthum: Seems like maybe this should be an internal interface anyway. I wouldn't expect the caller to separately check  keys are valid before inserting them.

deps/key-value-storage/src/postgres/mod.rs
fitzthum: I wonder if we should just leave out this feature entirely for now and add it later if we need it.

Currently I think we handle expiration logic for RVs in the caller, which might be simpler, especially if not all backends will support TTL.

mkulke: hmm, right, it's probably better to remove it if we don't implement it yet.

deps/key-value-storage/src/lib.rs
mkulke: yes, it can move from the trait to the crate root

deps/key-value-storage/src/postgres/mod.rs
Xynnn007: Good. Deleted.

deps/key-value-storage/src/lib.rs
Xynnn007: Nice name suggestion. Renamed to `SetParameters`

deps/key-value-storage/src/lib.rs
Xynnn007: Ok. Let me move it out and add `.` as a valid charactor.

deps/key-value-storage/src/postgres/README.md
Xynnn007: Yes. Updated,

deps/key-value-storage/src/postgres/README.md
Xynnn007: Yes. Fixed

deps/key-value-storage/src/postgres/mod.rs
Xynnn007: Cool feature! I always meet this situation and use field level `#[serde(default)]`. The whole struct level helps a lot. I added a test case.

deps/key-value-storage/src/postgres/mod.rs
Xynnn007: moved.

fitzthum:
Are you planning on updating docker compose to use this?

Xynnn007:
> Are you planning on updating docker compose to use this?

Yes. As this is not used by any crates (AS/KBS) in this PR, I will add it to later PRs. the commit is ready (in previous pushes. but to decouple the huge PR into small pieces I hide them)

fitzthum (approved):
LGTM

PR #1003, opened 2025-10-14
kbs/jwe: Add support for EC P-521 encryption keys

tylerfanelli:
Add support for use EC P-521 keys as the TEE public key.

### DIFF
--- a/kbs/src/jwe.rs
+++ b/kbs/src/jwe.rs
@@ -366,10 +366,11 @@ mod tests {
     };
     use kbs_types::TeePubKey;
     use openssl::rsa::Rsa;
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
good to see this. 

do you have any unit test case for this?

btw, it shares some common code style with p256. In future we could consider do a little refactor but not current work : )

tylerfanelli:
> Hi @tylerfanelli please take a rebase and the latest commit [a4c9d2a](https://github.com/confidential-containers/trustee/commit/a4c9d2a1533242d47c7c19cf12c9bbe57e1951f0) fixes the CI error

Thanks, I've rebased.

PR #1011, opened 2025-10-20
trustee-cli: add subject alternative name

spotlesstofu:
Add a Subject Alternative Name to the self-signed certificate, the kbs-client failed without it.

### DIFF
--- a/tools/trustee-cli/src/run.rs
+++ b/tools/trustee-cli/src/run.rs
@@ -6,6 +6,7 @@ use kbs::plugins::RepositoryConfig::LocalFs;
 use kbs::{ApiServer, KbsConfig};
 use log::{debug, info, warn};
 use openssl::asn1::Asn1Time;
+use openssl::bn::MsbOption;
 use openssl::hash::MessageDigest;
 use openssl::pkey::{PKey, Private};
 use openssl::rsa::Rsa;
@@ -208,6 +209,12 @@ fn build_x509(sockets: &[SocketAddr], private_key: PKey<Private>) -> Result<X509
         san_builder.ip(socket.ip().to_string().as_str());
     }
 
+    let serial = {
+        let mut bn = openssl::bn::BigNum::new()?;
+        bn.rand(128, MsbOption::ONE, false)?;
+        openssl::asn1::Asn1Integer::from_bn(bn.as_ref())?
+    };
+
     let mut builder = X509::builder()?;
     builder.set_version(2)?;
     builder.set_subject_name(name.as_ref())?;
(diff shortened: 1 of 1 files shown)


tools/trustee-cli/src/run.rs
```diff
@@ -217,6 +218,13 @@ fn write_new_https_cert(certificate_path: &Path, private_path: &Path) -> Result<
         builder.set_version(2)?;
         builder.set_subject_name(name.as_ref())?;
         builder.set_issuer_name(name.as_ref())?;
+        let subject_alternative_name = SubjectAlternativeName::new()
+            .dns("localhost")
+            .dns("localhost.localdomain")
+            .ip("127.0.0.1")
```
Xynnn007: Do we need to align this field with the [sockets](https://github.com/confidential-containers/trustee/blob/main/kbs/src/config.rs#L24) from given config?

mkulke: yes, I think so. you want to add the IP you are listening on, no?

"subjectAltName=IP:10.200.24.4"

spotlesstofu: Good point, I added that. This comes with additional caveats, please see the commit message.

tools/trustee-cli/src/run.rs
```diff
@@ -180,56 +194,31 @@ fn get_key_paths(base_dir: &Path, key_name: &str) -> (PathBuf, PathBuf) {
...
+    let mut builder = X509::builder()?;
+    builder.set_version(2)?;
+    builder.set_subject_name(name.as_ref())?;
+    builder.set_issuer_name(name.as_ref())?;
+    builder.append_extension(san_builder.build(&builder.x509v3_context(None, None))?)?;
+    builder.set_not_before(Asn1Time::days_from_now(0)?.as_ref())?;
+    builder.set_not_after(Asn1Time::days_from_now(365)?.as_ref())?;
+    builder.set_pubkey(private_key.as_ref())?;
+    builder.set_serial_number(
+        openssl::asn1::Asn1Integer::from_bn(openssl::bn::BigNum::from_u32(1)?.as_ref())?.as_ref(),
```
Xynnn007: This is just something copy-paste work and I just found it. The hardcode bn number 1 will cause potential collision. If we only use it for test, it's fine. Or, do we need to consider to generate a random bn here?

spotlesstofu: I added that in a new commit.

My initial understanding was that it's OK to have `1` as serial. Most use cases for the serial number are within a CA, not for self-signed certificates. Now, it looks like it's best practice to have it random anyways, especially to avoid some kinds of attacks.

PR #1012, opened 2025-10-20
Improvements to NVIDIA remote verifier

fitzthum:
A few simple improvements to the remote verifier. 

First, use the JwkSet struct from JsonWebToken to simplify loading the JWKs.

Second, store the NrasJwks inside the Nvidia verifier struct so that we don't make multiple requests when there is one guest with several devices. This does not provide any persistence across requests since the struct gets instantiated again, but it makes a single request with multiple devices more efficient.


mythi (approved):
LGTM

PR #1014, opened 2025-10-22
Add Pluggable Admin Backends to KBS

fitzthum:
I haven't fixed the tests or updated the docs yet, but I figured I would drop this draft here in case anyone has feedback on the general idea.

This is the first step in overhauling the the admin interface. Next I plan to add an admin authorization endpoint (for logging in and getting an admin token). Then we can make backends for things like OAuth2. Then we can develop more fine-grained roles.

### DIFF
--- a/integration-tests/src/common.rs
+++ b/integration-tests/src/common.rs
@@ -42,7 +42,7 @@ use tonic::transport::Server;
 
 const KBS_URL: &str = "http://127.0.0.1:8081";
 const RVPS_URL: &str = "http://127.0.0.1:51003";
-const WAIT_TIME: u64 = 3000;
+const WAIT_TIME: u64 = 10000;
 
 const ALLOW_ALL_POLICY: &str = "
     package policy


Xynnn007:
btw do we need to change the topic to admin specific?

kbs/src/admin/mod.rs
```diff
@@ -2,57 +2,86 @@
...
-            warn!("insecure admin APIs are enabled");
-            return Ok(Admin::default());
-        }
+// create a simple backend
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+pub enum AdminBackendType {
+    Simple(SimpleAdminConfig),
+    InsecureAllowAll,
+    DenyAll,
+}
```
Xynnn007, line 33: You can add a serde `tag` here
```suggestion
#[derive(Clone, Debug, Deserialize, PartialEq)]
#[serde(tag = "type")]
pub enum AdminBackendType {
    Simple(SimpleAdminConfig),
    InsecureAllowAll,
    DenyAll,
}
```

With this part (1/2) can have the form of config 

```toml
[admin]
type = "Simple"

[[admin.admin_backend.personas]]
id = "admin1"
public_key_path = "/opt/confidential-containers/trustee/admin1-pubkey.pem"
```

than

```toml
[[admin.admin_backend.Simple.personas]]
id = "admin1"
public_key_path = "/opt/confidential-containers/trustee/admin1-pubkey.pem"
```

Which means that if we choose one Admin type, e.g. `Simple`, all the configurations are implicitly Simple's, thus we do not explicitly show `Simple` in the config path.

kbs/src/admin/mod.rs
```diff
@@ -2,57 +2,86 @@
...
 
-impl Admin {
-    pub(crate) fn validate_auth(&self, request: &HttpRequest) -> Result<()> {
-        let Some(public_key) = &self.public_key else {
-            return Ok(());
-        };
+#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+#[serde(default)]
+pub struct AdminConfig {
+    pub admin_backend: AdminBackendType,
```
Xynnn007, line 47: With this change (2/2)
```suggestion
    #[serde(flatten)]
    pub admin_backend: AdminBackendType,
```

Furthurmore, the config can be generated into 
```toml
[admin]
type = "Simple"

[[admin.personas]]
id = "admin1"
public_key_path = "/opt/confidential-containers/trustee/admin1-pubkey.pem"
```

where wo do not have extra `admin_backend` layer.

kbs/src/admin/mod.rs
```diff
@@ -2,57 +2,86 @@
...
+    pub fn validate_admin_token(&self, request: &HttpRequest) -> Result<()> {
+        let res = self.backend.validate_admin_token(request);
+        if res.is_ok() {
+            info!("Allowing Admin access for {}", request.full_url().as_str());
+        } else {
+            info!(
+                "Not allowing Admin access for {}",
+                request.full_url().as_str()
+            );
+        }
```
Xynnn007: Shall the `res` (type  `Result<T, E>`)'s `E` carry some information that indicates why the admin access is denied? Probably we can print them to the log too

kbs/src/admin/simple.rs
```diff
@@ -0,0 +1,81 @@
...
+    public_key: Ed25519PublicKey,
+}
+
+impl SimpleAdminBackend {
+    pub fn new(config: SimpleAdminConfig) -> Result<Self> {
+        let mut personas = Vec::new();
+
+        for persona_config in &config.personas {
+            let user_public_key_pem =
+                std::fs::read_to_string(persona_config.public_key_path.clone())?;
```
Xynnn007:
```suggestion
                std::fs::read_to_string(&persona_config.public_key_path)?;
```

kbs/src/admin/simple.rs
```diff
@@ -0,0 +1,81 @@
...
+            let public_key = Ed25519PublicKey::from_pem(&user_public_key_pem)?;
+
+            personas.push(SimplePersona {
+                _id: persona_config.id.clone(),
```
Xynnn007:
```suggestion
                _id: persona_config.id,
```

kbs/src/admin/simple.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+#[serde(default)]
+pub struct SimpleAdminConfig {
+    pub personas: Vec<SimplePersonaConfig>,
+}
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+pub struct SimplePersonaConfig {
+    pub id: String,
```
Xynnn007, line 27: What does this id do? I originally thought there will be an id in the request to tell KBS which public key should be used, but there is not.

Xynnn007 (commented):
Thanks for this work! Some first comments

kbs/src/admin/simple.rs
fitzthum: This allows us to report which public key was used. I'll add a log message that prints the ID. In the future we can make the API richer to report who the admin is and what their privileges are.

kbs/src/admin/simple.rs
fitzthum: Doesn't compile

kbs/src/admin/mod.rs
fitzthum: It's the classic question of report errors from a loop. There could be several personas that fail for different reasons. How do we map that to the one error that we return?

kbs/src/admin/mod.rs
fitzthum: Really good suggestions

kbs/src/admin/simple.rs
Xynnn007: A log message sounds good for this PR

kbs/src/admin/mod.rs
Xynnn007: Oh, before we get too complex here. My first thought is that once a `request` comes, if it is banned by the `admin.backend`, there should be some error like "public key not recognized" returned by `validate_admin_token` function. Thus we can log the error as reason.

btw, we can use `inspect_err` for this. like
```rust
let res = self.backend.validate_admin_token(request)
    .inspect_err(|e| {
        info!(
                "Not allowing Admin access for {}, because:\n{e}",
                request.full_url().as_str()
            );
    })?;

info!("Allowing Admin access for {}", request.full_url().as_str());
Ok(res)
}
```

fitzthum: Ok, I made a change like this (but with a match statement instead). I also added some more logging to the simple backend and changed the error type a little bit to be more clear.

Xynnn007 (approved):

PR #1015, opened 2025-10-23
Add a unified policy engine implementation

Xynnn007:
Now both AS and KBS are sharing a very similar policy engine code base. This crate we bring in a new crate to combine them together to one to promote mantainance. Also, we use the key-value abstraction to store the policies. It is benefitial for k8s deployments - keep both read and write API exposable and do not require configmaps.

In later steps we will replace AS and KBS's with this crate.

### DIFF
--- a/.github/workflows/kbs-e2e-sample.yml
+++ b/.github/workflows/kbs-e2e-sample.yml
@@ -39,6 +39,7 @@ jobs:
     uses: ./.github/workflows/kbs-e2e.yml
     with:
       tee: sample
+      arch: aarch64
       runs-on-build: '["ubuntu-22.04-arm"]'
       runs-on-test: '["ubuntu-22.04-arm"]'
       tarball: kbs.tar.gz
--- a/.github/workflows/kbs-e2e.yml
+++ b/.github/workflows/kbs-e2e.yml
@@ -6,6 +6,9 @@ on:
       tee:
         type: string
         required: true
+      arch:
+        type: string
+        default: 'amd64'
       runs-on-build:
         type: string
         default: '["ubuntu-22.04"]'
@@ -79,7 +82,7 @@ jobs:
       with:
         path: test.tar.gz
         overwrite: true
-        name: artifacts-${{ inputs.tee }}
+        name: artifacts-${{ inputs.tee }}-${{ inputs.arch}}
 
   e2e-test:
     needs: build-binaries
@@ -91,7 +94,7 @@ jobs:
     - name: Download artifacts
       uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
       with:
-        name: artifacts-${{ inputs.tee }}
+        name: artifacts-${{ inputs.tee }}-${{ inputs.arch }}
 
     - name: Extract test folder
       run: tar xzf ./test.tar.gz


deps/policy-engine/src/engine/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+
+use std::sync::Arc;
+
+use crate::Result;
+use async_trait::async_trait;
+use serde::Deserialize;
+use serde_json::Value;
+use strum::EnumString;
+
+pub mod opa;
```
mkulke: is it more accurate to call this rego or regorus?

deps/policy-engine/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+pub use error::*;
+use key_value_storage::SetParameters;
+use serde::Deserialize;
+
+pub use key_value_storage::{KeyValueStorage, KeyValueStorageConfig};
+
+#[derive(Debug, Clone, Deserialize, Default, PartialEq)]
+pub struct PolicyEngineConfig {
+    #[serde(default)]
+    pub backend: KeyValueStorageConfig,
```
mkulke: I think storage is more accurate? 

serde annotation can go on the struct

deps/policy-engine/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+        self.engine.evaluate(data, input, &policy).await
+    }
+
+    /// Set a policy to the backend.
+    /// The policy is expected to be provided as string.
+    /// Concrete policy engine backend may handle the policy in different ways.
+    pub async fn set_policy(&self, policy_id: &str, policy: &str, overwrite: bool) -> Result<()> {
+        let params = SetParameters { overwrite };
+        self.backend
+            .set(policy_id.to_string(), policy.as_bytes().to_vec(), params)
```
mkulke: maybe this was an oversight in the storage PR, but does the set() fn need a Vec? usually you would accept a &[u8], so we can avoid allocation.

deps/policy-engine/src/engine/mod.rs
Xynnn007: yes. let's call it rego

deps/policy-engine/src/lib.rs
Xynnn007: ok. fixed

deps/policy-engine/src/lib.rs
Xynnn007: Yes, Let me change it in another commit

deps/policy-engine/src/policy/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+    /// - `input`: dynamic data that will help to enforce the policy.
+    /// - `policy`: the policy to be enforced.
+    async fn evaluate(&self, data: &str, input: &str, policy: &str) -> Result<EvaluationResult>;
+}
+
+#[derive(Debug, EnumString, Deserialize, Clone, Default, PartialEq)]
+#[strum(ascii_case_insensitive)]
+pub enum PolicyType {
+    #[default]
+    Rego,
```
Xynnn007, line 29: Please pay attention to here. Before this, we use `policy_engine` and `OPA` to reflect the policy engine we are using. But for users, they might not care about the policy engine conception but only what policy language will be used. Thus we use policy_type here and a new enum as `rego`. We hide the `regorus` to the backend.

deps/policy-engine/src/error.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    #[error("Illegal policy id. Only support alphabet, numeric, `-` or `_`")]
+    InvalidPolicyId,
+
+    #[error("Malformed policy: {0}")]
+    MalformedPolicy(#[source] anyhow::Error),
+
+    // Backend Related Errors
+    #[error("backend error: {0}")]
+    BackendError(#[from] key_value_storage::KeyValueStorageError),
```
fitzthum: Maybe this should be called StorageError rather than backend error.

Xynnn007: oh yes. let me fix

Xynnn007:
> You are also working on the PR to add reference value extensions to the AS policy. After this PR, does that mean that the AS and KBS policies will both have access to reference values?

Oh. Hopefully not. I think I need to do some change upon that PR

PR #1019, opened 2025-10-24
verifier: nvidia: change evidence encoding to base64

mythi:
nvidia-attester changes the evidence encoding to base64 to the verifier(s) and ITA backed need to be adjusted.

Depends on https://github.com/confidential-containers/guest-components/pull/1151 and I could add `kbs_protocol`crate bump here once merged.

To test:
```
-kbs_protocol = { git = "https://github.com/confidential-containers/guest-components.git", rev = "841b77f", default-features = false }
+kbs_protocol = { git = "https://github.com/mythi/guest-components.git", branch = "nv-2", default-features = false }
```

Question: we have talked about the backwards compatibility. I took a fast path here but could re-work this so that also the hex encoded evidence is supported.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -557,7 +557,7 @@ checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 [[package]]
 name = "attestation-agent"
 version = "0.1.0"
-source = "git+https://github.com/confidential-containers/guest-components.git?rev=841b77f#841b77fc36fac40054a81f261eb15ae95595c23d"
+source = "git+https://github.com/confidential-containers/guest-components.git?rev=5185b46#5185b46518397a6aa8add92ded73549baf5f629c"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -630,7 +630,7 @@ dependencies = [
 [[package]]
 name = "attester"
 version = "0.1.0"
-source = "git+https://github.com/confidential-containers/guest-components.git?rev=841b77f#841b77fc36fac40054a81f261eb15ae95595c23d"
+source = "git+https://github.com/confidential-containers/guest-components.git?rev=5185b46#5185b46518397a6aa8add92ded73549baf5f629c"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -653,7 +653,7 @@ dependencies = [
  "occlum_dcap",
  "picky-asn1 0.10.1",
  "picky-asn1-der 0.5.3",
- "picky-asn1-x509 0.14.6",
+ "picky-asn1-x509 0.15.2",
  "s390_pv",
  "scroll",
  "serde",
@@ -1072,7 +1072,7 @@ dependencies = [
  "hex",
  "hex-literal",
  "jsonwebtoken",
- "multimap 0.9.1",
+ "multimap",
  "openssl",
  "serde",
  "serde_json",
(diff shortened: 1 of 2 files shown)


deps/verifier/src/nvidia/mod.rs
```diff
@@ -133,7 +134,12 @@ impl Nvidia {
             _ => todo!(),
         };
 
-        let evidence_b64 = b64_engine.encode(hex::decode(device.evidence)?);
+        // Try hex::decode() to see if someone uses the original encoding still.
+        // An Err Result suggests the evidence is already base64 encoded and can be used as-is.
+        let evidence_b64 = match hex::decode(device.evidence.clone()) {
```
Xynnn007: we can use the reference to avoid memory copy
```suggestion
        let evidence_b64 = match hex::decode(&device.evidence) {
```

deps/verifier/src/nvidia/mod.rs
```diff
@@ -204,7 +210,12 @@ impl Nvidia {
         let b64_engine = base64::engine::general_purpose::STANDARD;
 
         let cert_chain_vec: Vec<u8> = b64_engine.decode(device.certificate)?;
-        let report_vec: Vec<u8> = hex::decode(device.evidence)?;
+        // Try hex::decode() to see if someone uses the original encoding still.
+        // An Err Result suggests the evidence is base64 encoded so re-try with that.
+        let report_vec: Vec<u8> = match hex::decode(device.evidence.clone()) {
+            Ok(evidence) => evidence,
+            Err(_) => b64_engine.decode(device.evidence)?,
```
Xynnn007: Do we need to print a `debug!` message here to tell the re-try happens?

mythi: I have no strong opinion here. I had that implemented while I was doing testing. `warn!()` might also be relevant for the "deprecated" paths?

Xynnn007: My first impression is that usual users (not experts) hardly care about what happens here. `warn!` is a high log level and almost always shows. Thus I suggest `debug!`, for those who are interested in this

mythi: Do you mean `debug!` for both arms? I don't expect people to use `nvidia-attester` from CoCo 0.16.0 too long so "almost always" is likely not going to be the case.

Xynnn007: not for both. Just for `Err(_) =>`.

What I mean by "almost always", is that usually people will set log level >= `info`. Thus this log will always be showed (if triggered).

Just show a log message here that "oh, the evidence is not hex, thus we tried to use b64 to decode" for better debugging.

mythi: > not for both. Just for `Err(_) =>`.

and for this "local verifier" case (I don't need to do anything for the "remote verifier") only?

fitzthum: yes definitely put a debug message here please

mythi: Added. It should now print something like:
```
2025-10-28T13:40:23Z DEBUG verifier::nvidia] Device evidence is not hex encoded (decoding failed with: Invalid character '/' at position 4). Trying base64 decode.
```

mythi:
> Thanks for this. I have retriggered the ci of rust suites. The link check gets a cough again and we can skip it.

thanks! maybe time to update `.lycheeignore` a bit again.

PR #1021, opened 2025-10-30
tpm: update report data size from 64 to 32 bytes

iroykaufman:
This change is for consistency with [PR#1093](https://github.com/confidential-containers/guest-components/pull/1093) in the guest component. All the explanation on this change are in this PR. 
Also, add some important note on the use of TPM

### DIFF
--- a/attestation-service/README.md
+++ b/attestation-service/README.md
@@ -84,7 +84,7 @@ Please refer to the individual verifiers for the specific format of the evidence
 - Hygon CSV: [CsvEvidence](../deps/verifier/src/csv/mod.rs)
 - IBM Secure Execution (SE): [SeEvidence](../deps/verifier/src/se/mod.rs)
 - NVIDIA: [NvDeviceEvidence](../deps/verifier/src/nvidia/mod.rs)
-- TPM: [TpmEvidence](../deps/verifier/src/tpm/mod.rs)
+- (v)TPM: [TpmEvidence](../deps/verifier/src/tpm/mod.rs)
 
 ## Output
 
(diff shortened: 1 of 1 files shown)


attestation-service/README.md
```diff
@@ -115,7 +115,11 @@ Supported Verifier Drivers:
 - `hygon-dcu`: Verifier Driver for Hygon Deep Computing Unit (DCU).
 - `se`: Verifier Driver for IBM Secure Execution (SE).
 - `nvidia`: Verifier Driver for NVIDIA Devices.
-
+- `TPM`: Verifier Driver for Trusted Platform Module (TPM)
```
dcmiddle: Not sure why we have two lists in the readme, but you probably want to add a similar bullet up on line 86 under the other verifier evidences. 

Also should this all be called vTPM? I don't think we are verifying a physical TPM. I can imagine different Physical TPM usages.

iroykaufman: A vTPM will likely be used in most cases, but the verifier should work the same for both TPM and vTPM, so I think keeping TPM is fine.

attestation-service/README.md
```diff
@@ -84,6 +84,7 @@ Please refer to the individual verifiers for the specific format of the evidence
 - Hygon CSV: [CsvEvidence](../deps/verifier/src/csv/mod.rs)
 - IBM Secure Execution (SE): [SeEvidence](../deps/verifier/src/se/mod.rs)
 - NVIDIA: [NvDeviceEvidence](../deps/verifier/src/nvidia/mod.rs)
+- TPM: [TpmEvidence](../deps/verifier/src/tpm/mod.rs)
```
Xynnn007: Probably we can name this as `(v)TPM`?

attestation-service/README.md
```diff
@@ -115,7 +116,11 @@ Supported Verifier Drivers:
 - `hygon-dcu`: Verifier Driver for Hygon Deep Computing Unit (DCU).
 - `se`: Verifier Driver for IBM Secure Execution (SE).
 - `nvidia`: Verifier Driver for NVIDIA Devices.
-
+- `TPM`: Verifier Driver for Trusted Platform Module (TPM)
```
Xynnn007: btw this should be `tpm` lowercase.

attestation-service/README.md
```diff
@@ -115,7 +116,11 @@ Supported Verifier Drivers:
...
+> **TPM Device Note**: TPM devices (except Azure vTPM series) are not bound to TEE endorsement.
```
Xynnn007, line 121: We can use github [markdown alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) 
```suggestion
> [!WARNING]  
> **TPM Device Note**: TPM devices (except Azure vTPM series) are not bound to TEE endorsement.
```

Xynnn007 (approved):
LGTM. cc @dcmiddle PTAL

Issue #926, opened 2025-08-26
RFC: Extend Admin API to support multiple personas


fitzthum:
Today the [admin configuration](https://github.com/confidential-containers/trustee/blob/main/kbs/src/admin/config.rs) of Trustee is very basic. It only has two fields; a path to a public key representing a singular Trustee admin, and a flag to disable checking the admin private key.

Enterprise use cases will require more sophisticated admin configurations with multiple personas, roles, identities, etc. I have some ideas about how we can introduce these features.

**Decoupled Admin Backend** 

Today what few admin configurations we have come from the Trustee config file. In the future, admin information could be specified in the config file, but it could also come from some other service such as LDAP. In many parts of Trustee we already have a modular backend pattern. We should adopt something similar for admin. The default can be a local admin provider where the admin config comes directly from the Trustee config. Other admin providers can use the Trustee config to store information about connecting to an external admin service.

The exact API of the admin backend isn't yet clear, but the next sections will help to understand some of the requirements.

**Multiple Admin Personas, Roles, and Identities**

There are a lot of things we could do with a more sophisticated admin config. Some things would require significant changes to other parts of Trustee, but some would not.

The simplest possible change would be to support multiple admins, which all have the same capabilities. As soon as we have multiple admins, we probably want to have some kind of admin ID so that we can track (at least in the logs) who did what.

We should probably introduce some kind of admin roles. In the short term these roles can simply allow access to combinations of API endpoints. There can be a role for resources, a role for policies, etc. In the longterm we can think about deeper integrations for roles and maybe even try to figure out some form of multi-tenancy. 

**Authentication Modes**

Today, authenticating to Trustee with the KBS client is a somewhat annoying. You must provide the admin private key to the KBS-Client. It can be annoying to specify this path every time and although it is secure, some people get a little bit scared by dealing with the private key.

In the short term, we could hide this by creating some local kbs-client config that stores your login information (like with Docker) . We should also think about how a modular admin backend would work with logging in. Authenticating a user will probably be delegated to the backend, meaning that the information required to login could be different for different backends. We'll need to make our authentication mechanism flexible enough to handle this.

bpradipt:
Would it make sense to add an oauth proxy layer and let the auth be handled via that for advanced usage?

fitzthum:
@bpradipt yeah I think we could have an oauth admin backend using scopes for different roles.

mkulke:
[...]

PR #1025, opened 2025-11-03
Allow admins to login with a username and password

fitzthum:
Requiring an admin to provide a signed JWT is a bit clunky. Instead, let's let them login with a username and password. 

To facilitate this, introduce a new kbs endpoint where an admin can provide their credentials and receive a token. This endpoint can support multiple login methods in the future, but for now introduce a simple password backend.

Now with the KBS-Client, you can do the following: 
```
$ ./kbs-client admin-login default_admin LYfmThesp8rGba
Login Succeeded
$ ./kbs-client config get-reference-values
"{}"
```

This PR updates the docker compose config to use the password backend by default, but it doesn't change any of the other configs (to avoid breaking various flows).

If the password backend is used but no personas are specified, one user will automatically be added. The username be `default_admin` and the password will be randomly generated and written to the KBS log at startup. This seems like a good way to balance ease-of-use and security.

### DIFF
--- a/.github/workflows/kbs-docker-e2e.yml
+++ b/.github/workflows/kbs-docker-e2e.yml
@@ -44,22 +44,23 @@ jobs:
       run: |
         cargo build --manifest-path tools/kbs-client/Cargo.toml --no-default-features --features sample_only --release
 
-    - name: Setup Keys
-      run: | 
-        openssl genpkey -algorithm ed25519 > kbs/config/private.key
-        openssl pkey -in kbs/config/private.key -pubout -out kbs/config/public.pub
-
     - name: Build KBS Cluster
       run: docker compose build --build-arg BUILDPLATFORM=${{ matrix.build_platform }} --build-arg ARCH=${{ matrix.target_arch }} --build-arg VERIFIER=${{ matrix.verifier }}
 
     - name: Start KBS cluster
-      run: docker compose up -d
+      run: docker compose up -d && sleep 5
+
+    - name: Admin Login
+      working-directory: target/release/
+      run: |
+        ADMIN_PASSWORD=$(docker compose logs kbs | grep "with password:" | awk '{print $NF}' | tr -d '\n')
+        ./kbs-client --url http://127.0.0.1:8080 admin-login default_admin "$ADMIN_PASSWORD"
 
     - name: Set Resource
       working-directory: target/release/
       run: |
         echo "$TEST_SECRET_CONTENT" > test-secret
-        ./kbs-client --url http://127.0.0.1:8080 config --auth-private-key ../../kbs/config/private.key  set-resource --path "$TEST_SECRET_PATH" --resource-file test-secret
+        ./kbs-client --url http://127.0.0.1:8080 config set-resource --path "$TEST_SECRET_PATH" --resource-file test-secret
 
     - name: Get Resource (negative)
       working-directory: target/release/
@@ -68,7 +69,7 @@ jobs:
 
     - name: Update policy
       working-directory: target/release/
-      run: ./kbs-client --url http://127.0.0.1:8080 config --auth-private-key ../../kbs/config/private.key  set-resource-policy --policy-file "$policy_path"
+      run: ./kbs-client --url http://127.0.0.1:8080 config set-resource-policy --policy-file "$policy_path"
       env:
         policy_path: ../../kbs/test/data/policy_2.rego
 


fitzthum:
Tests passing PTAL @confidential-containers/trustee-maintainers 

I have two more PRs coming after this. 1) OAuth2 support 2) Adding regex-based allowed paths to admin personas

fitzthum:
> would a random password per-process harmonize with kubernetes deployments? (replications, etc)

This is just the default for the most basic case where nothing is configured. There are a few options for handling multiple Trustee instances. First, define a persona with a password in the config file and share the config file between instances (as you normally would). Now this persona can login at any of the instances. Second, the password config allows you to specify a token key pair from a file. This is the key pair that signs and validates the token. So if you give all your instances the same one, a token from one instance can be used with any other instances.

Finally, in future PRs we will support external login services, all the instances could share one OAuth2 service or something similar.

kbs/src/admin/error.rs
```diff
@@ -25,6 +28,9 @@ pub enum Error {
     #[error("Failed to parse HTTP Auth Bearer header")]
     ParseAuthHeaderFailed(#[from] actix_web::error::ParseError),
 
+    #[error("Failed to parse admin login request")]
+    ParseAdminLoginFailed(#[from] serde_json::Error),
```
Xynnn007: This error enum is not related to this commit(3c97a8348e97a0645ebec3b5815a50355f9756d9)?

kbs/src/admin/password.rs
```diff
@@ -0,0 +1,176 @@
...
+const ADMIN_TOKEN_ISSUER: &str = "Trustee Admin";
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+#[serde(default)]
+pub struct PasswordAdminConfig {
+    pub personas: Vec<PasswordPersona>,
+    /// The number of hours that an admin token will be valid for.
+    /// After the token expires, an admin must login again and receive
+    /// a new token.
+    pub admin_token_life_hours: u64,
```
Xynnn007, line 38: Let's set this to seconds than hours for a better grained.

kbs/src/admin/password.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// The key pair used for signing and validating the admin tokens.
+    /// If a path is provided, the key pair will be loaded from a file.
+    /// This be used to share auth tokens between Trustee instances and
+    /// to allow Trustee to restart with invalidating existing tokens.
+    /// The key pair should be in PEM format.
+    /// If no path is provided, a random key pair will be generated.
+    pub key_pair_path: Option<PathBuf>,
```
Xynnn007, line 45: If this field is not used, let's delete it. Also, if we want to use the key pair, we can set pri and pub key separately with path.

kbs/src/admin/password.rs
```diff
@@ -0,0 +1,176 @@
...
+
+impl PasswordAdminBackend {
+    pub fn new(config: PasswordAdminConfig) -> Result<Self> {
+        let mut personas = HashMap::new();
+        for p in config.personas {
+            personas.insert(p.username.clone(), p.clone());
+        }
+
+        // TODO read key pair from file
+        let key_pair = Ed25519KeyPair::generate();
```
Xynnn007: This is what I meant. We can't prevent users from executing the built-in random generation logic during deployment. The best approach is to decouple this logic from the rest of the code, since the keypair is a stateful attribute, and it would be better to add it to the database or some other location.

We can have this initial logic in the db setting up script.

kbs/src/admin/password.rs
```diff
@@ -0,0 +1,176 @@
...
+            Err(e) => {
+                info!("Access not granted due to: \n{}", e);
+
+                return Err(Error::AdminAccessDenied)
+            }
+        }
+    }
+
+    fn generate_admin_token(&self, login_data: serde_json::Value) -> Result<String> {
+        let username = login_data
```
Xynnn007: Let's define a `LoginData` here and just deserialize it.

kbs/src/admin/password.rs
```diff
@@ -0,0 +1,176 @@
...
+
+use crate::admin::error::*;
+use crate::admin::AdminBackend;
+
+const ADMIN_TOKEN_ISSUER: &str = "Trustee Admin";
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+#[serde(default)]
+pub struct PasswordAdminConfig {
+    pub personas: Vec<PasswordPersona>,
```
Xynnn007, line 32: Before we get too complex in this, usually the admin panel can only be loginned by one persona, s.t. administrator.

I suggest clarifying the following questions before considering introducing multiple personas/roles:

1. What are the specific responsibilities of each persona? Why are multiple personas needed?

2. Will different personas affect the scope of allowed data/control plane interfaces or something else?

3. How should the different data and control plane interfaces allowed for different roles be defined?

4. What are some typical user stories or processes?

Xynnn007 (commented):
Some first comments for the work

kbs/src/admin/password.rs
fitzthum: Seconds seems too small. How about minutes? I am thinking that people probably won't want to login in very often; maybe every day or two.

kbs/src/admin/password.rs
fitzthum: This is used in the commit that adds support for the key_pair_path. I can move if it's too confusing.

kbs/src/admin/password.rs
fitzthum: See next commits. This is only done if you don't specify a keypair from a file.

kbs/src/admin/password.rs
fitzthum: I have a future PR to introduce the concept of roles for each persona. The simplest approach is to allow each persona to have a set of allowed paths defined by a regex. This allows one persona to be in charge of reference values while another handles policies or certain plugins and so on. You can also filter access to particular secrets with the regex, but this will require some other considerations.

We've had some specific requests internally for a better separation of privileges for admin. You can imagine a situation where one company has a Trustee deployment that is used my different teams. For instance, one team is in charge of building the workload images. They know what the RVs are. Another team might control access to secrets. Another team might handle compliance. Etc. 

Anyway before we get to any of that, it's still a step forward to support multiple admin personas even if they don't have different privileges. This is better than having to share a private key or a password when more than one person is configuring Trustee.

> usually the admin panel can only be loginned by one persona, s.t. administrator.

Even if there is just on type of admin (unlikely in the future), there are still going to be multiple people acting as that admin.

kbs/src/admin/password.rs
Xynnn007: Yes seconds are small. What matters here is grain. No matter minutes or seconds the user can configure a big number. I am ok with minutes.

kbs/src/admin/password.rs
Xynnn007: I suggest decoupling the functionality between PRs. This PR can simply introduce a single administrator account and password login method, because it's relatively clear, while you can fully implement multi-role functionality in another single PR, including:

1. Multi-role functionality

2. How to define the different APIs that different roles can access

3. Multi-role documentation and usage flow

kbs/src/admin/password.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        match res {
+            Ok(claims) => {
+                let username = claims.subject.ok_or(Error::AdminTokenInvalid)?;
+                info!("Admin access granted for {}", username);
+
+                Ok(())
+            }
+            Err(e) => {
+                info!("Access not granted due to: \n{}", e);
```
Xynnn007, line 141:
```suggestion
                warn!("Access not granted due to: \n{}", e);
```

tools/kbs-client/src/main.rs
```diff
@@ -234,11 +263,62 @@ async fn main() -> Result<()> {
...
+            let admin_token =
+                kbs_client::admin_login(cli.url, username, password, kbs_cert.clone()).await?;
+
+            // Write the admin data to the home directory.
+            // Windows is not supported.
+            let mut path = std::env::var_os("HOME")
+                .map(PathBuf::from)
+                .ok_or(anyhow!("Could not find home directory."))?;
+
+            path.push(KBS_CLIENT_DIRECTORY);
```
Xynnn007, line 276: I like this per-users config path.

kbs/src/admin/password.rs
```diff
@@ -0,0 +1,231 @@
...
+impl PasswordAdminBackend {
+    pub fn new(config: PasswordAdminConfig) -> Result<Self> {
+        let mut personas = HashMap::new();
+        for p in config.personas {
+            personas.insert(p.username.clone(), p.clone());
+        }
+
+        // If no personas are specified, create one admin persona
+        // with a random password.
+        // Print the password to the log.
```
Xynnn007, line 74: I suggest that to use a fixed default password like simple "admin" or so. This would help use to do test and debug things as the logging is sometimes not that easy to catch.

kbs/src/admin/password.rs
fitzthum: 2,3 are left to future work already. 

1 is trivial to support; just add a for loop. I put it in this PR since the new `simple` backend that uses client-signed tokens already supports multiple personas. I could move it to another PR, but seems reasonable to have it here. 2 and 3 are much more complicated.

kbs/src/admin/password.rs
fitzthum: Should it be a warning when there is a failed login? A warning would suggest that something bad or unusual has happened, but there is nothing inherently wrong with a failed login. That said, a bunch of failed logins could suggest something bad is happening.,

kbs/src/admin/password.rs
fitzthum: This would be easier to use, but it doesn't seem like a good idea to have a fixed password. I think we should move away from non-secure defaults whenever possible.

kbs/src/admin/error.rs
fitzthum: moved to next commit. i think the idea was that this was a generic error that other impls can use

kbs/src/admin/password.rs
fitzthum: Done

integration-tests/tests/admin.rs
```diff
@@ -196,3 +183,70 @@ async fn set_secret(#[case] test_config: KbsConfigType, #[case] valid_key: bool)
...
+    harness.wait().await;
+
+    info!("TEST: logging in with invalid credentials");
+    harness
+        .login("test1".to_string(), "banana".to_string())
+        .await
+        .map_or_else(
+            |_| Ok(()),
+            |_| Err(anyhow!("Able to log in with wrong credentials.")),
+        )?;
```
Xynnn007: As `fn login` returns `Result<()>`, we can easily use
```suggestion
        .context("Able to log in with wrong credentials.")?;
```

here.

integration-tests/tests/admin.rs
```diff
@@ -196,3 +183,70 @@ async fn set_secret(#[case] test_config: KbsConfigType, #[case] valid_key: bool)
...
+
+    harness
+        .login("test1".to_string(), "password2".to_string())
+        .await
+        .map_or_else(
+            |_| Ok(()),
+            |_| Err(anyhow!("Able to log in with wrong credentials.")),
+        )?;
```
Xynnn007:
```suggestion
        .context("Able to log in with wrong credentials.")?;
```

integration-tests/tests/admin.rs
```diff
@@ -196,3 +183,70 @@ async fn set_secret(#[case] test_config: KbsConfigType, #[case] valid_key: bool)
...
+
+    harness
+        .login("test2".to_string(), "password1".to_string())
+        .await
+        .map_or_else(
+            |_| Ok(()),
+            |_| Err(anyhow!("Able to log in with wrong credentials.")),
+        )?;
```
Xynnn007:
```suggestion
        .context("Able to log in with wrong credentials.")?;
```

integration-tests/tests/admin.rs
```diff
@@ -196,3 +183,70 @@ async fn set_secret(#[case] test_config: KbsConfigType, #[case] valid_key: bool)
...
+        .replace_range(3..5, "x");
+
+    info!("TEST: setting secret");
+    harness
+        .set_secret("a/b/c".to_string(), vec![0u8; 10])
+        .await
+        .map_or_else(
+            |_| Ok(()),
+            |_| Err(anyhow!("Able to log in with wrong credentials.")),
+        )?;
```
Xynnn007: ditto

integration-tests/tests/admin.rs
```diff
@@ -196,3 +183,70 @@ async fn set_secret(#[case] test_config: KbsConfigType, #[case] valid_key: bool)
...
+    info!("TEST: setting secret");
+    harness
+        .set_secret("a/b/c".to_string(), vec![0u8; 10])
+        .await?;
+
+    info!("TEST: tampering with admin token");
+    harness
+        .admin_token
+        .as_mut()
+        .unwrap()
```
Xynnn007: Is there any way to avoid `unwrap()` here? It's not good to me to mix `?` and `unwrap`s, as one is elagant error handling but another will panic directly

kbs/src/admin/password.rs
Xynnn007: Username and password login (corresponding to one administrator who can manage all APIs) is a clear and feasible feature. Including it as a whole in this PR is reasonable. Regarding multiple administrators, in the current PR, each administrator has the same permissions—all APIs, which looks somehow "redundant". You mentioned that permissions will differ in the future, so it's incomplete for now. Let's try to introduce individual sub-features in a decoupled manner. (Implicitly coupled features make reviews more difficult; later PRs would take time to review, and comparing and independently tracking the previous state each time is very time-consuming for reviewers. Independent features also make reviews easier :-) )

so please, decouple them into different PRs, if it's not a complete ability.

kbs/src/admin/password.rs
Xynnn007: This is a design consideration. We need to envision a scenario where multiple KBS instances are deployed simultaneously in a Kubernetes cluster. Regarding the deployer's choice of password, there are two possibilities:

1. They forgot or intentionally didn't provide a password: All services must initialize with the same default password - if each instance initializes different password, the launch will be fine, but the logining will be of great trouble as the deployers do not know which backend KBS will the load balancer lead to.

2. They set a password: All services use the same password.

1 requires the default password to be the same. Although we don't recommend using default passwords, it's unavoidable in testing or various scenarios. And we'd better mention the default password and DONOTs in the documents.

2 requires the KBS logic to be stateless, meaning the password must be stored in shared storage, not in individual instances' memory; otherwise, inconsistencies will easily arise.

kbs/src/admin/password.rs
Xynnn007: Yes. During ABS and AS processes, there are already a lot of info-level logs (including RCAR request arrivals, etc.). The severity of administrator login failure events is obviously higher than that of RCAR request arrivals, otherwise it would be overwhelmed by the massive amount of info logs.

kbs/src/admin/password.rs
Xynnn007: Hi @mkulke  you might be interested in this and please share some good practices upon this

mkulke: I am a bit puzzled by this. It's pretty unorthodox to log a password in clear text in the log, usually you'd be doing the opposite and make sure secrets are not being logged (accidentally). a security scanner in your environment might flag this behaviour.

The other thing that is unusual: the secret in this case is the service's own admin password (not some secret the service requires itself to access e.g. the database). Independent of the distributed/ephemeral state question, a service should not keep the clear text of a password anywhere, usually it should be a salted, slow hash.

we're not doing anything exotic here and I would advise against rolling our own security (see the last reported vuln). Trustee is conceptually a pretty vanilla CRUD service with AuthN/AuthZ requirements. Those are solved problems that should be well-covered by prior art in established libraries, frameworks, etc.

I understand we're not there yet and we want to be pragmatic to support some development scenario until then. But why not have an `--insecure-unauthorized-admin-api` flag instead of custom password logic?

fitzthum: I don't see how this has anything to do with "rolling our own" crypto. We will support external auth services in the next PRs.

The situation is fairly straightforward. We have a few ways to support admin authentication and a few ways to deploy Trustee. We are mainly talking about the docker-compose deployment here, which was originally for testing, but has become one of the most common ways for people to deploy the project. This deployment should be as simple as possible, but since people are using it for real stuff, it also needs to be secure.

We already have a config we can use to turn off admin authentication entirely. We should not do this for docker compose. Our next option would be to have one fixed default admin password. In this case rather than printing the password in plaintext in the log, we would be putting the password in plaintext in the repo.

The current approach is still incremental, but it avoids using a fixed, publicly known password, on publicly-accessible network endpoints.

Anyway, please give a concrete suggestion for how we should do authentication for docker compose. We could stick with the current approach (generate the admin keypair on-the-fly), but this makes the CLI more cumbersome, and it is hardly any more secure since the keypair is accessible to roughly the same people who can read the log.

fitzthum: @Xynnn007 let's not get too far down the road with complex k8s deployments. The scenario we're talking about here is only for the simplest situations, such as docker-compose. In a more complex environment, the personas will be setup, or ideally a separate, shared, authentication service will be used.

kbs/src/admin/password.rs
fitzthum: > so it's incomplete for now

No it isn't. Having multiple logins for the same persona/privilege is still an improvement. This means that multiple admin do not have to share credentials, which is especially important when using a username and password. If we had username and password support but could only have one username, that would be incomplete.

I don't think it's worth the time to split this into another PR, especially since the simple admin backend already supports admin personas in exactly the same way.

kbs/src/admin/password.rs
mkulke: > Anyway, please give a concrete suggestion for how we should do authentication for docker compose

admittedly, I have no experience with using docker compose for anything other than in a local dev environment. So maybe this pattern makes sense for compose. someone who knows more about docker compose should probably chime in.

Xynnn007: > Independent of the distributed/ephemeral state question, a service should not keep the clear text of a password anywhere, usually it should be a salted, slow hash.

Right. The hashed value should be set into the db, than the password plaintext.

kbs/src/admin/password.rs
Xynnn007: This is what I meant. You've introduced a multi-role login mechanism, but there's no difference between each role. Why do we need multiple roles? Why can't all administrators use the same administrator account and password? Just like using a Windows administrator account to log in to the operating system.

We only need additional design when we have a specific problem to solve. It's unwise to design multiple roles just for the sake of a seemingly great solution without defining a specific problem and differentiating the permissions for each role. I strongly suggest introducing multiple roles along with defining the capabilities of different roles as a PR; otherwise, I cannot accept this half-finished product.

kbs/src/admin/password.rs
fitzthum: The goal in this PR is to add the admin session abstraction and provide a simple implementation of it that doesn't rely on any external services. We can add more sophisticated auth integrations soon. I do want to do something good for docker-compose tho. If the password in the log seems too weird I can drop that commit, but I am not sure what to do instead.

kbs/src/admin/password.rs
pmores: (Not that it matters much but) I suspect not having personas would amount to password-sharing which, as a 90's survivor ;-), I'd consider an anti-pattern.  With credentials shared among multiple people someone will leak them sooner or later and it will not be possible to find out who.  It's not possible at all to find out who did what anyway since system logs only say it was done by a shared account.  I don't think this is good practice for a security-sensitive environment, credentials should generally map to individual people.

For that reason I think allowing for multiple personas does add considerable value even if all personas have the same actual rights, at least initially.

Also, I suspect that the "persona" terminology might be a bit misleading here.  In my experience persona is more of a concept from authorisation, rather than authentication which I believe we're dealing with here for now.  Personally I'd call "account" or even just "login" what's currently called persona.  A "RVSP admin" persona or a "policy admin" persona could later be assigned to these accounts.  However even if everybody agrees with me on this, seeing as #1014 introduced the persona term into the simple admin auth plugin already it might be better to allow this PR to go in as is and bulk-refactor later.

kbs/src/admin/password.rs
pmores: @Xynnn007 would you mind elaborating your objection?  I don't really see too much of a problem with this.  If I'm understanding correctly the randomly generated key pair only means that auth tokens can't be shared between KBS instances.  I wouldn't see the disabled functionality as fundamental (in fact I can imagine this could possibly even be desired) and it certainly doesn't look like a security problem to me either.

Xynnn007: Oh I am not against of a random key. I prefer a KBS core logic would better be stateless. Thus the key-generation logic would better to be outside of KBS logic. It will help to deploy multiple instances of KBS to serve as HA

kbs/src/admin/password.rs
pmores: I'm not familiar with the docker-compose scenario either atm so I can't speak to that.

I do agree though that printing a newly-generated passwd into log seems better than having a global default one.  What I'm not sure about though is the assumption that only KBS admins will have access to KBS logs anyway.  Am I wrong to think that host machine/cluster admins might well have access to KBS logs and that members of this group don't necessarily have to be also the folks authorised to make admin changes to KBS?

I think it's a good idea to follow established practices however I believe that when it comes to creating an account with login/password credentials, the established practice is basically to make the user create the account (think e.g. Linux OS installation).  I don't know how useful this is to us, I don't think we have a reliable opportunity to do that anywhere in the deployment process.

So if we can't do that, the next established practice would be not to create an account I suspect.  How bad would that be?  If KBS launches with the password auth mode but with no accounts defined, it will still run but admin operations will be impossible, is that correct?  How does this compare to printing an admin passwd into KBS log?  It seems
- KBS is launched without prior configuration
- admin has to know they need to check the log for their password, and
- admin needs to do so

vs
- KBS is launched without prior configuration
- admin gets an "access denied" when trying to admin, or perhaps is asked even earlier for credentials.  Both of these will likely remind them that they haven't configured any even without knowledge of KBS, or at worst
- admin has to know they need to create an account
- admin has to check the KBS config file and create an account.

Is not creating an account automatically so much worse (a honest question)?  It seems that in either case the admin is not able to do admin ops initially after launch and in either case they gain the ability by doing a small amount of work which is admittedly a bit greater in case of no-default-account which however seems overall much cleaner and more customary.

Xynnn007: Hi @pmores there are new thoughts in the issue https://github.com/confidential-containers/trustee/issues/926 which you might be interested in

kbs/src/admin/password.rs
pmores: I'm sure we all agree that statelessness and HA are important goals for KBS so the decision here seems to be more about whether the KBS needs to be stateless *by default* or not.

Apart from what I said before (i.e. I can imagine KBS deployments with a single KBS instance which don't care much about statelessness), I think it could also be said that by not inserting a key into the config file, the admin effectively states that they don't consider the key to be part of state.  But I understand this might depend on point of view.

Xynnn007: Yes, the reason statelessness is needed is to decouple state and computational logic in multi-replica deployment scenarios. Whether a stateless design is appropriate, or whether multi-replica deployment should be supported, is a question whether the project should address. While I personally support this direction, we still need to discuss and reach a consensus.

A compromise way is to support both, by some specific switches in the config to forward those state things to outer services.

PR #1029, opened 2025-11-05
Allow KBS Client to set init-data and add init-data integration test

fitzthum:
Allow the plaintext init-data to be specified via the KBS Client.

Add an integration test for the init-data. For now we have one simple test that uses a policy to check the init-data hash. In the future we can expand this.

I am thinking we should introduce an extension to parse the init-data toml into json so that it can be more easily accessed from inside the policy. I haven't added this extension yet because the policy stuff is being refactored. Once that is done I will add it and make some more tests. These tests should also help us work on the identifiers stuff too, which I am planning to start soon.

Unfortunately this will have some conflicts with #1025. Ideally we would merge that first, but it's more to review, so we could do this one instead.

### DIFF
--- a/integration-tests/src/common.rs
+++ b/integration-tests/src/common.rs
@@ -302,10 +302,20 @@ impl TestHarness {
         Ok(())
     }
 
-    pub async fn get_secret(&self, secret_path: String) -> Result<Vec<u8>> {
+    pub async fn get_secret(
+        &self,
+        secret_path: String,
+        init_data: Option<String>,
+    ) -> Result<Vec<u8>> {
         info!("TEST: Getting Secret");
-        let resource_bytes =
-            kbs_client::get_resource_with_attestation(KBS_URL, &secret_path, None, vec![]).await?;
+        let resource_bytes = kbs_client::get_resource_with_attestation(
+            KBS_URL,
+            &secret_path,
+            None,
+            vec![],
+            init_data,
+        )
+        .await?;
 
         Ok(resource_bytes)
     }
--- a/integration-tests/tests/init_data.rs
+++ b/integration-tests/tests/init_data.rs
@@ -1,0 +1,134 @@
+// Copyright (c) 2025 by NVIDIA.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::Result;
+use base64::{engine::general_purpose::STANDARD, Engine};
+use log::info;
+use serial_test::serial;
+use sha2::{Digest, Sha256};
+
+extern crate integration_tests;
+use crate::integration_tests::common::{KbsConfigType, PolicyType, TestHarness};
+
+const SECRET_BYTES: &[u8; 8] = b"shhhhhhh";
+const SECRET_PATH: &str = "default/test/secret";
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+#[serial]
+// Check if we can set a policy that expects a particular init-data hash
+// and get a resource.
+async fn get_resource_with_init_data_hash() -> Result<()> {
+    let _ = env_logger::try_init_from_env(env_logger::Env::new().default_filter_or("debug"));
+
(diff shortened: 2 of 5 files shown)


attestation-service/src/ear_token/broker.rs
```diff
@@ -207,6 +207,9 @@ impl EarAttestationTokenBroker {
             submods,
             extensions,
         };
+
+        debug!("EAR Token (unsigned) {:?}", &ear);
```
Xynnn007: You will love the new logging format as
```rust
debug!(
                    ear =? ear,
                    "The unsigned EAR token",
                );
```

integration-tests/src/common.rs
```diff
@@ -305,7 +305,26 @@ impl TestHarness {
...
+    ) -> Result<Vec<u8>> {
+        info!("TEST: Getting Secret with Init-Data");
+        let resource_bytes = kbs_client::get_resource_with_attestation(
+            KBS_URL,
+            &secret_path,
+            None,
+            vec![],
+            Some(init_data),
+        )
+        .await?;
```
Xynnn007: We can reuse the same API `get_resource_with_attestation` without defining this new one. Only need to raise the `info` log before calling it.

Xynnn007 (commented):
Some initial comments. But overall lgtm.

integration-tests/src/common.rs
fitzthum: yeah was trying to keep the existing function the same to minimize the merge conflict but i can update

integration-tests/tests/init_data.rs
```diff
@@ -0,0 +1,134 @@
...
+// and get a resource.
+async fn get_resource_with_init_data_hash() -> Result<()> {
+    let _ = env_logger::try_init_from_env(env_logger::Env::new().default_filter_or("debug"));
+
+    let harness = TestHarness::new(KbsConfigType::EarTokenBuiltInRvps.into()).await?;
+
+    // Set Secret
+    info!("TEST: setting secret");
+    harness
+        .set_secret(SECRET_PATH.to_string(), SECRET_BYTES.as_ref().to_vec())
```
Xynnn007: Not sure why this is not detected by lint checker.

```suggestion
        .set_secret(SECRET_PATH.to_string(), SECRET_BYTES.to_vec())
```

integration-tests/tests/init_data.rs
```diff
@@ -0,0 +1,134 @@
...
+        .await?;
+
+    // Set Policy
+    let policy = policy_with_init_data_hash(SIMPLE_INIT_DATA.to_string())?;
+
+    info!("TEST: policy with init-data hash: {}", policy.clone());
```
Xynnn007: I remembered that the logging can directly use original variable without memory copying.
```suggestion
    info!("TEST: policy with init-data hash: {policy}");
```

kbs/src/policy_engine/opa/mod.rs
```diff
@@ -50,6 +51,7 @@ impl PolicyEngineInterface for Opa {
             .map_err(|_| KbsPolicyEngineError::DataLoadError)?;
 
         // Add TCB claims as input
+        debug!("KBS Policy Input Claims: {}", &input_claims);
```
Xynnn007:
```suggestion
        debug!("KBS Policy Input Claims: {input_claims}");
```

tools/kbs-client/src/main.rs
```diff
@@ -55,6 +55,15 @@ enum Commands {
         /// and the public part of TEE Key should be consistent with tee-pubkey in the token.
         #[clap(long, value_parser)]
         attestation_token: Option<PathBuf>,
+
+        /// InitData plaintext
```
mythi, line 59: anything we should say about the expected format of this file here?

fitzthum: Added a comment. We can't be too prescriptive here since the spec supports toml and json. We don't really know what someone will be expecting when they use the KBS Client.

PR #1032, opened 2025-11-06
dependabot: bump open-pull-requests-limit to 3

mythi:
Currently, only one dependabot PR is allowed to be open at a time. A dependency that requires manual work can blocks other updates if it's not being worked on.

Bump open-pull-requests-limit to 3 to let dependabot to create parallel PRs and not get blocked.


fitzthum (approved):
Ok

PR #1035, opened 2025-11-06
k8s: add nvida remote verifier support to k8s script

fitzthum:
We're going to test the NVIDIA remote verifier in the Kata CI. To facilitate this, allow our kbs script to update the KBS config for the remote verifier.

### DIFF
--- a/kbs/config/kubernetes/deploy-kbs.sh
+++ b/kbs/config/kubernetes/deploy-kbs.sh
@@ -6,6 +6,8 @@ set -euo pipefail
 DEPLOYMENT_DIR="${DEPLOYMENT_DIR:-overlays}"
 OVERLAYS_DIR="overlays"
 
+NVIDIA_VERIFIER_MODE="${NVIDIA_VERIFIER_MODE:-local}"
+
 k8s_cnf_dir="$(dirname ${BASH_SOURCE[0]})"
 
 if [ "$(uname -m)" == "s390x" ] && [ -n "${IBM_SE_CREDS_DIR:-}" ]; then
@@ -31,6 +33,15 @@ kbs_cert="${k8s_cnf_dir}/base/kbs.pem"
     openssl pkey -in "${k8s_cnf_dir}/base/kbs.key" -pubout -out "${kbs_cert}"
 }
 
+# Enable the nvidia remote verifier if requested.
+if [[ "${NVIDIA_VERIFIER_MODE}" == "remote" ]]; then
+    cat >> ${k8s_cnf_dir}/base/kbs-config.toml << 'EOF'
+
+[attestation_service.verifier_config.nvidia_verifier]
+type = "Remote"
+
+EOF
+fi
 
 if [[ "${DEPLOYMENT_DIR}" == "nodeport" || "${DEPLOYMENT_DIR}" == "overlays" ]]; then
     kubectl apply -k "${k8s_cnf_dir}/${DEPLOYMENT_DIR}"


kbs/config/kubernetes/deploy-kbs.sh
```diff
@@ -31,6 +33,15 @@ kbs_cert="${k8s_cnf_dir}/base/kbs.pem"
...
 
+# Enable the nvidia remote verifier if requested.
+if [[ "${NVIDIA_VERIFIER_MODE}" == "remote" ]]; then
+    cat >> ${k8s_cnf_dir}/base/kbs-config.toml << 'EOF'
+
+[attestation_service.verifier_config.nvidia_verifier]
+type = "Remote"
+
+EOF
+fi
```
Xynnn007, line 44: Can we change the original config file than modify it in a script here? It could help avoid maintain config file in two places.

mythi: I was also thinking that updating the `.toml` file directly should be ok

fitzthum: Hmm you mean just always have the remote verifier enabled in K8S? That might be fine for the CI, but this script is also for general users to deploy Trustee in k8s. The remote verifier has some licensing implications so it might be best to require them to opt-in.

Also note that https://github.com/kata-containers/kata-containers/pull/12041 was already merged.

It would simplify things to make it always enabled tho. Wdyt?

Xynnn007: Modifying the original configuration file within the script can easily lead to forgetting that it might have been modified elsewhere simultaneously during subsequent maintenance. There are two feasible methods:

1. Keep only the suggested deployment method (Local) configuration file in the Trustee repository, and then integrate the Remote configuration file upstream, i.e., in Kata CI.

2. Keep both modes' configuration files in the Trustee repository, and then control which to use via environment variables.

Personally, I prefer method 1, because for Trustee, providing a single configuration instance is sufficient; Kata CI is essentially the consumer. However, since the upstream has already merged the environment variable options, I propose method 2.

fitzthum: I don't think duplicating the entire config file is a great idea. That creates more work whenever we change anything in the config, which is already a pain.

> Modifying the original configuration file within the script can easily lead to forgetting that it might have been modified elsewhere simultaneously during subsequent maintenance

Where else would it be modified? This config is only used for the k8s deployment. I think there are some improvements that we could make to `deploy-kbs.sh` in the future (like removing the logic for setting a key), but for now this seems like a straightforward change.

mythi: > The remote verifier has some licensing implications so it might be best to require them to opt-in.

ok, right. In that case I'm fine with this PR.

mkulke: in the future it would probably make some sense to have config.d folder à la containerd, in which small config snippets can be dropped into (and stitched together later, I don't think the config create supports that ootb sadly)

Xynnn007: > I don't think duplicating the entire config file is a great idea. That creates more work whenever we change anything in the config, which is already a pain.

Changing it in the script will also bring the same issue. If we use config files we can at least know that config files need to be modified. Once we use script we will almost forget this thing. But as you said, now the deployments are quite a messy for kubernetes. We need to refactor the code to make it more convenient for maintaince. I will take a look at it, before we get too trouble into it.

PR #1037, opened 2025-11-07
Http server: configurable number of worker threads

### LINKED ISSUE #1036: [Enhancement] Allow configuration of actix-web worker count for optimization
### Summary

Propose introducing a configuration option to allow users to explicitly set the number of `actix-web` workers spawned by the `trustee` process.

### Motivation

Currently, the `trustee` process relies on the default `actix-web` behavior, which spawns a number of workers equal to the number of logical CPU cores available on the machine.

On systems with a high core count and hyper-threading enabled, this can result in an extremely large number of workers (e.g., 256 workers).

This default is often unsuitable for **Kubernetes (k8s)** deployments where:
1. **Resource Limits are Applied:** The container may be allocated fewer CPU resources than the underlying host's core count.
2. **Multiple Replicas are Used:** Deploying multiple replicas of `trustee` with a high default worker count can lead to resource oversaturation and inefficient utilization across the cluster.

### Proposal

Introduce a mechanism (e.g., a new environment variable or a configuration file setting) to allow operators to specify a maximum or exact number for the `actix-web` workers.

**Example implementation idea:**

* **Configuration:** Read an environment variable, such as `TRUSTEE_ACTIX_WORKERS`.
* **Logic:**
    * If `TRUSTEE_ACTIX_WORKERS` is set, use that value for the worker count.
    * If not set, fall back to the current default (number of logical cores).

lmilleri:
Added `worker_count` attribute to the trustee `http_server` section for specifying the number of worker threads.
If not provided, defaults to the logical number of logical CPU cores

Fixes #1036

### DIFF
--- a/integration-tests/src/common.rs
+++ b/integration-tests/src/common.rs
@@ -218,6 +218,8 @@ impl TestHarness {
                 certificate: None,
                 insecure_http: true,
                 payload_request_size: 2,
+                worker_count: Some(4),
+            
             },
             admin: admin_config,
             policy_engine: PolicyEngineConfig {
--- a/kbs/docs/config.md
+++ b/kbs/docs/config.md
@@ -19,13 +19,14 @@ The following sections list the KBS properties which can be set through the conf
 
 The following properties can be set under the `[http_server]` section.
 
-| Property               | Type         | Description                                      | Required | Default              |
-|------------------------|--------------|--------------------------------------------------|----------|----------------------|
-| `sockets`              | String array | One or more sockets to listen on.                | No       | `["127.0.0.1:8080"]` |
-| `insecure_http`        | Boolean      | Don't use TLS for the KBS HTTP endpoint.         | No       | `false`              |
-| `private_key`          | String       | Path to a private key file to be used for HTTPS. | No       | None                 |
-| `certificate`          | String       | Path to a certificate file to be used for HTTPS. | No       | None                 |
-| `payload_request_size` | Integer      | Request payload size in mega bytes.              | No       | 2                    |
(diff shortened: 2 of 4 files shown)


kbs/docs/config.md
```diff
@@ -26,6 +26,7 @@ The following properties can be set under the `[http_server]` section.
 | `private_key`          | String       | Path to a private key file to be used for HTTPS. | No       | None                 |
 | `certificate`          | String       | Path to a certificate file to be used for HTTPS. | No       | None                 |
 | `payload_request_size` | Integer      | Request payload size in mega bytes.              | No       | 2                    |
+| `worker_count`         | Integer      | Number of HTTP actix worker threads              | No       | None                 |
```
fitzthum: What is the default value?

integration-tests/src/common.rs
```diff
@@ -218,6 +218,8 @@ impl TestHarness {
                 certificate: None,
                 insecure_http: true,
                 payload_request_size: 2,
+                worker_count: None,
```
fitzthum: In the tests, we could set this to be lower. We will generally only receive one request at a time.

kbs/docs/config.md
lmilleri: when not provided, defaults to the number of logical CPU cores. Change "None" to "Num of logical CPU cores" here?

mkulke (approved):

kbs/docs/config.md
mythi: > Change "None" to "Num of logical CPU cores" here?

+1

fitzthum (approved):
Thanks

PR #1038, opened 2025-11-07
Parse Config Files in InitData to JSON

fitzthum:
Today the AA and CDH config files are in the initdata claims, but they are toml strings that are difficult to access from the policy.

Let's parse these into json and add a test showing how you can write a policy that checks values from the config files.

This is another small step towards supporting identifiers, which may also be based on these config values.

See commit messages for details.


Xynnn007 (approved):
LGTM. Thanks for this!

PR #1042, opened 2025-11-10
fix: improve swtpm cleanup robustness

Xynnn007:
- Add check for swtpm.pid file existence before attempting cleanup
- Add informative messages during cleanup process
- Prevent errors when swtpm is not running

cc @bpradipt


bpradipt (approved):
/lgtm
Thanks @Xynnn007

PR #1044, opened 2025-11-10
Parse Kata Agent Policy from Init-Data and Add Integration Test

fitzthum:
Today, if the kata-agent policy is in the init-data, it will be carried through the Trustee attestation flow.

This sounds great, but things are actually quite unwieldy. The kata-agent policy is very large and difficult to parse. Rather than exposing the entire policy to the AS and KBS policy engines as a string (and including it in the attestation token as a string), let's extract the `policy_data` field from the OPA and parse that as JSON.

This way KBS and AS policies can check the agent policy data relatively easily. I add an integration test to illustrate this, although it is only a toy example.

Ideally the runtime will start separating the data and policy fields and we can remove this code (and not put the static parts of the policy in the init-data at all), but for now I think this is the best we can do.

This is another little step towards identifier support. See commit messages for details.

### DIFF
--- a/attestation-service/README.md
+++ b/attestation-service/README.md
@@ -92,6 +92,8 @@ If the verification of TEE evidence does not fail, the AS will return an Attesta
 
 Attestation Results token is an [EAT Attestation Results (EAR)](https://www.ietf.org/archive/id/draft-fv-rats-ear-05.html) in [JSON Web Token](https://datatracker.ietf.org/doc/html/rfc7519) format which contains the parsed evidence claims such as TCB status.
 
+More information about the attestation token is available [here](./docs/attestation_token.md).
+
 The claims part of the typical EAR token JWT is [here](./docs/example.token.json).
 
 ## Architecture
--- a/attestation-service/docs/attestation_token.md
+++ b/attestation-service/docs/attestation_token.md
@@ -1,0 +1,59 @@
+# The Attestation Token
+
+The Attestation Service generates an EAR attestation token, which contains many claims.
+
+For the general structure of the attestation token, refer to the [EAR specification](https://datatracker.ietf.org/doc/draft-ietf-rats-ear/).
+
+Generally speaking, the attestation token will include an Appraisal for each device
+that is part of the TCB (including the CPU).
+These appraisals can be accessed through the `submods` field in the EAR Token.
+Each submod is given a generic key constructed from the CPU class and the device count.
+For instance, the CPU submod will be called `cpu0`.
+If a GPU has been attested as part of the guest, there will be a `gpu0` submod.
+If there is more than one GPU, there will be additional `gpuN` submods.
+
+`cpu0` is considered to be the primary attester and has some special
+information associated with it.
+
+# ReportData and InitData
+
+ReportData and InitData are two key Trustee concepts.
+ReportData refers to data provided by a guest at attestation time.
+This is sometimes called user data. The KBS protocol uses this
+field to measure the the nonce and Tee Public Key.
+During attestation Trustee will ensure that the corresponding
+field in the hardware evidence matches the report data values
+that are expected for a given connection.
+
+InitData is a more powerful and subtle concept. See the [InitData Specification](../../kbs/docs/initdata.md)
+for more information.
+The basic idea is that InitData is a generalization over boot-time configuration
+fields such as HostData (on SNP) or MRConfig (on TDX).
+InitData is used to provision dynamic, measured, but not secret 
+configuration data to the guest.
+The InitData plaintext is a TOML or JSON file containing this configuration
+while the InitData hash is the hash of this file which is added to the measuremet.
+A client can optionally provide the InitData plaintext to Trustee.
+If so, Trustee will check the plaintext against the hardware evidence
+and expose the InitData plaintext to the policy engine and as part of the
+attestation token.
+
+Both the InitData and ReportData will usually be included the attestation token.
+These fields will be available under the `AnnotatedEvidence` extension of the
+`cpu0` Appraisal.
+
+If the plaintext InitData is provided, some transformations will be applied
+to the InitData to make it more easy to consume.
+
+As a result, the InitData section in the token may contain the following fields:
+- `cdh.toml` JSON version of the CDH config file from the InitData.
+- `aa.toml` JSON version of the AA config file from the InitData.
+- `agent_policy_claims` The `policy_data` claim from the Kata Agent policy (as JSON).
+
+These fields will only be present if the plaintext InitData contains the corresponding
+data.
+
+# Hardware Claims
+
+The annotated evidence extension will also include hardware-specific claims
+extracted by the verifiers. These are listed in the [TCB Claims](./tcb_claims.md) document.
--- a/attestation-service/docs/tcb_claims.md
+++ b/attestation-service/docs/tcb_claims.md
@@ -1,14 +1,10 @@
-# Parsed Claims
+# TCB Claims
 
-After CoCo-AS verifies every evidence, a token with _parsed claims_ will be returned.
-_parsed claims_ is a key-value map inside the token, and it reflects the environment
-information that the evidence contains. Different platforms will have different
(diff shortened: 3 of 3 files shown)


attestation-service/src/lib.rs
```diff
@@ -327,6 +327,17 @@ fn parse_init_data(data: Option<InitDataInput>) -> Result<(Option<Vec<u8>>, Valu
...
                     }
+
+                    // If there is a policy rego file, replace the full policy file
+                    // with some claims extracted from the file.
+                    if let Some(Value::String(rego_str)) = map.get("policy.rego").cloned() {
+                        if let Some(policy_data) = extract_policy_data(&rego_str) {
+                            // Since we found some policy claims, remove the original
+                            // policy rego from the claims.
+                            map.remove("policy.rego");
+                            map.insert("agent_policy_claims".to_string(), policy_data);
```
Xynnn007, line 338: Note that AA and CDH config still follow the original name `cdh.toml` and `aa.toml`. We'd better either use `policy.rego` also here, or

If new names are preferred, we should document them somewhere public. Or a user might get confused by the name changes.

attestation-service/src/lib.rs
```diff
@@ -335,6 +346,47 @@ fn parse_init_data(data: Option<InitDataInput>) -> Result<(Option<Vec<u8>>, Valu
...
+            '{' => depth += 1,
+            '}' => {
+                depth -= 1;
+                if depth == 0 {
+                    end_idx = i + 1;
+                    break;
+                }
+            }
+            _ => {}
+        }
```
Xynnn007: I like that we are trying to use some algorithms like the stack style to handle issues  here. There is a case not that "corner" - a `{` or a `}` occurs in a string part of the `policy_data`. This would make trouble to the calculation of nested counter.

Can we assert that the last part of `policy.rego` is always`policy_data`? If we can, we could just try to deserialize the contents starting with `policy_data := {` to the end with `serde_json::Value`?

attestation-service/src/lib.rs
fitzthum: The cases are a little bit different. For the config files, we are putting the exact same content into the claims, but we are changing the format. For the policy we are only putting part of the policy into the claims.

It might make sense to change `cdh.toml` to `cdh.json`. Either way is probably fine.  I think we probably should not call this `policy.rego` since it is not the policy.

attestation-service/src/lib.rs
fitzthum: Yeah if the policy data has a string that contains an unmatched bracket, this will break. It's not a perfect heuristic. We could assume that the policy_data is the end of the policy. Currently this is how genpol operates, but if it changes things would break. Checking the brackets seems less fragile than assuming that `policy_data` is the last thing in the file, but it is more complex. So idk I am open to either way wdyt?

Xynnn007: This is something we should align with the logic [here](https://github.com/kata-containers/kata-containers/blob/main/src/tools/genpolicy/src/policy.rs#L591-L597). Currently it's true that it's the end so it's fine to me.

checking the brackets is a big cost. A possible "ugly" workaround is to add a code comment at kata's `policy.rs` side to tell there should be changes upon trustee if kata side changes. 

cc @danmihai1 Could you help to ensure if "`policy_data` will always be the end of the `policy.json`"?

attestation-service/src/lib.rs
Xynnn007: Ok. We can consider the names later. before that, let's update the [document](https://github.com/confidential-containers/trustee/blob/main/attestation-service/docs/parsed_claims.md) to record what it is like now than showing it in the code.

attestation-service/src/lib.rs
mkulke: can we get some parsed interpretation (an AST) and work with that instead of relying on manual parsing gymnastics, e.g. what about comments and strings, that have escaped braces?

Xynnn007: I'm not sure if this is feasible, but we have already included the regorus library.

Hi @anakrish, we have met a case where the given input is a `rego` file like [this](https://github.com/confidential-containers/trustee/pull/1044/files#diff-abd569841a91381d208d4fcf801143d136da2e7c2dcc9b64bf95443d67eacb40R6-R2069), and we need to extract the last `policy_data` part into a JSON string. Is it possible by getting help from `regorus`?

fitzthum: Ok let me just assume that it is the end of the file. I think we could use regorus to parse the whole thing and extra a claim, but that seems excessive to me.

anakrish: Hey @Xynnn007, You could [eval_rule](https://docs.rs/regorus/latest/regorus/struct.Engine.html#method.eval_rule) `data.agent_policy.policy_data` to evaluate and fetch the value.

Otherwise you might need some marker within the file contents to reliably locate policy_data and then parse it as a json.

attestation-service/src/lib.rs
fitzthum: Ok, I added a commit to update `parsed_clalims`. I ended up changing a few more things there since the document is a little out of date.

attestation-service/docs/claims.md
```diff
@@ -1,14 +1,63 @@
...
+These appraisals can be accessed through the `submods` field in the EAR Token.
+Each submod is given a generic key constructed from the CPU class and the device count.
+For instance, the CPU submod will be called `cpu0`.
+If a GPU has been attested as part of the guest, there will be a `gpu0` submod.
+If there is more than one GPU, there will be additional `gpuN` submods.
+
+`cpu0` is considered to be the primary attester and has some special
+information associated with it.
+
+# ReportData and InitData
```
Xynnn007: It should be `RuntimeData`

attestation-service/docs/claims.md
```diff
@@ -1,14 +1,63 @@
...
+If the plaintext InitData is provided, some transformations will be applied
+to the InitData to make it more easy to consume.
+Specifically, if the InitData plaintext is provided as toml,
+some of the data fields in the InitData will be parsed into JSON.
+If the `cdh.toml` or `aa.toml` are present in the InitData,
+they will be parsed as JSON (although the keys will still be called `x.toml`).
+If `policy.rego` is present in the InitData and contains a `policy_data` claim
+at the end of the file, the `policy.rego` field will be removed from the InitData
+and replaced with the `policy_data` parsed as JSON. This will inserted
+with the key `agent_policy_claims`.
```
Xynnn007: Do you think it make sense to use three items to represent three different special cases (with Markdown hyphens at the beginning of each line) would be clearer and more organized. e.g.

```
Some special keys ...

- `cdh.toml`: if ....
- `aa.toml`: ...
- `agent_policy_claims`: ...
```

attestation-service/docs/claims.md
```diff
@@ -1,14 +1,63 @@
-# Parsed Claims
+# The Attestation Token
```
Xynnn007: This document is used to help people to write attestation policies. It tells users what keys(claims) can be used in the policy, so attestation token things can be moved to [here](https://github.com/confidential-containers/trustee/tree/main/attestation-service#output) or a separate document.

attestation-service/docs/claims.md
fitzthum: hmm we have some inconsistency here. In the source we use both RuntimeData and ReportData. In the attestation token we have `report_data` and `runtime_data_claims`, which we should align.

Which term do you prefer? Runtime data sounds similar to runtime measurements so maybe report data is more clear? Either way I can fixup the inconsistencies, but maybe in another PR.

attestation-service/docs/claims.md
fitzthum: I made a new doc called `attestation_token.md` and renamed this one to `tcb_claims.md`.

attestation-service/docs/claims.md
Xynnn007: Oh yes. Let's try to figure out them in another PR.

PR #1052, opened 2025-11-12
as/verifier: Enable s390x to verify amd64 attesters

BbolroC:
This PR enables KBS on s390x to fulfil attestation for amd64 confidential VMs via

- ITA for TDX
- Verifier for SNP and Nvidia

### DIFF
--- a/.github/actionlint.yml
+++ b/.github/actionlint.yml
@@ -2,3 +2,4 @@ self-hosted-runner:
   # Labels of self-hosted runner in array of string
   labels:
     - s390x
+    - s390x-large
--- a/.github/workflows/build-kbs-image.yml
+++ b/.github/workflows/build-kbs-image.yml
@@ -33,8 +33,6 @@ jobs:
           - rhel-ubi
         exclude:
           - target_arch: s390x
-            tag: kbs-ita-as
-          - target_arch: s390x
             tag: rhel-ubi
           - target_arch: aarch64
             tag: kbs-ita-as
--- a/.github/workflows/kbs-rust.yml
+++ b/.github/workflows/kbs-rust.yml
@@ -31,6 +31,8 @@ jobs:
           test_features: ""
         - instance: ubuntu-24.04-arm
           test_features: "coco-as-builtin,coco-as-grpc,intel-trust-authority-as,cca-attester"
+        - instance: s390x-large
+          test_features: "coco-as-builtin,coco-as-grpc,intel-trust-authority-as"
     runs-on: ${{ matrix.instance }}
 
     steps:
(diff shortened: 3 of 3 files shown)


deps/verifier/Cargo.toml
```diff
@@ -18,6 +18,11 @@ all-verifier = [
     "nvidia-verifier",
     "tpm-verifier",
 ]
+sel-verifier = [
+    "snp-verifier",
+    "nvidia-verifier",
+    "se-verifier",
+]
```
fitzthum: `se-verifier` vs `sel-verifier` is a little confusing. Is there a way to make it more clear which verifiers are for SE guests and which ones are capable of running on s390x? Maybe this could be handled with arch flags instead?

BbolroC: I've simply changed `sel-verifier` to `s390x-verifier` meaning verifiers which can run on s390x. Thanks!

Xynnn007: Hi @BbolroC If the name is expected to be changed, all the `se-verifier` in documents and codes should to be renamed to `s390x-verifier`.  Also for [guest-components side](https://github.com/confidential-containers/guest-components/tree/main/attestation-agent/attester/src/se)

Xynnn007: Oh sorry, I misunderstood. Could we rename as `verifiers-can-be-built-on-s390x` or something else. The pattern `*-verifier` can easily mislead people like me : (

mythi: I don't the need for a new feature especially since one of the verifiers in the new set is already enabled. How about just add the two other verifiers to the features list for `s390x` builds and add a comment there that these verifiers can run on IBM builds.

deps/verifier/Cargo.toml
```diff
@@ -65,9 +70,10 @@ scroll = { version = "0.13.0", default-features = false, features = [
 serde.workspace = true
 serde_json.workspace = true
 serde_with = { workspace = true, optional = true }
-sev = { version = "6.3.1", default-features = false, features = [
+sev = { git = "https://github.com/virtee/sev", rev = "d487809188c3c92e397bc92e2cb2a85eb982c980", default-features = false, features = [
```
mythi, line 68: There's no new release with the fix yet?

BbolroC: nope

deps/verifier/Cargo.toml
BbolroC: > How about just add the two other verifiers to the features list for s390x builds

Oh, it sounds great. I will give it a try. Thanks for suggestion!

PR #1054, opened 2025-11-13
Dependabot/cargo/jsonwebtoken 10.2.0 manual

Xynnn007:
Manual version for #1050

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,7 +106,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.106",
+ "syn 2.0.110",
 ]
 
 [[package]]
@@ -164,9 +164,9 @@ dependencies = [
 
 [[package]]
 name = "actix-tls"
-version = "3.4.0"
+version = "3.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac453898d866cdbecdbc2334fe1738c747b4eba14a677261f2b768ba05329389"
+checksum = "6176099de3f58fbddac916a7f8c6db297e021d706e7a6b99947785fee14abe9f"
 dependencies = [
  "actix-rt",
  "actix-service",
@@ -229,7 +229,7 @@ dependencies = [
  "serde_json",
  "serde_urlencoded",
  "smallvec",
- "socket2 0.6.0",
+ "socket2 0.6.1",
  "time",
  "tracing",
  "url",
@@ -244,7 +244,7 @@ dependencies = [
  "actix-router",
  "proc-macro2",
  "quote",
- "syn 2.0.106",
+ "syn 2.0.110",
 ]
 
 [[package]]
@@ -270,7 +270,7 @@ checksum = "b6ac1e58cded18cb28ddc17143c4dea5345b3ad575e14f32f66e4054a56eb271"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.106",
+ "syn 2.0.110",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -42,7 +42,7 @@ jwt-simple = { version = "0.12", default-features = false, features = [
 kbs_protocol = { git = "https://github.com/confidential-containers/guest-components.git", rev = "7be23a1", default-features = false }
 kbs-types = "0.14.0"
 kms = { git = "https://github.com/confidential-containers/guest-components.git", rev = "7be23a1", default-features = false }
-jsonwebtoken = { version = "10", default-features = false }
+jsonwebtoken = { version = "10", default-features = false, features = ["aws_lc_rs"] }
```
mkulke, line 45: will this link against the aws libcrypto c library? would rust-crypto be an option?

mythi: rust-crypto is an option too. not sure why they made this choice a breaking change instead of just adding adding aws-lc crypto as an additional feature

Xynnn007: This is the result of me compiling grpc-as using gnu (`cargo build -p attestation-service --bin grpc-as`) and the result of `ldd`
```
        linux-vdso.so.1 (0x00007ffc8406b000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fdb7d1ed000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fdb7d1c8000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fdb7d1a5000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fdb7b0b1000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fdb7d19f000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fdb7aebf000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fdb7d21d000)
```

also it can be built with `musl` (`cargo build -p attestation-service --bin grpc-as --target x86_64-unknown-linux-musl`)

thus it is `statically linked`.


Is this acceptable?

Currently, `jsonwebtoken` is imported by this crate and also by its dependency crate `rust-ear` with `aws_lc_rs`. If needed to change to `rust-crypto`, I need to first update the `jsonwebtoken` dependency on the `rust-ear` upstream library, switching from `aws_lc_rs` to rust-crypto and then go back to this PR.

> rust-crypto is an option too. not sure why they made this choice a breaking change instead of just adding adding aws-lc crypto as an additional feature

The upstream crate seems decouple the crypto suites into either `aws_lc_rs` or `rust-crypto` for developers to choose.

mkulke: there are competing requirements, sometimes you require static linking, sometimes static linking is forbidden. I think the former is mostly the case for constrained environments like an initrd in a TEE. for services it's acceptable to depend on a library I guess. If rust-ear already has made the decision to choose aws_lc_rs, we can just go with that, since JWK and ear are tightly coupled.

mkulke (approved):

Cargo.toml
Xynnn007: I must mis-deliver my idea. The dynamic linked information i showed reflects no aws libc is linked dynamically. Also it supports static link as I showed later. so it does not break the ststic link requirement we always head to.

kbs/src/token/jwk.rs
```diff
@@ -261,7 +261,7 @@ mod tests {
     )]
     #[case(
         "{\"keys\":[{\"kty\":\"oct\",\"alg\":\"COCO42\",\"kid\":\"coco123\",\"k\":\"foobar\"}]}",
-        true
+        false
```
mythi: This test intentionally writes an invalid JWK and we want to check an error is returned. With this change, the test case becomes useless and the test coverage for invalid algos gets lost. If we accept it here, we probably need a check for `UNKNOWN_ALGORITHM` elsewhere?

Xynnn007: sounds good. Let me do a little change here.

mythi (approved):
@Xynnn007 thanks for the updates! LGTM

Xynnn007:
There are some CI errors according to the links which is not related to this PR, thus it be acceptable to get this merged.

Link fix PR https://github.com/confidential-containers/trustee/pull/1080

PR #1061, opened 2025-11-14
snp-verifier: http_cache_reqwest to cache vceks

amd-aliem:
Use http_cache_reqwest to cache VCEKs. After review comments, this PR will cover only an in-memory performance cache, with a filesystem persistent cache to be added in a follow-up PR. 

---
Initial PR description:

> Support configuration option in snp verifier to use http_cache_reqwest to cache VCEKs. This allows users to pre-load them for offline environments if desired. Add tool to convert VCEK URLs to cache contents.
> 
> Follow up from https://github.com/confidential-containers/trustee/pull/989, implementing suggestion https://github.com/confidential-containers/trustee/pull/989#issuecomment-3400679521
> 
> Notes:
> - Main doc describing implementation: https://github.com/amd-aliem/trustee/blob/amd-cert-cache/attestation-service/docs/amd-offline-certificate-cache.md
> - Having this as an opt-in, despite performance benefits even in normal online situations. Reasoning is that I'm not 100% sure http_cache_reqwest ever evicts elements, and in default caching mode it appeared to only keep cache contents valid on the same request. Worried it will take up non-negligible disk space over time. 
> - Currently this new doc only describes how to enable w/ docker. Plan to add k8s operator instructions in the future but I haven't figured out how to test local changes w/ that yet.

### DIFF
--- a/deps/verifier/Cargo.toml
+++ b/deps/verifier/Cargo.toml
@@ -80,6 +80,8 @@ ccatoken = { git = "https://github.com/veraison/rust-ccatoken", rev = "6d5b8db",
 ear = { version = "0.4.0", optional = true }
 x509-parser = { version = "0.18.0", optional = true }
 reqwest.workspace = true
+reqwest-middleware = "0.4.2"
+http-cache-reqwest = { version = "0.16.0", default-features = false, features = ["manager-moka"] }
 bitflags = { version = "2.10.0", features = ["serde"] }
 time = "0.3.41"
 nvml-wrapper = { version = "0.11.0", optional = true, default-features = false, features = [
--- a/deps/verifier/src/snp/mod.rs
+++ b/deps/verifier/src/snp/mod.rs
@@ -5,11 +5,15 @@ use anyhow::anyhow;
 use asn1_rs::{oid, FromDer, Integer, OctetString, Oid};
 use async_trait::async_trait;
 use base64::{engine::general_purpose::STANDARD, Engine};
+use http_cache_reqwest::{
+    MokaManager, MokaCacheBuilder, Cache, CacheMode, HttpCache, HttpCacheOptions,
+};
 use openssl::{
     nid::Nid,
     x509::{self, X509},
 };
-use reqwest::{get, Response as ReqwestResponse, StatusCode};
+use reqwest::{Response as ReqwestResponse, StatusCode};
+use reqwest_middleware::ClientBuilder;
 use serde;
 use serde_json::json;
 use sev::{
@@ -19,10 +23,17 @@ use sev::{
         host::{CertTableEntry, CertType},
     },
 };
-use std::{collections::HashMap, hash::Hash, result::Result::Ok, sync::LazyLock};
+use std::{
+    collections::HashMap,
+    hash::Hash,
+    result::Result::Ok,
+    sync::{LazyLock, OnceLock},
+};
 use strum::{Display, EnumIter, EnumString, IntoEnumIterator};
 use tracing::{debug, instrument, warn};
 use x509_parser::prelude::*;
+
+use std::time::Instant;
 
 #[derive(Serialize, Deserialize)]
 pub struct SnpEvidence {
@@ -88,8 +99,129 @@ pub(crate) static CERT_CHAINS: LazyLock<HashMap<ProcessorGeneration, VendorCerti
         map
     });
 
-#[derive(Default, Debug)]
-pub struct Snp {}
+static VCEK_CACHE_MANAGER: OnceLock<MokaManager> = OnceLock::new();
+
+fn init_cache_manager() -> MokaManager {
+    MokaManager::new(MokaCacheBuilder::new(1024).build())
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct Snp {
+}
+
+impl Snp {
+    pub async fn new() -> Result<Self> {
+        Ok(Snp {})
+    }
+
+    pub fn build_vcek_client(&self) -> reqwest_middleware::ClientWithMiddleware {
+        let client_options = HttpCacheOptions {
+            cache_status_headers: true,
+            ..Default::default()
+        };
+
+        let manager = VCEK_CACHE_MANAGER
+            .get_or_init(init_cache_manager)
+            .clone();
+
+        let cache = Cache(HttpCache {
+            mode: CacheMode::Default,
+            manager,
+            options: client_options,
+        });
+
+        ClientBuilder::new(reqwest::Client::new()).with(cache).build()
+    }
+
+    /// Asynchronously fetches the VCEK from the Key Distribution Service (KDS) using the provided attestation report.
+    /// Returns the VCEK in DER format.
+    async fn fetch_vcek_from_kds(
+        &self,
+        att_report: AttestationReport,
+        proc_gen: ProcessorGeneration,
+    ) -> Result<Vec<u8>> {
+        // Use attestation report to get data for URL
+        if att_report.chip_id.as_slice() == [0; 64] {
+            bail!("Hardware ID is 0s on attestation report. Confirm that MASK_CHIP_ID is set to 0 to request VCEK from KDS.");
+        }
+
+        let hw_id = match proc_gen {
+                ProcessorGeneration::Turin => {
+                    let shorter_bytes: &[u8] = &att_report.chip_id[0..8];
+                    hex::encode(shorter_bytes)
+                }
+                _ => hex::encode(att_report.chip_id),
+            };
+
+        // Request VCEK from KDS
+        let vcek_url: String = match proc_gen {
+            ProcessorGeneration::Turin => {
+                let Some(fmc) = att_report.reported_tcb.fmc else {
+                    bail!("A Turin processor must have a fmc value");
+                };
+
+                format!(
+                    "{KDS_CERT_SITE}{KDS_VCEK}/{}/\
+                    {hw_id}?fmcSPL={:02}&blSPL={:02}&teeSPL={:02}&snpSPL={:02}&ucodeSPL={:02}",
+                    proc_gen,
+                    fmc,
+                    att_report.reported_tcb.bootloader,
+                    att_report.reported_tcb.tee,
+                    att_report.reported_tcb.snp,
+                    att_report.reported_tcb.microcode
+                )
+            }
+            _ => {
+                format!(
+                    "{KDS_CERT_SITE}{KDS_VCEK}/{}/\
+                    {hw_id}?blSPL={:02}&teeSPL={:02}&snpSPL={:02}&ucodeSPL={:02}",
+                    proc_gen,
+                    att_report.reported_tcb.bootloader,
+                    att_report.reported_tcb.tee,
+                    att_report.reported_tcb.snp,
+                    att_report.reported_tcb.microcode
+                )
+            }
+        };
+
+        let client = self.build_vcek_client();
+        let start = Instant::now();
+
+        let vcek_rsp: ReqwestResponse = client
+            .get(vcek_url.clone())
+            .send()
+            .await
+            .context("Unable to send request for VCEK")?;
+
+        let duration = start.elapsed();
+
+        // Log cache status if available, otherwise log direct fetch
+        if let Some(cache_status) = vcek_rsp.headers().get("x-cache") {
+            debug!(
+                "VCEK fetch: {:?} from {:?} ({:?})",
+                cache_status, vcek_url, duration
+            );
+        } else {
+            debug!(
+                "VCEK fetch: direct from KDS {:?} ({:?})",
+                vcek_url, duration
+            );
+        }
+
+        match vcek_rsp.status() {
+            StatusCode::OK => {
+                let vcek_rsp_bytes: Vec<u8> = vcek_rsp
+                    .bytes()
+                    .await
+                    .context("Unable to parse VCEK")?
+                    .to_vec();
+                Ok(vcek_rsp_bytes)
+            }
+
+            status => bail!("Unable to fetch VCEK from URL: {status:?}, {vcek_url:?}"),
+        }
+    }
+}
 
 #[derive(Clone, Debug)]
 pub(crate) enum VendorEndorsementKey {
@@ -221,7 +353,8 @@ impl Verifier for Snp {
             // No certificate chain provided, so we need to request the VCEK from KDS
             _ => {
                 // Get VCEK from KDS
-                let vcek_buf = fetch_vcek_from_kds(report, proc_gen.clone())
+                let vcek_buf = self
+                    .fetch_vcek_from_kds(report, proc_gen.clone())
                     .await
                     .context("Failed to fetch VCEK from KDS")?;
                 let vcek = Certificate::from_bytes(&vcek_buf)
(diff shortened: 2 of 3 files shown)


attestation-service/docs/amd-offline-certificate-cache.md
```diff
@@ -0,0 +1,247 @@
...
+ COPY --from=builder /usr/local/cargo/bin/grpc-as /usr/local/bin/grpc-as
++COPY ./attestation-service/vcek_cache /etc/test-vcek-cache
+```
+
+In the above example, the `cache-preloader` tool's target directory was `../../attestation-service/vcek_cache`:
+```
+cargo run -- -u test_urls.txt -c ../../attestation-service/vcek-cache
+```
+
+And `as-config.json` sets `cache_dir` to `/etc/test-vcek-cache`.
```
fitzthum: Not totally canonical to put a cache in `/etc` imo. Also I'm not sure that baking the cached certs into the image itself really makes sense, but I guess it's fine.

deps/verifier/src/snp/mod.rs
```diff
@@ -89,7 +94,187 @@ pub(crate) static CERT_CHAINS: LazyLock<HashMap<ProcessorGeneration, VendorCerti
     });
 
 #[derive(Default, Debug)]
-pub struct Snp {}
+pub struct Snp {
+    verifierconfig: SnpVerifierConfig,
```
fitzthum: nit: `verifierconfig` -> `verifier_config`

tools/cache-preloader/README.md
```diff
@@ -0,0 +1,48 @@
...
+```
+# VCEK URLs for Milan processors
+https://kdsintf.amd.com/vcek/v1/Milan/abc123?blSPL=03&teeSPL=00&snpSPL=08&ucodeSPL=115
+https://kdsintf.amd.com/vcek/v1/Milan/def456?blSPL=03&teeSPL=00&snpSPL=08&ucodeSPL=115
+
+# VCEK URLs for Genoa processors
+https://kdsintf.amd.com/vcek/v1/Genoa/789abc?blSPL=02&teeSPL=00&snpSPL=03&ucodeSPL=209
+```
+
+Lines starting with `#` are treated as comments and ignored.
```
fitzthum: Not necessary.

tools/cache-preloader/example_urls.txt
```diff
@@ -0,0 +1,14 @@
+# Example VCEK URLs for preloading
+# Lines starting with # are comments and will be ignored
```
fitzthum: Not necessary.

tools/cache-preloader/README.md
fitzthum, line 1: I guess this tool just deals with URLs so it is not specific to AMD even though the SNP verifier is the only place we use this type of caching so far. So I think it's fine to include this here.

attestation-service/docs/amd-offline-certificate-cache.md
amd-aliem: That's true, I'll update to outside of etc. & yeah I was thinking it was a way to not have to copy it upon every restart, but maybe a volume mount would make more sense. Something like this in docker-compose.yml?
```
$ git diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 3e4a72b..8fe2656 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -39,6 +39,7 @@ services:
     - ./kbs/config/as-config.json:/etc/as-config.json:rw
     - ./kbs/config/sgx_default_qcnl.conf:/etc/sgx_default_qcnl.conf:rw
     - ./kbs/config:/opt/confidential-containers/kbs/user-keys
+    - ./tools/cache-preloader/vcek_cache:/vcek_cache:rw
     command: [
       "grpc-as",
       "--socket",
```

fitzthum: Yeah I think that would be the most common way to set it up

amd-aliem: fixed! removed references to /etc and replaced section on building into image w/ shared directory

deps/verifier/src/snp/mod.rs
amd-aliem: fixed

tools/cache-preloader/README.md
amd-aliem: removed

attestation-service/docs/amd-offline-certificate-cache.md
```diff
@@ -0,0 +1,246 @@
...
+```json
+{
+    // ... other fields ...
+    "verifier_config": {
+        "snp_verifier": {
+            "cache": {
+                "type": "HttpCacheReqwest",
+                "offline_mode": true,
+                "cache_dir": "/path/to/cache"
+            }
```
mythi: following my comment in the other cache PR: could we try to make the cache (config) design such that it's not TEE specific. I could retrofit TDX/SGX CollateralService setup under the same Trustee configs

the default for `cache_dir` should probably be aligned with the base path discussion (#992) . 

need any cache bust settings to check for fresh content every now and then?

amd-aliem: Sure, how about:
- Rename `VCEKCacheConfig` -> `ClientCacheConfig` and move into deps/verifier/src/lib.rs, for TEEs to add to their config as needed. 
- Rename field `offline_mode` -> `cache_mode`
  - In this HttpCacheReqwest type, would just mirror HttpCacheReqwest enums: https://docs.rs/http-cache-reqwest/latest/http_cache_reqwest/enum.CacheMode.html
- Update default for `cache_dir` to `/opt/confidential-containers/client-cache` ? Or do you mean we'll want to utilize the base path code when that's merged? (or both haha)

Would the cache_bust settings be like a set of URLs that we disallow from caching?

amd-aliem: Ah I see the comment from @mkulke about just enabled/disabled instead of surfacing options - reviewing that with my team

fitzthum:
fwiw I don't think the distinction between caching and "pre-loading" makes much sense. We're going to add code that checks a cache but then later we're going to add code that first checks a pre-load store before checking the cache? Those are the same thing. Caching is a general term. I think it's totally reasonable to pre-set the cache with some values.

There is some feedback here asking for the cache to be more robust (i.e. support TTL), more generic, and not require external tooling (just put the artifacts in a config file). I made a PR that did these things, but the feedback was that it was too generic.

Either way, let's settle on something because users are asking for this feature.

deps/verifier/src/snp/mod.rs
```diff
@@ -89,7 +94,187 @@ pub(crate) static CERT_CHAINS: LazyLock<HashMap<ProcessorGeneration, VendorCerti
...
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[serde(tag = "type")]
+#[derive(Default)]
+pub enum VCEKCacheConfig {
+    #[default]
+    None,
+
+    /* HTTP cache using reqwest with cacache backend (disk)
+    Fields: offline_mode, cache_dir
+    WARNING: cacache has NO automatic eviction */
```
Xynnn007:
```suggestion
    /// HTTP cache using reqwest with cacache backend (disk)
    /// Fields: offline_mode, cache_dir
    /// WARNING: cacache has NO automatic eviction
```

deps/verifier/src/snp/mod.rs
```diff
@@ -89,7 +94,187 @@ pub(crate) static CERT_CHAINS: LazyLock<HashMap<ProcessorGeneration, VendorCerti
...
+}
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+pub struct SnpVerifierConfig {
+    #[serde(default)]
+    vcek_cache: VCEKCacheConfig,
+}
+
+fn default_cache_dir() -> PathBuf {
+    PathBuf::from("./cache")
```
Xynnn007: Could we use an absolute path than a relative one? As the relative one can be usually hard to get which is the base. A possible path can be under `/opt/amd/vceks` or somewhere else

amd-aliem: Sure! This is similar to request - https://github.com/confidential-containers/trustee/pull/1061#discussion_r2536424640
I think in that PR the root base is "/opt/confidential-containers/" - @mythi is defaulting to `/opt/confidential-containers/vcek-cache` what you meant here?

amd-aliem:
Hey @mkulke @fitzthum @mythi,

@Xynnn007 has kindly offered to help me out with a follow-on PR after this one. Just wanted to align on direction to make sure we're ok with that - reiterating on my comment https://github.com/confidential-containers/trustee/pull/1061#issuecomment-3554346647:
1. Do we want to continue with this implementation method? Option to revisit the more generic cache that @fitzthum implemented in https://github.com/confidential-containers/trustee/pull/989
2. If so, should the pre-loader tool be moved into a subcommand of `trustee-cli`? Would it be preferable to do that here or in a follow-up PR?

I'm guessing there's not a commitment to not breaking configs here if it doesn't go into a release yet, but wanted to make sure we're not stuck supporting the format in this PR if something changes.

@Xynnn007 also shared a diagram - here's a bit more detailed one that I had created for my team last week with the current implementation:
image
Where middleware can be swapped out for any other certificate fetching function in the future. You can see the main draw here in my opinion is that the same URL can be used to fetch the same certs, and the certs are all managed in the same place. Plus resilience to network disruption even without the user preloading certs, with the correct settings.

deps/verifier/src/snp/mod.rs
mythi: > @mythi is defaulting to `/opt/confidential-containers/vcek-cache` what you meant here?

I didn't have any specific path in mind. We have used `/opt/confidential-containers/` so that'd be consistent with what is there today.

amd-aliem:
Hey @mythi, ok I can remove all of the offline references for now and then open another PR after this one.

And I can make this fully transparent so it's always enabled, but should I still use cacache as the manager? 
- In current impl I'm using the `ForceCache` setting ([docs](https://docs.rs/http-cache-reqwest/latest/http_cache_reqwest/enum.CacheMode.html)). I'm not certain this will pay attention to `Cache-Control: no-cache`. I'll see if I can test that.
- I'm using ForceCache because when using the `Default` setting w/ cacache, cache elements appear to only be fresh for a few seconds, yet stale elements are never evicted from the filesystem. In addition to the issue above, there's this issue from the parent crate also pointing that out: https://github.com/06chaynes/http-cache/issues/118. 
  - In testing I see a single vcek entry is about 12K, which could grow to over 1G in less than a month on a busy kbs, even with just one certificate being cached. Perhaps that's unreasonable to be concerned about?

Would it maybe be better to use the in-memory moka one for the 'transparent' cache? That one appears to support a max cache size, and we can set it appropriately small to not affect users. 

I would still like to use cacache in the next PR to allow the user to set a config to enable it, for example wanting a persistent cache, wanting offline support, etc.

amd-aliem:
Implemented the in-memory option - performance is worse than previous filesystem cache implementation, now that it's adhering to the `Cache-Control: no-cache` directive to still confirm validity against KDS if it's found in the cache. However still a significant performance gain and I'm not hitting 429s anymore.

Some example timing results:
```
2025-12-15T22:04:53.865019Z DEBUG attestation_evaluate{request_id=115336be-6d4c-40e1-bf86-33e2f968472a}:AMD SNP: verifier::snp: VCEK fetch: "MISS" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (1.260074931s)
2025-12-15T22:04:54.201726Z DEBUG attestation_evaluate{request_id=654e59e3-61df-4cdc-8d77-97906282920a}:AMD SNP: verifier::snp: VCEK fetch: "HIT" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (231.777146ms)
2025-12-15T22:04:54.471756Z DEBUG attestation_evaluate{request_id=a7363ebb-5415-4b17-9853-4aaf59932fb4}:AMD SNP: verifier::snp: VCEK fetch: "HIT" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (228.408276ms)
2025-12-15T22:05:18.769272Z DEBUG attestation_evaluate{request_id=2e11e642-eb7e-4605-989f-bfd3d119ce29}:AMD SNP: verifier::snp: VCEK fetch: "MISS" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (1.575616906s)
2025-12-15T22:05:19.049419Z DEBUG attestation_evaluate{request_id=9b5356f3-8a32-4d3c-baf4-948e0794c8ff}:AMD SNP: verifier::snp: VCEK fetch: "HIT" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (239.98664ms)
2025-12-15T22:05:19.326548Z DEBUG attestation_evaluate{request_id=0f85a344-a13f-47fa-bc84-4d581695873a}:AMD SNP: verifier::snp: VCEK fetch: "HIT" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (236.823039ms)
2025-12-15T22:05:23.748519Z DEBUG attestation_evaluate{request_id=c60a06bb-bf1f-4f43-812c-26899c456bd2}:AMD SNP: verifier::snp: VCEK fetch: "HIT" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (241.359907ms)
2025-12-15T22:05:24.022639Z DEBUG attestation_evaluate{request_id=a595d4bd-bae9-4b0d-a2f8-8a06790c2843}:AMD SNP: verifier::snp: VCEK fetch: "HIT" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (233.20828ms)
2025-12-15T22:05:24.297647Z DEBUG attestation_evaluate{request_id=6cee5775-5006-4c29-9e62-f481ac093d24}:AMD SNP: verifier::snp: VCEK fetch: "HIT" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (236.815188ms)
2025-12-15T22:09:22.030739Z DEBUG attestation_evaluate{request_id=6ae3ee65-b9fb-4fcd-854d-063e2466760a}:AMD SNP: verifier::snp: VCEK fetch: "MISS" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (1.36134411s)
2025-12-15T22:09:22.300330Z DEBUG attestation_evaluate{request_id=ed8ba02f-32b0-4773-9f19-d3b13f942cbb}:AMD SNP: verifier::snp: VCEK fetch: "HIT" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (230.787416ms)
2025-12-15T22:09:22.574272Z DEBUG attestation_evaluate{request_id=9d88cdfa-c967-4b58-9049-5c88dcfca76c}:AMD SNP: verifier::snp: VCEK fetch: "HIT" from "https://kdsintf.amd.com/vcek/v1/Genoa/f6fefdd6c54410ab3a335b9318d9567d9486b980c43bf9696f13758584a07aa527f4ef762ccac1b6e9da168c31b9d6129ee58255060aa9761af301882da8e2e2?blSPL=10&teeSPL=00&snpSPL=27&ucodeSPL=25" (234.799584ms)
```

If we change from Default to "IgnoreRules" it goes down to microseconds, but then it's not technically honoring the server's directive. I heard from my team that certs are not really revoked often, so it may be reasonable to ignore and manually eject a cache entry after some number of seconds or minutes?

deps/verifier/Cargo.toml
```diff
@@ -80,6 +80,8 @@ ccatoken = { git = "https://github.com/veraison/rust-ccatoken", rev = "6d5b8db",
 ear = { version = "0.4.0", optional = true }
 x509-parser = { version = "0.18.0", optional = true }
 reqwest.workspace = true
+reqwest-middleware = "0.4.2"
+http-cache-reqwest = { version = "0.16.0", features = ["manager-moka"] }
```
mythi: maybe disable `cacache`:
```suggestion
http-cache-reqwest = { version = "0.16.0", default-features = false, features = ["manager-moka"] }
```

deps/verifier/src/snp/mod.rs
```diff
@@ -88,8 +99,130 @@ pub(crate) static CERT_CHAINS: LazyLock<HashMap<ProcessorGeneration, VendorCerti
...
+                    proc_gen,
+                    att_report.reported_tcb.bootloader,
+                    att_report.reported_tcb.tee,
+                    att_report.reported_tcb.snp,
+                    att_report.reported_tcb.microcode
+                )
+            }
+        };
+
+        let client = self.build_vcek_client();
```
mythi, line 187: would there be any benefits in re-using the client between queries?

amd-aliem: I wasn't sure how to handle that w/ multi-threading - wouldn't want a >1s network request to block other requests

deps/verifier/Cargo.toml
amd-aliem: fixed!

deps/verifier/src/snp/mod.rs
```diff
@@ -88,8 +99,130 @@ pub(crate) static CERT_CHAINS: LazyLock<HashMap<ProcessorGeneration, VendorCerti
...
+
+    /// Asynchronously fetches the VCEK from the Key Distribution Service (KDS) using the provided attestation report.
+    /// Returns the VCEK in DER format.
+    async fn fetch_vcek_from_kds(
+        &self,
+        att_report: AttestationReport,
+        proc_gen: ProcessorGeneration,
+    ) -> Result<Vec<u8>> {
+        // Use attestation report to get data for URL
+        let hw_id: String = if att_report.chip_id.as_slice() != [0; 64] {
```
mkulke: nit: we can use an early return, since we bail:

```
if att_report.chip_id.as_slice() == [0; 64] {
  bail!(...);
}
let hw_id = match ...
```

deps/verifier/src/snp/mod.rs
```diff
@@ -88,8 +99,130 @@ pub(crate) static CERT_CHAINS: LazyLock<HashMap<ProcessorGeneration, VendorCerti
...
+        };
+
+        // Request VCEK from KDS
+        let vcek_url: String = match proc_gen {
+            ProcessorGeneration::Turin => {
+                let fmc = if let Some(fmc) = att_report.reported_tcb.fmc {
+                    fmc
+                } else {
+                    bail!("A Turin processor must have a fmc value");
+                };
```
mkulke: nit: same here. we can do let/else and early return:

```
let Some(fmc) = att_report.reported_tcb.fmc else {
  bail!(...)
}
```

amd-aliem: fixed!

deps/verifier/src/snp/mod.rs
amd-aliem: fixed!

Issue #1056, opened 2025-11-13
ccel fails during attestation.


wangkpot:
### Describe the bug

 when using the project confidential-containers/guest-components/tree/main/attestation-agent to extend rtmr_3 and obtain evidence, the value of rtmr_3 is not written into the CCEL. As a result, attestation fails during verification.

### How to reproduce

 when using the project confidential-containers/guest-components/tree/main/attestation-agent to extend rtmr_3 and obtain evidence, the value of rtmr_3 is not written into the CCEL. As a result, attestation fails during verification.

### CoCo version information

trustee: main

### What TEE are you seeing the problem on

None

### Failing command and relevant log output

```shell

```

mythi:
> to extend rtmr_3 and obtain evidence, the value of rtmr_3 is not written into the CCEL. As a result, attestation fails during verification.

Do you have any logs to share what the error is? How do you extend RTMR3 and do you have eventlog enabled in AA config?

mythi:
CoCo guest-components and guest kernel versions?

wangkpot:
> > 1. error log: "CCEL eventlog does not pass RTMR [4] check", Actually, the actual location where the error occurs is rmtr[3].
> >    2.I use AA interface "AttestationAgentService.ExtendRuntimeMeasurement" to  extend RTMR3, and i enabled in AA config file.
> 
> Oh. This error message is a bug. It should be `"CCEL eventlog does not pass CCMR [4] check"`. The mapping between CCMR and RTMR is [here](https://uefi.org/specs/UEFI/2.10/38_Confidential_Computing.html#intel-trust-domain-extension).
> 
> However, it is unreasonable for the integrity check to still fail. Please also provide the AS version along side with [@mythi](https://github.com/mythi) mentioned.



There is no CCMR [4] in [document](https://uefi.org/specs/UEFI/2.10/38_Confidential_Computing.html#intel-trust-domain-extension).

And in [here](https://github.com/confidential-containers/trustee/blob/458a0603768117132d15139ea34e0f7006224da1/deps/verifier/src/tdx/mod.rs#L92C2-L119C19), the correspondence between CCMR and RTMR is inconsistent with the documentation you provided.

My AS version is on the main branch, with the latest commit ID being https://github.com/confidential-containers/trustee/commit/beb4ce9346bc7f9a7ff0f686eb868472767ad8fb, which dates back to approximately five months ago.

wangkpot:
[...]

PR #1068, opened 2025-11-19
eventlog: fix error message for mixing CCMR and RTMR

Xynnn007:
Eventlog cares about the CC Event Log Measurement Register which is not the same as Runtime Measurement Register, but the two has some relationship depending on the underlying platform.

The mapping can be found here

https://uefi.org/specs/UEFI/2.10/38_Confidential_Computing.html#vendor-specific-information

related to #1056 

cc @pawelpros

### DIFF
--- a/deps/eventlog/src/lib.rs
+++ b/deps/eventlog/src/lib.rs
@@ -72,6 +72,12 @@ pub struct EventDetails {
 
 #[derive(Debug, Clone)]
 pub struct ReferenceMeasurement {
+    /// The index of the CC Event Log Measurement Register.
+    /// Different TEE platform may have platform specific measurement
+    /// registers to map to the CC Event Log Measurement Register.
+    ///
+    /// See <https://uefi.org/specs/UEFI/2.11/38_Confidential_Computing.html#vendor-specific-information>
+    /// for platform specific mapping.
     pub index: u32,
     pub algorithm: TcgAlgorithm,
     pub reference: Vec<u8>,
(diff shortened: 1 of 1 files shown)


deps/eventlog/src/lib.rs
```diff
@@ -117,12 +117,12 @@ impl CcEventLog {
         let digest_map = collect_digests_by_index(&self.log);
 
         for item in data.iter() {
-            let calculated_ccel_rtmr = replay(&digest_map, item.index, item.algorithm)?;
-            if calculated_ccel_rtmr != item.reference {
+            let calculated_ccel_ccmr = replay(&digest_map, item.index, item.algorithm)?;
+            if calculated_ccel_ccmr != item.reference {
                 bail!(
-                    "CCEL eventlog does not pass RTMR [{}] check. CCEL value: {}, Quote value: {}",
+                    "CCEL eventlog does not pass CCMR [{}] check. CCEL value: {}, Quote value: {} \nThe relation ship between CCMR and RTMR can be found in https://uefi.org/specs/UEFI/2.10/38_Confidential_Computing.html#vendor-specific-information",
```
mythi: But the error is not related to vendor mapping here, is it? You could extend an MR without writing a log entry for it and it'll also fail here.

Xynnn007: You're correct. The reason I'm updating this error information isn't because it's vendor-related, but because the original error used the term "RTMR," which is vendor-specific. Specifically, only TDX and CSV now use RTMR to indicate some specific registers in the remote attestation context. The register originally linked to Eventlog integrity in the UEFI standard documentation should be CCMR. Different platforms have different hardware implementations aligned with CCMR. I'm providing a link here so users can use the CCMR number to find the corresponding register for their platform to troubleshoot the problem.

mythi: Maybe I misunderstood that this refers to that PCR to MR "relationship" / mapping. Could we just say MR or measurement register? For example the kernel uses TDX measurement register (MR) wording.

Xynnn007: The relationship mentioned here is "CCMR" to platform-specific RTMR, rather than PCR to "CCMR". Sorry I did not get your point quite well. Do you mean that we should rename "RTMR" to "MR"?

mythi: > Do you mean that we should rename "RTMR" to "MR"?

basically yes. Or some other re-wording like "Eventlog does not pass measurement replay [index={]] .." to make it fully acronym free.

pawelpros: Hmm I cannot find useful docs beside: https://trustedcomputinggroup.org/wp-content/uploads/TCG-PC-Client-Platform-Firmware-Profile-Version-1.06-Revision-52_pub-3.pdf 3.3.4 PCR Usage - when I was implementing this feature. I remember that somewhere was quite nice description which pointed that PCR's are common cross vendor but for any of them "understanding" what is stored where is up to vendor that's why in this (related to Intel and CSV) method:

```
pub fn replay_and_match(&self, data: Vec<ReferenceMeasurement>) -> Result<()> {
```

I used vector map which maps PCR to RMTR.

I don't think that pinning exactly Intel related link in description is best way - maybe in documentation or comment rather than expection ?

Xynnn007: I added a comment upon the `index` field of `ReferenceMeasurement` struct.

PR #1079, opened 2025-11-24
Add extension support to Attestation Service policy

fitzthum:
Allow extensions to be specified via the AS policy. These will be add to the EAR token.

This will make it a little more complicated to merge our policy engines together, but this is not the first/only extension that we are using.


attestation-service/src/policy_engine/mod.rs
```diff
@@ -75,7 +75,10 @@ type PolicyDigest = String;
 
 #[derive(Debug)]
 pub struct EvaluationResult {
-    pub rules_result: Value,
+    pub trust_claims: Value,
+    /// Extensions to be added to the attestation token.
+    /// The extension name, extension index, and the Value.
+    pub extensions: Vec<(String, i32, Value)>,
```
Xynnn007: Can we define a single struct for this? btw, could we provide an example rego to include the extension, in either document or a unit test?

fitzthum: Yes, I can make it a struct.

I have some tests locally. They introduce some additional logic with identifiers so I was going to leave that to the next PR, but I can add them here. I will update the docs in this PR at least.

fitzthum: Btw would you prefer the return value to the policy to also be a struct? As in the policy returns a JSON object with `{"extension_name": "foo", "extension_index": -245, "extensio_value" : {}}`. Currently the policy represents an extension as `["food", -245, {}]`

Xynnn007: Yes it's a good idea. The way in my mind (half finished) is that the [policy module](https://github.com/confidential-containers/trustee/blob/main/deps/policy-engine/src/policy/rego.rs#L13) will try to parse the only result `"data.policy.result"`. Two ways we can make the returned value in reg, s.t.

## way 1: put along with other AR4SI vectors

```
result := {
   "file_system": 2,
   "executables": 2,
    ...
   "extensions": {
       ...
   }
}
```

## way 2: have single claims

```
result := {
   "tiers": {
      "file_system": 2,
      "executables": 2,
       ...
   },
   "extensions": {
       ...
   }
}
```

Looks like way 2 is better?

fitzthum: Ok, I changed to a struct and updated the docs. 

I prefer having `trust_claims` and `extensions` as two totally separate claims.

attestation-service/docs/policy.md
```diff
@@ -19,7 +19,23 @@ These dimensions can be used to return a detailed, but generic description of th
...
+Any extensions that are specified via this claim will be added to the attestation token.
+The extensions claim should contain an array of extension objects.
+An extension object should look like this:
+```json
+{
+    "name": "<name-of-extension">,
+    "index": "<index-of-extension">,
+    "value": ...
+}
+```
```
Xynnn007, line 32: There are two design patterns here. One is to allow generic extensions to be added to the AS policy, which requires the user to know the index and name. The advantage is strong extensibility, but the disadvantage is that it's cumbersome for users to write the policy and they also need to know which extensions are available.

The other pattern is to register an extension, named something like ext.confidential-containers-claims or ext.trustee-claims, and the user only needs to enter any value. The code of AS hardcodes the name and index. The advantage is that it's convenient for users to write, but the disadvantage is that it limits the scope of the extension.

Both makes sense and probably both can survive together.

fitzthum: Yeah adding some built-in extensions could make sense in the future, but let's start with this and see what people use.

attestation-service/docs/policy.md
```diff
@@ -19,7 +19,23 @@ These dimensions can be used to return a detailed, but generic description of th
...
-AS policies will be provide `input`.
+A policy can also optionally define a claim called `extensions`.
+Any extensions that are specified via this claim will be added to the attestation token.
+The extensions claim should contain an array of extension objects.
+An extension object should look like this:
+```json
+{
+    "name": "<name-of-extension">,
+    "index": "<index-of-extension">,
+    "value": ...
```
mkulke:
```suggestion
    "name": "<name-of-extension>",
    "index": "<index-of-extension>",
    "value": ...
```

does the index represent a ClaimKey?

fitzthum: Updated. Not totally clear from the spec if these are claim keys in the JWT sense. It seems to outline additional key claims in section 9, which doesn't include anything about extensions.

btw the crate calls this value a key rather than an index, so let me change our terminology here.

Xynnn007: This used for serialization of EAR token. The `name` is for JSON ([human readable](https://github.com/veraison/rust-ear/blob/main/src/ear.rs#L475)) and `key` is for like CBOR ([not human readable](https://github.com/veraison/rust-ear/blob/main/src/ear.rs#L491)).

Actually the `key` field might not be used in our case as we only serialize into JWT format than CWT. We could somehow hide the `key` field in the policy extension.

fitzthum: I considered using random values for the extensions, but that didn't seem like a good idea. Probably best to give the crate what it asks for even though we don't use CBOR.

Xynnn007: I understand the need to stay consistent with the lower-level library. Personally I’d prefer not to expose parameters that don’t have any effect, as they can be confusing to users. If we keep it, let’s document clearly that this parameter is currently unused and only kept for compatibility, so users know not to rely on it.

PR #1086, opened 2025-12-01
Replace derivative dependency with educe

Cropi:
This commit replaces the usage of the derivative crate with educe across the entire workspace. The derivative crate is no longer actively maintained [1] (upstream is effectively dead - distros prefer not to ship unmaintained dependencies), whereas educe provides a modern, maintained, and feature-rich alternative for deriving traits with custom behavior. This does not affect the functionality.

https://github.com/mcarton/rust-derivative/issues/117

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1861,17 +1861,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "derivative"
-version = "2.2.0"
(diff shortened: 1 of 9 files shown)


mythi:
also related is RUSTSEC-2024-0388 which is shown on our scorecard. perhaps add a reference to it in the commit message?

PR #1089, opened 2025-12-01
tests: tweak integration test cleanup

fitzthum:
Update the init_data test to test the result after calling the cleanup code. This way if the result is bad the test harness will still be cleaned up and future tests will work.

There are still some return paths before the cleanup code, but they are unlikely to be triggered, so for now just make this simple adjustment.

### DIFF
--- a/integration-tests/tests/init_data.rs
+++ b/integration-tests/tests/init_data.rs
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use base64::{engine::general_purpose::STANDARD, Engine};
 use log::info;
 use serial_test::serial;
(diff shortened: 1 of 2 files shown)


integration-tests/tests/init_data.rs
```diff
@@ -45,11 +45,15 @@ async fn get_resource_with_init_data_hash() -> Result<()> {
...
+    let res = harness
         .get_secret(SECRET_PATH.to_string(), Some(SIMPLE_INIT_DATA.to_string()))
-        .await?;
+        .await;
 
     harness.cleanup().await?;
+
+    if res? != SECRET_BYTES {
+        bail!("Secret retrieved, but secret has wrong value");
+    }
```
Xynnn007, line 56: Let me understand. This means that in any cases, the harness needs to be cleaned up once it is not used anymore. Is it possible to make `cleanup` the destructor ([Drop](https://doc.rust-lang.org/rust-by-example/trait/drop.html)) of the harness object, so that it's automatically cleaned up whenever the object's lifecycle ends, avoiding the need for us to be overly concerned with when to explicitly call `cleanup`?

fitzthum: Drop doesn't work with async (can't put async code in `Drop`, which is what we would need). There are some more complex workarounds that I am looking at, but this is an easy improvement in the short term.

Xynnn007: Ok. I did not notice that `cleanup` is async 😅

PR #1098, opened 2025-12-04
Kbs/fix resource path policy

### LINKED ISSUE #1097: docs and testing is wrong for KBS resource policy `resource-path`
### Describe the bug

This issue is primarily about a mismatch in all our docs vs KBS (resource) policy evaluation for `resource-path`. With the move to the plugin model, the `resource-path` passed to `evaluate()` takes:

```rust
let endpoint = format!("{base_path}{additional_path}"); 
```

This means that with `path := split(data["resource-path"], "/")`, we get:

```
path[0] == "resource"
```

and the "three segment path format:`<TOP>/<MIDDLE>/<TAIL>`"  follows starting from `path[1]`.

The gap with testing is that while we have a lot of tests running `evaluate()` with `resource-path` set and policies written for it, they follow the old path structure which is not what API server users get.

Suggested fixes:
- update all our docs to `resource/<TOP>/<MIDDLE>/<TAIL>` model
- add e2e to test that structure
- update unit tests that also reflect this new `endpoint` structure
- (update the docs to highlight that plugins can use `path[0]` to identify themselves(?))
- (extra: cleanup `tcb-status` from .rego files since it's from Simple token)

### How to reproduce

Pick any policy that tests `path[0] == <TOP>`. This gives `PolicyDeny`.

### CoCo version information

trustee v0.15.0

None

```shell

```

Xynnn007:
Fixes #1097 

Please see concrete commit messages

~~User interface change:~~
~~1. rename all kbs resource policy to kbs request policy~~
~~2. add a new `set-request-policy` subcommand for kbs-client, also for compability keep `set-resource-policy`. The two have same behavior~~

### DIFF
--- a/kbs/config/kubernetes/ita/policy.rego
+++ b/kbs/config/kubernetes/ita/policy.rego
@@ -35,7 +35,7 @@
 #    ```
 #    {
 #        "plugin": <plugin-name>,
-#        "resource-path": </.../<END>>,
+#        "resource-path": [<...>, <sections>, <END>],
 #        "query": {
 #            "a": "b",
 #            ...
@@ -47,7 +47,7 @@
 #    ```
 #    {
 #        "plugin": "resource",
-#        "resource-path": "/myrepo/License/key",
+#        "resource-path": ["myrepo", "License", "key"],
 #        "query": {}
 #    }
 #    ```
@@ -55,7 +55,7 @@
 #    ```
 #    {
 #        "plugin": "plugin1",
-#        "resource-path": "/para/meters",
+#        "resource-path": ["para", "meters"],
 #        "query": {
 #           "version": "1.0.0"
 #        }	
@@ -63,7 +63,7 @@
 #    ```
 #
 #    For the "resource" plugin specifically:
-#      - resource-path format: /<repo>/<type>/<tag>
+#      - resource-path format: three items slice
 #      - query: {}
 #
 # Policy Rules
--- a/kbs/src/api_server.rs
+++ b/kbs/src/api_server.rs
@@ -187,12 +187,11 @@ pub(crate) async fn api(
     // if the path parts is equal to 1, return an empty vector
     let plugin = path_parts[0];
 
-    let trailing_resource_path = match &path_parts[..] {
+    let resource_path = match &path_parts[..] {
         [_, rest @ ..] => rest,
         _ => &[],
-    }
-    .join("/");
-    let resource_path = format!("/{}", trailing_resource_path);
+    };
+
     let query = query.into_inner();
     let policy_data = json!(
         {
@@ -228,10 +227,10 @@ pub(crate) async fn api(
         // GET /reference-value/<reference_value_id>
         "reference-value" if request.method() == Method::GET => {
             core.admin.validate_admin_token(&request)?;
-            let reference_value_id = resource_path.trim_start_matches('/');
+            let reference_value_id = resource_path.join("/");
             let reference_values = core
                 .attestation_service
-                .query_reference_value(reference_value_id)
+                .query_reference_value(&reference_value_id)
                 .await
                 .map_err(|e| Error::RvpsError {
                     message: format!("Failed to get reference_values: {e}").to_string(),
@@ -288,14 +287,14 @@ pub(crate) async fn api(
 
             let body = body.to_vec();
             if plugin
-                .validate_auth(&body, &query, &resource_path, request.method())
+                .validate_auth(&body, &query, resource_path, request.method())
                 .await
                 .map_err(|e| Error::PluginInternalError { source: e })?
             {
                 // Plugin calls need to be authorized by the admin auth
                 core.admin.validate_admin_token(&request)?;
                 let response = plugin
-                    .handle(&body, &query, &resource_path, request.method())
+                    .handle(&body, &query, resource_path, request.method())
                     .await
                     .map_err(|e| Error::PluginInternalError { source: e })?;
 
--- a/kbs/src/policy_engine/opa/default_policy.rego
+++ b/kbs/src/policy_engine/opa/default_policy.rego
@@ -35,7 +35,7 @@
 #    ```
 #    {
 #        "plugin": <plugin-name>,
-#        "resource-path": </.../<END>>,
+#        "resource-path": [<...>, <sections>, <END>],
 #        "query": {
 #            "a": "b",
 #            ...
@@ -47,7 +47,7 @@
 #    ```
 #    {
 #        "plugin": "resource",
-#        "resource-path": "/myrepo/License/key",
+#        "resource-path": ["myrepo", "License", "key"],
 #        "query": {}
 #    }
 #    ```
@@ -55,7 +55,7 @@
 #    ```
 #    {
 #        "plugin": "plugin1",
-#        "resource-path": "/para/meters",
+#        "resource-path": ["para", "meters"],
 #        "query": {
 #           "version": "1.0.0"
 #        }	
@@ -63,7 +63,7 @@
 #    ```
 #
 #    For the "resource" plugin specifically:
-#      - resource-path format: /<repo>/<type>/<tag>
+#      - resource-path format: three items slice
 #      - query: {}
 #
 # Policy Rules
--- a/kbs/src/policy_engine/opa/mod.rs
+++ b/kbs/src/policy_engine/opa/mod.rs
@@ -173,7 +173,7 @@ mod tests {
         "test/data/policy_1.rego",
         json!({
             "plugin": "resource",
-            "resource-path": "/my_repo/Alice/key",
+            "resource-path": ["my_repo", "Alice", "key"],
             "query": {}
         }),
         "Alice",
@@ -184,7 +184,7 @@ mod tests {
         "test/data/policy_1.rego",
         json!({
             "plugin": "another-plugin",
-            "resource-path": "/my_repo/Alice/key",
+            "resource-path": ["my_repo", "Alice", "key"],
             "query": {}
         }),
         "Alice",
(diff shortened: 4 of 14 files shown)


kbs/src/policy_engine/opa/mod.rs
```diff
@@ -43,8 +43,8 @@ impl PolicyEngineInterface for Opa {
 
         // Add resource path as data
         let resource_path_object =
-            regorus::Value::from_json_str(&format!("{{\"resource-path\":\"{}\"}}", resource_path))
-                .map_err(|_| KbsPolicyEngineError::ResourcePathError)?;
+            regorus::Value::from_json_str(&format!("{{\"request_path\":\"{}\"}}", request_path))
+                .map_err(|_| KbsPolicyEngineError::RequestPathError)?;
```
mythi: we could do this but another idea would be to split the path here to what it was:

```
      {{
            "resource-path": "..."
            "plugin-path": "<plugin>"
        }}
```

Xynnn007: This would be a good plan, while `resource-path` is only for `resource` plugin. What about using
```json
{
  "plugin": "plugin",
  "parameter": "a/b/c",
  "quey": ["d=1", "e", "f=3"]
}
```
for 

`plugin/a/b/c?d=1&e&f=3`

kbs/docs/kbs.yaml
```diff
@@ -284,14 +295,14 @@ components:
...
-    ResourcePolicy:
+    Policy:
       required:
         - policy
       properties:
         policy:
           type: string
           description: >-
-            Base64 encoded resource distribution policy.
+            Base64 encoded resource distribution policy. (original ResourcePolicy object)
```
pawelpros:
```suggestion
           Base64-encoded policy for resource distribution, formerly referred to as ResourcePolicy.
```

kbs/config/kubernetes/ita/policy.rego
```diff
@@ -1,42 +1,52 @@
-# Resource Policy
+# KBS Request Policy
```
pawelpros:
```suggestion
# KBS Request Policy
```

kbs/src/policy_engine/opa/mod.rs
pawelpros: For me it's good idea (typo in "query")

mythi: I was thinking `resource-path` for backwards compatibility and smoother transition to whatever the new keys would be.

fitzthum: Yeah I think it's fine to keep it as resource path and even to keep referring to everything the KBS does as resources even if they're coming from plugins.

Xynnn007: Yes, under the semantics of RESTful, practically everything can be called a resource. I'll revert it to a resource policy, while retaining the fine-grained parameters introduced by the PR in the policy engine.

kbs/src/policy_engine/opa/default_policy.rego
```diff
@@ -1,41 +1,67 @@
...
+#    }
+#    ```
+#
+#    The original path string format is: <plugin-name>/.../<END>[?a=b&...&<QUERY>]
+#    Examples:
+#      - "resource/myrepo/License/key"
+#      - "<plugin-name>/para/meters?version=1.0.0"
+#
+#    For the "resource" plugin specifically:
+#      - subResourcePath format: /<repo>/<type>/<tag>
```
mythi: As far as I can see, `subResourcePath` is available to *all* plugins but the structure is not defined. Should we say: 
```suggestion
#      - "<plugin-name>/<subResourcePath>?version=1.0.0"
#
#    For the "resource" plugin specifically:
#      - subResourcePath must have format: /<repo>/<type>/<tag>
```

kbs/src/policy_engine/opa/mod.rs
fitzthum: If we want to rename something we might consider renaming the resource plugin.

kbs/src/policy_engine/opa/default_policy.rego
Xynnn007: It's defined on line 36. but I get your point to have a regular representation here. As `<subResourcePath>` starts with `/`, so `<plugin-name><subResourcePath>` is right, but it's confusing. Do you have some ideas?

kbs/src/policy_engine/opa/mod.rs
```diff
@@ -175,35 +169,141 @@ mod tests {
...
         1,
-        Err(KbsPolicyEngineError::ResourcePathError)
+        Ok(false)
     )]
     #[case(
         "test/data/policy_invalid_2.rego",
-        "my_repo/Alice/key",
+        json!({
+            "plugin": "resource",
+            "subResourcePath": "/my_repo/Alice/key",
```
mythi: some leftover here?

kbs/src/policy_engine/opa/default_policy.rego
```diff
@@ -1,41 +1,83 @@
...
+#        "plugin": "plugin1",
+#        "resource-path": "/para/meters",
+#        "query": {
+#           "version": "1.0.0"
+#        }	
+#    }
+#    ```
+#
+#    For the "resource" plugin specifically:
+#      - resource-path format: /<repo>/<type>/<tag>
```
mythi: The `/` prefix for `<repo>` makes Rego `split()` to give 4 elements?

kbs/src/policy_engine/opa/mod.rs
mythi: leftover?

Xynnn007: Oh yes. Let me fix

Xynnn007: Let me fix this.

kbs/src/policy_engine/opa/default_policy.rego
Xynnn007: Yes.

mythi: OK, I guess we have to accept this change since the path is no longer for the KBS resources but generic to any plugin.

kbs/src/api_server.rs
```diff
@@ -125,7 +130,7 @@ impl ApiServer {
                         (1024 * 1024 * http_config.payload_request_size) as usize,
                     ))
                     .service(
-                        web::resource([kbs_path!("{base_path}{additional_path:.*}")])
+                        web::resource([kbs_path!("{plugin}{resource_path:.*}")])
```
fitzthum: I always find this macro a little confusing. Can you add a comment to describe what is happening (like how is it splitting the path) or maybe it would be clearer to split the path later using more standard rust?

kbs/src/policy_engine/mod.rs
```diff
@@ -24,13 +24,13 @@ pub const DEFAULT_POLICY_PATH: &str = "/opt/confidential-containers/kbs/policy.r
 pub(crate) trait PolicyEngineInterface: Send + Sync {
     /// Determine whether there is access to a specific path based on the input claims.
     /// Input parameters:
-    /// request_path: Required to be a string in segments path format:<FIRST>/.../<END>, for example: "my'repo/License/key".
-    /// input_claims: Parsed claims from Attestation Token.
+    /// data: data to be evaluated for regorus (rego) engine, in JSON format.
```
fitzthum, line 27: I find `data` to be a little vague, but I guess this is all going to change with the upcoming unified policy engine stuff.

kbs/src/policy_engine/opa/default_policy.rego
```diff
@@ -1,41 +1,83 @@
...
+# below for specific conditions.
 
 package policy
 
 default allow = false
 
+plugin = data.plugin
+
 allow if {
+	plugin == "resource"
```
fitzthum, line 84: I'm not sure this is really necessary, but I am fine with it.

kbs/test/data/policy_1.rego
```diff
@@ -2,9 +2,10 @@ package policy
 
 default allow = false
 
-path := split(data["resource-path"], "/")
+subpaths := split(data["resource-path"], "/")
```
fitzthum: A lot of policies end up splitting the resource path into different fields. Maybe we should do this automatically and have the data contain a list of individual fields rather than one string? Maybe that is more cumbersome to work with in some circumstances idk

kbs/src/api_server.rs
Xynnn007: Yes it's somehow a magic. Let me replace with rust logic

kbs/src/policy_engine/mod.rs
Xynnn007: Yes. Data is an OPA conception, alongside with input.

kbs/src/policy_engine/opa/default_policy.rego
Xynnn007: This statement is to ensure semantic consistency with previous behavior - by explicitly show it.

kbs/test/data/policy_1.rego
Xynnn007: This would introduce some more changes. Let me add a separate commit and see if it is acceptable.

kbs/src/policy_engine/opa/default_policy.rego
mythi: how about we explain this to users in the commented section too (`For the "resource" plugin specifically`) so that it's easier to see why it's in here?

kbs/test/data/policy_1.rego
Xynnn007: https://github.com/confidential-containers/trustee/pull/1098/commits/c6d3e747d56ffaab1e31a4b5fb29ee5964275ae3

kbs/src/policy_engine/opa/default_policy.rego
Xynnn007: Added to `Policy Rules`

PR #1103, opened 2025-12-09
Add some validated identifiers

fitzthum:
It's time to add some validated identifiers. It will take a few iterations to figure out exactly what we want, but let's start with some simple ones. This PR provides some concrete identifiers to see how the system works.

Also, introduce an integration test that makes sure we are correctly extracting these identifiers from an init-data with a policy.

### DIFF
--- a/integration-tests/src/common.rs
+++ b/integration-tests/src/common.rs
@@ -30,8 +30,11 @@ use reference_value_provider_service::rvps_api::reference::reference_value_provi
 use reference_value_provider_service::storage::{local_json, ReferenceValueStorageConfig};
 use reference_value_provider_service::{server::RvpsServer, Rvps};
 
-use anyhow::{anyhow, Result};
-use base64::{engine::general_purpose::STANDARD, Engine};
+use anyhow::{anyhow, bail, Result};
+use base64::{
+    engine::general_purpose::{STANDARD, URL_SAFE_NO_PAD},
+    Engine,
+};
 use log::info;
 use openssl::pkey::PKey;
 use serde_json::json;
@@ -339,5 +342,24 @@ impl TestHarness {
         rvps_client::register(RVPS_URL.to_string(), message.to_string()).await?;
 
         Ok(())
+    }
+
+    pub async fn get_attestation_token_payload(
+        &self,
+        init_data: Option<String>,
+    ) -> Result<serde_json::Value> {
+        let token = kbs_client::attestation(KBS_URL, None, vec![], init_data).await?;
+
+        let parts: Vec<&str> = token.split('.').collect();
+        if parts.len() != 3 {
+            bail!("Invalid attestation token.");
+        }
+
+        let payload_base64 = parts[1];
+        let payload_bytes = URL_SAFE_NO_PAD.decode(payload_base64)?;
+
+        let payload: serde_json::Value = serde_json::from_slice(&payload_bytes)?;
+
+        Ok(payload)
     }
 }
--- a/integration-tests/tests/init_data.rs
+++ b/integration-tests/tests/init_data.rs
@@ -2,9 +2,10 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, Result};
 use base64::{engine::general_purpose::STANDARD, Engine};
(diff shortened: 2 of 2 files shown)


attestation-service/docs/policy.md
```diff
@@ -37,6 +37,14 @@ If the extension is registered in the EAR specification, the key will be positiv
 Otherwise, the key should be negative.
 The extension value can be any valid JSON, including nested objects.
 
+The default policy includes the extension `ear.trustee.identifiers`, which is used
```
Xynnn007, line 40: Note that only in CoCo case where agent policy is also enabled, the extension exists.

attestation-service/src/ear_token/ear_default_policy_cpu.rego
```diff
@@ -297,3 +297,45 @@ configuration := 0 if {
 }
 
 ##### SE TODO
+
+
+#################################
+# EXTENSIONS
+#
+# Extensions are added to the EAR Appraisal
```
Xynnn007, line 305: For trustee is not only used in CoCo, but also in other cases. We can have many different extension definitions. For these, we can add a context here as note that the following extensions will be added if the request is from CoCo, also with agent-policy enabled.

attestation-service/docs/policy.md
```diff
@@ -37,6 +37,14 @@ If the extension is registered in the EAR specification, the key will be positiv
...
+to store fields that identify the workload.
+The `identifiers` extension includes two subfields.
+
+* `validated` identifiers are those that are bound to the hardware evidence via
+attestation and that the guest runtime binds to the workload.
+* `unvalidated` identifiers are anything else that describe the workload.
```
bpradipt, line 49: Any examples on how do you see the validated identifiers being used?

attestation-service/src/ear_token/ear_default_policy_cpu.rego
```diff
@@ -297,3 +297,45 @@ configuration := 0 if {
...
+
+extensions := [
+	{"name": "ear.trustee.identifiers",
+	 "key": -18,
+	 "value": {
+		"validated": validated_identifiers
+	}
+	}
```
Xynnn007, line 322: let's do some format work upon these lines

attestation-service/docs/policy.md
fitzthum: As long as you are using the default policy the extension will be there, but it will be empty if you don't have any identifiers, which currently depend on the agent policy. I am hoping to add some that use the CDH config, the eventlog, and maybe information from Trustee itself.

attestation-service/src/ear_token/ear_default_policy_cpu.rego
fitzthum: Let me add something to this comment about the kata agent policy.

attestation-service/docs/policy.md
fitzthum: This is still in discussion, but the basic idea is that the KBS (or some other consumer of the attestation token) will use the identifiers plus some policy to come up with one top-level identity for the pod, such as a SPIFFE ID. For instance, you could use selectors over the identifiers.

I plan to add some more information about this and about the details of each identifier as things evolve.

attestation-service/docs/policy.md
Xynnn007: Good. We might need a separate document section to show all the information that default policy will do

attestation-service/docs/policy.md
```diff
@@ -37,6 +37,17 @@ If the extension is registered in the EAR specification, the key will be positiv
...
 The extension value can be any valid JSON, including nested objects.
 
+The default policy includes the extension `ear.trustee.identifiers`, which is used
+to store fields that identify the workload.
+With Confidential Containers these identifiers will usually be based on the Kata
+Agent Policy or other configuration fields in the InitData.
+Other runtimes might use other mechanisms, such as the event log.
+The `identifiers` extension includes two subfields.
+
+* `validated` identifiers are those that are bound to the hardware evidence via
```
Xynnn007, line 47: A question is that are there anything that is not bound to hardware evidence, but still can be captured by this policy?

fitzthum: Maybe. There could be identifiers that are indirectly connected to the evidence or that actually come from Trustee (see newi dentifiers issue). Still, I think this is the most concise way of describing validated identifiers at least for now.

Issue #1092, opened 2025-12-02
Refactor trustee to be stateless (decouple compute and storage)


Xynnn007:
Currently, trustee components keep some state in memory or local storage (e.g. policies, reference values, and sessions). This makes horizontal scaling and Kubernetes-friendly deployment harder.

We want to refactor trustee to be stateless, by fully decoupling compute and storage.

We already have a Rust crate for `key-value-storage` that provides a storage abstraction. This issue aims to standardize and extend the use of that abstraction so that all stateful data goes through it.

## Goals

- Make trustee components stateless from the compute point of view
- Make trustee easier to deploy and scale on Kubernetes
- Support multiple storage backends (e.g. in-memory, local file, database, local json)
- Ensure all persistent data is accessed via the key-value-storage abstraction

## Scope

Decouple storage for at least the following modules:

- AS policies
- KBS policies
- RVPS reference values
- KBS session state

## High-level requirements

- All persistent or shared data must be read/written via the key-value-storage abstraction
- It must be possible to plug different storage backends (in-memory, file-based, DB, etc.) without changing upper layer logic
- Stateless behavior must be supported:
    - Multiple instances can run in parallel without relying on local state
- Pods/instances can be replaced without data loss
- For simple or standalone deployments, it should still be possible to use an in-memory or local-file backend

## Sub-Tasks

- [x] Add two sub-modules, local JSON and local file, to key-value storage.
- [x] Replace RVPS's reference value storage with key-value storage.
- [x] Replace AS's policy engine with policy-engine, and also replace its policy storage with key-value storage.
- [x] Replace KBS's policy storage with key-value storage, replace some resource backends with key-value storage.
- [ ] Replace KBS's session state with key-value storage
- [ ] Resolve issues with integration tests, kbs-client, and trustee-cli.
- [ ] Update documentation, configuration files, and deployment scripts.

PR #1108, opened 2025-12-11
[kv-storage] add localfs/localjson plugin and rename simple to Memory

Xynnn007:
Please see the concrete commit messages.

This PR is part of #1093

### DIFF
--- a/deps/key-value-storage/src/lib.rs
+++ b/deps/key-value-storage/src/lib.rs
@@ -17,6 +17,8 @@ pub mod memory;
 #[cfg(feature = "postgres")]
 pub mod postgres;
 
+pub mod local_json;
+
 pub mod local_fs;
 
 #[derive(Default)]
@@ -53,6 +55,9 @@ pub enum KeyValueStorageConfig {
     #[default]
     Memory,
 
+    #[serde(alias = "LocalJson")]
+    LocalJson(local_json::Config),
+
     #[serde(alias = "LocalFs")]
     LocalFs(local_fs::Config),
 }
@@ -69,6 +74,9 @@ impl KeyValueStorageConfig {
                     })?,
             )),
             KeyValueStorageConfig::Memory => Ok(Arc::new(memory::MemoryKeyValueStorage::default())),
+            KeyValueStorageConfig::LocalJson(config) => {
+                Ok(Arc::new(local_json::LocalJson::new(config.clone())?))
+            }
             KeyValueStorageConfig::LocalFs(config) => {
                 Ok(Arc::new(local_fs::LocalFs::new(config.clone())?))
             }
--- a/deps/key-value-storage/src/local_json/README.md
+++ b/deps/key-value-storage/src/local_json/README.md
@@ -1,0 +1,65 @@
+# Local JSON Storage
+
+A key-value storage implementation using local JSON file format.
+
+## Overview
+
+`LocalJson` is a key-value storage backend implementation based on local JSON files, implementing the `KeyValueStorage` trait. It stores all key-value pairs in a single JSON file, suitable for single-machine environments or scenarios requiring simple persistent storage.
+
+## Features
+
+- **Persistent Storage**: Data is stored in a local JSON file, preserving data across program restarts
+- **Concurrency**: Uses `RwLock` for concurrent read-write control
+- **Auto Initialization**: Automatically creates directories and empty file if the specified file path doesn't exist
+
+## Configuration
+
+### Config Structure
+
+```rust
+pub struct Config {
+    pub file_path: String,  // Path to the JSON file
+}
+```
+
+### Default Configuration
+
+- **Default File Path**: `/opt/confidential-containers/storage/key_value.json`
+
+If `file_path` is not specified, the above default path will be used.
+
+## Implementation Details
+
+### File Format
+
+Data is stored in JSON object (HashMap) format, for example:
+
+```json
+{
+  "key1": "MTIzCg==",
+  "key2": "MjM0Cg=="
+}
+```
+
+Values are stored as base64 encoded string (with URLSAFE alphabet) in JSON.
+
+## Notes
+
+1. **Performance**: Each write operation reads and writes the entire file, making it unsuitable for high-frequency write scenarios
+2. **Concurrency**: Although thread-safe, multiple write operations execute serially, which may become a performance bottleneck
+3. **File Size**: All data is stored in a single file; excessively large files may impact performance
+4. **Data Consistency**: In exceptional circumstances (such as program crashes), there may be a risk of data inconsistency
+
+## Suitable Scenarios
+
+- Single-machine environments
+- Low-frequency write scenarios
+- Applications requiring simple persistent storage
+- Development and testing environments
+
+## Unsuitable Scenarios
+
+- High-frequency write scenarios
+- Distributed environments
+- Production environments requiring high availability
+- Applications requiring transaction support
--- a/deps/key-value-storage/src/local_json/mod.rs
+++ b/deps/key-value-storage/src/local_json/mod.rs
@@ -1,0 +1,208 @@
+// Copyright (c) 2025 by Alibaba.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Local file of JSON format for the key-value storage.
+//!
+//! All key-value pairs are stored in a single JSON file.
+
+use std::{collections::HashMap, fs, path::PathBuf};
+
+use async_trait::async_trait;
+use base64::{engine::general_purpose::URL_SAFE, Engine};
+use serde::Deserialize;
+use tokio::sync::RwLock;
+use tracing::{debug, instrument};
+
+use crate::{KeyValueStorage, KeyValueStorageError, Result, SetParameters};
+
+/// Default file path for the local JSON file.
+const FILE_PATH: &str = "/opt/confidential-containers/storage/local_json/key_value.json";
+
+pub struct LocalJson {
+    file_path: String,
+    lock: RwLock<i32>,
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[serde(default)]
+pub struct Config {
+    pub file_path: String,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            file_path: FILE_PATH.to_string(),
+        }
+    }
+}
+
+impl LocalJson {
+    pub fn new(config: Config) -> Result<Self> {
+        let mut path = PathBuf::new();
+        path.push(&config.file_path);
+
+        let parent_dir =
+            path.parent()
+                .ok_or_else(|| KeyValueStorageError::InitializeBackendFailed {
+                    source: anyhow::anyhow!(
+                        "Illegal `file_path` for LocalJson's config without a parent dir."
+                    ),
+                })?;
+        debug!(path =? parent_dir, "create file path for LocalJson backend");
+        fs::create_dir_all(parent_dir)
+            .map_err(|e| KeyValueStorageError::InitializeBackendFailed { source: e.into() })?;
+
+        if !path.exists() {
+            debug!(path =? path, "creating empty file for LocalJson backend");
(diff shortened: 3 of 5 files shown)


deps/key-value-storage/src/local_fs/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+const ALLOWED_KEY_CHARS: &str =
+    "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_./";
+
+/// Check if the key is valid.
+///
+/// The key is valid if it only contains ASCII alphanumeric characters, `-`, `_` or `.`.
+/// No spaces and other special characters are allowed to prevent SQL injection.
+fn is_valid_key(key: &str) -> bool {
+    key.chars().all(|c| ALLOWED_KEY_CHARS.contains(c)) && !key.starts_with('.')
+}
```
mkulke: do we want to have a common requirement for valid keys that applies to all backends? It would be inconvenient if you have to rename your keys because you are migrating data across backends.

deps/key-value-storage/src/local_fs/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+        if !is_valid_key(key) {
+            return Err(KeyValueStorageError::SetKeyFailed {
+                source: anyhow::anyhow!("key contains invalid characters"),
+                key: key.to_string(),
+            });
+        }
+
+        let _ = self.lock.write().await;
+        let file_path = self.dir_path.join(key.replace('/', "@"));
+        if parameters.overwrite || !file_path.exists() {
```
mkulke, line 68: what happens if the file exists and we haven't set overwrite, do we really want to ignore the write or should we bail out with an error (I think that would make sense).

deps/key-value-storage/src/local_fs/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+    async fn set(&self, key: &str, value: &[u8], parameters: SetParameters) -> Result<()> {
+        if !is_valid_key(key) {
+            return Err(KeyValueStorageError::SetKeyFailed {
+                source: anyhow::anyhow!("key contains invalid characters"),
+                key: key.to_string(),
+            });
+        }
+
+        let _ = self.lock.write().await;
+        let file_path = self.dir_path.join(key.replace('/', "@"));
```
mkulke: why don't we disallow `/` from the valid key names?

deps/key-value-storage/src/local_fs/README.md
```diff
@@ -0,0 +1,67 @@
...
+A key-value storage implementation using the local file system.
+
+## Overview
+
+`LocalFs` is a key-value storage backend implementation based on the local file system, implementing the `KeyValueStorage` trait. Each key-value pair is stored in a separate file in the file system, making it suitable for single-machine environments or scenarios requiring simple persistent storage with file-level granularity.
+
+## Features
+
+- **Persistent Storage**: Data is stored in individual files on the local file system, preserving data across program restarts
+- **Thread Safety**: Uses `RwLock` for concurrent read-write control
```
mkulke, line 12: w/ async and tokio there's also concurrency on a single thread.

```suggestion
- **Concurrency**: Uses `RwLock` for concurrent read-write control
```

deps/key-value-storage/src/local_json/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+            file_path: config.file_path,
+            lock: RwLock::new(0),
+        })
+    }
+}
+
+#[async_trait]
+impl KeyValueStorage for LocalJson {
+    #[instrument(skip_all, name = "LocalJson::set", fields(key = key))]
+    async fn set(&self, key: &str, value: &[u8], parameters: SetParameters) -> Result<()> {
```
mkulke, line 73: do we want a restriction on key for this impl? see the above suggestion to have a common restricted character set for keys.

deps/key-value-storage/src/local_json/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+        let _ = self.lock.write().await;
+        let file = tokio::fs::read(&self.file_path).await.map_err(|e| {
```
mkulke, line 75: If there's a lot of k/v pairs, then this method of loading/parsing/modifying/serializing/writing will be a bottleneck soon, since due to the w-lock the set fn will become sequential.

there are some low-hanging fruits for optimization, like using an internal generation counter and caching the parsed file in the getter.

personally, I think we should not allow writes on the local_json backend. the json makes sense if your trustee instance is read-only and you manage your state elsewhere (e.g. via kubernetes configmaps), but not for much else.

deps/key-value-storage/src/local_json/README.md
```diff
@@ -0,0 +1,65 @@
...
+Data is stored in JSON object (HashMap) format, for example:
+
+```json
+{
+  "key1": [118, 97, 108, 117, 101, 49],
+  "key2": [118, 97, 108, 117, 101, 50]
+}
+```
+
+Values are stored as byte arrays (`Vec<u8>`), represented as number arrays in JSON.
```
mkulke: does it make sense to use a base64_url representation?, it's a bit less verbose and easier to produce from other tools.

deps/key-value-storage/src/local_fs/mod.rs
Xynnn007: Good idea. Let me add another commit in the PR for this.

deps/key-value-storage/src/local_fs/mod.rs
Xynnn007: Yes. This logic will be the same as local json. Let me do the fix. Togetherly the postgres needs fix too.

deps/key-value-storage/src/local_fs/mod.rs
Xynnn007: The KBS resource backend can still use the crate, thus we need to handle resource paths like `a/b/c`

deps/key-value-storage/src/local_json/mod.rs
Xynnn007: Yes, I reuse the lib function of judge keys and also adds a top level README.

deps/key-value-storage/src/local_json/mod.rs
Xynnn007: Yes, the purpose of implementing local JSON was simply for compatibility with the previous system. I don't recommend doing much (or any) optimization work here.  If there's a need for more robust functionality, it's recommended to use a database or other backend solution more suitable for production.

deps/key-value-storage/src/local_json/README.md
Xynnn007: Makes sense. Let me change this.

deps/key-value-storage/src/local_fs/mod.rs
mkulke: can we maybe escape forward slashes like this `\x2F`? it's a bit more obvious what's going on, since e.g. systemd also uses this escaping in e.g. unit files:

https://donsnotes.com/tech/charsets/ascii.html

deps/key-value-storage/src/local_json/mod.rs
mkulke: I think the local json was used for reference values, no? but it's not compatible anymore, because the reference_values.json file has a different schema than just simple key/value pairs.

is this accurate?

if yes, then I am wondering: if there is no compatibility with previous json files, is there a need of a json backend?

there might still be a point for it (like maybe having a full state "snapshot" in a single file is sort of useful). however, a flat generic key/value array in json is really a machine format, not something that is immediately useful for people without more tooling (like you can do: kubectl edit configmap reference-values.json`).

Xynnn007: Yes it was once used for RVPS.

fwiw I just try to bring all original plugins into the new crate and keep it as is. The usages might be like what you've mentioned - some pre-configured, static values.

I am ok with either removing it or keeping it.

mkulke: In this case I don't see any use case that wouldn't be covered by the file-based k/v and hence not include it.

Regarding the reference-values.json, we probably want to deprecate that, once there is a comprehensive storage implementation. 

I understand it doesn't cover the full reference value semantics/schema (afair it was introduced as a lo-fi solution to get something working).

But we should deprecate it in an orderly fashion, e.g. announce deprecation with a warning in release notes and logs, when being used and remove it in the next release.

Xynnn007: Ok. I will mark this plugin as "to be deprecated". 

The next PR I will replace RVPS logic to use this crate.

Does this make sense? Or, another way is to directly not include this plugin, and we just go ahead to replace RVPS logic to use this crate too.

mkulke: > Does this make sense? Or, another way is to directly not include this plugin, and we just go ahead to replace RVPS logic to use this crate too.

yes, this makes more sense.

I guess want to provide a path forward that supports existing reference-values.json files while we provide an alternative. (both options) in the next release reference-values.json can be removed (one option). I wouldn't invest any work in making reference-values.json mutable.

cc @bpradipt

Xynnn007: Added a deprecation warning in README.

mkulke: but why add a new json storage implementation to deprecate it in the next release? or does the new json backend somehow support existing reference-values.json files?

Xynnn007: i deleted local json plugin but only add local fs. I remembered that @fitzthum once mentioned some folks still want this way

fitzthum: @lmilleri @bpradipt do you have a preference between localfs and local json?

Xynnn007: Some context: I am currently working on a unified plug-in design for all storage systems to address [this issue](https://github.com/confidential-containers/trustee/issues/1092). This means we will support a variety of different storage backends to meet stateless deployments, as well as existing deployments (such as configmaps based on static files).

Therefore, I would like to confirm whether both localFS and LocalJson need to be retained? Are there any other ideas?

fitzthum: honestly I would just keep both for now so we don't get blocked on this PR, which is otherwise close to being merged

deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -125,15 +125,32 @@ impl KeyValueStorage for PostgresClient {
             });
         }
 
-        let mut sql = format!(
-            "INSERT INTO {} ({KEY_COLUMN}, {VALUE_COLUMN}) VALUES ( $1, $2 )",
+        if !parameters.overwrite {
```
mkulke: this part seems unrelated, can we remove it from this PR?

deps/key-value-storage/src/lib.rs
```diff
@@ -12,11 +12,15 @@ use std::sync::Arc;
...
 
 #[cfg(feature = "postgres")]
 pub mod postgres;
 
+pub mod local_json;
+
+pub mod local_fs;
+
 #[derive(Default)]
 pub struct SetParameters {
```
fitzthum, line 25: I don't really like this struct, especially since it only has one field. We could just pass an overwrite argument to `set`. The logic around overwriting stuff is a little convoluted anyway.

We can tweak this in a follow-up though, since it's not actually added in this PR.

deps/key-value-storage/src/postgres/mod.rs
Xynnn007: Oh yes. It's inspired by the comment https://github.com/confidential-containers/trustee/pull/1108#discussion_r2610411076 . If it's not good I can make a separate PR

deps/key-value-storage/src/lib.rs
Xynnn007: Oh yes. Originally there would be some other parameters than overwrite for the set context due to https://github.com/confidential-containers/trustee/pull/1000#discussion_r2444051093 and this is a room for them. But we can change it if we do not need this

PR #1111, opened 2025-12-16
Add Admin Roles

fitzthum:
This is a powerful feature, but it's relatively simple to implement. 

First, make each admin backend return a role string. This allows us to implement the role logic not in each individual backend, but in the common admin code. This logic is fairly simple. We have a list of admin roles which specify an id and a regular expression. If no roles are specified, all admin will have access to all requests. This provides backwards compatibility.

Since this is a security-critical configuration, I did some defensive programming. For example, the `AdminRoles` struct does not allow unknown fields when deserializing. The admin id is case insensitive. Etc 

I created some unit tests. No integration tests yet due to merge conflicts with the previous admin PR.

### DIFF
--- a/kbs/src/admin/allow_all.rs
+++ b/kbs/src/admin/allow_all.rs
@@ -13,7 +13,7 @@ pub struct InsecureAllowAllBackend {}
 
 impl AdminBackend for InsecureAllowAllBackend {
     fn validate_admin_token(&self, _request: &HttpRequest) -> Result<String> {
-        warn!("Allow All admin backend is set. Anyone can access admin APIs");
+        warn!("Allow All admin backend is set. All admin tokens are valid.");
         Ok("Anonymous".to_string())
     }
 }
--- a/kbs/src/admin/simple.rs
+++ b/kbs/src/admin/simple.rs
@@ -66,13 +66,16 @@ impl AdminBackend for SimpleAdminBackend {
                 .verify_token::<NoCustomClaims>(token, Some(VerificationOptions::default()));
             match res {
                 Ok(_claims) => {
-                    info!("Admin access granted for {}", persona.id);
+                    info!("Admin token validated (simple) for persona: {}", persona.id);
 
                     // Return the first matching persona
                     return Ok(persona.id.clone());
                 }
                 Err(e) => {
-                    info!("Access not granted for {} due to: \n{}", persona.id, e);
+                    info!(
+                        "Admin token not validated for {} due to: \n{}",
+                        persona.id, e
+                    );
                 }
             }
         }


kbs/src/admin/allow_all.rs
```diff
@@ -12,8 +12,8 @@ use crate::admin::AdminBackend;
 pub struct InsecureAllowAllBackend {}
 
 impl AdminBackend for InsecureAllowAllBackend {
-    fn validate_admin_token(&self, _request: &HttpRequest) -> Result<()> {
+    fn validate_admin_token(&self, _request: &HttpRequest) -> Result<String> {
         warn!("Allow All admin backend is set. Anyone can access admin APIs");
-        Ok(())
+        Ok("Anonymous".to_string())
```
Xynnn007, line 17: What about
1. have a separate policy file (no need to be rego, but if rego is good we can use rego) to define admin authZ things (What this PR is intending to do)
2. the input to the the id/subject or claims parsed from the input JWT of header's bearer.
3. The output is true/false.

We can by default give a allow all/deny all policy.

In this way we can have a simple authZ system that is code-policy-decoupled.

The policy can includes all the "role" names and its acceptable paths (e.g. in regex)

fitzthum: I think that's too complicated. Users are already confused by the two policies we have. Let's only add another one if we absolutely have to. What is implemented here is basically a policy mechanism, but it's made as simple as possible. Let's not do anything more complex unless users explicitly ask for it.

Xynnn007: Yup, what we are aiming for right now in the admin module is a simple authorization (AuthZ) mechanism, i.e., defining which APIs can be accessed.

The solution proposed in this PR enables a quick way to define roles and their accessible APIs, and it is ready for deployment. This approach is very useful for some internal or short-term projects where simplicity and fast iteration are the primary goals.

However, from a longer-term evolution perspective—especially if we expect more frequent policy changes or additional authorization scenarios - there are a few concerns worth discussing:

1. Authorization rules are embedded in startup configuration

Currently, the rules are defined in the KBS configuration at startup time and are not independently versioned or reloadable. As a result, updating or modifying authorization rules requires restarting the entire KBS, which may become inconvenient as policies evolve.

2.Mixed abstraction between token verification and authorization decision

The current validate_admin_token abstraction mixes two different concerns:

- Token Verification / Identity Resolution (e.g., extracting credentials, verifying signatures, validating issuers), and
- Authorization Decision (allow / deny, or mapping to accessible APIs).

In retrospect, this separation was not clearly enforced in earlier designs (including my own previous review in https://github.com/confidential-containers/trustee/pull/1014/changes#diff-57ec221f4d45eb2d52007f10e72821894545bdea23a2ca94924f4a55fe39a44a), and this PR makes the boundary more visible. It may be beneficial to clarify this abstraction going forward.

From a directional and longer-term design point of view (not necessarily a requirement for this PR), one possible evolution could be:

1. Move client authentication toward token-based credentials
As a follow-up direction, the kbs-client could rely on tokens rather than private keys. For example, docker-compose or Kubernetes deployments could generate tokens at launch time and mount them at a well-defined path for the kbs-client to consume. This would be a change that we once talked about and can be considered independently of this PR.

2. Conceptually split `admin::validate_admin_token` (the module's than the plugin's) into two stages

- Token Verification / Identity Resolution: pluggable implementations that support bearer tokens, JWTs, or  other credential formats, and return structured identity or claims.
- Authorization Decision: a separate step that evaluates whether the resolved identity is allowed to access a given API.

This separation could exist conceptually first, even if the initial implementation remains simple.

3. Introduce a simple, standalone authorization policy
The authorization decision could be driven by a lightweight policy definition (e.g., file-based), instead of being embedded in startup configuration. KBS would still not perform identity issuance or user management; it would only verify externally issued credentials and apply a local authorization policy. More complex authentication or identity logic could remain out of band.

This direction certainly has drawbacks:

1. Higher implementation complexity in the initial iteration.

2. Increased learning cost for users, who would need to understand the policy model.
That said, this could be mitigated by providing a default policy (e.g., allow_all) for simple use cases.

Overall, the current PR provides a practical and usable solution, while the points above are intended as considerations for future evolution rather than blockers for the current implementation.

wdyt?

fitzthum: Runtime configuration of admin roles is out of scope. Maybe sometime we could add an API to update the roles (and store them in the kv backend), but this is complicated, and currently we don't even have an admin login interface.

For 2, I will just rename the upper function to be more generic. We already have a split between checking the token,  which is handled by the admin backends, and allowing access to the endpoint. It isn't perfect since the allow-all backend could be specified and roles could block things on top of that, but I've updated the docs to explain the different.

fitzthum: @Xynnn007 let's revisit this before the PR gets stale. i renamed a few things and updated the docs. I think it's best to use a simple approach for now. We have some internal people asking for this feature. I am hoping to get some feedback from them after we implement something and iterate if needed.

Xynnn007: What about this? I will do some changes upon your current PR after getting it merged. including moving key pair to token and some other changes about code structure

kbs/src/admin/mod.rs
```diff
@@ -45,6 +47,25 @@ impl Default for AdminBackendType {
...
+/// An admin role is a rule that grants access for some roles to some endpoints.
+#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct AdminRole {
+    /// The admin role that this rule applies to.
+    /// The id is case insensitive.
+    pub id: String,
+    /// A regular expression selecting request paths this rule allows.
+    /// In other words, the paths that the above role can access.
+    #[serde(default)]
```
Xynnn007, line 67: That means by default no endpoints can be accessed?

fitzthum: Yes. That seems less error-prone.

kbs/src/admin/allow_all.rs
fitzthum: Sure, but what do you have in mind generally?

Xynnn007: Yes. Almost [the PR](https://github.com/fitzthum/kbs/pull/2)

fitzthum: Ok. I have some comments about that PR but we can save for when it is actually posted. It seems fine in general although it's much more complex.

fitzthum:
@mythi @mkulke ptal. this is not a very big PR

kbs/docs/config.md
```diff
@@ -203,6 +211,23 @@ If the `Simple` backend is used, a list of admin personas can be provided, each
...
+| id                | String | The role that this rule applies to                       | Yes      |
+| allowed_endpoints | String | A regular expression matching endpoints that are allowed | Yes      |
+
+After the admin token is validated, the role extracted from the token will be checked against
+the roles specified in the config file.
+If a matching role is found, the `allowed_endpoints` regular expression is used to determine
+which endpoints the admin can access.
+If no matching role is found (and at least one role is specified), the admin will not be able
+to access any endpoints.
+Roles and role ids are case insensitive.
```
mkulke: is this something we do consisently for identifiers and strings in trustee or is there a reason why would want to do this here?

kbs/src/admin/mod.rs
```diff
@@ -59,23 +80,47 @@ impl TryFrom<AdminConfig> for Admin {
...
-        let res = self.backend.validate_admin_token(request);
-        match res {
-            Ok(()) => info!("Allowing Admin access for {}", request.full_url().as_str()),
-            Err(ref e) => info!(
-                "Not allowing Admin access for {} due to: \n{}",
-                request.full_url().as_str(),
-                e
-            ),
+    pub fn check_admin_access(&self, request: &HttpRequest) -> Result<()> {
+        if let Ok(role) = self.backend.validate_admin_token(request) {
```
mkulke: nit: exit early with let ... else to save an indentation.

kbs/docs/config.md
fitzthum: afaik the uris, policies, and whatever else is all case-sensitive. I was thinking it would be nice to make this not case sensitive to make it easier for admins to be sure that the proper roles are captured, especially since the role will often be extracted from a third-party token.

I was thinking this would reduce the chance of errors in this critical config, but since this is an allow-list (as long as at least one role is specified) maybe it actually makes things more permissive and increases the chance of errors. idk. I can drop it if you think that would be best.

mkulke:
> Not sure I totally grasp this but at the moment we are only matching on the URI. There is no distinction between POST and GET.

ah, I meant that this is a reasonable use case that will probably come up and is not expressible in the current scheme. e.g.  "policy_reader" is allowed to "^GET /policies.*$" while "policy_contributor" is also allowed to do "^POST /policies$"

If we want to stick with regexes, I'd suggest making a preceding verb mandatory. or, if we want to add support at a later point, make a leading `/` mandatory (so we can reserve the empty verb as "all verbs")

kbs/docs/config.md
mkulke: > afaik the uris, policies, and whatever else is all case-sensitive. I was thinking it would be nice to make this not case sensitive to make it easier for admins to be sure that the proper roles are captured, especially since the role will often be extracted from a third-party token.
> 
> I was thinking this would reduce the chance of errors in this critical config, but since this is an allow-list (as long as at least one role is specified) maybe it actually makes things more permissive and increases the chance of errors. idk. I can drop it if you think that would be best.

I would prefer less ambiguity over some degree of convenience/easiness. I think it's fair to restrict the character set that is allowed ([a-zA-Z_-]), but I wouldn't implicitly convert anything.

kbs/src/admin/mod.rs
```diff
@@ -59,23 +80,47 @@ impl TryFrom<AdminConfig> for Admin {
...
+                    info!("Allowing Request to {}", request.full_url().as_str());
+                    return Ok(());
+                }
+            }
         }
 
-        res
+        info!(
+            "Not allowing Admin access to {}",
+            request.full_url().as_str()
```
mkulke: should we use requets.uri here as well, so it's more obvious what being matched?

kbs/docs/config.md
```diff
@@ -203,6 +211,23 @@ If the `Simple` backend is used, a list of admin personas can be provided, each
...
+One role entry looks like this.
+
+| Property          | Type   | Description                                              | Required |
+|-------------------|--------|----------------------------------------------------------|----------|
+| id                | String | The role that this rule applies to                       | Yes      |
+| allowed_endpoints | String | A regular expression matching endpoints that are allowed | Yes      |
+
+After the admin token is validated, the role extracted from the token will be checked against
+the roles specified in the config file.
+If a matching role is found, the `allowed_endpoints` regular expression is used to determine
```
mkulke, line 226: should we define/provide examples what an endpoint is? request.uri seems to be like [defined](https://docs.rs/actix-web/latest/actix_web/http/struct.Uri.html) like this:

```
abc://username:password@example.com:123/path/data?key=value&key2=value2#fragid1
|-|   |-------------------------------||--------| |-------------------| |-----|
 |                  |                       |               |              |
scheme          authority                 path            query         fragment
```

in the tests we seem to imply an endpoint is "path", but are query and maybe even fragment allowed?

btw, I would expect the anchored test-case to fail since it would start with "http://....", am I misreading the docs or why does this pass?

fitzthum: Hmm I think because I am setting the URI directly in the tests cases with what seems like not correct values. Let me add an integration test for this that will use real requests made to Trustee.

kbs/docs/config.md
fitzthum: Ok, removed the case stuff

kbs/docs/config.md
fitzthum: Ok, per the integration tests (which do a full server and client flow), the URI that actix gives us looks like `/kbs/v0/attestation-policy`. I updated the unit tests to use a similar format.

With this in mind, I actually changed things to require that the regex start with `^/kbs`. Hopefully that will help clue people into what these paths look like.

The actix behavior does seem a little weird, but this is what I had in my head when I was first making this PR.

Also note that in any case the `method` is not included and is out of scope for this approach.

kbs/src/admin/mod.rs
```diff
@@ -59,23 +80,52 @@ impl TryFrom<AdminConfig> for Admin {
             AdminBackendType::DenyAll => Arc::new(DenyAllBackend::default()) as _,
         };
 
-        Ok(Admin { backend })
+        // Parse roles to ensure valid regexes and no duplicates.
+        let mut roles = HashMap::new();
+        for role in value.roles {
+            if !role.allowed_endpoints.starts_with("^/kbs") {
```
mkulke: anchoring should also check for ends_with("$"), no?

fitzthum: I guess that term is a bit ambiguous. I added a check for `$` at the end.

mkulke (approved):

PR #1113, opened 2025-12-17
deps/kv: add error for postgres when set existing key

Xynnn007:
If the overwrite flag is not set, set to an existing key will raise an error that the key exists.

### DIFF
--- a/deps/key-value-storage/src/postgres/mod.rs
+++ b/deps/key-value-storage/src/postgres/mod.rs
@@ -125,24 +125,41 @@ impl KeyValueStorage for PostgresClient {
             });
         }
 
-        let mut sql = format!(
-            "INSERT INTO {} ({KEY_COLUMN}, {VALUE_COLUMN}) VALUES ( $1, $2 )",
-            self.table
-        );
-
         if parameters.overwrite {
-            sql = format!("{sql} ON CONFLICT ({KEY_COLUMN}) DO UPDATE SET {VALUE_COLUMN} = $2");
+            let sql = format!(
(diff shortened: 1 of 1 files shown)


deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -125,15 +125,32 @@ impl KeyValueStorage for PostgresClient {
             });
         }
 
-        let mut sql = format!(
-            "INSERT INTO {} ({KEY_COLUMN}, {VALUE_COLUMN}) VALUES ( $1, $2 )",
+        if !parameters.overwrite {
```
mythi: from what I'm reading about SQL, `DO NOTHING ... RETURNING` can help to detect if an error should be returned

Xynnn007: yes. Let me fix this.

deps/key-value-storage/src/postgres/mod.rs
```diff
@@ -125,25 +125,42 @@ impl KeyValueStorage for PostgresClient {
...
+            let result = query(&sql)
+                .bind(key)
+                .bind(value)
+                .fetch_optional(&*self.pool)
+                .await
+                .map_err(|e| KeyValueStorageError::SetKeyFailed {
+                    source: e.into(),
+                    key: key.to_string(),
+                })?;
+            if result.is_none() {
```
mkulke, line 156: wouldn't the result be `Some(row)` on conflict?

Xynnn007: The original SQL is
```
INSERT INTO table (key, value)
VALUES ($1, $2)
ON CONFLICT (key) DO NOTHING
RETURNING *
```

when there is not conflict, the insertion succeeds and will return the inserted row. When there is a conflict, it will return nothing thus none. The logic doesn't seem to go against intuition but the unit test succeeds

mkulke: maked sense, thx

PR #1127, opened 2025-12-23
dcap: skip supplemental data test

fitzthum:
Since updating to 1.24, dcap expects the supplemental data field to be larger.

We have a test that has hard-coded supplemental data. We should update this, but for now let's just skip the test as we figure out how to generate new supplemental data.


Xynnn007:
This is blocking CI tests. Let me get this merged and we can fix it back at anytime.

cc @pawelpros

Xynnn007 (approved):

PR #1131, opened 2025-12-25
Update Key-value-storage and policy-engine

Xynnn007:
This PR mainly lays the groundwork for the dependency in #1092; please see the specific commit message.

Pre of #1129

### DIFF
--- a/deps/policy-engine/src/lib.rs
+++ b/deps/policy-engine/src/lib.rs
@@ -19,43 +19,36 @@ pub use key_value_storage::{KeyValueStorage, KeyValueStorageConfig};
 pub struct PolicyEngineConfig {
     /// The storage to store the policies.
     pub storage: KeyValueStorageConfig,
+}
 
-    /// The type of policy engine to use.
-    /// Currently, only Rego is supported.
-    pub policy_type: PolicyType,
+pub trait EngineTrait {
+    /// The suffix of the policy file.
+    /// Concrete policy engine backend may handle the policy in different ways.
+    /// For example, the policy engine may store the policy in a different format.
+    /// In this case, the policy engine may need to add a suffix to the policy id to distinguish the policy.
+    /// This is also for compatibility with the existing policy setting and getting
+    /// APIs. Concretely, users do not need to specify the `.rego` suffix.
+    fn policy_suffix() -> &'static str {
+        ""
+    }
 }
 
 #[derive(Clone)]
-pub struct PolicyEngine {
+pub struct PolicyEngine<T: Send + Sync + EngineTrait> {
     pub storage: Arc<dyn KeyValueStorage>,
-    pub engine: Arc<dyn Engine>,
+    pub engine: T,
 }
 
-impl PolicyEngine {
-    pub async fn new(config: PolicyEngineConfig) -> Result<Self> {
-        let storage = config.storage.to_key_value_storage().await?;
-        let engine = config.policy_type.to_engine();
-        Ok(Self { storage, engine })
-    }
-
-    pub async fn evaluate(
-        &self,
-        data: &str,
-        input: &str,
-        policy_id: &str,
-    ) -> Result<EvaluationResult> {
-        let policy = self.get_policy(policy_id).await?;
-        self.engine.evaluate(data, input, &policy).await
-    }
-
+impl<T: Send + Sync + EngineTrait> PolicyEngine<T> {
     /// Set a policy to the backend.
     /// The policy is expected to be provided as string.
     /// Concrete policy engine backend may handle the policy in different ways.
     pub async fn set_policy(&self, policy_id: &str, policy: &str, overwrite: bool) -> Result<()> {
         let params = SetParameters { overwrite };
+        let policy_id = format!("{}{}", policy_id, T::policy_suffix());
         let _ = self
             .storage
-            .set(policy_id, policy.as_bytes(), params)
+            .set(&policy_id, policy.as_bytes(), params)
             .await
             .map_err(PolicyError::from)?;
         Ok(())
@@ -63,14 +56,20 @@ impl PolicyEngine {
 
     /// List all policies in the backend.
     pub async fn list_policies(&self) -> Result<Vec<String>> {
-        self.storage.list().await.map_err(From::from)
+        let policies = self.storage.list().await?;
+        let policies = policies
+            .into_iter()
+            .map(|policy| policy.strip_suffix(T::policy_suffix()).map(|p|p.to_string()).ok_or(PolicyError::MalformedPolicy(anyhow::anyhow!("There is at least one policy in the storage with invalid name. The policy name should contain the policy suffix {}.", T::policy_suffix()))))
+            .collect::<Result<Vec<String>>>()?;
+        Ok(policies)
     }
 
     /// Get a policy from the backend.
     /// The policy is expected to be provided as string.
     /// Concrete policy engine backend may handle the policy in different ways.
     pub async fn get_policy(&self, policy_id: &str) -> Result<String> {
-        let policy_str = self.storage.get(policy_id).await?;
+        let policy_id = format!("{}{}", policy_id, T::policy_suffix());
+        let policy_str = self.storage.get(&policy_id).await?;
 
         match policy_str {
             Some(policy_str) => {
--- a/deps/policy-engine/src/policy/rego.rs
+++ b/deps/policy-engine/src/policy/rego.rs
@@ -2,28 +2,48 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-use async_trait::async_trait;
-use serde_json::Value;
-use tracing::instrument;
+use std::collections::HashMap;
 
-use crate::{Engine, EvaluationResult, PolicyError};
+use regorus::Extension;
+use serde_json::Value;
+use tracing::{info, instrument};
+
+use crate::{EngineTrait, EvaluationResult, PolicyEngine, PolicyEngineConfig, PolicyError, Result};
 
 /// The rule to evaluate the policy.
 /// Note that only the result of this rule will be returned.
 pub const EVAL_RULE: &str = "data.policy.result";
 
+pub struct RegorusExtension {
+    pub name: String,
+    pub id: u8,
+    pub extension: Box<dyn Extension + Send + Sync>,
+}
+
 #[derive(Debug, Clone, Default)]
 pub struct Regorus {}
 
-#[async_trait]
-impl Engine for Regorus {
+impl EngineTrait for Regorus {
+    fn policy_suffix() -> &'static str {
+        ".rego"
+    }
+}
+
+impl Regorus {
+    /// The inputs to an policy engine. Inspired by OPA, we divided the inputs
+    /// into three parts:
+    /// - `data`: static data that will help to enforce the policy.
+    /// - `input`: dynamic data that will help to enforce the policy.
+    /// - `policy`: the policy to be enforced.
     #[instrument(skip_all, name = "Regorus")]
-    async fn evaluate(
+    pub async fn evaluate(
         &self,
-        data: &str,
+        data: Option<&str>,
         input: &str,
         policy: &str,
-    ) -> Result<EvaluationResult, PolicyError> {
+        eval_rules: Vec<&str>,
+        extensions: Vec<RegorusExtension>,
+    ) -> Result<EvaluationResult> {
         let mut engine = regorus::Engine::new();
 
         let policy_hash = {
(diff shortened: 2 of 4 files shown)


deps/key-value-storage/src/lib.rs
```diff
@@ -27,10 +27,15 @@ pub struct SetParameters {
     pub overwrite: bool,
 }
 
+pub enum SetResult {
+    Inserted,
+    AlreadyExists,
+}
```
fitzthum, line 33: Some std stuff uses an option for this i.e. return Some(val) if val was already there. Might be simpler. Up to you.

deps/key-value-storage/src/local_json/mod.rs
```diff
@@ -82,10 +82,7 @@ impl KeyValueStorage for LocalJson {
             .map_err(|e| KeyValueStorageError::MalformedValue { source: e.into() })?;
         let value_b64 = URL_SAFE.encode(value);
         if parameters.overwrite && items.contains_key(key) {
-            return Err(KeyValueStorageError::SetKeyFailed {
-                source: anyhow::anyhow!("key already exists"),
-                key: key.to_string(),
-            });
+            return Ok(SetResult::AlreadyExists);
```
fitzthum, line 85: What about making a new error variant for a key that already exists? Maybe I am missing the point of this change slightly.

deps/policy-engine/src/lib.rs
```diff
@@ -19,58 +19,61 @@ pub use key_value_storage::{KeyValueStorage, KeyValueStorageConfig};
...
-    /// Currently, only Rego is supported.
-    pub policy_type: PolicyType,
+pub trait EngineTrait {
+    /// The suffix of the policy file.
+    /// Concrete policy engine backend may handle the policy in different ways.
+    /// For example, the policy engine may store the policy in a different format.
+    /// In this case, the policy engine may need to add a suffix to the policy id to distinguish the policy.
+    /// This is also for compatibility with the existing policy setting and getting
+    /// APIs. Concretely, users do not need to specify the `.rego` suffix.
+    fn policy_suffix() -> &'static str {
```
fitzthum, line 31: Is this really necessary? Why don't we just store the policy with whatever name the user provides?

deps/policy-engine/src/lib.rs
```diff
@@ -19,58 +19,61 @@ pub use key_value_storage::{KeyValueStorage, KeyValueStorageConfig};
...
-        self.engine.evaluate(data, input, &policy).await
-    }
-
+impl<T: Send + Sync + EngineTrait> PolicyEngine<T> {
     /// Set a policy to the backend.
     /// The policy is expected to be provided as string.
     /// Concrete policy engine backend may handle the policy in different ways.
     pub async fn set_policy(&self, policy_id: &str, policy: &str, overwrite: bool) -> Result<()> {
         let params = SetParameters { overwrite };
+        let policy_id = format!("{}{}", policy_id, T::policy_suffix());
```
fitzthum, line 48: Are we doing anything to make sure that there are no collisions between policies and other things in KV storage like resources?

deps/policy-engine/src/policy/rego.rs
```diff
@@ -2,28 +2,48 @@
...
+use tracing::{info, instrument};
 
-use crate::{Engine, EvaluationResult, PolicyError};
+use crate::{EngineTrait, EvaluationResult, PolicyEngine, PolicyEngineConfig, PolicyError, Result};
 
 /// The rule to evaluate the policy.
 /// Note that only the result of this rule will be returned.
 pub const EVAL_RULE: &str = "data.policy.result";
 
+pub struct ExtensionWrapper {
```
fitzthum: nit: I think you can just call this `RegorusExtension` (to distinguish it from EAR extensions)

deps/key-value-storage/src/lib.rs
Xynnn007: You're right, to be honest, I've been thinking about this problem all day. I considered several influencing factors:

1. Currently, the parameters allow for non-overwriting writes.

2. How to demonstrate that the value hasn't been overwritten.

3. If returning the existing value, we need to add another logic to get the current value, and this logic should be atomic with the existence check


The main issue, specifically regarding scenario 3 in the database context, led me to abandon the approach of returning the original value. I believe this is an imperfect outcome, requiring that if the caller wants to retrieve a value, they should call `get`, not `set`, but still it meets our needs of following PRs.


Let's leave it as it is in the PR, and welcome any ideas that require changes—as long as the upper-level caller needs.

deps/key-value-storage/src/local_json/mod.rs
Xynnn007: The reason for this is that, in scenarios where no overwriting is required, inserting an existing value should not result in an error, but rather a notification that the value was not written.

This is used when setting the default policy in AS. When the user's policy storage backend is not configured with a policy, the policy configuration is written; if it is configured, the write will not overwrite and will not report an error.

deps/policy-engine/src/lib.rs
Xynnn007: I also do not like this design too much tbo, mainly for compatibility with higher-level systems. Let me describe the scenario and I'd like to hear your opinion:

1. (Advantage, compatibility) For traditional localFS, if we only use keys for storage, a policy originally named `default_cpu` will directly use the name `default_cpu` instead of `default_cpu.rego` if we do not use this suffix.

2. (Disadvantage, strange for non-localFS systems) For a database as the backend storage, for example, if we have a policy, it would be strange to represent it as `policy.rego` in the table.

deps/policy-engine/src/lib.rs
Xynnn007: Good question. Actually, for the database as the backend, each specific configuration point can define a table (with default values ​​if not configured). We can store AS policy, KBS policy, resources, etc., on different tables within the same database to prevent cross-domain conflicts.

Xynnn007:
> The change in the abstraction is a little confusing. I guess the properties of the Rego backend (setting extensions, getting data and input), etc) are too specific to implement in one generic policy trait.

Yes. Especially extensions.

mythi:
note: this PR does not trigger `CoCo-AS Rust Suites` tests at all.

deps/policy-engine/src/lib.rs
```diff
@@ -19,56 +19,61 @@ pub use key_value_storage::{KeyValueStorage, KeyValueStorageConfig};
...
-        self.storage.list().await.map_err(From::from)
+        let policies = self.storage.list().await?;
+        let policies = policies
+            .into_iter()
+            .filter_map(|policy| {
+                policy
+                    .strip_suffix(T::policy_suffix())
+                    .map(|policy| policy.to_string())
+            })
+            .collect();
```
mkulke: if we want to drop all suffixes regardless of whether they exist or not we could use:

```
policies = policies.iter().map(|p| p.trim_end_matches(T::policy_suffix()).into()).collect();
```

this is slightly different than discarding all policies that do not have a suffix. I'm not sure either approach is a good idea.

if we encounter something that we do not expect (there's is a policy that should have a suffix, but there's none) we should bail out with an error, in my opinion. being forgiving (silently dropping) on faulty input data is almost never a good idea, especially for security-sensitive applications.

Xynnn007: Good point. Let me take your advice (or not) once we have a conclusion on https://github.com/confidential-containers/trustee/pull/1131#discussion_r2654934465 - could you share some idea on that?

deps/policy-engine/src/lib.rs
mkulke: In abstract terms I would expect every storage implementation to handle that without leaking storage specifics to upper layers. For a database it would be id=$POLICY_NAME in table "policies", for a the filesystem a path /policies/$POLICY_NAME.rego, for redis it might be HGET policies $POLICY_NAME etc.

This implies that the storage has some awareness of the objects it is managing, which is fine, we don't build a generic k/v store here. there are various ways to do this, using generics and putting the suffix in a Trait:

```
trait Storage {
    fn get<T: Storable>(&self, id: &str) -> Result<T>;
    ...
}

trait Storable {
    fn collection_name() -> &'static str;
    fn suffix() -> Option<&'static str> {
        None
    }
}
```

I would also recommend fixing the table names (i.e. making the collection names non-configurable: "policies", "reference_values", etc) and, similarly, hashmap names and subfolders on the fs. this will help with migrations later on and I'm not sure there is an upside this. in a database the schema belongs to an application (you'd use database objects, tablespaces or similar namespacing features to host multiple schemas in a DB)

fitzthum: It might be simpler to just escape/replace the `.` as needed for each backend and not have any suffix abstraction.

Either way, I agree that the we probably don't want to monkey around with the names too much in the policy code. Ideally the storage backend would handle this.

Another approach would be to add method to the storage interface to store rego specifically. This is less generic than what Magnus suggested. We might also use an enum for the collection type rather than allowing any value? Possibly the collection type and the suffix could be combined into one? Either way, I think it's a good suggestion.

deps/policy-engine/src/lib.rs
fitzthum: I like what Magnus suggested regarding a collection on the other thread. One interesting idea would be to use a URI (like the resource URI) to point to things in kv storage. We could use the first part to delineate the type of resource. Not sure if this is optimal, but just throwing it out there.

deps/key-value-storage/src/local_json/mod.rs
fitzthum: why not just return `Ok` if the policy already exists (and overwrite is set to false)? Maybe log an error

Xynnn007: this is a way for upper caller to know whether the key exists while logging cannot

deps/policy-engine/src/lib.rs
Xynnn007: > I would also recommend fixing the table names (i.e. making the collection names non-configurable: "policies", "reference_values", etc) and, similarly, hashmap names and subfolders on the fs. this will help with migrations later on and I'm not sure there is an upside this. in a database the schema belongs to an application (you'd use database objects, tablespaces or similar namespacing features to host multiple schemas in a DB)

What about still keeping the configurable table names, while providing a standard configuration file template and an SQL file to initialize the entire database? The reason is that the current lib is unaware of the upper layer, and it's somewhat intrusive when we can only define table names from the upper layer.

I believe that for ordinary users, whether using a database (db) or a filesystem (fs), the only thing they need to care about is the actual content of the files stored on the backend. Certainly, it maintains some flexibility. 


> It might be simpler to just escape/replace the `.` as needed for each backend and not have any suffix abstraction.

Now users does not directly touch `.`, it will be automatically added to the `<policy-name>` as `.rego`. So we might not need to handle `.` explicitly, but only maintain a "allowed" alphabet for key names. Then the backend will handle the actual storage things transparently.

> Another approach would be to add method to the storage interface to store rego specifically. This is less generic than what Magnus suggested. We might also use an enum for the collection type rather than allowing any value? Possibly the collection type and the suffix could be combined into one? Either way, I think it's a good suggestion.

Another interfae to store rego is not a good idea, tbo. As common key-value interface can meet the needs enough. What do you mean by "collection type"? Where do we keep them and what does it do? Can you share some code examples?

mkulke: > The reason is that the current lib is unaware of the upper layer, and it's somewhat intrusive when we can only define table names from the upper layer.

i know it's tempting, but IMO we can (and should) couple the storage implementation to the application schema (that means that the code interfacing with the db, filesystem is aware of the application objects). the database or the filesystem is a storage abstraction, we don't need an abstraction on top of those (a generic key/value storage, applicable to everything).

A database schema is an implementation detail of an application that a user shouldn't be concerned about. Making it a configuration surface will increase complexity (without a good use case, at least I'm not aware of one). Migration scripts (*.up.sql, etc) now have to be templated and process configuration to enforce constraints (you cannot use the same table for policy and reference_values, etc...)

Xynnn007: Let me imagine a user configuration process following your line of thought. For an all-in-one KBS, the user only needs to configure a data-backend or similar option in the KBS configuration file that specifies the db (address/port/pswd/..) or fs(path). And the table names and directory names of underlying items will be hardcoded.

For microservice deployment models, RVPS, AS, and KBS will all have such an option. Is it reasonable for me to push things forward in this direction?

mkulke: yes, however the configuration of the storage all-in-one vs microservices is a discrete issue.

in theory every (micro)service should have its own schema and maybe a user might want to use a different storage backend on KBS and on RVSP. but, that might also be a valid choice for all-in-one. I think it's fair to then prescribe a microservice deployment model to the user.

A user might also want to have their KBS secrets backed by a file storage (mounted from some secret vault), but the KBS policies should be in a database. so... now we have per-class storage configuration...

I don't think this all this has to be addressed initially; we can iterate from restricted to more flexible configuration options later, when concrete use cases are being requested by users. as long as we don't have a large api/config surface to support this is easy, because it's an internal refactoring. once an api or config surface is exposed it's harder to iterate.

deps/key-value-storage/src/local_json/mod.rs
fitzthum: Why does the caller need to know?

Xynnn007: Mostly there are two reasons

1. The design principle of separation of concerns: how to respond to successful/incomplete write operations should be driven by the caller, and the underlying library should faithfully return the processing result.

2. A typical scenario, such as when we add Prometheus indicators at a higher level, it also facilitates the statistical analysis of write operations.

fitzthum: Well doesn't really seem necessary to me, especially since overwrite is only set in a tiny number of cases, but I guess it doesn't matter much either way.

Xynnn007: So let me keep the code as-is,  and welcome any changes later if it makes sense to upper caller and necessary.

deps/policy-engine/src/lib.rs
Xynnn007: I can implement this in the next integration PR.

deps/policy-engine/src/lib.rs
Xynnn007: @mkulke Ah, I finally get your point - sorry for my "jet lag" - to raise an error if the policies stored inside the storage do not have an expected suffix, rather than skip them. Right? I think it makes sense, oce the concrete storage instance is dedicated to the policy engine.

mkulke (approved):

PR #1133, opened 2025-12-28
Update supplemental data test to generate dynamic data

iroykaufman:
Replaced hard-coded supplemental data in the test with a new function that generates the data dynamically. This change ensures the test remains valid and adaptable to future updates in the data structure.

### DIFF
--- a/deps/verifier/src/intel_dcap/claims.rs
+++ b/deps/verifier/src/intel_dcap/claims.rs
@@ -146,20 +146,10 @@ mod tests {
     use serde_json::json;
 
     #[test]
-    #[ignore]
     fn parse_supplemental_data() {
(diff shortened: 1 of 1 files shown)


deps/verifier/src/intel_dcap/claims.rs
```diff
@@ -202,4 +192,40 @@ mod tests {
 
         supp
     }
+    fn create_supp_data() -> String {
```
Xynnn007, line 196: can you add a blank line here between the last code block and the function?

Xynnn007 (approved):
Thanks @iroykaufman ! Only one nit

mythi:
@iroykaufman thanks!

PR #1138, opened 2025-12-31
CI: fix rvps rust check

Xynnn007:
use ** to capture any change under rvps directory


mythi (approved):
LGTM.

Perhaps we could revisit the `paths` filter for the workflows. I remember this hitting us in the past as well.

PR #1142, opened 2026-01-05
snp-verifier: offline vcek cache support

amd-aliem:
Implement a filesystem based cache option that enables specifying a directory which may be preloaded with certs in http-cache-reqwest format for offline trustee deployments.

Benefits of enabling the cache:
- Allows normal function of trustee even if KDS is unreachable. This applies for completely air-gapped situations, as well as transient KDS or external network outages.
- Improves timing performance of KDS requests, as the bottleneck is generally reaching out to KDS to fetch a certificate or verify that the certificate is still valid.

Drawbacks of enabling the cache:
- There is no checking of certificate revocation lists. Certificate revocation events are generally very rare and certificates normally are valid for years - but be aware that revoked certificates must be manually ejected by the administrator.

Notes:
- [AMD Offline Certificate Cache](https://github.com/amd-aliem/trustee/blob/amd-offline-cache/attestation-service/docs/amd-offline-certificate-cache.md) is the main doc describing the steps to get an air-gapped deployment with the appropriate vceks
- Tested KDS + filesystem store priorities by blocking traffic to my trustee container and/or removing certificates from filesystem. I saw it correctly falling back to secondary option if it fails in both cases.

### DIFF
--- a/attestation-service/docs/amd-offline-certificate-cache.md
+++ b/attestation-service/docs/amd-offline-certificate-cache.md
@@ -1,0 +1,185 @@
+# Offline AMD VCEK Caching Guide
+
+This document describes the process to pre-load VCEKs (Versioned Chip
+Endorsement Keys) into the trustee environment to allow normal attestation
+without connection to the AMD KDS. This guide mainly targets air-gapped
+environments.
+
+> [!Note]
+> Currently this guide shows a method that builds on using
+> [docker](https://confidentialcontainers.org/docs/attestation/installation/docker/)
+> to manage kbs services. Additional deployment methods will be covered in future releases.
+
+## Enabling Offline VCEK Store in Attestation Service
+
+### 1. Set VCEK Sources Configuration Option
+
+Update the attestation configuration file to use a predefined vcek store:
+
+`kbs/config/as-config.json`:
+
+```json
+{
+    // ... other fields ...
+    "verifier_config": {
+        "snp_verifier": {
+          // Configure VCEK sources to try, in order. Defaults to [KDS].
+            "vcek_sources": [
+                {
+                  "type": "OfflineStore",
+                },
+                // Optionally add a fallback to KDS. Leave out if KDS will not
+                // be reachable.
+                //{
+                //  "type": "KDS",
+                //}
+            ]
+        }
+    }
+}
+```
+
+With the `OfflineStore` configuration specified, Trustee will inspect the
+configured directory for VCEK values following the following format:
+
+```
+/opt/confidential-containers/attestation-service/kds-store/vcek/{hwid}/vcek.der
+```
+
+Where `{hwid}` is the hardware ID of the AMD EPYC server in lowercase hexadecimal.
+
+### 2. Create and Populate VCEK directory
+
+Create a `vcek` directory populated with the following structure:
+
+```
+vcek/
+├── <Hex hardware ID>/ (ID must be lowercase)
+│   └── vcek.der (cert pre-downloaded from kdsintf.amd.com)
+├── <Hex hardware ID>/
+│   └── vcek.der
+├── <Hex hardware ID>/
+    └── vcek.der
+```
+
+Trustee requires one unique certificate per physical AMD EPYC server that the
+KBS will be servicing. The server's hardware ID and certificate's URL can be
+fetched using the [snphost tool](https://github.com/virtee/snphost).
+
+On each AMD SNP host, with root/admin access run:
+```
+sudo snphost show vcek-url
+```
+
+You should get a URL of the form:
+```
+https://kdsintf.amd.com/vcek/{version}/{machine}/{product_name}/{hwid}?{params}
+```
+
+You may download it from a browser by pasting that URL, or you can run
+the following command on any server with network access to AMD's KDS.
+```
+sudo snphost fetch vcek der .
+```
+
+> [!IMPORTANT]
+> - Note that the VCEK URL is specific to the hardware AMD firmware of the
+> machine. If the firmware is updated, the VCEK URL will change. See the
+> [AMD VCEK documentation](https://docs.amd.com/api/khub/documents/dWGhwYpo1Wv51rJN4d~47g/content)
+> for more information about the VCEK URL format.
+
+### 3. Install `vcek` directory into trustee deployment
+
+- **Install in running trustee deployment**
+
+Use docker commands to copy your `vcek` folder into the configured directory:
+```
+sudo docker exec -it trustee-as-1 mkdir -p /opt/confidential-containers/attestation-service/kds-store/
+sudo docker cp ./vcek/ trustee-as-1:/opt/confidential-containers/attestation-service/kds-store/vcek/
+```
+
+- **Mount shared directory**
+
+You may also mount a shared directory from the host into the container by
+updating `docker-compose.yml` with a specified directory mapping:
+
+```yaml
+  as:
+    ... <existing configuration>
+    volumes:
+    ... <existing volumes>
+    - ./vcek:/opt/confidential-containers/attestation-service/kds-store/vcek:rw
+```
+
+### Example:
+
+For some number of AMD EPYC servers you wish for trustee to service:
+```
+ssh privileged-user@epyc-host "sudo snphost show vcek-url" >> urls.txt
+```
+
+On a system with network access to AMD KDS:
+```bash
+
+# Create the vcek folder that will be copied to trustee
+mkdir vcek
+
+# Fetch certificates using the above URLs file
+# There should be one line for each EPYC host that kds will service
+cat urls.txt | while read line; do
+  hwid="$(echo "$line" | cut -d/ -f7 | cut -d'?' -f1 | tr '[:upper:]' '[:lower:]')"
+  mkdir vcek/$hwid
+  cd vcek/$hwid
+  sudo snphost fetch vcek der .
+  cd ../..
+done
+
+# Copy the archive to your air-gapped trustee attestation-service deployment
+scp -r vcek user@target-host:/path/to/trustee
+```
+
+On the air-gapped trustee attestation-service host:
+```bash
+# Update as-config.json to enable Disk Caching
+cd /path/to/trustee
+vi kbs/config/as-config.json
+# Update the verifier_config section to:
+#     "verifier_config": {
+#        "snp_verifier": {
+#            "vcek_sources": [
+#              {
+#                "type": "OfflineStore",
+#              }
+#            ]
+#        }
+#    }
+
+# Start trustee
+docker compose up -d
+
+# Copy the vcek store into the running attestation service container
+sudo docker exec -it trustee-as-1 mkdir -p /opt/confidential-containers/attestation-service/kds-store/
+sudo docker cp ./vcek/ trustee-as-1:/opt/confidential-containers/attestation-service/kds-store/vcek/
+
+# Alternative to docker copy would be to add the following shared mount to
+# docker-compose.yml under as.volumes section:
+# - ./vcek:/opt/confidential-containers/attestation-service/kds-store/vcek:rw
+```
+
+## Limitations
+
+VCEK stores must be updated/rebuilt in the following events:
+- AMD EPYC host added to serviced cluster
+- AMD EPYC host firmware components updated
+- VCEK certificate revoked by CA
+
+## Troubleshooting
+
+Enable debug and check logs for `vcek` keyword. Ensure configured sources match
+expected values.
+
+```
+echo "RUST_LOG=debug" > debug.env
+docker compose --env-file debug.env up -d
+docker logs trustee-as-1 | grep -i vcek
+```
--- a/deps/verifier/src/snp/mod.rs
+++ b/deps/verifier/src/snp/mod.rs
@@ -64,6 +64,9 @@ pub(crate) const FMC_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .9)
 const KDS_CERT_SITE: &str = "https://kdsintf.amd.com";
 const KDS_VCEK: &str = "/vcek/v1";
 
+// KDS Offline Store
+const KDS_OFFLINE_STORE_PATH: &str = "/opt/confidential-containers/attestation-service/kds-store";
+
 /// Attestation report versions supported
 const REPORT_VERSION_MIN: u32 = 3;
 const REPORT_VERSION_MAX: u32 = 5;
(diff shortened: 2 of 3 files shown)


deps/verifier/src/snp/mod.rs
```diff
@@ -119,17 +123,28 @@ impl Snp {
...
+                let cache = Cache(HttpCache {
+                    mode: CacheMode::ForceCache,
+                    manager: CACacheManager::new(path.clone().into(), true),
+                    options: client_options,
+                });
+                ClientBuilder::new(reqwest::Client::new())
+                    .with(cache)
+                    .build()
+            }
+            _ => {
```
fitzthum: So if `vcek_cache` is set to `None` we still have a cache? That doesn't seem quite right. Maybe we want three options or at least a different name.

tools/trustee-cli/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+### **Example:**
+The following examples create a cache-preload directory `./cache` filled with
+the URLS in `./urls.txt`.
+
+
+1. Create a url file with desired urls:
+```bash
+echo "http://github.com
+https://github.com/confidential-containers/trustee" > urls.txt
```
fitzthum: May want to use a url for a vcek here for a more useful example.

tools/trustee-cli/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+The Trustee CLI (`trustee`) is a unified management tool that deploys and manages
+trustee components: the Key Broker Service (KBS), Attestation Service (AS), and
+Reference Value Provider Service (RVPS).
+
+# Table of Contents
+
+- [Quick Start](#quick-start)
+- [Installation](#installation)
+  - [Docker (Recommended)](#docker-recommended)
```
fitzthum: Why is docker recommended?

tools/trustee-cli/README.md
```diff
@@ -0,0 +1,147 @@
+# Trustee CLI
```
fitzthum: This doc might be beyond the scope of this PR.

tools/trustee-cli/src/cache_preload.rs
```diff
@@ -0,0 +1,138 @@
...
+        .get(url)
+        .send()
+        .await
+        .context(format!("Failed to fetch URL: {}", url))?;
+
+    let status = response.status();
+    let duration = start.elapsed();
+
+    if status.is_success() {
+        println!("✓ {} ({}ms)", url, duration.as_millis());
```
fitzthum: We don't seem to use `println` elsewhere. Probably better to do `info!` here.

attestation-service/docs/amd-offline-certificate-cache.md
```diff
@@ -0,0 +1,246 @@
...
+
+# Copy the cache into the running attestation service container
+sudo docker cp ./vcek-cache/ trustee-as-1:/opt/confidential-containers/vcek-cache/
+
+# Alternative to above copy would be to add to following to
+# docker-compose.yml under as.volumes section:
+# - ./vcek-cache:/opt/confidential-containers/vcek-cache:rw
+```
+
+## Understanding the Cache Structure
```
fitzthum: Not necessary.

attestation-service/docs/amd-offline-certificate-cache.md
```diff
@@ -0,0 +1,246 @@
...
+  certificates) indexed by content hash
+- **index-v5/**: Stores cache metadata including URL-to-content mappings,
+  headers, and cache control information
+
+When transferring the cache to an air-gapped environment, you must preserve
+this exact directory structure. The entire cache directory should be copied
+as-is to `/opt/confidential-containers/vcek-cache` in the attestation service
+container.
+
+## Limitations
```
fitzthum, line 169: This section can be like one sentence.

fitzthum (commented):
Looks ok. A few comments on the code. 

Also tbh I don't really like the docs. They can be a lot more concise.

tools/trustee-cli/README.md
amd-aliem: Fair, removed for now! Whenever someone starts one I can review the cache-preload subcommand

tools/trustee-cli/README.md
amd-aliem: Perhaps I was misunderstanding the purpose of the tool - I thought it was supposed to also be running trustee containerized as well.

Also it took me the better part of a day to figure out how to build it on x86 epyc without docker - I kept running into the intel package missing panic, then found there was no reasonable way to pass the `--snp-verifier` feature flag due to how the dependencies worked, then installed the intel packages but for whatever reason the ones available to me on ubuntu 25.04 via the instructions did not include the most recent one... I did finally get the right ones, but it was quite a headache. So I thought from an amd user perspective it would also be much easier to use docker.

But anyways I'll leave it to someone else to create docs for this :)

fitzthum:
As I mentioned on the other PR(s) I don't think the points against caching make much sense. That said, the vcek sources config looks fine and it does allow us to have multiple fallbacks. This is more complex, but the design seems fine. The AMD KDS should probably be a default option so that no extra configuration is required in basic cases.

attestation-service/docs/amd-offline-certificate-cache.md
amd-aliem: shortened!

deps/verifier/src/snp/mod.rs
```diff
@@ -222,6 +282,28 @@ impl Snp {
...
+pub struct SnpVerifierConfig {
+    #[serde(default)]
+    pub vcek_sources: Vec<VCEKSource>,
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[serde(tag = "type")]
+pub enum VCEKSource {
+    OfflineStore {
+        path: String,
```
mkulke: Maybe Option<String> so we can use the default base directory, so the users do not have to specify path = "..." if they want to use the default directory

attestation-service/docs/amd-offline-certificate-cache.md
```diff
@@ -0,0 +1,185 @@
...
+        }
+    }
+}
+```
+
+With the `OfflineStore` configuration specified, Trustee will inspect the
+configured directory for VCEK values following the following format:
+
+```
+/${CONFIGURED_VCEK_BASE}/vcek/{hwid}/vcek.crt
```
mkulke: $CONFIGURED_VCEK_BASE is not introduced

attestation-service/docs/amd-offline-certificate-cache.md
```diff
@@ -0,0 +1,185 @@
...
+# Update as-config.json to enable Disk Caching
+cd /path/to/trustee
+vi kbs/config/as-config.json
+# Update the verifier_config section to:
+#     "verifier_config": {
+#        "snp_verifier": {
+#            "vcek_sources": [
+#              {
+#                "type": "OfflineStore",
+#                "path": "/opt/confidential-containers/"
```
mkulke:
```suggestion
```

deps/verifier/src/snp/mod.rs
```diff
@@ -132,25 +134,83 @@ impl Snp {
...
+        );
+        bail!("Failed to fetch VCEK from any configured source")
+    }
+
+    fn fetch_vcek_from_offline_store(
+        &self,
+        att_report: AttestationReport,
+        proc_gen: &ProcessorGeneration,
+        path: String,
+    ) -> Result<Vec<u8>> {
```
mkulke, line 175:
```suggestion
    fn fetch_vcek_from_offline_store(
        &self,
        att_report: AttestationReport,
        proc_gen: &ProcessorGeneration,
        path: Option<String>,
    ) -> Result<Vec<u8>> {
        // default dir should contain the /vcek segment
        let path = pathh.unwrap_or_default(DEFAULT_AMD_VCEK_PATH);
```

mkulke (approved):
looks good. some nits about handling a default vcek directory

deps/verifier/src/snp/mod.rs
amd-aliem: good idea, fixed (and changed the `base_url` option as well to match)

attestation-service/docs/amd-offline-certificate-cache.md
amd-aliem: fixed!

deps/verifier/src/snp/mod.rs
```diff
@@ -132,25 +137,83 @@ impl Snp {
...
 
+    /// Fetches VCEK by trying each configured source in order until one succeeds
+    async fn fetch_vcek_from_sources(
+        &self,
+        att_report: AttestationReport,
+        proc_gen: ProcessorGeneration,
+    ) -> Result<Vec<u8>> {
+        if self.verifier_config.vcek_sources.is_empty() {
+            return self.fetch_vcek_from_kds(att_report, proc_gen, None).await;
+        }
```
mythi: how about making `VCEKSource::KDS` as the serde default. With that, this extra check is not needed.

deps/verifier/src/snp/mod.rs
```diff
@@ -132,25 +137,83 @@ impl Snp {
...
+            _ => hex::encode(att_report.chip_id),
+        }
+    }
+
     /// Asynchronously fetches the VCEK from the Key Distribution Service (KDS) using the provided attestation report.
     /// Returns the VCEK in DER format.
     async fn fetch_vcek_from_kds(
         &self,
         att_report: AttestationReport,
         proc_gen: ProcessorGeneration,
```
mythi: this could be made a reference too

deps/verifier/src/snp/mod.rs
```diff
@@ -222,6 +287,19 @@ impl Snp {
...
+    pub vcek_sources: Vec<VCEKSource>,
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[serde(tag = "type")]
+pub enum VCEKSource {
+    OfflineStore { path: Option<String> },
+    KDS { base_url: Option<String> },
+}
+
```
mythi: these are rather generic configs that could serve other TEEs too (e.g., TDX we could map "enable cache", "collateral url" to what TDX uses). The question is do we want to keep the config snp-verifier specific for now or try to have a global config entry

mkulke (approved):
thanks for considering the suggestions, IMO it's fine to merge the change as is, we can internally refine the solution later to use a storage backend and consolidate it with other certs stores and collateral caches without breaking the config/user surface.

attestation-service/docs/amd-offline-certificate-cache.md
```diff
@@ -0,0 +1,184 @@
...
+            "vcek_sources": [
+                {
+                  "type": "OfflineStore",
+                },
+                // Optional fallback to KDS. Switch order if KDS fetch should
+                // be attempted first.
+                //{
+                //  "type": "KDS",
+                //}
+            ]
```
fitzthum, line 36: Seems like this is now a default option, so maybe this comment should change a little bit.

deps/verifier/src/snp/mod.rs
```diff
@@ -64,6 +64,9 @@ pub(crate) const FMC_SPL_OID: Oid<'static> = oid!(1.3.6 .1 .4 .1 .3704 .1 .3 .9)
 const KDS_CERT_SITE: &str = "https://kdsintf.amd.com";
 const KDS_VCEK: &str = "/vcek/v1";
 
+// KDS Offline Store
+const KDS_OFFLINE_STORE_PATH: &str = "/opt/confidential-containers";
```
fitzthum: not sure this is the best path. maybe something a little more specific would be good. like `opt/confidential-containers/attestation-service/kds-store`. Also note that in other places we use a shared `work-dir` variable to keep all the paths together. We should use that here too. Also potentially could make this configurable but not really necessary.

deps/verifier/src/snp/mod.rs
```diff
@@ -132,25 +137,83 @@ impl Snp {
             .build()
     }
 
+    /// Fetches VCEK by trying each configured source in order until one succeeds
+    async fn fetch_vcek_from_sources(
```
fitzthum: can probably just call this `fetch_vcek`.

amd-aliem: sounds good, updating

deps/verifier/src/snp/mod.rs
amd-aliem: Sure I can update the default to `/opt/confidential-containers/attestation-service/kds-store`. About using `work_dir`, do you mean accessing the attestation config's work_dir and using it as the base path for kds-store?

```
cat kbs/config/as-config.json
{
    "work_dir": "/opt/confidential-containers/attestation-service",
    "policy_engine": "opa",
    ....
    "verifier_config": {
        "snp_verifier": {
            	"vcek_sources": [
				{
            	    "type": "OfflineStore"
				}
            ]
        }
    }
}
```
I would maybe have to pass it into the snp_verifier when creating it 🤔 https://github.com/confidential-containers/trustee/blob/15ae6593c390d8c8d597e40758a877e6362a1aa0/attestation-service/src/config.rs#L13-L19

The OfflineStore type does allow a configurable path rn, which I think we'd expect to be an absolute path that doesn't necessarily point to something in the work_dir:
```
pub enum VCEKSource {
    OfflineStore { path: Option<String> },
    KDS { base_url: Option<String> },
}
```

attestation-service/docs/amd-offline-certificate-cache.md
amd-aliem: Reworked wording about it - trying to convey the ordering rules (add KDS if you want a fallback, otherwise KDS will never be attempted).

deps/verifier/src/snp/mod.rs
amd-aliem: fixed!

deps/verifier/src/snp/mod.rs
amd-aliem: fixed!

deps/verifier/src/snp/mod.rs
fitzthum: I guess relatively easy way to handle is to import the default work dir variable into this file and concat a path to the end of that. Not perfect but that's what we do in some other places. It's fine for the configurable path to be absolute.

amd-aliem: Also does it make sense for `kds-store` to be under `attestation-service` 🤔 
Full path is now:
```
/opt/confidential-containers/attestation-service/kds-store/vcek/{hwid}/vcek.crt
```

fitzthum: Yes the verifiers are generally part of the attestation service. Although they're in a separate crate so you probably won't actually be able to import from the AS (without causing a circular dependency).

attestation-service/docs/amd-offline-certificate-cache.md
fitzthum: Unless nothing is specified, then the KDS will be used, right?

deps/verifier/src/snp/mod.rs
amd-aliem: Yeah having trouble importing. I remember someone had linked this: https://github.com/confidential-containers/trustee/pull/992 that could help in the future. Perhaps I can just hardcode it right now until something like that is implemented? I see that's being done in some other RVPS code: https://github.com/confidential-containers/trustee/blob/15ae6593c390d8c8d597e40758a877e6362a1aa0/rvps/src/storage/local_fs/mod.rs#L18

attestation-service/docs/amd-offline-certificate-cache.md
amd-aliem: Correct

deps/verifier/src/snp/mod.rs
fitzthum: Yeah not ideal but not worth fixing in this PR I guess.

deps/verifier/src/snp/mod.rs
amd-aliem: Perhaps in the future the same struct definitions can be used within the implementation-specific configs, so they can reject unsupported options as needed?

amd-aliem: Maybe can move the VCEK enums to `deps/verifier/src/lib.rs`

PR #1144, opened 2026-01-07
nvidia: make unit test generic and add cases

fitzthum:
Make the main nvidia verifier unit test more generic so we can use rstest to test different hardware evidence.

To add a test case, you just need the report, the expected report data, and the cert chain from a guest.

Add a test case that uses evidence pulled from a coco CI run.

Tests using the remote verifier continue to be skipped. I ran them locally. They might be unskipped in the future.


deps/verifier/src/nvidia/mod.rs
```diff
@@ -333,25 +333,35 @@ mod tests {
...
+    #[case::remote_verifier_coco(false, "87d8e24ab336adafe228d49e83d745f6dba4ae505372b6a5704820856b343fece279b616efefc2aae21da80cf5581250", include_str!("../../test_data/nvidia/hopper_coco_report1.txt"), include_str!("../../test_data/nvidia/hopper_coco_certs1.txt"))]
+    // The local verifier does not currently work with this report, which is from a newer device
+    // that has some unknown fields in opaque data.
+    #[ignore]
+    #[case::local_verifier_coco(true, "87d8e24ab336adafe228d49e83d745f6dba4ae505372b6a5704820856b343fece279b616efefc2aae21da80cf5581250", include_str!("../../test_data/nvidia/hopper_coco_report1.txt"), include_str!("../../test_data/nvidia/hopper_coco_certs1.txt"))]
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
-    async fn test_evaluation(#[case] local_verifier: bool) {
+    async fn test_evaluation(
+        #[case] local_verifier: bool,
+        // The the report data (as hex) that was used to create the evidence.
```
```suggestion
        // The expected report data (as hex) that was used to create the evidence.
```

deps/verifier/src/nvidia/mod.rs
```diff
@@ -333,25 +333,35 @@ mod tests {
...
+        #[case] expected_report_data: &str,
+        // Report data as a hex string
```
mythi: The second arg talks about "report data" too which is a different thing than this. Maybe
```suggestion
        // Evidence as a hex string
```

deps/verifier/src/nvidia/mod.rs
```diff
@@ -333,25 +333,35 @@ mod tests {
...
+    // Tests with the remote verifier are ignored to avoid putting strain on NRAS.
+    // Please run these tests if you make any changes to the verifier.
     #[ignore]
-    #[case::remote_verifier(false)]
+    #[case::remote_verifier(false, "931d8dd0add203ac3d8b4fbde75e115278eefcdceac5b87671a748f32364dfcb", include_str!("../../test_data/nvidia/hopperAttestationReport.txt"), include_str!("../../test_data/nvidia/hopper_cert_chain_case1.txt"))]
+    // Use the remote verifier with evidence from a CoCo CI run
+    #[ignore]
+    #[case::remote_verifier_coco(false, "87d8e24ab336adafe228d49e83d745f6dba4ae505372b6a5704820856b343fece279b616efefc2aae21da80cf5581250", include_str!("../../test_data/nvidia/hopper_coco_report1.txt"), include_str!("../../test_data/nvidia/hopper_coco_certs1.txt"))]
+    // The local verifier does not currently work with this report, which is from a newer device
+    // that has some unknown fields in opaque data.
```
mythi, line 345: something that needs to be fixed in the local verifier?

fitzthum: Yes. I think @cclaudio was looking into it at some point.

PR #1147, opened 2026-01-12
Add nvswitch support to the nvidia verifier

fitzthum:
Unfortunately I don't have public testing data for this (yet), but I have tested it internally. See commit messages for details.

Also, throw in Blackwell support.

### DIFF
--- a/attestation-service/src/ear_token/ear_default_policy_switch.rego
+++ b/attestation-service/src/ear_token/ear_default_policy_switch.rego
@@ -1,0 +1,72 @@
+package policy
+
+import rego.v1
+
+default hardware := 97
+
+default executables := 33
+
+default configuration := 36
+
+default file_system := 0
+
+default instance_identity := 0
+
+default runtime_opaque := 0
+
+default storage_opaque := 0
+
+default sourced_data := 0
+
+hardware := 2 if {
+	input.sampledevice.svn in data.reference.device_svn
+}
+
+trust_claims := {
+	"executables": executables,
+	"hardware": hardware,
+	"configuration": configuration,
+	"file-system": file_system,
+	"instance-identity": instance_identity,
+	"runtime-opaque": runtime_opaque,
+	"storage-opaque": storage_opaque,
+	"sourced-data": sourced_data,
+}
+
+# Switches verified by NRAS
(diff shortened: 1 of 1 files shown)


deps/verifier/src/nvidia/mod.rs
```diff
@@ -186,11 +200,14 @@ impl Nvidia {
         }
 
         let claims = response.claims()?;
+        println!("{:?}", serde_json::to_string(&claims));
```
Xynnn007: This is a leftover?

deps/verifier/src/nvidia/mod.rs
```diff
@@ -69,9 +70,21 @@ struct NvDeviceEvidence {
     device_evidence_list: Vec<NvDeviceReportAndCert>,
 }
 
+// NVML Wrapper does not know about switches,
+// so create our own enum for the device architecture.
```
mythi, line 74: could we keep it `DeviceArchitecture` and just stop using `nvml_wrapper::enums::device::DeviceArchitecture;`?

attestation-service/src/ear_token/ear_default_policy_switch.rego
```diff
@@ -0,0 +1,72 @@
...
+	"hardware": hardware,
+	"configuration": configuration,
+	"file-system": file_system,
+	"instance-identity": instance_identity,
+	"runtime-opaque": runtime_opaque,
+	"storage-opaque": storage_opaque,
+	"sourced-data": sourced_data,
+}
+
+# GPUs verified by NRAS
```
mythi:
```suggestion
# Switches verified by NRAS
```

deps/verifier/src/nvidia/mod.rs
fitzthum: Currently the local verifier uses the `DeviceArchitecture` struct in a few places. I kept both enums to avoid changing this. I could fixup the local verifier, but I'm not sure what it's future is so I left it for now.

mythi: Makes sense.

Xynnn007 (approved):

Issue #350, opened 2024-03-13
[RVPS] RVPS Update Design


Xynnn007:
Currently, we have Reference Value Provider Service (RVPS) as a black box aiming to provide the following functionalities
1. `register_reference_value(manifest)` Receive different format of reference value manifests, e.g. Sample format, in-toto (developing), CoRIM (TBD) and store the related reference values
2. `query_reference_value(claim_id)` Provide a querying API for users e.g. attestation service to query reference value to a specific key

The reference value inside the RVPS are some key-value pairs. The key is a measurement claim name, such as `mr_td` for TDX and `measurement` for SNP.
The value is the reference values of that field.

@thomas has raised a core issue named compound reference value problem upon current design.

The concrete model is described [here](https://github.com/confidential-containers/trustee/issues/238#issuecomment-1825517010)

Thus, a new concept name `Target Environment (TE)` is brought as defined [in RATS](https://www.rfc-editor.org/rfc/rfc9334.html#section-3.1).

A reference value can comprise one or more measurement claims, this means that only if the all claims
in the reference value and TE's measurements are matched altogether, the verification against the reference value
succeeds. A reference value applies to a given TE.

Reference values can be provided in "manifests" comprising one or more reference value.

The updated RVPS will update the API like the following
1. `register_reference_value(manifest)` will not change.
2. `query_reference_value(claim_id)` will be updated to `query_reference_value(TE_id)`. This means a reference value will be indexed by
an id of TE. The result reference value should be matched altogether.

Thus we need some related work along with the road.
1. How to generate `TE_id` from a given reference value manifest.
2. Some reference values ​​themselves have complex logic that goes beyond an exact match of a field.
For example, a security version number field is greater than a specific number. We might also do some change to facilitate the current OPA integration in CoCoAS.

## Roadmaps & Thoughts

- [ ] A better way to help reference values and OPA policy inside AS work together. We might need an intermediate strategy language for users to facilitate configuring attestation policy.
- [ ] Support CoRIM format as one of the reference value manifest in RVPS.
- [ ] Publish CoCo's TDX stack reference value in CoRIM and finish e2e test.
- [ ] Publish CoCo's Arm stack reference value in CoRIM and finish e2e test.

fitzthum:
It seems like one of the main concrete suggestions here is to update the RVPS to allow reference values to be grouped in terms of the so-called Target Environment. This seems like a good idea, but I"m still a bit fuzzy on how the RVPS should be used in practice.

[...]

Issue #238, opened 2023-11-23
RVPS limitations


thomas-fossati:
Today I started reviewing the RVPS implementation and have a few observations that I wanted to share.

## Semantic squashing

Whilst I get the ethos behind the choice of a uniform format for the ref-value type, I also think the assumption that ref-values are simple, stand-alone (id, hash) pairs hurts the usability of the service.

As a concrete example, Arm CCA workloads have ref-values with the following shape:

```json
{
  "initial-measurement": "48b2245b81dc07b9de288872ee8b68d3cde1a0d1a4ac47540d03a5ae3bcaaf4f",
  "extensible-measurements": [
    "2459b963cbba21426c35d2d14696e1276937a43904c20f879ea843219cfa91fe",
    "de9dd6746a35d665e0ba9c4764c05fb732d6da82982fd1cb6e037884044c484d",
    "a8b7369d7e326138a2ef3f830196d43d43c3db922a5fc94da8aaeb93559e1a80",
    "6ae61d7f4603304adf786dca761990c70599848e27bc2b548f10abb23189e736"
  ],
  "personalization-value": "00fa9623e7d40d168e652e0f8b9d651112697adec1e7b227c738c22137611f0f58b5d5de4a6e624076b083395df9a41b7993abb4a9026f1174b56183e5a94650"
}
```

where:

* `initial-measurement` measures the boot bits
* `extensible-measurements` is for the run-time
* `personalisation-value` contains user data supplied at launch

I wouldn't know how to fit that compound structure into the current RVPS data model unless I either make drastically simplifying (and lossy) assumptions on what information I decide to cut off (with the consequent deterioration of the quality/fidelity of the verification process), or I cram the whole JSON above into the hash value ;-).
Both outcomes are highly undesirable.

An alternative approach is to maintain the simple key-value interface for the verifier, skip normalizing in the RVPS, and let the scheme-specific code in the verifier interpret the fetched ref-values.

## Sharding of the store

From an initial (and admittedly superficial) code inspection, it is unclear if/how the sharding of the potentially multi-tenant and multi-scheme store is being addressed.
The keys are not explicitly namespaced and it looks like the extractors will just use what they've been supplied.  Assuming that the keys will be globally unique is obviously unsafe.

[...]

PR #1148, opened 2026-01-12
rvps: add reference value uuid

fitzthum:
We use names to identify reference values, but this doesn't capture a particular instance of a reference value. For example, a reference value name might be "firmware_hash", but there could be many reference values with this name. Even if the name is more specific (i.e. includes a version) this is still capturing the target context of the reference value (as in what it represents in the TCB) rather than the source instance.

As such, let's introduce an rv id to uniquely identify an rv instance. To begin with, this doesn't really do anything, but in the future we can generate reference value reports that capture exactly which rv instances were used to validate a guest.

To support backwards compatibility, create a new UUID when deserializing an RV that doesn't have one. Still, update the RV version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5223,6 +5223,7 @@ dependencies = [
  "tonic 0.14.2",
  "tonic-prost",
  "tonic-prost-build",
+ "uuid",
  "walkdir",
 ]
 
--- a/rvps/Cargo.toml
+++ b/rvps/Cargo.toml
@@ -44,6 +44,7 @@ tempfile.workspace = true
 tokio.workspace = true
 tonic.workspace = true
 tonic-prost.workspace = true
+uuid = { version = "1", features = ["v4", "serde"] }
 
 [build-dependencies]
 shadow-rs.workspace = true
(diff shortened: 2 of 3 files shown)


Xynnn007:
Nice catch to the problem. The `name` field works the same way as an identifier. My original idea is to use a URI like format to replace `name` field. I have implemented in this way in [the PR](https://github.com/kata-containers/kata-containers/pull/12285/changes#diff-c6fc7d1c7e0c83285fee8286d16678f15c7335c68a230e4d8a442a72f77c046cR607) deeply in the code.

The idea came from the following materials
1. @thomas-fossati 's [example](https://github.com/confidential-containers/trustee/issues/637#issuecomment-2556566339) and you once mentioned with a similar format [here](https://github.com/confidential-containers/trustee/issues/637#issuecomment-2557270273)
2. In more older time, the idea of Target Environment ID (TE ID) [here](https://github.com/confidential-containers/trustee/issues/238#issuecomment-1825517010) and [here](https://github.com/confidential-containers/trustee/issues/350)

Anyway, I think we can define a URI format to index a reference value, using path separators to semantically distinguish scopes. Furthermore, we can define a specific URI naming convention for CoCo and then unify the entire reference value publishing process.


And this PR will be the first important step on our road to final shape I believe.

PR #1149, opened 2026-01-13
toolchain: move to Rust 1.90.0

mythi:
Move to build using a more recent Rust toolchain. With `.is_multiple_of()` the minimum version required to build becomes 1.87.0

### DIFF
--- a/deps/verifier/src/nvidia/spdm_response.rs
+++ b/deps/verifier/src/nvidia/spdm_response.rs
@@ -286,7 +286,7 @@ impl OpaqueData {
     fn decode_switch_pdi(bytes: &[u8]) -> Result<Value> {
         const PDI_DATA_SIZE: usize = 8;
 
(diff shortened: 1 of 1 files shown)


BbolroC:
> I think I covered all the fixes needed. `s390x` runner is not in a good shape, @BbolroC

Resolved, all should go green again. Thanks for reporting! ❤️

dcmiddle:
This all looks good to me, with one question about dead code.
I see your commit message:

> For the TdShim type definitions, we keep them in case they end
> up getting used later (in other places than tests too).
> 

But would it be better to wrap them in #[cfg(test)] until such time that they get used?
Not blocking feedback, just a mild preference if it makes sense to you.

dcmiddle:
fyi, guest components is updated https://github.com/confidential-containers/guest-components/pull/1285

PR #1150, opened 2026-01-15
Add helm chart deployment

Xynnn007:
This PR mainly introduces the helm chart for trustee, and also some fixes met during tests. Please see the concrete commits for details.

Two issues to be created once merged
- Helm | Have a strategy for release, image tags
- HTTPS support for Helm

Updated:
- [ ] ways for release
- [x] relies on https://github.com/confidential-containers/trustee/pull/1291
- [ ] Ways to enable HTTPS, connections between pods.
- [x] Think about LB
- [ ] ways to enable ITA
- [x] replace postgres with https://artifacthub.io/packages/helm/bitnami/postgresql
- [ ] add redis support
- [ ] optimize key genetation logic
- [x] add dcap verifier config
- [ ] e2e test in trustee repo for helm
- [ ] provisioned snp/dcap/csv/rvps configs


deployment/helm-chart/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.16.0
```
fitzthum: Will we have to update this during the release process? Also is this related to the container images that are used?

deployment/helm-chart/README.md
```diff
@@ -0,0 +1,202 @@
...
+- `reference-value-provider-service`: Reference Value Provider Service
+
+All services use StatefulSets with persistent storage to maintain state across pod restarts.
+
+### Design Decision: StatefulSet vs Deployment
+
+Currently, all services are deployed as StatefulSets because **compute and storage are not decoupled**. The services require direct access to persistent storage volumes, which makes StatefulSets the appropriate choice for maintaining stable storage identities and ordered pod management.
+
+> [!NOTE]
+> There is an [open issue](https://github.com/confidential-containers/trustee/issues/1092) tracking the decoupling of compute and storage. Once this is implemented, we may migrate to Deployments with external storage backends, which would provide better scalability and flexibility.
```
fitzthum: Fine as it is. We could also try to get the db support in first, but maybe that's more work.

deployment/helm-chart/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+| Parameter | Description | Default |
+|-----------|-------------|---------|
+| `kbs.replicaCount` | Number of KBS replicas | `1` |
+| `kbs.image.repository` | KBS image repository | `ghcr.io/confidential-containers/staged-images/kbs-grpc-as` |
+| `kbs.image.tag` | KBS image tag | `latest` |
+| `kbs.service.type` | KBS service type | `LoadBalancer` |
+| `kbs.service.port` | KBS service port | `8080` |
+| `kbs.userKeysSecretName` | Secret name for user keys (empty for auto-generation) | `""` |
+| `kbs.autoGenerateKeys` | Enable auto-generation of keys via initContainer | `true` |
```
fitzthum: We'll need something else for Trustee in coco but this is fine for now. Just making a note for myself.

deployment/helm-chart/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Configuration
+
+The following table lists the configurable parameters and their default values:
+
+### Global Parameters
+
+| Parameter | Description | Default |
+|-----------|-------------|---------|
+| `log_level` | Log level for all services | `info` |
```
fitzthum: Logging for Trustee in CoCo will be intereseting, but this is good.

deployment/helm-chart/README.md
```diff
@@ -0,0 +1,202 @@
...
+  userKeysSecretName: ""  # Empty = auto-generate
+  autoGenerateKeys: true
+```
+
+The initContainer generates:
+- `private.key` and `public.pub` (ED25519 keys for KBS authentication)
+- `ca.key` and `ca-cert.pem` (RSA 2048 CA certificate for token signing)
+- `token.key`, `token-cert.pem`, and `token-cert-chain.pem` (EC prime256v1 token keys for attestation token signing)
+
+**Note**: With auto-generation, keys are stored in an `emptyDir` volume and will be regenerated on pod restart. This is suitable for development/testing environments.
```
fitzthum: I guess for Trustee in coco each of these keys will have to be a sealed secret and we won't do any shared dir.

fitzthum (commented):
This is great. cc @lmilleri 

I made a few notes. Mostly they are just notes for myself about Trustee in CoCo. We can ignore those for now.

deployment/helm-chart/README.md
Xynnn007: Yes, I originally planned to complete the database support first and then start helm chart. However, recently there has been a lot of interest in deployment models, so I submitted this PR first to show that we are working on it.

deployment/helm-chart/files/kbs-config.toml.template
```diff
@@ -0,0 +1,22 @@
+[http_server]
+sockets = ["0.0.0.0:{{ .Values.kbs.service.port }}"]
+insecure_http = true
```
lmilleri: support https?

deployment/helm-chart/files/kbs-config.toml.template
```diff
@@ -0,0 +1,22 @@
+[http_server]
```
lmilleri: is all-in-one supported?

lmilleri:
Very nice. Any plans to add support for single image with all the components (kbs-as-rvps)? Or do you think deploying separate services is definitely the way to go?

Xynnn007:
> Very nice. Any plans to add support for single image with all the components (kbs-as-rvps)? Or do you think deploying separate services is definitely the way to go?

Yes, we can add it later if needed. Currently, it completely mimics the logic of docker-compose as first draft. Personally I prefer micro service state.

vans163:
Autocreate seems not working

```
helm install trustee ./deployment/helm-chart --namespace coco-trustee --create-namespace
Error: INSTALLATION FAILED: StatefulSet.apps "attestation-service" is invalid: [spec.template.spec.volumes[4].secret.secretName: Required value, spec.template.spec.containers[0].volumeMounts[3].name: Not found: "user-keys"]
```

Xynnn007:
Folks. Update this PR and test locally with default (localfs) and postgres on alibaba cloud container service and [KIND](https://kind.sigs.k8s.io/)

PTAL

cc @fidencio @beraldoleal

Xynnn007:
@fitzthum #1239 will influence this PR a lot. It makes sense for either to get merged first and I can help to take rebase in another.

Xynnn007:
Let me play with Copilot for a while.

deployment/helm-chart/templates/as-deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+              mountPath: /etc/attestation-service
+            - name: sgx-config
+              mountPath: /etc/sgx_default_qcnl.conf
+              subPath: sgx_default_qcnl.conf
+            - name: policy
+              mountPath: /opt/confidential-containers/attestation-service/token/ear/policies/opa
+            - name: user-keys
+              mountPath: /opt/confidential-containers/kbs/user-keys
+            - name: data
+              mountPath: /opt/confidential-containers/attestation-service
```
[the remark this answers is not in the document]
Xynnn007: Added comments in readme.md

rvps/src/lib.rs
```diff
@@ -25,7 +25,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::extractors::ExtractorsConfig;
 
-pub const REFERENCE_VALUE_STORAGE_NAMESPACE: &str = "reference-value";
+pub const REFERENCE_VALUE_STORAGE_NAMESPACE: &str = "reference_value";
 
 /// Default version of Message
```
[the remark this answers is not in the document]
Xynnn007: This is still during the release cycle. Thus no breaking changes.

deployment/helm-chart/templates/as-deployment.yaml
```diff
@@ -0,0 +1,120 @@
...
+          imagePullPolicy: {{ .Values.as.image.pullPolicy }}
+          command:
+            - "grpc-as"
+            - "--socket"
+            - "0.0.0.0:{{ include "coco-trustee.port.as" . }}"
+            - "--config-file"
+            - "/etc/attestation-service/config.json"
+          env:
+            - name: RUST_LOG
+              value: {{ .Values.log_level }}
```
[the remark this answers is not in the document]
Xynnn007: The values are fixed, no YAML-significant

deployment/helm-chart/values.yaml
```diff
@@ -0,0 +1,123 @@
...
+    limits:
+      cpu: 500m
+      memory: 512Mi
+
+rvps:
+  image:
+    repository: ghcr.io/confidential-containers/staged-images/rvps
+    pullPolicy: IfNotPresent
+    tag: "latest"
+  service:
```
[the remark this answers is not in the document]
Xynnn007: Let me make an issue for this

kbs/src/token/jwk.rs
```diff
@@ -156,7 +163,10 @@ impl JwkAttestationTokenVerifier {
         }
 
         let pem = x5c[0].split('\n').collect::<String>();
-        let der = URL_SAFE_NO_PAD.decode(pem).context("Illegal x5c cert")?;
+        let der = STANDARD
+            .decode(pem.trim())
+            .or_else(|_| URL_SAFE_NO_PAD.decode(pem.trim()))
+            .context("Illegal x5c cert")?;
```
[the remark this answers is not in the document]
Xynnn007: x5c has paddings

deployment/helm-chart/files/kbs-config.toml.template
```diff
@@ -0,0 +1,48 @@
+{{- if include "coco-trustee.kbs.sessionStorageType" . }}
+session_storage_type = "{{ include "coco-trustee.kbs.sessionStorageType" . }}"
+{{- end }}
+
+[http_server]
+sockets = ["0.0.0.0:{{ include "coco-trustee.port.kbs" . }}"]
+insecure_http = true
```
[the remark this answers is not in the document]
Xynnn007: Let me mark this is a TODO issue

deployment/helm-chart/values.yaml
```diff
@@ -0,0 +1,123 @@
...
+bootstrapUserKeysJob:
+  keygenImage:
+    repository: alpine/openssl
+    tag: "latest"
+    pullPolicy: IfNotPresent
+  kubectlImage:
+    repository: quay.io/kata-containers/kubectl
+    tag: "latest"
+    pullPolicy: IfNotPresent
+  resources:
```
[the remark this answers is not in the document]
Xynnn007: This is an issue for quay.io/kata-containers/kubectl I think.

attestation-service/src/rvps/grpc.rs
```diff
@@ -85,3 +90,27 @@ impl RvpsApi for Agent {
...
+    address: String,
+}
+
+#[async_trait::async_trait]
+impl Manager for GrpcManager {
+    type Connection = ReferenceValueProviderServiceClient<Channel>;
+    type Error = anyhow::Error;
+
+    async fn connect(&self) -> std::result::Result<Self::Connection, Self::Error> {
+        let channel = Channel::from_shared(self.address.clone())?
```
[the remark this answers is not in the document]
Xynnn007: Problem acked. Refactored with a different solution

attestation-service/src/rvps/grpc.rs
```diff
@@ -26,6 +28,8 @@ fn default_address() -> String {
     "127.0.0.1:50003".into()
 }
 
+const DEFAULT_RVPS_POOL_SIZE: u64 = 100;
+
```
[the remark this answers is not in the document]
Xynnn007: Let's use a hardcore value for ease.

deployment/helm-chart/values.yaml
```diff
@@ -0,0 +1,123 @@
...
+# The keygen image is used to generate the demo keys, and the kubectl image is used to create the release-scoped Secret.
+bootstrapUserKeysJob:
+  keygenImage:
+    repository: alpine/openssl
+    tag: "latest"
+    pullPolicy: IfNotPresent
+  kubectlImage:
+    repository: quay.io/kata-containers/kubectl
+    tag: "latest"
+    pullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
Xynnn007: leave as todo

deployment/helm-chart/Chart.yaml
Xynnn007: Yes. I will leave this as an issue and make it perfect in following PR.

deployment/helm-chart/README.md
Xynnn007: I agree to do some changes for trustee in coco on the basic usage now

attestation-service/src/rvps/grpc.rs
```diff
@@ -26,26 +27,35 @@ fn default_address() -> String {
...
-    #[error("tonic transport error: {0}")]
-    TonicTransport(#[from] tonic::transport::Error),
-}
+/// The default pool size for the RVPS client.
+const DEFAULT_RVPS_POOL_SIZE: u64 = 10;
 
 pub struct Agent {
-    client: Mutex<ReferenceValueProviderServiceClient<tonic::transport::Channel>>,
+    pool: Pool<GrpcManager>,
 }
```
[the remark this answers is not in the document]
Xynnn007: Pool supports concurrency thus we do not need Mutex to control.

Xynnn007:
I think I have finished fixing spots reported by copilot, because it's starting talking nonsense. So PTAL

attestation-service/src/ear_token/broker.rs
```diff
@@ -197,30 +194,40 @@ impl EarAttestationTokenBroker {
...
-
-                            tx.send(res)
+                            rt.block_on(async move {
+                                match tokio::time::timeout(
+                                    Duration::from_secs(10),
+                                    fut(rvps_client, reference_value_id),
+                                )
+                                .await
+                                {
+                                    Ok(res) => res,
```
fitzthum: I don't really like matches with an empty arm like this. Can use map_err or similar instead.

deployment/helm-chart/files/sgx_default_qcnl.conf
```diff
@@ -0,0 +1 @@
+{"collateral_service": "https://api.trustedservices.intel.com/sgx/certification/v4/"}
```
zvonkok: Nitpick no EOF

deployment/helm-chart/templates/as-configmap.yaml
```diff
@@ -0,0 +1,21 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "coco-trustee.names.sgxConfig" . }}
+  labels:
+    {{- include "coco-trustee.labels" . | nindent 4 }}
+data:
+  sgx_default_qcnl.conf: |
+{{ tpl (.Files.Get "files/sgx_default_qcnl.conf") . | nindent 4 }}
```
zvonkok: No EOF

deployment/helm-chart/templates/kbs-configmap.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "coco-trustee.names.kbsConfig" . }}
+  labels:
+    {{- include "coco-trustee.labels" . | nindent 4 }}
+data:
+  "kbs-config.toml": |
+{{ tpl (.Files.Get "files/kbs-config.toml.template") . | nindent 4 }}
```
zvonkok: EOF

deployment/helm-chart/templates/rvps-configmap.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "coco-trustee.names.rvpsConfig" . }}
+  labels:
+    {{- include "coco-trustee.labels" . | nindent 4 }}
+data:
+  rvps.json: |
+{{ tpl (.Files.Get "files/rvps.json.template") . | nindent 4 }}
```
zvonkok: EOF

zvonkok:
Since we're deploying a k8s Service, do we want to think about the ingress part, like load balancing?

deployment/helm-chart/Chart.yaml
```diff
@@ -0,0 +1,26 @@
...
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.18.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "0.18.0"
+
+# TODO: Have a mechanism for app version
```
zvonkok: EOF nitpick

zvonkok:
Do you want to extract the commits for the Rust parts in their own PR? I get that those commits are to enable the Helm Chart but it overloads the current PR for easier review.

fidencio:
Requested myself a review, and I will go through this Tomorrow.

deployment/helm-chart/files/kbs-config.toml.template
fitzthum, line 1: Any reason to mix toml and json configs for the different components.

deployment/helm-chart/files/sgx_default_qcnl.conf
fitzthum, line 1: I thought we no longer needed this config? (But we do need to set the related env var)

deployment/helm-chart/files/kbs-config.toml.template
```diff
@@ -0,0 +1,48 @@
...
+
+[[plugins]]
+name = "resource"
+type = "kvstorage"
+
+{{- $st := include "coco-trustee.storage.type" . | trim }}
+{{- $sst := include "coco-trustee.kbs.sessionStorageType" . | trim }}
+{{- $needLocalJson := or (eq $st "LocalJson") (eq $st "local_json") (eq $sst "LocalJson") (eq $sst "local_json") }}
+{{- $needLocalFs := or (eq $st "LocalFs") (eq $st "local_fs") (eq $sst "LocalFs") (eq $sst "local_fs") }}
+{{- $needPostgres := or (eq $st "Postgres") (eq $st "postgres") (eq $sst "Postgres") (eq $sst "postgres") }}
```
fitzthum, line 38: So we're supporting three different options for storage? I wonder if we should just pick one

fitzthum (commented):
A couple small questions

Are we still waiting for https://github.com/confidential-containers/trustee/pull/1239

deployment/helm-chart/files/sgx_default_qcnl.conf
Xynnn007: The current config does not support to set PCCS service address -  this is useful at least for Alibaba Cloud. But I agree with @zvonkok 's idea to get the rust part code in a separate PR. I will adopt the comment you raised in that PR.

deployment/helm-chart/files/kbs-config.toml.template
Xynnn007: Ah TBO I also agree to choose one, s.t. postgres as it's most case and easy to maintain. Do you think other folks will agree with us?

Xynnn007:
Hey @zvonkok @fidencio Let me get the rust code part into another PR and leave this pure helm.

deployment/helm-chart/files/sgx_default_qcnl.conf
fitzthum: cc @mythi

deployment/helm-chart/files/kbs-config.toml.template
Xynnn007: Actually for docker-compose, we are always using JSON for AS/RVPS, TOML for KBS. Do you think we need to have a converge in this PR?

fidencio: Can be a follow-up, IMHO.

deployment/helm-chart/files/kbs-config.toml.template
fidencio: I'd default to postgres, but then drop the LocalFs and LocalJson from the codebase entirely (not only from the helm).  My experience with kata-containers is that if there's an option, someone will open an issue to have that exposed in the helm. :-/

mythi:
> This PR mainly introduces the helm chart for trustee,

I may have missed the answer but why trustee repo and not charts repo?

fidencio:
> I may have missed the answer but why trustee repo and not charts repo?

It will be linked / referred to from the charts repo, in the same way that kata / peer-pods helms already are.

deployment/helm-chart/files/kbs-config.toml.template
Xynnn007: @fidencio TBO I like your idea. The maintaince effort is huge. I think whether we could do this heavily depending on @bpradipt and @lmilleri 's idea.

fitzthum: Yeah gotta coordinate with Trustee operator before dropping the local support, but worth discussing.

Xynnn007: Oh I forgot to mention this during the weekly meeting.

fidencio:
@Xynnn007, I started experimenting with this Today, but just went through what we do on the Kata Containers side right now with the kustomize overlay, trying to map how would be the usage of this chart there ... right now this is **NOT** a review, but rather a **braindump**.

There's one incompatibility that I found already and I'd like to loop @BbolroC and @stevenhorsman in.

**For the s390x, what the kustomize does (@stevenhorsman / @BbolroC, please, correct me here):**
* Mounts IBM SE credentials into the KBS pod -- creates a local PV/PVC backed by ${IBM_SE_CREDS_DIR} (a host directory containing certs, CRLs, HKD keys, RSA keys) and mounts it at /run/confidential-containers/ibmse/ in the KBS container
* Sets SE_SKIP_CERTS_VERIFICATION env var on the KBS container (the kata CI script then sed's false to true)
* Generates a key.bin secret via kustomize secretGenerator (same as the default overlay, just moved into ibm-se/)

**Then what kata's prepare_credentials_for_qemu_se() does before deployment:**
* Generates RSA keys (encryption key pair)
* Copies the SE image header (kata-containers-se.img) into hdr/hdr.bin
* Copies certs into the credentials directory
* All of this material ends up on-disk at ${IBM_SE_CREDS_DIR}, then gets mounted via the PV/PVC

**In order to have something similar we'd need to:**
* Add a way of mounting arbitrary host-path credentials for IBM SE verification
* Add a way to configure SE_SKIP_CERTS_VERIFICATION env var on the KBS deployment
* Add some PV/PVC for a local credentials directory

I **think** (as this is not tested yet) that the changes proposed below would do, but do not add this to your PR yet ... as Tomorrow I'll keep working on this / on testing this integrated with the Kata Containers CI.
```
diff --git a/deployment/helm-chart/templates/kbs-deployment.yaml b/deployment/helm-chart/templates/kbs-deployment.yaml
index 7e9092d..8b6680d 100644
--- a/deployment/helm-chart/templates/kbs-deployment.yaml
+++ b/deployment/helm-chart/templates/kbs-deployment.yaml
@@ -63,6 +63,9 @@ spec:
           env:
             - name: RUST_LOG
               value: {{ .Values.log_level }}
+            {{- with .Values.kbs.extraEnvVars }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
             {{- $pgurl := include "coco-trustee.postgresUrlSecretName" . | trim }}
             {{- $storageBackendIsPostgres := include "coco-trustee.storage.storageBackendIsPostgres" . | trim }}
             {{- $kbsSessionStorageIsPostgres := include "coco-trustee.kbs.sessionStorageIsPostgres" . | trim }}
@@ -96,6 +99,9 @@ spec:
             {{- end }}
             - name: nebula-ca
               mountPath: /opt/confidential-containers/kbs/nebula-ca
+            {{- with .Values.kbs.extraVolumeMounts }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
             {{- range .Values.kbs.resourceRepository }}
             - name: "{{ .repoName }}-{{ .type }}"
               mountPath: "/opt/confidential-containers/kbs/repository/{{ .repoName }}/{{ .type }}"
@@ -120,6 +126,9 @@ spec:
         {{- end }}
         - name: nebula-ca
           emptyDir: {}
+        {{- with .Values.kbs.extraVolumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
         {{- range .Values.kbs.resourceRepository }}
         - name: "{{ .repoName }}-{{ .type }}"
           secret:
diff --git a/deployment/helm-chart/values.yaml b/deployment/helm-chart/values.yaml
index 1a1a91e..8df958f 100644
--- a/deployment/helm-chart/values.yaml
+++ b/deployment/helm-chart/values.yaml
@@ -71,6 +71,29 @@ kbs:
       memory: 512Mi
   resourceRepository: []

+  # Extra environment variables to inject into the KBS container.
+  # Example:
+  #   extraEnvVars:
+  #     - name: SE_SKIP_CERTS_VERIFICATION
+  #       value: "true"
+  extraEnvVars: []
+
+  # Extra volumes to attach to the KBS pod.
+  # Example:
+  #   extraVolumes:
+  #     - name: ibmse-creds
+  #       hostPath:
+  #         path: /path/to/credentials
+  #         type: Directory
+  extraVolumes: []
+
+  # Extra volume mounts for the KBS container.
+  # Example:
+  #   extraVolumeMounts:
+  #     - name: ibmse-creds
+  #       mountPath: /run/confidential-containers/ibmse/
+  extraVolumeMounts: []
+
 as:
   image:
     repository: ghcr.io/confidential-containers/staged-images/coco-as-grpc
@@ -133,3 +156,4 @@ secrets:
   # Required when useEphemeralGeneratedKeys is false. Secret must contain keys:
   # KBS_ADMIN_PRIVATE_KEY, KBS_ADMIN_PUBKEY, AS_TOKEN_SIGNING_PRIVATE_KEY, AS_TOKEN_VERIFICATION_PUBLIC_KEY_CERT_CHAIN
   existingSecretName: ""
+
```

fidencio:
@zvonkok asked about Ingress / load balancing ... I think load balancing here can be premature and can come as a follow-up.  However, ingress setup is a **must** for the Kata Containers CI and that could be added with something like:
```
index 4057ccf..3aa6bf8 100644
--- a/deployment/helm-chart/values.yaml
+++ b/deployment/helm-chart/values.yaml
@@ -140,3 +140,21 @@ secrets:
   # Required when useEphemeralGeneratedKeys is false. Secret must contain keys:
   # KBS_ADMIN_PRIVATE_KEY, KBS_ADMIN_PUBKEY, AS_TOKEN_SIGNING_PRIVATE_KEY, AS_TOKEN_VERIFICATION_PUBLIC_KEY_CERT_CHAIN
   existingSecretName: ""
+
+# Optional Kubernetes Ingress for the KBS service.
+ingress:
+  enabled: false
+  className: ""
+  annotations: {}
+  # Host-based routing. Leave empty to match all hosts (IP-only access).
+  host: ""
+  tls: []
+  #  - secretName: kbs-tls
+  #    hosts:
+  #      - kbs.example.com
+
+# Expose the KBS service via a NodePort.
+nodePort:
+  enabled: false
+  # If empty, Kubernetes assigns a random port from the NodePort range.
+  port: ""
```

Please, as the previous suggestion, do not add this yet to your PR, as I want to do a full end-to-end test on my side.

fidencio:
@Xynnn007, this is something I can open a PR soon (as soon as I do the end to end tests) ... but I'd like to have healthz endpoints on AS / RVPS / KBS. That would make things way more reliable / robust.

cc @fitzthum

fidencio:
@Xynnn007, I have patches atop of your PR here: https://github.com/fidencio/trustee/tree/feat-helm-chart ... and I'm testing those on the Kata Containers side with: https://github.com/kata-containers/kata-containers/pull/12982

Let's see how things go.

BbolroC:
> There's one incompatibility that I found already and I'd like to loop @BbolroC and @stevenhorsman in.

> For the s390x, what the kustomize does (@stevenhorsman / @BbolroC, please, correct me here):

Sorry for the late answer. Your explanation is correct and your patch looks good. I will support the testing. Thanks!

Xynnn007:
@fidencio Some updates about the PR
1. I rebased with upstream code. Now the admin tokens are used instead of admin keys.

Nothing else is added.

deployment/helm-chart/files/kbs-config.toml.template
```diff
@@ -0,0 +1,52 @@
+{{- if include "coco-trustee.kbs.sessionStorageType" . }}
+session_storage_type = "{{ include "coco-trustee.kbs.sessionStorageType" . }}"
+{{- end }}
+
+[http_server]
+sockets = ["0.0.0.0:{{ include "coco-trustee.port.kbs" . }}"]
+insecure_http = true
```
fitzthum, line 7: So we don't even have the option to enable https?

fidencio:
@Xynnn007, I've pushed 4 commits atop of yours (all helm related) and that's what I used to get it to work on Kata Containers side.

Please, review those, and then squash those into your original commit, adding also my `Signed-off-by: ...`.

This series LGTM with those changes and was tested end-to-end on the Kata Containers side.
I'd love to get it merged ASAP, so we can switch Kata Containers to using it, and ensure others will have an easier time improving it than I did. :-)

fidencio:
cc @fitzthum as well

deployment/helm-chart/templates/as-deployment.yaml
```diff
@@ -73,6 +73,34 @@ spec:
             - name: tcp
               containerPort: {{ include "coco-trustee.port.as" . }}
               protocol: TCP
+          # Intel TDX/SGX DCAP verification makes blocking FFI + collateral
+          # calls that can stall the gRPC health endpoint for several seconds
+          # under attestation load. The previous tight probes (timeout 5s,
```
mythi: needs to be verified after https://github.com/confidential-containers/trustee/pull/1368

deployment/helm-chart/files/kbs-config.toml.template
```diff
@@ -8,24 +8,23 @@ insecure_http = true
 
 [attestation_token]
 trusted_certs_paths = ["/opt/confidential-containers/kbs/user-keys/token-cert-chain.pem"]
+insecure_key = true
```
fitzthum: This field is `insecure_header_jwk` now, but we do not want that enabled. Fortunately the line above configures things securely, so you can just drop this.

deployment/helm-chart/files/kbs-config.toml.template
```diff
@@ -8,24 +8,23 @@ insecure_http = true
...
 
 [admin]
-type = "Simple"
-
-[[admin.personas]]
-id = "admin"
-public_key_path = "/opt/confidential-containers/kbs/user-keys/public.pub"
+# Insecure test posture (matches insecure_http / insecure_key above): allow
+# anonymous admin requests so kbs-client can set resources without a token.
+authorization_mode = "InsecureAllowAll"
```
fitzthum: We can't enable this for the helm chart. It's handy for tests, but the helm chart isn't just for tests. We want this to be the main way that people deploy Trustee in k8s, so gotta be as secure as possible. Fortunately, I think Ding already added the secure config.

deployment/helm-chart/values.yaml
```diff
@@ -87,18 +94,25 @@ as:
     type: ClusterIP
     port: 50004
     loadBalancerAnnotations: {}
+  # Attestation is CPU-bound (signature/cert-chain verification, regorus policy
+  # evaluation) and the all-verifier image loads TDX/SGX DCAP and NVIDIA verifier
+  # libraries. Tight limits here caused the pod to be OOMKilled / throttled into
+  # liveness-probe failures mid-/attest, resetting the in-flight gRPC call.
```
fitzthum: nit: kinda weird LLM comment. I would just drop this

deployment/helm-chart/templates/as-deployment.yaml
fitzthum: We're referencing the previous timeouts, but this is a new feature, right? Not sure you really need this comment at all.

fitzthum (commented):
A few comments. I think we gotta have all the security features that we can turned on. That might not be necessary for testing, but the main purpose of the Helm chart is for people to use it.

I had been thinking about enabling HTTPS by default for the Helm chart. I think we should do that, but not in this PR.

deployment/helm-chart/README.md
```diff
@@ -0,0 +1,202 @@
...
+helm uninstall trustee -n coco-trustee
+
+# If useEphemeralGeneratedKeys is set to true, you need to delete the ephemeral keys manually
+# kubectl delete secret -n coco-trustee trustee-bootstrap-user-keys
+```
+
+> [!NOTE] Intel DCAP: `sgx_default_qcnl.conf`
+>
+> The chart embeds **`files/sgx_default_qcnl.conf`** into a release-scoped ConfigMap (name `<Helm fullname>-sgx-config`, e.g. **`trustee-sgx-config`** when the release is **`trustee`**) and mounts it on the **gRPC Attestation Service** deployment at **`/etc/sgx_default_qcnl.conf`** (see `templates/as-configmap.yaml` and `templates/as-deployment.yaml`).
+> This configuration is **only relevant when the Attestation Service verifies Intel TDX/SGX** (Intel DCAP stack reads this path while resolving TDX quotes). It is **not** used for KBS or RVPS, and it is **not** required when you only use non-TDX verifiers (for example sample or non-Intel evidence paths) or AS images built without Intel TDX DCAP support. To use a different collateral endpoint or corporate proxy, replace the contents of **`files/sgx_default_qcnl.conf`** in your chart fork or overlay before `helm install` / `helm upgrade`.
```
mythi: can we get rid of all this?

deployment/helm-chart/templates/postgres-deployment.yaml
```diff
@@ -0,0 +1,88 @@
+{{- $pg := (.Values.storageBackend.postgres | default dict) }}
+{{- $pgi := ($pg.internal | default dict) }}
+{{- $initKvTables := (ternary $pgi.initKvTables true (hasKey $pgi "initKvTables")) }}
+{{- if eq (include "coco-trustee.postgres.useBundled" . | trim) "true" }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "coco-trustee.names.postgres" . }}
```
mythi: why do we need to manage our own prostgres? seem error prone and adds maintenance overhead.

deployment/helm-chart/files/kbs-config.toml.template
Xynnn007: Yes. Let's do it in a following PR, with simple commit lines

deployment/helm-chart/files/kbs-config.toml.template
Xynnn007: Yes. I will do some rebase work to overwrite this

deployment/helm-chart/files/kbs-config.toml.template
Xynnn007: I will handle this after a rebase

deployment/helm-chart/values.yaml
Xynnn007: The comments make sense, but agree to remove as we do not need too much context in this file

deployment/helm-chart/templates/as-deployment.yaml
Xynnn007: Removed ir-relavant comments

deployment/helm-chart/templates/postgres-deployment.yaml
Xynnn007: The idea of helm is to deploy things quickly and out-of-box.

This will help those who do not have postgres. Also this provides ways to connect to an external postgres instance.

deployment/helm-chart/README.md
Xynnn007: Yes. SGX QCNL things is being optimizing quickly based on your great work. This PR was draft ~ 6 months ago and at that time this way worked.

Do you agree to firstly get the current shape in and I will make another PR to fix qcnl config things? I will help to test things.

Now Fabiano has already test things in kata-ci and it works fine. I don't want to drag helm work out for too long and block Fabiano's work.

Xynnn007:
Note: DO NOT MERGE. TESTING ON KATA SIDE WITH ADMIN TOKEN THINGS.

deployment/helm-chart/templates/postgres-deployment.yaml
mythi: I don't think we want to maintain a postgres deployment

deployment/helm-chart/templates/bootstrap-user-keys-job.yaml
```diff
@@ -0,0 +1,113 @@
+{{- if include "coco-trustee.kbs.useBootstrapKeysJob" . | trim }}
```
mythi, line 1: We must be missing something fundamental if we have to do all this. How the operator handles this?

deployment/helm-chart/files/kbs-config.toml.template
```diff
@@ -0,0 +1,52 @@
...
+
+[http_server]
+sockets = ["0.0.0.0:{{ include "coco-trustee.port.kbs" . }}"]
+insecure_http = true
+
+[attestation_token]
+trusted_certs_paths = ["/opt/confidential-containers/kbs/user-keys/token-cert-chain.pem"]
+
+[attestation_service]
+type = "coco_as_grpc"
```
mythi, line 13: is ITA configurable?

deployment/helm-chart/templates/postgres-deployment.yaml
Xynnn007: maybe there must be other ways i do not know. how can handle a scenario where users want a backend than local* but out of the box?

deployment/helm-chart/templates/bootstrap-user-keys-job.yaml
Xynnn007: https://github.com/confidential-containers/trustee-operator/blob/f177616dced4fa68a7bc9ec443a85e025edc1a4f/internal/controller/trusteeconfig_controller.go#L605-L606

it can generate. in helm we explicitly tell users whether they want to BYOK or generate.

deployment/helm-chart/files/kbs-config.toml.template
Xynnn007: it can but this PR we do not support yet. Can we delay it in a following PR?

deployment/helm-chart/templates/postgres-deployment.yaml
Xynnn007: Let me add a TODO to replace current with https://artifacthub.io/packages/helm/bitnami/postgresql one

deployment/helm-chart/templates/bootstrap-user-keys-job.yaml
mythi: Didn't we plan to use `trustee-cli` for the same? Perhaps this job can be moved to a conditional initContainer that calls our trustee-cli. I've recently used `kube-rs` quite a bit and I think it's easy to have the tool to provision the secret directly (if we accept the dependency here, but it could be in the planned coco-tools repo also).

deployment/helm-chart/files/kbs-config.toml.template
mythi: That's probably OK given it's not used by Kata anymore (it was until recently still)

mythi: Is there a reason we move away from `builtin` to `as_grpc`? I had understood `builtin` was the preferred mode.

Xynnn007: There is always nowhere to mention that builtin is preferred actually and this just follows what docker-compose is doing. Are there any hard blockers for grpc as?

deployment/helm-chart/templates/bootstrap-user-keys-job.yaml
Xynnn007: Now Trustee cli's capabilities are mostly the same as kbs-client. One possible different ability might be to provision a trustee inside TEE.

It can be hard for `trustee-cli` to achieve this goal as the admin token configurations are fixed before KBS launches. The signed token and the public key should be generated before the KBS launches. Now the logic is in a conditional job container and we can optimize it later.

deployment/helm-chart/files/kbs-config.toml.template
mythi: IIRC @fitzthum mentioned it once but I might also remember it wrong. I don't have issues with grpc as.

deployment/helm-chart/templates/bootstrap-user-keys-job.yaml
mythi: > It can be hard for `trustee-cli` to achieve this goal as the admin token configurations are fixed before KBS launches.

This is true for the initContainer approach. We could still have some tool doing the work instead of the long shell script args?

deployment/helm-chart/README.md
mythi: `sgx_default_qcnl.conf` has not been used since early February after `dcap_verifier` was added to verifier config.

Do you think it will take long time still to get #1368 merged?

deployment/helm-chart/templates/bootstrap-user-keys-job.yaml
Xynnn007: can we optimize it in later PRs? let me add another todo for this.

deployment/helm-chart/files/kbs-config.toml.template
fitzthum: Doesn't really matter imo. The connections between components must not be exposed to the outside network ofc.

mythi: Old discussion in #375 when I proposed splitting the services

fitzthum: not really a concern anymore as long as communication between the components is not exposed.

deployment/helm-chart/README.md
Xynnn007: i think no big blockers for #1368.

let's firstly get this in. 

then i will go back to help with #1368. that pr has no overlapping to helm, thus no conflicts imo.

then, let me update helm with qcnl configs. 

does this sound good to you?

deployment/helm-chart/files/kbs-config.toml.template
mythi: it's `http` in-cluster comms between all services (each component runs their on `Deployment`)

deployment/helm-chart/README.md
mythi: It's not a good deal for me because this PR might still take time to get merged.

fidencio: @mythi, let's get your stuff in, but I must say it's unfair.

I've spent a lot of time getting the TDX CI to behave as expected with the chart and I did that without any help or interest from someone from Intel to do so.

If having your work merged is the goal, then let's do it, even taking the risk of adding more work on my back to make sure the chart works with your CI ...

deployment/helm-chart/files/kbs-config.toml.template
fitzthum: that is interesting. it's not what we do for the existing k8s deployment. i guess the ingress only points to the kbs, though. this is going to be a problem for trustee in coco, though. it probably would be best if we had everything in one pod. potentially we could address this later, but i'm on the fence about that

Xynnn007: now the containers are in the same pod, and only kbs service is exposed. They are all in-bound dataflow and cannot be tampered by others.

fitzthum: You mean today or in this PR? It looks lik there are different deployments for the difference services in this PR. Also I see that the load balancer can be enabled not just for the kbs but for all the services. Looks like it is internal by default (If i am reading all this yaml correctly). Not sure there is ever a case where we would want these to be externally reachable.

Anyway @Xynnn007 I'll leave it up to you if we want to handle now or later.

Xynnn007: Yes. Let me handle this later.

Xynnn007:
@fidencio @fitzthum @mythi 

Updated the PR:
1. Reorganize the codes
2. Add DCAP verifier config
3. Use `bitnamicharts` to provide internal postgres deployment
4. add deletion logic to ephemeral keys

I'd prefer to leave other work items given by the reviews in later PRs. All of them are listed in the https://github.com/confidential-containers/trustee/pull/1150#issue-3816849895 box. I will open issues once this PR gets merged.

Test locally with sample kbs-client.

Btw I'd like to help @fidencio with kata-ci integration work, who has made **VERY HUGE EFFORT** to the work already and I really appreciate it.

deployment/helm-chart/files/kbs-config.toml.template
mythi: I believe that means extra work on Kata side because the deployment is going to be different (the current Kata implementation checks all pods are there which is not the case in the future). Would be good to get this in place before merge

fitzthum: I am fine either way. There will be some changes down the road in either case.

deployment/helm-chart/Chart.yaml
```diff
@@ -0,0 +1,32 @@
...
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.18.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "0.18.0"
```
fitzthum, line 24: Are we planning to keep this in sync with the Trustee version? It's a bit tricky to do that with our current release process.

fitzthum (commented):
Ok, one more comment.

Also, what is the plan for testing the Helm chart in this repo? Is that going to be part of this PR?

deployment/helm-chart/files/kbs-config.toml.template
Xynnn007: +1 for @fitzthum . In both way, we still need extra work. So from my personal preference, this PR can be merged firstly, and then do ITA patch.

We do not need to make helm perfect on this one PR. And we can start kata side integration work later. We are in between development cycles so it's fine to make fixups later.  @fidencio  what's your opinion for this?

Xynnn007:
> Also, what is the plan for testing the Helm chart in this repo? Is that going to be part of this PR?

I tried to use kind to set-up a light kubernetes cluster on GHA runner to test with kbs-client. However, the commit seems large, and also there is some timeout issues. I prefer to add it in a following PR and needs some optimization.

deployment/helm-chart/files/kbs-config.toml.template
Xynnn007: @mythi I understand the concerns, and I already get some of the stuff ready in this PR. For ITA it's hard for me to test as I do not have access to it.

Given that this PR has already been open for quite some time, we would prefer to move forward with a multi-PR way so that the current work can land first, and then address the remaining items iteratively.

This is not a sign that we are unwilling to support Intel TDX-related work. On the contrary, we remain committed to supporting it - as always. and I am happy to continue contributing in subsequent PRs.

I hope this makes sense to you, and shall we proceed with this PR first and follow up on the remaining enhancements in later PRs?

deployment/helm-chart/Chart.yaml
Xynnn007: It's tricky from my side too. So I added a TODO mark [in the code ](https://github.com/confidential-containers/trustee/pull/1150/changes#diff-10222a3ba50866ae465b80945b8f31bb675ddf5ed02d98b7f09eab222795c71bR32)

I think this is somehow coupled with [the discussion](https://github.com/confidential-containers/trustee/issues/1355)


My initial idea is that if we can integrate Trustee's Helm deployment and testing into an end-to-end CI within the Trustee repository, then we can solve the puzzle of "first Kata CI, then release Trustee," enabling independent releases of Trustee, with Kata referencing the tag?

Xynnn007: Before that, it may be fine to leave the appVersion number some number, as it does not make much sense?

deployment/helm-chart/files/kbs-config.toml.template
mythi: ITA is not my concern here. My concern is the model that we are ready accept changes that we know need to be changed/removed and have user impact. I'd rather get the baseline correct (e.g., that this needs to run within a single pod) and make incremental (small) additions on top of it.

mythi:
> 2\. Add DCAP verifier config

thanks! The commit message looks odd. I don't think they need all to be set (at least I was testing air-gapped with just `collateral_service = file:// ...` added).

deployment/helm-chart/files/kbs-config.toml.template
Xynnn007: @mythi Putting all containers in one pod and separating them is a trade-off.

Separating them allows for fine-grained disaster recovery, while keeping them together ensures that all traffic flows within the pod.

> I believe that means extra work on Kata side because the deployment is going to be different (the current Kata implementation checks all pods are there which is not the case in the future). Would be good to get this in place before merge

In either way, it's enough for the kata side to probe the kbs health api, as helm logic has dependency launch logic, which can ensure other services are ready before kbs is health.

fidencio: I already have it working on kata with this in the way it is here. I even got fixes for the health API merged exactly due to this. :-)

Xynnn007:
> thanks! The commit message looks odd. I don't think they need all to be set (at least I was testing air-gapped with just `collateral_service = file:// ...` added).

@mythi 

Fixed, PTAL. Also I think https://github.com/confidential-containers/trustee/pull/1150#discussion_r3389834136 is resolved.

Also, to push forward the work, let's make a due time of the next trustee meeting (until 6.15). If no other points raised, I will get this PR merged. Feel free to give a review until that time.

Also, keep in mind that we can fix things any time before actually integrate this into kata-ci.

deployment/helm-chart/Chart.yaml
fitzthum: ok. i guess Kata tracks the versions of the components separately so we don't have to figure this out quite yet.

deployment/helm-chart/README.md
```diff
@@ -0,0 +1,202 @@
...
+  --path my_repo/resource_type/demo --resource-file /tmp/demo-resource.txt
+```
+
+Fetch a resource by KBS URI path (`get-resource` is a top-level subcommand; it follows the normal attestation / token flow for your client build and policy):
+
+```bash
+kbs-client --url http://127.0.0.1:8080 get-resource --path my_repo/resource_type/demo
+```
+
+## Configuration (`values.yaml` and overrides)
```
mythi: How about using some tool to handle the README and the values in particular. One example I know we have used: https://github.com/norwoodj/helm-docs

deployment/helm-chart/templates/_helpers.tpl
```diff
@@ -121,6 +122,20 @@ AS verifier config JSON fragment
...
 {{- end -}}
+{{- if and $dcap $nv }},
+{{- end }}
+{{ if $dcap -}}
+{{- $_ := required "as.verifier.dcap.collateral_service must be set when as.verifier.dcap is set" (trim (default "" $dcap.collateral_service)) -}}
+"dcap_verifier": {
+    "collateral_service": "{{ $dcap.collateral_service }}"
+    {{- if hasKey $dcap "use_secure_cert" }},
+    "use_secure_cert": {{ $dcap.use_secure_cert }}
+    {{- end }}
```
mythi: this is not used by the code at the moment so the chat does not need it IMO. I will submit a PR to drop it from the docs.

Xynnn007:
@mythi helm-docs updated. `use_secure_cert` deleted. PTAL.

fitzthum (approved):
LGTM

We still have work to do after this PR, but I think we should probably get this in and iterate on it.

Xynnn007:
I think most comments have been resolved. Others are recorded as issues with `Helm Chart |` name prefix. Let me get this in and start with other work.

PR #1151, opened 2026-01-15
verifier: add versioned evidence for az-{snp,tdx}-vtpm TEEs 

mkulke:
corresponding PR on GC: https://github.com/confidential-containers/guest-components/pull/1267

Introduce versioned evidence handling for az-snp-vtpm and az-tdx-vtpm modules to enable schema evolution between attester and verifier.

V0 evidence (no version field) is automatically detected and normalized to V1 format, maintaining backward compatibility with existing attesters.

- Non-versioned evidence will be considered as v0 and normalized into a versioned evidence.
- Inlined TPM verification functions (verify_tpm_*). They are somewhat trivial, and it makes to understand and review the verification if we don't branch out into external crates. We might want to consolidate them further with the TPM verifier.
- Added tests to confirm V0 and V1 evidence formats can be verified.

- Update az-{snp,tdx}-vtpm crates to 0.8.0: 

There is a problem with the released version of the sev crate (7.1.0 currently), it currently fails to build properly as the feature flag combination used in trustee breaks the sev build.

Hence the sev crate was bumped to a git revision post 7.1.0 that contains the requried fix. Now, since the az-snp-vtpm crate also imports sev w/ version=7, there will be mismatches of sev versions being used.

The next release of the sev crate should fix this, in the meantime a project-wide override for the sev crate has been added. Once the sev crate has been releaased, we are able to remove the override.


deps/verifier/Cargo.toml
```diff
@@ -65,7 +65,10 @@ scroll = { version = "0.13.0", default-features = false, features = [
 serde.workspace = true
 serde_json.workspace = true
 serde_with = { workspace = true, optional = true }
-sev = { git = "https://github.com/virtee/sev", rev = "d487809188c3c92e397bc92e2cb2a85eb982c980", default-features = false, features = [
+# note: this one is overriden in the workspace Cargo.toml, because the
+# 7.1.0 version doesn't build w/ sev + serde features enabled. The
+# override can be removed once 7.2.0 is released
+sev = { version = "7", default-features = false, features = [
```
Xynnn007, line 71: Do you think we should wait for the release or this patch is ok?

deps/verifier/src/az_tdx_vtpm/compat.rs
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) Microsoft Corporation.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+//! Compatibility layer for legacy (v0) evidence format conversion.
```
mythi, line 6: could the ITA code use this compat layer too?

https://github.com/confidential-containers/trustee/blob/f342a12115c67ea3af769fcd57e450f569ff7377/kbs/src/attestation/intel_trust_authority/mod.rs#L44-L48

deps/verifier/Cargo.toml
mkulke: we were already using the unreleased github version. I was just moving it to the global Cargo.tom, so also the transitive dependencies are using it and we don't get into incompatibilities.

deps/verifier/src/az_tdx_vtpm/compat.rs
mkulke: right, good catch, I haven't thought about ITA. We probably want to do address it in this PR. If the ITA integration is parsing the evidence, then it might not be able to handle V1 evidences. the v1 evidence represents the data bodies as base64, so the ITA parsing would fail, right?

```rust
#[derive(Serialize, Deserialize)]
pub(super) struct EvidenceV0 {
    pub(super) tpm_quote: Quote,
    pub(super) hcl_report: Vec<u8>,
    pub(super) td_quote: Vec<u8>,
}

/// Attestation evidence for Azure TDX vTPM (v1).
#[serde_as]
#[derive(Serialize, Deserialize)]
pub(super) struct EvidenceV1 {
    pub(super) version: u32,
    pub(super) tpm_quote: TpmQuote,
    #[serde_as(as = "Base64<UrlSafe>")]
    pub(super) hcl_report: Vec<u8>,
    #[serde_as(as = "Base64<UrlSafe>")]
    pub(super) td_quote: Vec<u8>,
}
```

mythi: ITA takes these as base64 encoded so in case of V1, we can take `td_quote` as is (V0 we encode it before sending it). From `hcl_report` we take var data and encode it.

mythi: https://github.com/confidential-containers/trustee/blob/f342a12115c67ea3af769fcd57e450f569ff7377/kbs/src/attestation/intel_trust_authority/mod.rs#L135-L144

mkulke: Hey Mikko, I pushed a commit for that that would remove the ITA module's dep to the az-cvm-vtpm crate with a dep on the local "verifier/?az-tdx-vtpm-verifier" crate. PTAL, it might look a bit clunky, but I think it represents the requirements adequately. otherwise, we would have to duplicate version-aware parsing in the ITA module, which is probably not what we want.

mythi: cc @pawelpros

mkulke: due to the entanglement with platform-specific code in the verifier (tpm, sev) that will prohibit building ITA w/ s390x and arm targets, I came to the conclusion that we indeed want to inline the parsing of the azure evidence in the ITA code

Cargo.toml
```diff
@@ -77,3 +77,9 @@ tonic-prost = "0.14"
 tonic-prost-build = "0.14"
 tracing = "0.1.44"
 tracing-subscriber = { version = "0.3.22", features = ["env-filter"] }
+
+[patch.crates-io]
+az-snp-vtpm = { git = "https://github.com/kinvolk/azure-cvm-tooling", rev = "83cd2a3" }
+az-tdx-vtpm = { git = "https://github.com/kinvolk/azure-cvm-tooling", rev = "83cd2a3" }
+az-cvm-vtpm = { git = "https://github.com/kinvolk/azure-cvm-tooling", rev = "83cd2a3" }
```
fitzthum, line 84: We have been using release versions here, right? Are we waiting for a release on this?

deps/verifier/src/az_tdx_vtpm/mod.rs
```diff
@@ -50,29 +46,46 @@ impl Verifier for AzTdxVtpm {
         let evidence = serde_json::from_value::<Evidence>(evidence)
             .context("Failed to deserialize Azure vTPM TDX evidence")?;
 
-        let hcl_report = HclReport::new(evidence.hcl_report)?;
-        verify_tpm_signature(&evidence.tpm_quote, &hcl_report)?;
+        // Normalize to V1 format
+        // let evidence = evidence.try_normalize()?;
```
fitzthum: Is this supposed to be commented out?

mkulke: nope, leftover. thx

Cargo.toml
mkulke: yup. the latest release of those crates reference sev v7.  when using its default-features it break compilation on non x86_64 targets (but s390x KBS wants to also verify also e.g. SNP evidence).

I opened an issue on the sev repo, maybe we get this addressed in a 7.2 release, until then the features are tweaked in a branch that will allow unbreak s390x and arm.

deps/verifier/Cargo.toml
mkulke: we already had to use a patched version, I only moved the override to a central place.

mkulke:
let's merge this after the v0.17 release, it wouldn't break anything or change behaviour. but it would be better to have less unreleased deps in a release.

fitzthum:
Release is basically done now. I think we can merge.

PR #1152, opened 2026-01-15
kbs: metrics endpoint documentation

pmores:
This PR proposes documentation of the Prometheus endpoint to finish issue #543.

Please note that the final two sections (Prometheus config example & k8s deployment) are just what I and my AI :-) _think_ should work.  I'm proposing those sections since having them has been deemed useful but I **haven't tested them** since I have no access to a suitable environment at the moment.  Unless someone can verify them during review I believe they'd better be dropped from the proposal.


Xynnn007 (approved):
LGTM. Thanks!

PR #1155, opened 2026-01-23
as: DCAP configuration and DCAP TCB reference values

mythi:
The main purpose is to move to use "early" TCB reference values. While that is possible even with the existing `/etc/sgx_qcnl_default.conf` I hope to get rid of it in favor of "verifier config" in Trustee. It's a bit problematic
with QCNL: see f403032 commit for details.

### DIFF
--- a/attestation-service/docs/config.md
+++ b/attestation-service/docs/config.md
@@ -18,9 +18,11 @@ section:
 | `work_dir`                 | String                      | The location for Attestation Service to store data. | False      | Firstly try to read from ENV `AS_WORK_DIR`. If not any, use `/opt/confidential-containers/attestation-service`       |
 | `rvps_config`              | [RVPSConfiguration][2]      | RVPS configuration                                  | False      | -       |
 | `attestation_token_broker` | [AttestationTokenBroker][1]  | Attestation result token configuration.            | False      | -       |
+| `verifier_config`          | [VerifierConfig][3]          | TEE verifier-specific configuration.               | False      | -       |
 
 [1]: #attestationtokenbroker
 [2]: #rvps-configuration
+[3]: #verifier-configuration
 
 #### AttestationTokenBroker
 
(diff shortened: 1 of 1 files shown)


deps/verifier/src/intel_dcap/mod.rs
```diff
@@ -7,14 +7,73 @@ use intel_tee_quote_verification_rs::{
...
+use std::fs::File;
+use std::io::{ErrorKind, Write};
 use std::mem;
 use std::time::{Duration, SystemTime};
 use tracing::{debug, warn};
 
 mod claims;
 mod error;
 
+const INTEL_PCS_URL: &str = "https://api.trustedservices.intel.com/sgx/certification/v4/";
```
fitzthum, line 22: `PCS` or `PCCS` I guess I've actually seen it both ways...

deps/verifier/src/lib.rs
```diff
@@ -111,6 +127,10 @@ pub async fn to_verifier(
         Tee::Sgx => {
             cfg_if::cfg_if! {
                 if #[cfg(feature = "sgx-verifier")] {
+                    let c = _config.map(|c| c.dcap_verifier).unwrap_or(None);
+                    if intel_dcap::set_qcnl_config(c).is_ok() {
+                        debug!("DCAP QCNL configuration was written to $QCNL_CONF_PATH.");
+                    }
```
fitzthum: I feel like this could be less repetitive. Maybe move the debug print into the set `set_qcnl_config` method.  Can you calculate the `c` variable outside of the match? Also, better name for `c` would be good.

mythi: updated. is it any better?

deps/verifier/src/intel_dcap/mod.rs
mythi: `PCS` is the public Intel service. `PCCS` is the `C`aching service.

deps/verifier/src/intel_dcap/mod.rs
```diff
@@ -7,14 +7,74 @@ use intel_tee_quote_verification_rs::{
...
+    env::var("QCNL_CONF_PATH")
+        .map_err(std::io::Error::other)
+        .and_then(File::create_new)
+        .and_then(|mut f| {
+            f.write_all(
+                serde_json::to_string(&c.unwrap_or_default())
+                    .map_err(|_| std::io::Error::from(ErrorKind::InvalidInput))?
+                    .as_bytes(),
+            )
+        })
```
mkulke, line 61: the chained combinators are valid, but a bit undiomatic in modern rust, is there a reason why the ? operator is not used?

attestation-service/docs/config.md
```diff
@@ -141,3 +235,143 @@ Configurations for token signer
...
+    "rvps_config": {
+        "type": "GrpcRemote",
+        "address": "127.0.0.1:50003"
+    },
+    "attestation_token_broker": {
+        "duration_min": 5
+    },
+    "verifier_config": {
+        "dcap_verifier": {
+            "collateral_service": "https://api.trustedservices.intel.com/sgx/certification/v4/",
```
mkulke, line 371: I suppose we need to introduce a config option to specify the azure collateral service?

docker-compose.yml
```diff
@@ -31,13 +31,15 @@ services:
     image: ghcr.io/confidential-containers/staged-images/coco-as-grpc:latest
     environment:
       - RUST_LOG
+      - QCNL_CONF_PATH=/run/dcap/qcnl.conf
```
mkulke, line 34: we probably also want to drop this from the e2e tests?

attestation-service/docs/config.md
mythi: Last time I checked Azure THIM did not support pulling TDX verification collateral. It would use the same `collateral_service` config still. I'm also learning that `pccs_url` is not used for verification collateral so for Trustee it's only `collateral_service`.

docker-compose.yml
mythi: you mean to use `/etc/sgx_default_qcnl.conf`?

deps/verifier/src/intel_dcap/mod.rs
mythi: no particular reasons other than IIRC my thinking was to have all possible errors printed at the end.

docker-compose.yml
mkulke: ah, nevermind. misread. it's probably still good to exercise the path and set a custom config, but this can be done another time.

attestation-service/docs/config.md
```diff
@@ -141,3 +235,143 @@ Configurations for token signer
...
+                    "type": "KDS",
+                    "base_url": "https://kdsintf.amd.com"
+                }
+            ]
+        }
+    }
+}
+```
+
+Configuration with AMD SEV-SNP verifier using offline store:
```
fitzthum, line 287: Might want to link here https://github.com/confidential-containers/trustee/blob/main/attestation-service/docs/amd-offline-certificate-cache.md

attestation-service/docs/config.md
```diff
@@ -141,3 +235,143 @@ Configurations for token signer
...
+                    "type": "OfflineStore",
+                    "path": "/var/lib/attestation-service/kds-store"
+                }
+            ]
+        }
+    }
+}
+```
+
+Configuration with NVIDIA GPU verifier (local verification):
```
fitzthum, line 313: Might want to link to https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/nvidia/README.md

deps/verifier/src/lib.rs
fitzthum: a bit

mythi: I thought about moving this outside the match too but that would have required additional `cfg()` checks and currently match covers the whole function...

attestation-service/docs/config.md
mythi: added.

deps/verifier/src/lib.rs
fitzthum: Yeah, i think it is fine, although people might not totally expect to find this config setup code in the `to_verifier` method.

mythi: Right, not ideal I guess but it saved me from having to move "new()" or similar for the three verifiers.

PR #1156, opened 2026-01-29
Key-Value Storage and Policy Engine changes for Trustee Stateless

Xynnn007:
Thread #1092 

I made some modifications to my original stateless trustee PR, mainly to address @mkulke 's point about keeping configuration files in one place as much as possible. To this end, I made changes to the policy-engine and key-value libraries, as well as the kbs/as/rvps libraries. I split this into two PRs; this first PR primarily involves changes to the underlying libraries. The overall presentation is in [the PR](https://github.com/confidential-containers/trustee/pull/1129)

To sum up, the final config looks like (e.g. kbs)

```toml
[http_server]
sockets = ["0.0.0.0:8080"]
insecure_http = true

[admin]
type = "InsecureAllowAll"

[attestation_token]

# Unified storage backend configuration
# This single configuration will be used for:
# - KBS policy engine (instance: "kbs")
# - Resource plugin storage (instance: "resource")
# - Built-in AS policy storage (instance: "attestation-service-policy")
# - Built-in AS RVPS storage (instance: "reference-value")
[storage_backend]
storage_type = "LocalFs"

[storage_backend.backends.local_fs]
dir_path = "/opt/confidential-containers/storage"

[attestation_service]
type = "coco_as_builtin"
work_dir = "/opt/confidential-containers/attestation-service"

duration_min = 5
# Note: policy_storage will be overridden by storage_backend if provided

[attestation_service.rvps_config]
type = "BuiltIn"
# Note: storage will be overridden by storage_backend if provided
# Optional: configure extractors
# [attestation_service.rvps_config.extractors]
# swid_extractor = {}

[[plugins]]
name = "resource"
backend = "kvstorage"
# Note: When using kvstorage backend, storage will be overridden by storage_backend if provided
```

### DIFF
--- a/deps/policy-engine/src/lib.rs
+++ b/deps/policy-engine/src/lib.rs
@@ -11,7 +11,7 @@ pub mod error;
 pub use error::*;
 use key_value_storage::SetParameters;
 
-pub use key_value_storage::{KeyValueStorage, KeyValueStorageConfig};
+pub use key_value_storage::KeyValueStorage;
 
 pub trait EngineTrait {
     /// The suffix of the policy file.
--- a/deps/policy-engine/src/policy/rego.rs
+++ b/deps/policy-engine/src/policy/rego.rs
@@ -118,9 +118,9 @@ impl Regorus {
 }
 
 impl PolicyEngine<Regorus> {
-    pub async fn new(storage: KeyValueStorageInstance) -> Result<Self> {
+    pub fn new(storage: KeyValueStorageInstance) -> Self {
         let engine = Regorus::default();
-        Ok(Self { storage, engine })
+        Self { storage, engine }
     }
 
     pub async fn evaluate_rego(


deps/key-value-storage/src/lib.rs
```diff
@@ -77,18 +82,101 @@ impl KeyValueStorageConfig {
...
+    local_fs: Option<local_fs::ShimConfig>,
+}
+
+impl KeyValueStorageStructConfig {
+    /// Convert the configuration to a client with the given instance name.
+    ///
+    /// The instance name is used to interpreted by the concrete backend, e.g.
+    /// the PostgreSQL backend will use the instance name as the table name.
+    ///
+    /// The
```
fitzthum: The...

deps/key-value-storage/src/local_fs/mod.rs
```diff
@@ -18,13 +18,42 @@ use crate::{
...
+impl Default for ShimConfig {
+    fn default() -> Self {
+        Self {
+            dir_path: DEFAULT_DIR_PATH.to_string(),
+        }
+    }
+}
+
+impl ShimConfig {
+    pub fn to_instance_config(&self, instance: &str) -> Config {
```
fitzthum: I guess one of the key ideas here is to have the shim vs the instance. I don't think these terms are very clear, but my fundamental question is why not just set the instance at runtime. As in when we make a request to the kv store, we could provide the "instance." Then we wouldn't need any shim config. We would just use the various config structs directly (but remove the instance field from them).

deps/key-value-storage/src/lib.rs
```diff
@@ -177,6 +139,18 @@ impl KeyValueStorageStructConfig {
...
+/// Unified storage backend configuration
+#[derive(Debug, Clone, Deserialize, PartialEq, Default)]
+#[serde(default)]
+pub struct StorageBackendConfig {
+    // Note: this field is aiming to serving all the persistent storages in KBS.
+    /// Storage backend type
+    pub storage_type: KeyValueStorageType,
+
+    /// Backend-specific configuration
+    pub backends: KeyValueStorageStructConfig,
```
fitzthum, line 146: Can we combine these two fields into one? We've often used an enum with a config in each variant for something like this. Either way, this should be `backend` not `backends`, right?

deps/key-value-storage/README.md
```diff
@@ -50,6 +50,122 @@ Stores key-value pairs in a PostgreSQL database table.
 
 See [Postgres README](src/postgres/README.md) for more details.
 
+## Unified Storage Backend Configuration
+
+The `StorageBackendConfig` provides a unified way to configure storage backends that can be shared across multiple components in an application. This is particularly useful for applications like KBS (Key Broker Service) that need to manage multiple storage instances for different purposes.
+
+### What is an Instance?
```
fitzthum: I don't really like this term. To me an instance sounds like it would be one independent instance of a backend, like an entire database. What we're really talking about here is just a subset of the database, right? This is more like a category or a context. Not sure the best term tho yet.

deps/key-value-storage/src/local_fs/mod.rs
Xynnn007: There is a [thread](https://github.com/confidential-containers/trustee/pull/1131#discussion_r2661232806).  

In short, my original PR version allowed precise configuration of the "instance" for each storage and each place. Following @mkulke's suggestion, I hardcoded the instance name. This way, users only need to configure one backend storage, and the subsequent instances are automatically generated, eliminating complex maintenance. The benefits of sacrificing this configuration flexibility could outweigh its drawbacks.

deps/key-value-storage/src/lib.rs
Xynnn007: As you can see, the configuration of which storage we use is separated from all user-defined storage. This is because I also want to decouple the session state of RCAR in the future. In my opinion, storing session state in an in-memory database like Redis is better than storing it in a traditional database because access and changes are more frequent. I can simply add a field to specify the storage it needs.

deps/key-value-storage/README.md
Xynnn007: What about "namespace"?

fitzthum: That's good.

deps/key-value-storage/src/lib.rs
fitzthum: If the idea is to use different storage types for different components, why not specify the configuration for those for each component. We can have one global storage backend enum with a config inside it and then each component can have an optional storage backend enum.

Having multiple storage backends centrally is a bit confusing and what if you wanted to have two components use the same type of backend but with different configurations? This seems to assume that there would be at most one of each.

deps/key-value-storage/src/local_fs/mod.rs
fitzthum: Well this seems like a more complicated way to have something that is less powerful. I guess I am fine with it if you can find a better name for the `ShimConfig` or figure out way to not need that struct at all (like an optional field or something).

Xynnn007: In fact We can fully get rid of original `Config` structs and rename the `ShimConfig`s to `Config`s, and change the `new` function like

```rust
impl PostgresClient {
    pub async fn new(config: Config, namespace: String) -> Result<Self> {
        ...
    }
```

where the namespace is table for postgres, and the new `Config` is the current `ShimConfig` which does not have a table field. wdyt?

deps/key-value-storage/src/lib.rs
Xynnn007: Yes. You are right that in this way we do not support different components to use different configurations of a same backend. I believe both methods have their pros and cons. I'll make examples then ask @bpradipt @lmilleri - who might be the direct consumer to trustee for their comments.

## Way 1 - Fine grained configuration

```toml
[http_server]
sockets = ["0.0.0.0:8080"]
insecure_http = true

[admin]
type = "InsecureAllowAll"

[attestation_token]

[attestation_service]
type = "coco_as_builtin"

duration_min = 5
# point 1: attestation service policy storage
[attestation_service.attestation_token_broker.policy_storage]
storage_type = "LocalJson"
file_path = "/opt/confidential-containers/attestation-service/ear-policies.json"

[attestation_service.rvps_config]
type = "BuiltIn"
# point 2: reference values storage
[attestation_service.rvps_config.reference_values_storage]
storage_type = "LocalFs"
file_path = "/opt/confidential-containers/attestation-service/reference_values"

[[plugins]]
name = "resource"
backend = "kvstorage"
# point 3: resource plugin storage
storage_type = "Memory"

# point 4: policy engine storage
[policy_engine]
storage_type = "Postgres"
db = "test"
table = "test"
username = "user"
port = 5432
host = "1270.0.1"
```

## Way 2 - An Overall configuration point

```toml
# Unified storage backend configuration
# This single configuration will be used for:
# - KBS policy engine (namespace: "kbs")
# - Resource plugin storage (namespace: "resource")
# - Built-in AS policy storage (namespace: "attestation-service-policy")
# - Built-in AS RVPS storage (namespace: "reference-value")
[storage_backend]
storage_type = "LocalFs"

[storage_backend.backends.local_fs]
dir_path = "/opt/confidential-containers/trustee"

[http_server]
sockets = ["0.0.0.0:8080"]
insecure_http = true

[admin]
type = "InsecureAllowAll"

[attestation_token]

[attestation_service]
type = "coco_as_builtin"

duration_min = 5

[attestation_service.rvps_config]
type = "BuiltIn"

[[plugins]]
name = "resource"
backend = "kvstorage"
```

Obviously,

| Ability | Way 1 | Way 2 | 
|---|---|---|
| Set different storage point with different backends and different configurations | N | Y |
| When there is a change upon backend storage, we only need to change one position / The configuration is smaller | Y | N|

deps/key-value-storage/README.md
Xynnn007: I will push the whole changes once the configuration preference is decided.

deps/key-value-storage/src/local_fs/mod.rs
Xynnn007: Let me make a separate commit for this for a view.

deps/key-value-storage/src/lib.rs
Xynnn007: btw after creating this table, I interviewed some users within Alibaba Cloud who use Trustee. They preferred Option 2 – they preferred the simpler configuration and they are not developers thus do not care about the complex cases... :-)

mkulke: maybe I'm misunderstanding, but could we offer a generic option that is overridable in the individual subsections? somewhat like e.g. `working_directory`  field in github actions that you can define on job level and then override on individual steps?

Obviously, this only works if we have no relations between the entities that you would like to define in a schema. We should stop and be sure that we rule this out as a concept (e.g. a user_id on a policy_table). Hand-crafting transactions over relational data that doesn't sit in a common schema is very annoying and bug prone.

If we want to do this in the future, then I'd be leaning towards always prescribing a database and removing flexibility:

app: sqlite for local deployments, postgres otherwise. sessions: sled local development , redis otherwise

deps/key-value-storage/src/local_fs/mod.rs
fitzthum: that's fine

deps/key-value-storage/src/lib.rs
fitzthum: Yeah, similar to @mkulke it seems like the easiest for users would be to have one global entry for the kvstorage. In most cases this is the only thing that users would adjust. All of our different components would default to using kvstorage backend, which would use the global config. It seems like this is the situation that we would want most people to use.

Then, optionally components like the rvps and resource plugin would accept a storage config. If this is set, a separate kvstorage would be created for this component with whatever config is specified. I guess this is closer to option 2 but not quite the same.

btw is the plan to keep supporting other storage backends besides kvstore, e.g. the existing json storage in the RVPS? In the examples above it's a little unclear when we are talking about the kvstore (and whatever backend it uses) and when we're talking about a different storage backend entirely.

Xynnn007: Got it. I will think about the override logic. @mkulke up to now we do not need this kind of abstraction, but we can explicitly get rid of non-db storages in future if we need.

btw @fitzthum do you mean the `kvstorage` under `plugins` section of way 2? It might not be a good name and it means to use the unified storage backend configuration. The other options for the `backend` can be pkcs11, etc.
And the original JSON storage is included as a plugin `LocalJson` in the `key-value-storage` crate.

Xynnn007: btw, does it make sense to add the override logic after the current series of stateless refactoring work? I do not think it would require huge change, but prefer to finish the stateless work first.

fitzthum: Yes, the plugin backend abstraction should stay since we have other implementations. For RVPS and other things I don't think we need that intermediate abstraction. They can simply always use a kv storage of some kind. Either the global one or one specified for that component.

I think it's probably fine to refactor the config later, although that is in some ways a breaking change.

deps/key-value-storage/src/local_fs/mod.rs
Xynnn007: @fitzthum PTAL. I have fixup the commit in the tree as it's somehow straightforward. This [commit](https://github.com/confidential-containers/trustee/pull/1156/changes/352f0a6e0cfc7102538eeebf2c7adbfe1288f9e7)

Issue #1160, opened 2026-01-30
trustee-cli cannot be built without verifier dependencies


mythi:
### Describe the bug

https://github.com/confidential-containers/trustee/commit/9abce393febd42c811ee69c4ec1da5b992888d81 dropped the verifier dependencies (TPM and DCAP libs) based on checking the code that these are not needed. The PR CI did not catch there are issues. However, a post-merge action shows issues and currently the image cannot be built.

Two problems:

- [ ] rework trustee-cli to not depend on `kbs::attestation::config::AttestationServiceConfig::CoCoASBuiltIn;` for  config paths because this expects KBS is built using `coco_as_builtin` feature which pulls in verifier dependencies.
- [ ] add trustee-cli image to PR checks

### How to reproduce

```bash
$ podman build . -f tools/trustee-cli/Dockerfile
```

### CoCo version information

HEAD

### What TEE are you seeing the problem on

None

### Failing command and relevant log output

```shell

```

PR #1167, opened 2026-02-11
Revert "tools: drop verifier dependencies from trustee-cli image"

fitzthum:
It turns out we need these deps for the trustee-cli because the trustee-cli actually depends on the verifiers (trustee-cli can spin up a full trustee as a process, including the verifiers). This should unbreak the trustee-cli post-merge test.

Fixes: https://github.com/confidential-containers/trustee/issues/1160

cc @mythi 

This reverts commit 9abce393febd42c811ee69c4ec1da5b992888d81.


mythi:
The actions are back to normal https://github.com/confidential-containers/trustee/actions/runs/21976125844

PR #1169, opened 2026-02-12
verifier: bump az-{snp,tdx}-vtpm crates to 0.8.1

mkulke:
This crate bump will make the global patch overrides not required any more because the new revision only depends on the `snp` feature in the sev crate, which unbreaks compilation on non x86_64 targets.

Other smaller changes:

- we can remove some nuisance json roundtripping, because the crate exposes the signature for a tpm quote
- we also verify the VCEK is valid at the time of verification.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -813,8 +813,9 @@ dependencies = [
 
 [[package]]
 name = "az-cvm-vtpm"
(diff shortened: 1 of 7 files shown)


mythi (approved):
LGTM

PR #1171, opened 2026-02-13
agents: refine agent instructions

mkulke:
clarify some aspects from observing an agent session when it worked on PR #1170 I think it's mostly there for trivial fixes, but some aspects need tweaking


mythi (approved):
nit: the PR link is a bit off but LGTM otherwise

PR #1177, opened 2026-02-18
kbs: add missing cfg guard for coco-as-builtin import in tests

Cropi:
The `local_fs` and `ReferenceValueStorageConfig` imports in the config test module are only used by  `coco-as-builtin` test cases but were missing the corresponding `#[cfg(feature = "coco-as-builtin")]` guard, producing unused import warnings when building with other feature sets.
This is a cosmetic change. Does not affect functionality.

### DIFF
--- a/kbs/src/config.rs
+++ b/kbs/src/config.rs
@@ -136,6 +136,7 @@ mod tests {
         rvps::{grpc::RvpsRemoteConfig, RvpsConfig, RvpsCrateConfig},
     };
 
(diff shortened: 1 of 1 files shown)


Cropi:
> LGTM
> 
> are we missing some configuration in our workflows to not pick this up in the CI?

I was experimenting with the project when I bumped into this. By the time I started to prepare changes the problem disappeared but it made sense to me to submit the patch.

PR #1181, opened 2026-02-19
verifier: extend NVIDIA SPDM opaque data parser for newer firmware

josegarvayo-oc:
Decode additional fields introduced in later versions of NVIDIA firmware. 

Found this gap while testing on Azure Confidential VM with H100 GPU. 

No new test added, but unignored local_verifier_coco test that was failing. Now it pass with this change.

Field definitions taken from NVIDIA nvtrust repo:

https://github.com/NVIDIA/nvtrust/blob/main/guest_tools/gpu_verifiers/local_gpu_verifier/src/verifier/attestation/spdm_msrt_resp_msg.py

### DIFF
--- a/deps/verifier/src/nvidia/mod.rs
+++ b/deps/verifier/src/nvidia/mod.rs
@@ -360,9 +360,6 @@ mod tests {
     // Use the remote verifier with evidence from a CoCo CI run
     #[ignore]
     #[case::remote_verifier_coco(false, "87d8e24ab336adafe228d49e83d745f6dba4ae505372b6a5704820856b343fece279b616efefc2aae21da80cf5581250", include_str!("../../test_data/nvidia/hopper_coco_report1.txt"), include_str!("../../test_data/nvidia/hopper_coco_certs1.txt"), Architecture::Hopper)]
-    // The local verifier does not currently work with this report, which is from a newer device
-    // that has some unknown fields in opaque data.
-    #[ignore]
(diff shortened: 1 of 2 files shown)


deps/verifier/src/nvidia/spdm_response.rs
```diff
@@ -316,6 +321,28 @@ impl OpaqueData {
 
         Ok(Value::Object(values))
     }
+
+    fn decode_feature_flag(bytes: &[u8]) -> Result<Value> {
+        let value = Self::decode_le_u64(bytes)?;
```
mkulke, line 326: maybe we can use u64::from_le_bytes?

and do we really need padding? I would assume the input bytes in hardware reports are of a fixed size (and we should assert that).

```
let mut padded = [0u8; 8];
let len = slice.len().min(8);
padded[..len].copy_from_slice(&slice[..len]);
let val = u64::from_le_bytes(padded);
```

mkulke (commented):
thanks, someone from nvidia should take a glance (cc @fitzthum @dcmiddle @fidencio). I assume by un-ignoring the test case we are asserting the newly added enums? If not, can we cover those in a unit test?

deps/verifier/src/nvidia/spdm_response.rs
josegarvayo-oc: u64::from_le_bytes() seems better; I would change that. 

I would add the padding, just in case.

deps/verifier/src/nvidia/mod.rs
dcmiddle, line 357: @fitzthum not sure if 'run these tests' is relevant for this PR. Thought you would know best.

fitzthum: Yeah should be fine. This PR concerns the local verifier and increases coverage. Still might be good to give this test a spin locally tho to make sure nothing weird happened. @josegarvayo-oc

josegarvayo-oc: We tested on a Standard_NCC40ads_H100_v5 instance on azure, without this change AS always failed GPU verification

PR #1184, opened 2026-02-20
verifier: az_snp_vtpm now validates VCEK against any AMD CPU gen.

josegarvayo-oc:
Previously az_snp_vtpm only used AMD Milan cert to validate VCEK, resulting in verifier errors when testing using AMD Genoa CPU, which is the current available in Azure.

Now it tries VCEK against all AMD vendor certs, and only fails if all of them fail.

I could not find a way to get CPU Processor Generation from evidence, and Milan generation seems to be deprecated.  Genoa is the one current available in Azure (https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/gpu-accelerated/nc-family).


josegarvayo-oc:
> thanks, good initiative. I think we can use the snp verifiers extractor for processor generation (make fn pub(crate)), and query this in the verifier:
> 
> ```rust
> ...
>         let proc_gen: ProcessorGeneration = get_processor_generation(&snp_report)?;
> 
>         // Get vendor certs for specific processor type
>         let Some(vendor_certs) = CERT_CHAINS.get(&proc_gen) else {
>             bail!("Vendor certs not found for processor type: {proc_gen:?}");
>         };
> 
>         let ask = (&vendor_certs.ask).into();
>         let ark = (&vendor_certs.ark).into();
> 
>         let az_vendor_certs = AzVendorCertificates {
>             ca_chain: AmdChain { ask, ark },
>         };
> 
>         az_vendor_certs
>             .ca_chain
>             ...
> ```
> 
> in this case we don't need constructor for the `AzSnpVtpm` struct anymore. we can annotate it with `#[derive(Default)]`
> 
> and invoke it in verifier/src/lib.rs with ::default(), like the other verifiers.

Thanks for the suggestion.

 I tried but  test_evaluate() fails to evaluate evidence:

    #[rstest]
    #[case::v0(EVIDENCE_V0_JSON)]
    #[case::v1(EVIDENCE_V1_JSON)]
    #[tokio::test]
    async fn test_evaluate(#[case] evidence_json: &str) {

 looks like evidence v0-v1 versions does not provide a cpu family/model id. 

From this line in get_processor_generation I get  comes in v3+, which is the only one accepted for snp verifier I guess:

 let cpu_fam = att_report
        .cpuid_fam_id
        .ok_or_else(|| anyhow::anyhow!("Attestation report version 3+ is missing CPU family ID"))?;

Shall I assume report are always in version 3+, and set these tests to be #ignore?

deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -149,12 +125,21 @@ impl Verifier for AzSnpVtpm {
...
         let vcek = Vcek(vcek_x509);
 
-        self.vendor_certs
-            .ca_chain
-            .validate()
-            .context("Failed to validate CA chain")?;
-        vcek.validate(&self.vendor_certs.ca_chain)
-            .context("Failed to validate VCEK")?;
+        let proc_gen = get_processor_generation(&snp_report).context("Failed to get CPU Generation")?;
+        if let Some(vendor_certs) = CERT_CHAINS.get(&proc_gen) {
```
mkulke: nit: use let Some ... else { } to bail out early

Issue #836, opened 2025-06-26
KBS: Missing Features for Production


zvonkok:
Here is a list of things Trustee doesn't do: 
 
- [ ] KBS is not close to FIPS 140-3 level 1 (software module) nor is it aligned with the relevant protection profiles for HSMs (including soft HSMs).
- [ ] It is not well suited for a system of system attestation. 
- [ ] It’s not compliant with KMIP either
- [ ] Does not plug into OSCP
- [ ] Handling of the RIM database
- [ ] Customizable attestation policies
- [ ] Sealed secrets
- [ ] No authorization for resource retrieval (There should be authorization policies so users can only retrieve their own resources)
- [ ] No authentication for resource retrieval 
- [ ] Designed for just one project or single build (multi-project or mult-build or namespace separation)
- [ ] There is only one resource policy. If two projects both set up different resource policies, the new policy can overwritten the previous one 
- [ ] Resource identification only by path
- [ ] High availability/Scalabitly
- [ ] HSM requirements / PKCS#11 ?
- [ ] Rate limitation to other services like AMD KDS, caching service? 

Should an OTS solution like Vault be preferred? (or a cloud HSM so long as you’re not single infrastructure vendor) and Trustee brokers it to the backend?

Xynnn007:
Hi @zvonkok Nice topic! Some naive questions and thanks in advance

> It is not well suited for a system of system attestation.

What do you mean by system attestation?

> Customizable attestation policies

What do you mean by "Customizable"?

zvonkok:
> What do you mean by system attestation?



Aiming here for a composite, hierarchical attestation flow that produces one verifiable statement for the **entire** distributed platform, not just the individual parts, also referring to layered attestation, where each part depends on the layer below it, throughout the complete stack, host + guest + container.

fitzthum:
> There is no way for a relying party to inject a Verifier policy for attestation at runtime with multiple different policies.

Ah right. This should be pretty easy to add once we have plaintext init data, which is coming soon. We can define a special (optional) init-data key which is use to select the policy.

Xynnn007:
> Aiming here for a composite, hierarchical attestation flow that produces one verifiable statement for the **entire** distributed platform, not just the individual parts, also referring to layered attestation, where each part depends on the layer below it, throughout the complete stack, host + guest + container.

This is the aim we are going to. We are perfecting the guest stack attestation framework and hopefully it is a right way to go
[...]

PR #1188, opened 2026-02-24
External plugin support for KBS

bpradipt:
Creating this as a draft PR to make it easier to discuss and collaborate.

### DIFF
--- a/.github/workflows/test-e2e-kbs.yml
+++ b/.github/workflows/test-e2e-kbs.yml
@@ -139,6 +139,57 @@ jobs:
           VERIFIER: ${{ matrix.verifier }}
         run: make test-kbs-docker-e2e
 
+  # External plugin e2e
+  ext-plugin-e2e-test:
+    name: External Plugin e2e
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    env:
+      OS_VERSION: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        persist-credentials: false
+
+    - name: Install Rust toolchain
+      uses: actions-rust-lang/setup-rust-toolchain@150fca883cd4034361b621bd4e6a9d34e5143606 # v1.15.4
+      with:
+        rustflags: ""
+        cache: false
+
+    - name: Set up rust build cache
+      uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
+      continue-on-error: false
+      with:
+        path: |
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          target/
+        key: rust-${{ runner.arch }}-${{ env.OS_VERSION }}-${{ hashFiles('./Cargo.lock') }}
+
+    - name: Install build dependencies
+      run: sudo apt-get update && sudo apt-get install -y protobuf-compiler libprotobuf-dev libtss2-dev
+
+    - name: Install TDX build dependencies
+      run: |
+        curl -L https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo gpg --dearmor --output /usr/share/keyrings/intel-sgx.gpg
+        echo 'deb [arch=amd64 signed-by=/usr/share/keyrings/intel-sgx.gpg] https://download.01.org/intel-sgx/sgx_repo/ubuntu noble main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list
+        sudo apt-get update
+        sudo apt-get install -y libsgx-dcap-quote-verify-dev libsgx-dcap-default-qpl
+
+    - name: Run external plugin unit tests
+      working-directory: kbs
+      run: make check TEST_FEATURES="external-plugin"
+
+    - name: Install e2e test dependencies
+      working-directory: kbs/test
+      run: make install-dev-dependencies
+
+    - name: Run external plugin e2e tests
+      working-directory: kbs/test
+      run: make e2e-ext-plugin
+
   # Azure vTPM TEE e2e (manual trigger only)
   tdx-e2e-test:
     name: KBS e2e (Azure TDX vTPM)


go-sdk/README.md
fitzthum, line 1: If we're providing a well documented grpc interface, I don't think we need to include Go / Python implementations in the repo. This is more maintenance for limited benefit imo.

deps/verifier/src/intel_dcap/claims.rs
```diff
@@ -114,7 +114,7 @@ pub(crate) fn prepare_custom_claims_map(
     claims_map
 }
 
-fn get_sa_list(sa_list: &[c_char; 450]) -> Value {
+fn get_sa_list(sa_list: &[c_char; 320]) -> Value {
```
fitzthum: Unrelated.

kbs/src/plugins/implementations/external_plugin.rs
```diff
@@ -0,0 +1,511 @@
...
+    async fn validate_auth(
+        &self,
+        _body: &[u8],
+        _query: &HashMap<String, String>,
+        _path: &[&str],
+        _method: &Method,
+    ) -> Result<bool> {
+        // Return config value directly (no gRPC call per Phase 1 decision)
+        Ok(self.config.validate_auth)
+    }
```
fitzthum, line 314: I would consider forwarding this method and the below to the proxy. These methods take the full request, so they can dynamically require admin authentication. This is more granular than what we can capture in the config file.

fitzthum (commented):
A couple comments. Generally I think this is a fine approach, but I would try to reduce the scope as much as possible. Probably don't need the example implementations (in this repo at least). 

Also, why are we modifying the existing plugins?

go-sdk/README.md
bpradipt: Ok.

kbs/src/plugins/implementations/external_plugin.rs
bpradipt: Ok

protos/plugin.proto
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// ValidateAuthRequest carries the HTTP request details for the ValidateAuth RPC.
+// Fields mirror PluginRequest; a distinct message type satisfies the
+// RPC_REQUEST_RESPONSE_UNIQUE lint rule and allows the two RPCs to evolve
+// independently in future versions.
+message ValidateAuthRequest {
+    bytes body = 1;
+    map<string, string> query = 2;
+    repeated string path = 3;
```
Xynnn007, line 49: this means `kbs/v0/a/b/c` will be handled in `["kbs","v0","a","b","c"]`?

kbs/src/attestation/backend.rs
```diff
@@ -431,6 +431,67 @@ impl AttestationService {
...
     }
 
+    /// Build a PluginContext from the current session state.
+    ///
+    /// Checks for a session cookie first, then falls back to a Bearer token.
+    /// For cookie-based sessions the TEE type and attestation status are read
+    /// from the session store.  For Bearer token auth, KBS has already verified
+    /// the token before reaching the plugin handler, so is_attested is set to
+    /// true; tee_type is not available without re-parsing the JWT claims.
+    pub async fn get_plugin_context(&self, request: &HttpRequest) -> crate::plugins::PluginContext {
```
Xynnn007: I like the idea of pluginContext as it shares a same idea with mine to transform the complex attestation context into a simple one for different plugins to use.

I want to do an abstraction upon this `PluginContext` here as a core component working in kbs logic, probably naming to `TrustContext`.

Currently the `is_attested` only means it has an attestation token binding to the session, while an attestation token only means the endorsement is verified based on the current CoCoAS blackbox. This can be perfect-ted in future.

kbs/src/attestation/backend.rs
```diff
@@ -431,6 +431,67 @@ impl AttestationService {
...
+
+        if session.is_expired() {
+            return crate::plugins::PluginContext::default();
+        }
+
+        let (is_attested, tee_type) = match session {
+            SessionStatus::Authed { request: req, .. } => {
+                let tee_type = serde_json::to_string(&req.tee)
+                    .ok()
+                    .map(|s| s.trim_matches('"').to_owned());
```
Xynnn007: Not related to this PR. We are always trying to do `trim` options upon the `Tee` type's serialization. We need to impl a `ToString` or some trait for `Tee` someday.

kbs/src/plugins/plugin_manager.rs
```diff
@@ -16,8 +36,250 @@ use super::{NebulaCaPlugin, NebulaCaPluginConfig};
...
+        // Secure-by-default: external plugins are a new feature with no existing
+        // configs to preserve, so omitting tls_mode should require a CA cert
+        // rather than silently falling back to plaintext.
+        TlsMode::Tls
+    }
+}
+
+#[cfg(feature = "external-plugin")]
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+pub struct ExternalPluginConfig {
```
Xynnn007: I saw a lot of feature flags `external-plugin`. Can we move the definitions into a separate module `implementations/external_plugin.rs`

kbs/src/plugins/plugin_manager.rs
```diff
@@ -16,8 +36,250 @@ use super::{NebulaCaPlugin, NebulaCaPluginConfig};
...
+///
+/// External gRPC plugins are stored directly as `Arc<GrpcPluginProxy>` and
+/// dispatched via `handle_with_context()`, which injects session/TEE metadata
+/// into gRPC request headers.
+#[derive(Clone)]
+pub enum PluginInstance {
+    BuiltIn(Arc<dyn ClientPlugin>),
+    #[cfg(feature = "external-plugin")]
+    External(Arc<GrpcPluginProxy>),
+}
```
Xynnn007: I noticed that the [user-side toml definitions](https://github.com/confidential-containers/trustee/pull/1188/changes#diff-82526e5a414e702f8136dd640c5c16614b427e7106f170858701066468a8664fR25-R30) still follows [the previous style](https://github.com/confidential-containers/trustee/blob/main/kbs/config/kbs-config-grpc.toml#L16-L19), s.t. `name` item is used to choose `external`. Thus we can organize the code in a style that `external` is treated as a same-level plugin as current `pkcs11`, `sample`, etc. Thus we might not need this wrapper. Just add a new `implementations/external.rs` - or if more files are needed, `implementations/external/mod.rs` to have the external plugins forwarder.

Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "kbs",
+    "rust-sdk",
```
Xynnn007: I wonder if we need this rust-sdk. IMOO, a protobuf is enough. But we can keep this for some time and see if the maintaince is good.

kbs/src/plugins/plugin_manager.rs
bpradipt: This is intentional to avoid changing the existing ClientPlugin trait. External plugins need to forward request context (session ID, TEE type,  attestation status) as gRPC metadata, which required a separate dispatch path (handle_with_context). Adding &PluginContext toClientPlugin::handle() would have been a breaking change for all existing plugin implementations.

Infact my earlier implementation was around the approach you described but resulted in changing the existing plugins. So after @fitzthum comment I went with the current approach.

Happy to revisit if there's a preferred way to extend the trait without breaking existing plugins.

protos/plugin.proto
bpradipt: The /kbs/v0/ prefix is stripped by the actix route before reaching this code. For /kbs/v0/a/b/c, the first segment (a) is the plugin name and the path field receives ["b",  "c"]. 
PluginRequest has a comment explaining this but ValidateAuthRequest and NeedsEncryptionRequest doesn't. I'll add the same doc comment to both.

kbs/src/plugins/implementations/external_plugin.rs
```diff
@@ -0,0 +1,900 @@
...
+            .context("Plugin failed initial health check")?;
+
+        // Start background health monitoring
+        let health_task = tokio::spawn(Self::health_monitor(
+            config.clone(),
+            state.clone(),
+            config.plugin_name.clone(),
+        ));
+
+        let timeout = config.timeout_ms.map(Duration::from_millis);
```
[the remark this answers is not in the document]
bpradipt: Looks like a false positive. Wrongly attributing `ensure_auth_key_pair` to external_plugin.rs

kbs/src/plugins/implementations/external_plugin.rs
```diff
@@ -0,0 +1,900 @@
...
+use tonic_health::pb::health_check_response::ServingStatus;
+use tonic_health::pb::health_client::HealthClient;
+use tonic_health::pb::HealthCheckRequest;
+use tracing::{debug, info, warn};
+
+use crate::error::PluginCallError;
+use crate::plugins::external::plugin_api::{
+    kbs_plugin_client::KbsPluginClient, GetCapabilitiesRequest, NeedsEncryptionRequest,
+    PluginRequest, ValidateAuthRequest,
+};
```
Xynnn007: We can also include the code of `crate::plugins::external` into this file and delete `external.rs`?

kbs/src/plugins/plugin_manager.rs
Xynnn007: In fact, in my view, the only difference between external plugin calls and internal integrations is whether or not a gRPC forwarder is involved; they are essentially the same. Both calls need to accept the same parameters, regardless of how the parameters are designed. There is no problems upon the trust relations. 

```rust
        body: &[u8],
        query: &HashMap<String, String>,
        path: &[&str],
        method: &Method,
        context: &PluginContext,
```

kbs/src/plugins/implementations/external_plugin.rs
bpradipt: Good suggestion @Xynnn007

kbs/src/plugins/plugin_manager.rs
```diff
@@ -17,7 +28,131 @@ use super::{NebulaCaPlugin, NebulaCaPluginConfig};
...
+///
+/// External gRPC plugins are stored directly as `Arc<GrpcPluginProxy>` and
+/// dispatched via `handle_with_context()`, which injects session/TEE metadata
+/// into gRPC request headers.
+#[derive(Clone)]
+pub(crate) enum PluginInstance {
+    BuiltIn(Arc<dyn ClientPlugin>),
+    #[cfg(feature = "external-plugin")]
+    External(Arc<GrpcPluginProxy>),
+}
```
Xynnn007: I think we agree to make `GrpcPluginProxy` to a separate `impl ClientPlugin`, w/o introducing the distinguishment of builtin/external here? 

cc @fitzthum

bpradipt: My suggestion will be to avoid it as it'll require (imho) significant refactoring of internal plugins.  Need plugincontext and http status code. We can let the external plugin implementation be separate for now, wait for few iterations to see the adoption and then decide.  Purely a personal opinion

Xynnn007: Ok. Refactoring is not a small work. As the config can look the same as what it is refactored, so internal change does not influence the user side. I am ok with this. Let's do it later if we need.

rust-sdk/src/server.rs
```diff
@@ -0,0 +1,248 @@
...
+        let adapter = PluginHandlerAdapter::new(handler);
+        let plugin_server = KbsPluginServer::new(adapter);
+
+        // Build tonic server, optionally with TLS
+        let mut server_builder = if let Some(tls) = self.tls_config {
+            let tls_config = tls.into_server_tls_config()?;
+            Server::builder().tls_config(tls_config)?
+        } else {
+            Server::builder()
+        };
```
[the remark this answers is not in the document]
Xynnn007: I remembered that we do not need the rust sdk also? Need to ensure
cc @bpradipt @fitzthum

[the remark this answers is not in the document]
bpradipt: Oh. I thought it was only about the other sdks. Would having an example help adopters or it's maintenance headache ?
I'm open for suggestions.
cc @fitzthum @Xynnn007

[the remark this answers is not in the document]
fitzthum: I would lean towards not having it in tree, at least not to start with. Can always add later if people want it.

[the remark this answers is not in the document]
bpradipt: @fitzthum @Xynnn007 I have removed the sdk. There is an example in the test directory that is used for e2e tests.
PTAL

kbs/src/plugins/implementations/external_plugin.rs
```diff
@@ -0,0 +1,904 @@
...
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+pub struct ExternalPluginConfig {
+    pub plugin_name: String,
+    pub endpoint: String,
+    #[serde(default)]
+    pub timeout_ms: Option<u64>,
+
+    // TLS configuration fields
+    #[serde(default)]
+    pub tls_mode: TlsMode,
```
Xynnn007, line 61: I am wondering if this config is needed for a user?

e.g. I am a user, if I
1. want to use http, then I prefer to config only `endpoint` with `http://` prefix.
2. want to use TLS, then I prefer to config only `endpoint` with `https://` prefix, and optional the server's trust anchor (cert (chain))
3. want to use TLS, then I prefer to config `endpoint` with `https://` prefix and the client key/cert. Optionally the server's trust anchor (cert (chain)).


This will not block current PR from my side, but I am considering to do a refactor upon the code. Does it make sense to you?

kbs/src/plugins/implementations/external_plugin.rs
```diff
@@ -0,0 +1,904 @@
...
+pub(crate) enum PluginState {
+    Healthy,
+    Unavailable,
+}
+
+pub(crate) struct GrpcPluginProxy {
+    pool: Pool<PluginGrpcManager>,
+    timeout: Option<Duration>,
+    config: ExternalPluginConfig,
+    state: Arc<RwLock<PluginState>>,
```
Xynnn007: We can use `tokio`'s Rwlock here for async runtime. If we mix sync and async primitives there will be some strange unexpected behavior.

kbs/src/plugins/implementations/external_plugin.rs
```diff
@@ -0,0 +1,904 @@
...
+            config,
+            state,
+            _health_task: health_task,
+            supported_methods,
+        })
+    }
+
+    /// Background task that periodically checks plugin health and updates state.
+    /// Runs every 10 seconds for the lifetime of the proxy.
+    async fn health_monitor(
```
Xynnn007: I am not sure if this health checker is required. As I envision it, the KBS gRPC plugin should forward requests to the target only when a user explicitly initiates a request. If the target is inaccessible, the plugin should—on one hand—return an error to the caller, and—on the other—simply log the error internally within KBS. 

It works more like "lazily" and low cost. 

But this will not block the PR. I am thinking about to change the logic to a admin api to check all plugin status, which will lazily run when the request comes, than heartbeat all life in the following PR. wdyt?

kbs/test/config/external-plugin-attest.toml
```diff
@@ -0,0 +1,45 @@
...
+[[admin.personas]]
+id = "admin"
+public_key_path = "./work/kbs.pem"
+
+[[plugins]]
+name = "external"
+plugin_name = "echo-test"
+endpoint = "http://127.0.0.1:50051"
+tls_mode = "insecure"
+timeout_ms = 5000
```
Xynnn007: There are some changes for config in the [doc](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/config.md#rvps-configuration). Let me give a config with the same aim you want to express in the PR and you can replace the whole.

```toml
[http_server]
sockets = ["127.0.0.1:8087"]
private_key = "./work/https.key"
certificate = "./work/https.crt"
insecure_http = false

[attestation_token]
trusted_certs_paths = ["./work/ca-cert.pem"]

[attestation_service]
type = "coco_as_builtin"

duration_min = 5

[attestation_service.attestation_token_broker.signer]
key_path = "./work/token.key"
cert_path = "./work/token-cert-chain.pem"

[attestation_service.rvps_config]
type = "BuiltIn"

[admin]
type = "Simple"

[[admin.personas]]
id = "admin"
public_key_path = "./work/kbs.pem"

[[plugins]]
name = "external"
plugin_name = "echo-test"
endpoint = "http://127.0.0.1:50051"
tls_mode = "insecure"
timeout_ms = 5000
```

and note that you need to set kbs resource policy (./work/ext-plugin-attest-policy.rego) manually in the test workflow using kbs-client.

kbs/test/config/external-plugin-resource-tls.toml
```diff
@@ -0,0 +1,41 @@
...
+[policy_engine]
+policy_path = "./work/ext-resource-policy.rego"
+
+[[plugins]]
+name = "external"
+plugin_name = "resource"
+endpoint = "https://127.0.0.1:50054"
+timeout_ms = 5000
+tls_mode = "tls"
+ca_cert_path = "./work/echo-plugin-ca.pem"
```
Xynnn007: The same for KBS resource policy.

```toml
[http_server]
sockets = ["127.0.0.1:8089"]
insecure_http = true

[admin]
type = "InsecureAllowAll"

# Token verifier: trust the CA that signed the attestation token signer cert.
[attestation_token]
trusted_certs_paths = ["./work/ca-cert.pem"]

[attestation_service]
type = "coco_as_builtin"

duration_min = 5

[attestation_service.attestation_token_broker.signer]
key_path = "./work/token.key"
cert_path = "./work/token-cert-chain.pem"

[attestation_service.rvps_config]
type = "BuiltIn"

[[plugins]]
name = "external"
plugin_name = "resource"
endpoint = "https://127.0.0.1:50054"
timeout_ms = 5000
tls_mode = "tls"
ca_cert_path = "./work/echo-plugin-ca.pem"
```

kbs/test/config/external-plugin-resource.toml
```diff
@@ -0,0 +1,50 @@
...
+# name lets the test exercise the full kbs-client roundtrip: admin-gated POST
+# to store a secret, then attested GET with JWE decryption to retrieve it.
+#
+# This is not a conflict with the built-in resource plugin: both cannot be
+# registered simultaneously (KBS enforces unique plugin names at startup), and
+# this config intentionally omits the built-in plugin entry.
+plugin_name = "resource"
+endpoint = "http://127.0.0.1:50053"
+tls_mode = "insecure"
+timeout_ms = 5000
```
Xynnn007: The same,
```toml
[http_server]
sockets = ["127.0.0.1:8088"]
insecure_http = true

[admin]
type = "InsecureAllowAll"

# Token verifier: trust the CA that signed the attestation token signer cert.
# The AS signs tokens with token.key/token-cert-chain.pem (same CA used by kbs.toml).
[attestation_token]
trusted_certs_paths = ["./work/ca-cert.pem"]

[attestation_service]
type = "coco_as_builtin"

duration_min = 5

[attestation_service.attestation_token_broker.signer]
key_path = "./work/token.key"
cert_path = "./work/token-cert-chain.pem"

[attestation_service.rvps_config]
type = "BuiltIn"

[[plugins]]
name = "external"
# plugin_name is intentionally "resource" — the same name used by the built-in
# resource plugin — because kbs-client hardcodes /kbs/v0/resource/ in both
# set-resource (POST) and get-resource (GET via kbs:// URI scheme). Using this
# name lets the test exercise the full kbs-client roundtrip: admin-gated POST
# to store a secret, then attested GET with JWE decryption to retrieve it.
#
# This is not a conflict with the built-in resource plugin: both cannot be
# registered simultaneously (KBS enforces unique plugin names at startup), and
# this config intentionally omits the built-in plugin entry.
plugin_name = "resource"
endpoint = "http://127.0.0.1:50053"
tls_mode = "insecure"
timeout_ms = 5000
```

kbs/test/config/external-plugin-tls.toml
```diff
@@ -0,0 +1,31 @@
...
+[policy_engine]
+policy_path = "../../sample_policies/allow_all.rego"
+
+[[plugins]]
+name = "external"
+plugin_name = "echo-test"
+endpoint = "https://127.0.0.1:50052"
+timeout_ms = 5000
+tls_mode = "tls"
+ca_cert_path = "./work/echo-plugin-ca.pem"
```
Xynnn007:
```toml
[http_server]
sockets = ["127.0.0.1:8086"]
insecure_http = true

[admin]
type = "InsecureAllowAll"

[attestation_service]
type = "coco_as_builtin"

[attestation_service.rvps_config]
type = "BuiltIn"

[[plugins]]
name = "external"
plugin_name = "echo-test"
endpoint = "https://127.0.0.1:50052"
timeout_ms = 5000
tls_mode = "tls"
ca_cert_path = "./work/echo-plugin-ca.pem"
```

kbs/test/config/external-plugin.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[policy_engine]
+policy_path = "../../sample_policies/allow_all.rego"
+
+[[plugins]]
+name = "external"
+plugin_name = "echo-test"
+endpoint = "http://127.0.0.1:50051"
+tls_mode = "insecure"
+timeout_ms = 5000
```
Xynnn007:
```toml
[http_server]
sockets = ["127.0.0.1:8085"]
insecure_http = true

[admin]
type = "InsecureAllowAll"

[attestation_service]
type = "coco_as_builtin"

[attestation_service.rvps_config]
type = "BuiltIn"

[[plugins]]
name = "external"
plugin_name = "echo-test"
endpoint = "http://127.0.0.1:50051"
tls_mode = "insecure"
timeout_ms = 5000
```

kbs/test/test_data/configs/external-plugin-test.toml
```diff
@@ -0,0 +1,40 @@
...
+name = "sample"
+item = "test-value"
+
+# External gRPC echo plugin
+[[plugins]]
+name = "external"
+plugin_name = "echo-plugin"
+endpoint = "http://127.0.0.1:50051"
+tls_mode = "insecure"
+timeout_ms = 5000
```
Xynnn007:
```toml
# KBS configuration for external plugin integration testing

[http_server]
sockets = ["127.0.0.1:8080"]
insecure_http = true

[admin]
type = "InsecureAllowAll"

# Built-in attestation service with local work directory
# Before running: mkdir -p /tmp/kbs-as-work
[attestation_service]
type = "coco_as_builtin"

[attestation_service.rvps_config]
type = "BuiltIn"

# Compiled-in sample plugin
[[plugins]]
name = "sample"
item = "test-value"

# External gRPC echo plugin
[[plugins]]
name = "external"
plugin_name = "echo-plugin"
endpoint = "http://127.0.0.1:50051"
tls_mode = "insecure"
timeout_ms = 5000
```

kbs/Cargo.toml
```diff
@@ -122,6 +123,11 @@ reference-value-provider-service.path = "../rvps"
 serial_test.workspace = true
 toml.workspace = true
 
+[[bin]]
+name = "external_plugin_test_server"
```
Xynnn007, line 125: This review comment is for commit b4310b5a36b3b00894280b547b55bfe2470ced8a. 
I am amazing the CI can pass because the config format has been changed while the PR does not. Please ensure the tests will run during the PR CI. Thanks!

.github/workflows/kbs-e2e-ext-plugin.yml
```diff
@@ -0,0 +1,81 @@
...
+      working-directory: kbs
+      run: make check TEST_FEATURES="external-plugin"
+
+    - name: Install e2e test dependencies
+      working-directory: kbs/test
+      run: make install-dev-dependencies
+
+    - name: Run external plugin e2e tests
+      working-directory: kbs/test
+      run: make e2e-ext-plugin
```
Xynnn007: For CI e2e: I'd like to combine this e2e test into the e2e workflow in a way, wdyt?

Xynnn007 (commented):
There are some comments. Overall I like the logic, and we might need to ensure the CI runs well and correctly covers what we want.

I will not block the current code from getting merged, however I'd like to help to do some refactor upon the architecture. Does it sound good to you?

kbs/docs/PLUGIN_GUIDE.md
```diff
@@ -0,0 +1,721 @@
...
+KBS (Key Broker Service) external plugins are standalone gRPC services that
+extend KBS with custom endpoints. When a client sends a request to
+`/kbs/v0/<plugin-name>/...`, KBS forwards it to the corresponding external
+plugin over gRPC, and returns the plugin's response to the client.
+
+This architecture lets you add new capabilities -- secret injection, policy
+evaluation, certificate issuance, custom key management -- without modifying
+the KBS codebase. Each plugin runs as an independent process or container,
+communicating with KBS via the `kbs.plugin.v1.KbsPlugin` gRPC service
+defined in [`protos/plugin.proto`](../../protos/plugin.proto).
```
mythi: Do we really need a custom protocol for this? It becomes another contract we will have to deal with. Instead, could we say external plugins must be KMIP servers and we can have our external plugins support to be a KMIP client?

Some level of KMIP support is on #836 as well.

bpradipt: @mythi as per my understanding KMIP and this protocol serve different layers.  KMIP for the resource/key operations whereas this plugin protocol extends the KBS HTTP routing layer. KBS receives an HTTP request at /kbs/v0/<plugin-name>/... and proxies it to the plugin. 
Did I get it wrong ?

kbs/Cargo.toml
bpradipt: The new GHA will not work from the fork so it's understandable. But strangely the local make test worked. So I need to figure out how did it work with the older config

kbs/docs/PLUGIN_GUIDE.md
mythi: Right, so KMIP would be a plugin similar to vault or PCKS11 and this is a plugin for plugins that could, e.g., offload functionality to external services. 

Two questions I have in mind are:
- do we want to extend the api to arbitrary functions we cannot control/define and that go beyond "secret delivery through attestation"
- what "admin" (i.e., no attestation) functions external plugins can do

bpradipt: > do we want to extend the api to arbitrary functions we cannot control/define and that go beyond "secret delivery through attestation"

Yes, the design intentionally supports more than secret delivery. The idea is to have well defined plugin APIs and leave the featureset to the plugin implementors to decide what features they want to provide. It can be a secrets (eg. KMIP based), certificate manager, license management, extended policy check (eg. using geolocation) etc.

> what "admin" (i.e., no attestation) functions external plugins can do

Admin-gated plugin requests go through the same path as all KBS admin operations (setting policies, registering reference values, etc.). The plugin decides which endpoints are admin-gated and what operations they want to perform.

kbs/Cargo.toml
bpradipt: The old fields were silently ignored. KBS started with defaults for those fields.

kbs/src/plugins/implementations/external_plugin.rs
bpradipt: I will remove the health check for now.

.github/workflows/kbs-e2e-ext-plugin.yml
bpradipt: The CI will get triggered when key files are changed. So I think we can keep it separate. Will be easier to debug and improve.

kbs/docs/PLUGIN_GUIDE.md
mythi: > Yes, the design intentionally supports more than secret delivery. The idea is to have well defined plugin APIs and leave the featureset to the plugin implementors to decide what features they want to provide. It can be a secrets (eg. KMIP based), certificate manager, license management, extended policy check (eg. using geolocation) etc.

Is Trustee responsible for the API compatibility of those extensions?

> Admin-gated plugin requests go through the same path as all KBS admin operations (setting policies, registering reference values, etc.). The plugin decides which endpoints are admin-gated and what operations they want to perform.

The question was more about what external admin functions you have in mind.

bpradipt: The functions that were in my mind were same as internal plugins. IoW I'm seeing the external plugin in same way as internal plugin. Just that it stays outside of this tree.

kbs/src/plugins/implementations/external_plugin.rs
bpradipt: Yes, this actually makes lots of sense. Should I do this in this PR or can handle this as a follow-up post creating a tracker issue?

Xynnn007: no need to do in this PR.  can be done later

.github/workflows/kbs-e2e-ext-plugin.yml
bpradipt: @Xynnn007 is this ok for now?

Xynnn007: yes. and for code organization, I want to reorg this into current e2e workflow, which will be triggered on all PR. wdyt?

kbs/docs/PLUGIN_GUIDE.md
```diff
@@ -0,0 +1,685 @@
...
+
+### Who Is This For?
+
+Developers who want to extend KBS with custom endpoints. You should be
+comfortable with Rust and have a basic understanding of gRPC concepts
+(services, messages, metadata).
+
+### Prerequisites
+
+- Rust 1.70+
```
Xynnn007: Is the Rust version correct? Probably we do not need this `### Prerequisites` section?

Xynnn007: Is the Rust version correct? Probably we do not need this `### Prerequisites` section?

kbs/docs/PLUGIN_GUIDE.md
```diff
@@ -0,0 +1,685 @@
...
+
+### Accessing Request Context via gRPC Metadata
+
+KBS forwards session and attestation context to plugins via gRPC metadata
+headers on every `Handle` RPC. Plugins can read these to make authorization
+decisions:
+
+| Header | Type | Description |
+|--------|------|-------------|
+| `kbs-session-id` | string | Client session identifier (cookie-based auth only) |
```
Xynnn007: Do we need to forward the `kbs-session-id`? As it's some randomly generated bytes and representation inside KBS to distinguish different request clients, and almost cannot be known by any external services

kbs/docs/PLUGIN_GUIDE.md
```diff
@@ -0,0 +1,685 @@
...
+| `kbs-tee-type` | string | TEE type from attestation (e.g., `"SevSnp"`, `"Tdx"`); cookie-based auth only |
```
Xynnn007:
```suggestion
| `kbs-tee-type` | string | TEE type from attestation (e.g., `"snp"`, `"tdx"`); cookie-based auth only |
```

kbs/docs/PLUGIN_GUIDE.md
```diff
@@ -0,0 +1,685 @@
...
+        -f path/to/Dockerfile .
+  - name: Push to registry
+    run: |
+      docker push my-plugin:${{ github.sha }}
+```
+
+Adapt the build context and Dockerfile path for your SDK and project
+structure.
+
+---
```
Xynnn007: We might not need these secitons?

kbs/docs/PLUGIN_GUIDE.md
```diff
@@ -0,0 +1,685 @@
...
+
+In the KBS config, point to the Kubernetes Service:
+
+```toml
+[[plugins]]
+name = "external"
+plugin_name = "my-plugin"
+endpoint = "http://my-plugin.default.svc.cluster.local:50051"
+tls_mode = "insecure"  # or use mTLS with cert-manager
+```
```
Xynnn007: We might not need these detailed sections?

kbs/docs/PLUGIN_GUIDE.md
```diff
@@ -0,0 +1,685 @@
...
+
+```
+kbs_plugin_requests_total{plugin_name="my-plugin"} 142
+kbs_plugin_request_duration_seconds_bucket{plugin_name="my-plugin",le="0.01"} 130
+kbs_plugin_request_duration_seconds_bucket{plugin_name="my-plugin",le="0.1"} 140
+kbs_plugin_errors_total{plugin_name="my-plugin"} 2
+```
+
+Use these metrics in Grafana dashboards or Prometheus alerting rules to
+monitor plugin health and performance.
```
Xynnn007: Maybe we can have all KBS metrics in a separate document, than ad-hoc? We can do that in another PR, and do not need in this PR.

kbs/docs/PLUGIN_GUIDE.md
```diff
@@ -0,0 +1,685 @@
...
+the `KbsPlugin` service with four RPCs:
+
+```protobuf
+service KbsPlugin {
+    rpc Handle(PluginRequest) returns (PluginResponse) {};
+    rpc GetCapabilities(GetCapabilitiesRequest) returns (GetCapabilitiesResponse) {};
+    rpc ValidateAuth(ValidateAuthRequest) returns (ValidateAuthResponse) {};
+    rpc NeedsEncryption(NeedsEncryptionRequest) returns (NeedsEncryptionResponse) {};
+}
+```
```
Xynnn007: We might not need this

kbs/docs/PLUGIN_GUIDE.md
```diff
@@ -0,0 +1,685 @@
...
+
+**TlsMode values:**
+
+| Value | KBS Side | Plugin Side | Certificate Requirements |
+|-------|----------|-------------|--------------------------|
+| `insecure` | No TLS | No TLS | None |
+| `tls` | Verifies server cert | Presents server cert | `ca_cert_path` on KBS; server cert/key on plugin |
+| `mtls` | Verifies server cert + presents client cert | Verifies client cert + presents server cert | All cert paths on both sides |
+
+---
```
Xynnn007: Let's update these parts in `kbs/docs/config.md`

kbs/src/plugins/plugin_manager.rs
```diff
@@ -17,7 +28,131 @@ use super::{NebulaCaPlugin, NebulaCaPluginConfig};
...
-type ClientPluginInstance = Arc<dyn ClientPlugin>;
+/// Request context forwarded to plugins via gRPC metadata.
+///
+/// Provides session, TEE, and attestation information so that plugins
+/// can make authorization decisions and produce audit logs without
+/// receiving raw tokens (trust boundary preserved by omission).
+#[derive(Clone, Debug, Default)]
+pub struct PluginContext {
+    /// Session ID from KBS session cookie (if available)
+    pub session_id: Option<String>,
```
Xynnn007: The same question for session_id

protos/buf.yaml
```diff
@@ -0,0 +1,46 @@
+version: v2
```
Xynnn007: why do we need this buf.yaml?

kbs/src/plugins/implementations/external_plugin.rs
```diff
@@ -0,0 +1,708 @@
...
+//! External gRPC plugin — configuration and proxy implementation.
+//!
+//! This module contains the configuration types ([`TlsMode`], [`ExternalPluginConfig`])
+//! and the gRPC forwarding proxy ([`GrpcPluginProxy`]) for external KBS plugins.
+
+use actix_web::http::Method;
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+use backoff::{future::retry, Error as BackoffError, ExponentialBackoff};
+use mobc::{Manager, Pool};
```
Xynnn007: A question is whether we need gRPC pool for now? I lean a little to keep the foundation abilities in this PR - try as straightforward

kbs/docs/PLUGIN_GUIDE.md
bpradipt: Yeah right. Not really needed.

kbs/docs/PLUGIN_GUIDE.md
bpradipt: ok. I'll remove it

kbs/src/plugins/implementations/external_plugin.rs
bpradipt: I'm ok to remove it for now.

kbs/docs/PLUGIN_GUIDE.md
bpradipt: I'll keep some of them. Also rename this file to ext_plugin.md to be specific.

protos/buf.yaml
bpradipt: It's needed by the proto-breaking.yml CI workflow. I have added a comment to make it clear

kbs/docs/PLUGIN_GUIDE.md
bpradipt: This section is not needed. I'll remove it

bpradipt: Removed

kbs/src/plugins/plugin_manager.rs
bpradipt: I'll remove it.

.github/workflows/kbs-e2e-ext-plugin.yml
bpradipt: @Xynnn007 I think I misunderstood your comment.
Do you want me to add the e2e-ext-plugin target to test-e2e-kbs.yml and remove this kbs-e2e-ext-plugin.yml ?

Xynnn007: yes!

bpradipt: Got it. Thanks. Will re-push with the changes

Xynnn007 (approved):
Thanks for this @bpradipt LGTM!

kbs/src/plugins/implementations/external_plugin.rs
```diff
@@ -0,0 +1,665 @@
...
+/// The `api_server.rs` match arms guard specific (name, method) pairs, but
+/// other methods or unguarded combinations fall through to plugin lookup.
+/// Rejecting these names at startup prevents silent misbehaviour.
+pub(crate) const RESERVED_PLUGIN_NAMES: &[&str] = &[
+    "auth",
+    "attest",
+    "attestation-policy",
+    "reference-value",
+    "resource-policy",
+];
```
fitzthum: Don't really like this hard-coded approach. Can't we prefix all external plugins with `/external` or something to avoid collisions.

kbs/src/plugins/implementations/external_plugin.rs
```diff
@@ -0,0 +1,665 @@
...
+
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum TlsMode {
+    /// Mutual TLS: both client and server authenticate with certificates
+    Mtls,
```
fitzthum: Do we have a requirement for mTLS? Seems like we could leave this out in the first iteration.

kbs/src/plugins/plugin_manager.rs
```diff
@@ -17,7 +28,129 @@ use super::{NebulaCaPlugin, NebulaCaPluginConfig};
 #[cfg(feature = "pkcs11")]
 use super::{Pkcs11Backend, Pkcs11Config};
 
-type ClientPluginInstance = Arc<dyn ClientPlugin>;
+/// Request context forwarded to plugins via gRPC metadata.
+///
+/// Provides TEE and attestation information so that plugins can make
+/// authorization decisions and produce audit logs without receiving raw
+/// tokens (trust boundary preserved by omission).
```
fitzthum: This comment doesn't make much sense to me. What does "trust boundary preserved by omission" mean?

More generally, I'm confused by this approach. Why does the plugin need access to the Tee? What does that have to do with the remote plugins you are adding here? As Ding mentioned, we are thinking about how to expose more information to plugins (potentially the entire token). Just adding the Tee type in this PR seems kind of random.

Furthermore, why do we need the `is_attested` field? With existing plugins, the plugin specifies whether an endpoint needs to be attested/authorized. If attestation is required, the KBS checks this before calling the endpoint. So by the time a plugin is executing one of these endpoints, it can assume that attestation has happened. No need for it to check the attestation status itself. Why are things different here?

kbs/src/plugins/plugin_manager.rs
```diff
@@ -17,7 +28,129 @@ use super::{NebulaCaPlugin, NebulaCaPluginConfig};
...
+/// the original `handle()` signature (no context, returns `Vec<u8>`).
+///
+/// External gRPC plugins are stored directly as `Arc<GrpcPluginProxy>` and
+/// dispatched via `handle_with_context()`, which injects session/TEE metadata
+/// into gRPC request headers.
+#[derive(Clone)]
+pub(crate) enum PluginInstance {
+    BuiltIn(Arc<dyn ClientPlugin>),
+    #[cfg(feature = "external-plugin")]
+    External(Arc<GrpcPluginProxy>),
```
fitzthum: I guess this works, but I would prefer if we created a new plugin for serving external requests. Rather than reclassify all our existing plugins as external or internal, just create a new plugin called `external` and have it implement its own self-contained logic for proxying the requests. Wouldn't that be simpler?

protos/buf.yaml
fitzthum, line 1: This seems unrelated to the PR. Maybe we should handle protobuf hygiene separately.

protos/plugin.proto
```diff
@@ -0,0 +1,144 @@
...
+    // and returns a response with optional status code and content type hints.
+    // KBS may override these hints based on its own policies.
+    rpc Handle(PluginRequest) returns (PluginResponse) {};
+
+    // GetCapabilities returns metadata about the plugin's supported features.
+    //
+    // This is called once at plugin startup to register the plugin with KBS.
+    // The response includes the plugin's name, version, supported HTTP methods,
+    // and optional key-value attributes for operational visibility.
+    rpc GetCapabilities(GetCapabilitiesRequest) returns (GetCapabilitiesResponse) {};
```
fitzthum: Do we really need this? Can't we just set the name and uri of a plugin in the config file and then forward requests to it when they come in at runtime?

protos/plugin.proto
```diff
@@ -0,0 +1,144 @@
...
+    // true  = KBS must verify admin credentials before calling Handle.
+    // false = KBS requires a valid attestation token and policy evaluation.
+    bool requires_admin_auth = 1;
+}
+
+// NeedsEncryptionResponse is returned by the NeedsEncryption RPC.
+message NeedsEncryptionResponse {
+    // true  = KBS JWE-encrypts the Handle response with the TEE public key.
+    // false = KBS returns the Handle response body as-is.
+    bool encrypt = 1;
```
fitzthum: nit: `encrypted` doesn't really match `requires_admin_auth` above. maybe should be `requires_payload_encryption` or something like that.

kbs/docs/ext_plugin.md
```diff
@@ -0,0 +1,252 @@
...
+           │                             │
+           ▼                             ▼
+    plugin.Handle()               plugin.Handle()
+           │                             │
+           ▼                             ▼
+    Return response           plugin.NeedsEncryption()
+                                         │
+                              ├── true → JWE response
+                              └── false → raw response
+```
```
fitzthum, line 39: mermaid?

fitzthum (commented):
Still seems more complicated than it needs to be tbh. See comments.

kbs/src/plugins/plugin_manager.rs
bpradipt: @fitzthum nice suggestion. This might be even simpler to maintain and extend.
I'll look into it.

protos/buf.yaml
bpradipt: ok. I included it primarily because of the external plugin. Will handle it in subsequent PR.

bpradipt:
@fitzthum @Xynnn007 I pushed an update which is a redesign of the implementation.
@fitzthum this follows your suggestion (which was excellent imho). I add an in-tree external plugin stub/gateway and that abstracts everything that is needed to support this functionality. 
Let me know your thoughts. Happy to make additional changes based on your feedback.

kbs/docs/ext_plugin.md
Xynnn007: looks like this is not changed @bpradipt

bpradipt: I made it as a plain asciichart. Do we want mermaid specifically ?

Xynnn007: It's preferred. Usually mermaid can be easily maintained via "coding"

bpradipt: got it.. I'll update it

bpradipt:
Looks like the CI failures are unrelated

kbs/Cargo.toml
```diff
@@ -48,6 +51,7 @@ aes-gcm = "0.10.1"
 aes-kw = "0.2.1"
 anyhow.workspace = true
 async-trait.workspace = true
+backoff = { version = "0.4", features = ["tokio"], optional = true }
```
Xynnn007: Looks like [backoff](https://github.com/ihrwein/backoff) is not actively maintained for ~ 4years. Do you have better candidates?

bpradipt: I think this is well maintained - https://github.com/Xuanwo/backon https://docs.rs/backon/latest/backon/
I'll adapt the code.

bpradipt:
@Xynnn007 @fitzthum the CI is green now..

mythi:
I had one question earlier that is not yet discussed: the readme should probably say these plugins are not part of Trustee core API

bpradipt:
> I had one question earlier that is not yet discussed: the readme should probably say these plugins are not part of Trustee core API

One thing to note is that the design of the ext plugin changed. The external plugin is a stub/gateway in-tree plugin and is bound by the same core KBS API as any other in-tree plugin.  

Does adding something like this helps to address your concern ?
"The KBS external plugins are out-of-tree plugins and not part of core Trustee"

kbs/docs/ext_plugin.md
```diff
@@ -0,0 +1,256 @@
...
+
+- **In-tree stub plugin** — `external_plugin` is an in-tree KBS plugin (enabled
+  with `--features external-plugin`) and acts a routing stub only. It handles
+  auth, attestation, and JWE, then forwards requests over gRPC to whichever
+  backend the URL names.
+
+- **Out-of-tree plugins** — the actual plugin implementations are separate
+  processes or containers that live **outside the Trustee repository**. You
+  write and maintain them; KBS talks to them over the `KbsPlugin` gRPC
+  interface. Nothing in core Trustee knows or cares about their internals.
```
bpradipt, line 30: @mythi does the above text addresses your concern?

mythi: my concern is more than just "internals". Trustee brokers the REST requests to the configured plugin but each plugin is responsible for their API versioning/backwards compatibility. Bugs should also be reported to the respective plugin repositories.

The gRPC interface is more difficult as this is what Trustee defines. Are we ready to say it's "generally available" or keep external plugin support with some experimental status still.

bpradipt: @mythi it's good to see that this plugin has initiated a thinking on plugin status. 

Should defining the status of a Trustee plugin be a separate dedicated issue ?
Aspects like test coverage, CI, maintainers etc should be included to decide on the plugin status imho. 

Regarding your comment on "internals", doesn't this statement "You  write and maintain them. Nothing in core Trustee knows or cares about their internals" implies the intent?
Or better you please share me a suggested text that you are comfortable with so that we can move ahead with this PR.

mythi: I was not clear but the suggested addition was already there:

> Nothing in core Trustee knows or cares about their internals. Trustee brokers the REST requests to the configured plugin but each plugin is responsible for their API versioning/backwards compatibility. Bugs should also be reported to the respective plugin repositories.

bpradipt: @mythi I force-pushed with your suggested addition. Thanks.

kbs/docs/config.md
```diff
@@ -333,6 +333,33 @@ with the `[plugins.self_signed_ca]` properties.
...
+```
+
+Each backend is reachable at `/kbs/v0/external/<name>/...`.
+
+**Per-backend fields:**
+
+| Field | Type | Required | Default | Description |
+|---|---|---|---|---|
+| `name` | string | Yes | — | Sub-plugin name used in URL routing |
+| `endpoint` | string | Yes | — | gRPC endpoint (`http://` for insecure, `https://` for TLS) |
```
mythi, line 356: did you consider local sockets `unix://`?

bpradipt: No. Do you see use cases for local ?

bpradipt: The use cases which triggered the need for ext plugin had the actual plugins running (and operated) remotely in different env. So adding local socket support felt unnecessary addition.
But if you see use cases, we can consider it.

mythi: > No. Do you see use cases for local ?

Nothing concrete. Maybe some local attester plugin for admins could help with the Trustee TEE provisioning challenge.

bpradipt: Hmm. Let's think about this.

PR #1189, opened 2026-02-25
RVPS RFC: add Reference Value URI spec

### LINKED ISSUE #1159: [RFC] Reference Value ID Proposal
## Background

Now have the function `query_reference_value` in AS policy, which provides a strong ability to get the reference values during the policy enforcement.

The input to this function is a general reference value id. With more and more work about reference value happens, for example the reference value publishment for CoCo Release in Kata build pipeline, we need a mechanism to well organize and indexing different reference values. This RFC aims to propose a URI format to identity the different reference values.

## Needs

To sum up, there are some related work and needs upon this topic. At least including
1. to efficiently index different reference values
2. to have "namespace/tag" ability to index different reference values for different usages/domains
3. potentiallty, support the reference value pull model. That means this URI tells the RVPS where and how to get  the real reference value
4. This ID works between AS and RVPS, and can be referred in an Attestation Service policy
5. This ID can also be defined by a reference value publisher. For example, CoCo/Kata could maintain a reference value registry with different reference value ids and their values according to different release versions.

## Design

We provide the following reference value URI format:

```
rvps://[method]/<namespace-1>/<namespace-2>/.../<namespace-n>[:tag]
```

Concretely,
1. `<namespace-x>`s can be any number and any level, used by the definer to distinguish different namespaces.
2. A `tag` with a prefix `:` is optional. Usually for users to make different versions for a same logical reference value

## More Discussion

We did not discuss the `[method]` field, because I did not get a complete and whole story about this. The basic idea is that `method` is a hint about how the reference value can be got. If we do not set this field, the URI looks like `rvps:///a/b/c` will treat the backend of RVPS as a simple Key-Value storage and `rvps:///a/b/c` will be the key to index the reference value. 

If we set this field, the `method` field will be translated into a plugin name and the other parts can be given to the concrete plugin as parameters to get the reference value. For example, abstractly we can define `rvps://link/confidentialcontainers.org/v0.18.0` that means the reference value will be get from the URL `confidentialcontainers.org/v0.18.0` by HTTP GET or so. How to interpret  `confidentialcontainers.org/v0.18.0` should be defined in `link` plugin.

Overall, hopefully we could reach an agreement on the parts except `[method]` part, and discuss more about the pull model, how it interacts with RV ID and how it works in future.

Xynnn007:
Close #1159

### DIFF
--- a/rvps/docs/reference-value-uri.md
+++ b/rvps/docs/reference-value-uri.md
@@ -1,0 +1,159 @@
+# Reference Value URI
+
+## Background
+
+[Attestation Service](../../attestation-service/) policy currently provides `query_reference_value` extension, which allows policy enforcement to retrieve reference values (RVs) from RVPS during evaluation. As RV usage grows (e.g., publishing RVs for CoCo/Kata releases in build pipelines), we need a recommended and interoperable way to organize and index RVs so that they are easy to manage across different projects/domains and versions.
+
+At the same time, today’s RVPS storage interface is essentially a key-value model where the key can be any arbitrary string. This means the system is already functional without a standardized key format. Therefore, this RFC defines a convention/spec **recommendation** for RV identifiers, primarily to improve consistency and operability, while keeping backward compatibility with arbitrary keys.
+
+> [!NOTE]  
+> During [discussion](https://github.com/confidential-containers/trustee/issues/1159), a future direction is identified: the "pull model" (RVPS fetching RVs from external sources), which is the main driver for URIs that can locate external resources. However, because “external location URI design” is tightly coupled with pull-model design, we explicitly scope this RFC to a minimal format first, and leave the authority/external-pointer story for a follow-up proposal.
+
+## Specifications
+
+### Scope and Non-Goals
+
+#### In scope
+
+1. Define a recommended RV identifier format to be consumed by `query_reference_value`.
+2. Provide namespace/path + optional tag semantics suitable for organizing RVs and versions.
+3. Define expected behavior around missing tags and latest-like usage (as recommendations).
+
+#### Out of scope (future work)
+1. Encoding RVPS address / remote location / plugin mechanism into the URI (pull model).
+2. Standard push/update APIs
+
+### Semantics: "Instance-local" Identifier, "Globally meaningful" Convention
+
+An RV URI is resolved within a specific RVPS instance. Concretely:
+
+`query_reference_value(rv_uri)` is evaluated against the RVPS configured/used by the AS deployment.
+The same `rvps:///...` identifier may resolve to different contents on different RVPS instances, depending on what has been set there.
+
+Although resolution is instance-local today, it's recommended that reference value publishers adopt stable and unique URI naming conventions so that:
+
+1. Different components are clearly separated by namespace paths.
+2. Different versions are clearly separated by tags.
+3. The same logical RV can be referred to consistently across tooling, documentation, and policies—even if multiple RVPS instances mirror the same naming scheme.
+
+### RV Identifier Format (Recommended)
+
+#### Canonical form
+```
+rvps:/// <namespace-1> / <namespace-2> / ... / <namespace-n> [ : <tag> ]
+```
+
+Here,
+1. The prefix MUST be `rvps:///` (scheme + empty authority).
+2. There **MUST** be at least one namespace segment.
+3. `:tag` is **OPTIONAL**.
+
+Examples:
+
+- `rvps:///confidential-containers/kata/measurements:v0.18.0`
+- `rvps:///org/project/component`
+- `rvps:///coco/kata/tdx/rtmr0:latest`
+
+#### Namespaces (`/<namespace-x>/...`)
+
+1. Namespace segments are hierarchical and purely organizational.
+2. The spec does not mandate segment meanings; publishers/operators define their own conventions.
+3. Implementations **SHOULD** treat the entire /<...> part (plus tag if present) as the lookup key.
+
+> [!NOTE]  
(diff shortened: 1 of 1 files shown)


rvps/docs/reference-value-uri.md
```diff
@@ -0,0 +1,120 @@
...
+1. Different components are clearly separated by namespace paths.
+2. Different versions are clearly separated by tags.
+3. The same logical RV can be referred to consistently across tooling, documentation, and policies—even if multiple RVPS instances mirror the same naming scheme.
+
+### RV Identifier Format (Recommended)
+
+#### Canonical form
+```
+rvps:/// <namespace-1> / <namespace-2> / ... / <namespace-n> [ : <tag> ]
+```
```
fitzthum, line 44: resource URI only supports 3 parts. Do we want these to be consistent? the 3 parts of the resource URI are somewhat arbitrary.

rvps/docs/reference-value-uri.md
```diff
@@ -0,0 +1,120 @@
...
+
+1. This RFC does not require all reference value keys to follow this format.
+2. `query_reference_value` and RVPS MUST continue to work with arbitrary string keys (legacy behavior).
+3. The `rvps:///...` format is a recommended convention to improve consistency and enable future extensions.
+
+### CoCo Naming Convention
+
+For the Confidential Containers (CoCo) ecosystem, we define a recommended common prefix to standardize RV naming across components and releases:
+
+1. Common prefix: `rvps:///github.com/confidential-containers/`
```
fitzthum, line 102: Not sure if we want to have `github.com` in the example here. That might be getting to the questions about the pull model.

rvps/docs/reference-value-uri.md
Xynnn007: They can be if needed. 3 parts is somehow an accidental standard

rvps/docs/reference-value-uri.md
Xynnn007: I suggest to have a URI like github.com or confidentialcontainers.org or something. It's clear where the reference value comes from

rvps/docs/reference-value-uri.md
fitzthum: Yeah I think maybe we should adopt RVPS URI with N parts and then update the resource URI to N parts later.

rvps/docs/reference-value-uri.md
fitzthum: Ok, although I wonder if that will be confusing in the future if we have a domain in the first part of that URI as well. i.e. `rvps:://rvps.something/github.com/a/b/c`. I guess we can worry about that later.

rvps/docs/reference-value-uri.md
```diff
@@ -0,0 +1,120 @@
...
+
+#### Canonical form
+```
+rvps:/// <namespace-1> / <namespace-2> / ... / <namespace-n> [ : <tag> ]
+```
+
+Here,
+1. The prefix MUST be `rvps:///` (scheme + empty authority).
+2. There **MUST** be at least one namespace segment.
+3. `:tag` is **OPTIONAL**.
```
bpradipt, line 49: nit: It's better to mention the default tag is `latest` when omitted to avoid any ambiguity?

rvps/docs/reference-value-uri.md
```diff
@@ -0,0 +1,120 @@
...
+2. Standard push/update APIs
+
+### Semantics: "Instance-local" Identifier, "Globally meaningful" Convention
+
+An RV URI is resolved within a specific RVPS instance. Concretely:
+
+`query_reference_value(rv_uri)` is evaluated against the RVPS configured/used by the AS deployment.
+The same `rvps:///...` identifier may resolve to different contents on different RVPS instances, depending on what has been set there.
+
+Although resolution is instance-local today, it's recommend that reference value publishers adopt stable and unique URI naming conventions so that:
```
bpradipt: nit: s/recommend/recommended/

rvps/docs/reference-value-uri.md
bpradipt: It's better to add a brief note on the difference between KBS resource URI (3 parts) and RVPS here.

rvps/docs/reference-value-uri.md
```diff
@@ -0,0 +1,120 @@
...
+This convention aims to ensure that different components and different releases are naturally separated by path and tag, and makes it easier to build shared tooling (publishers, CI pipelines, mirroring RVPS instances) around a consistent identifier scheme.
+
+### Future Extension Point (Pull Model)
+
+The part between `://` and the first `/` (i.e., URI authority / location / method) is intentionally left unused in this RFC.
+A follow-up proposal **MAY** define how to reference external RV sources as part of a pull model, potentially introducing:
+1. authority-based RVPS selection, and/or
+2. source-locator encoding, and/or
+3. structured metadata mapping from RV IDs to pull sources.
+Any future design MUST remain compatible with identifiers of the form `rvps:///path[:tag]`.
```
bpradipt, line 158: Any thoughts about adding an example of how this uri format will be used in the policy file?

rvps/docs/reference-value-uri.md
bpradipt: Also a brief explanation on the reasoning behind empty authority might be helpful.

rvps/docs/reference-value-uri.md
fitzthum: We should be careful about the scope here. Currently our reference value names are very generic (e.g. snp-launch-measurement). We should think about whether we want the default policy to remain generic or if we want it to be more clearly scoped to coco and reference coco RVs.

rvps/docs/reference-value-uri.md
fitzthum: At the moment the uri is just a string, so I don't think we have any default behavior. No tag and latest are not equivalent unless we introduce some logic for that.

bpradipt: We should document this to avoid ambiguity

rvps/docs/reference-value-uri.md
Xynnn007: makes sense.

rvps/docs/reference-value-uri.md
Xynnn007: Added

bpradipt (approved):
/lgtm
Thanks @Xynnn007

Issue #1132, opened 2025-12-26
[KBS] Abstraction layer for different Attestation Services Backends


Xynnn007:
### Background

Currently, we allow KBS to interact with multiple ASs by implementing [the trait](https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/backend.rs#L90-L125) (not required all, but at least [verify](https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/backend.rs#L98C14-L98C20)). To manage API access based on the attestation, KBS currently uses a unified resource policy.

### What We are Facing?

- We assume that all different ASes are returning a JWT (namely, attestation token), including CoCoAS, ITA and Keylime(WIP). Potentially more AS types.
- The input to the resource policy is directly the claims field of the JWT.
- Different attestation token types will have different fields in the claims.
- The example KBS resource applies only for CoCoAS' EAR Token.
- So, if any developer wants to use KBS to connect to AS services other than CoCoAS, they will need to customize the KBS policy based on the concrete token format they choose to receive from the AS. In other words, currently KBS resource policy is **tightly coupled** with the concrete backend AS.

This issue raises this point primarily to encourage everyone to assess whether it's a problem.

IMPO, since KBS and AS are separate components, it must provide a common abstraction for the returned tokens across different ASs.

### How we can deal with it?

This means abstracting a semantic set for AS tokens in different contexts (whether it's CoCoAS's EAR or ITA's JWT), which should at least include the following:

- attestation_summary: A minimal, authorization-oriented view derived from attestation results, such as the coarse-grained trust outcomes. (we can discuss about the concrete in a separate thread if needed)
- an evidence-bound tee pubkey: This is the core value that KBS with its RCAR handshake provides - a TEE key pair that is bound to the concrete client platform.
- custom_claims: any AS-specific extensions (e.g. extensions including validated identifiers in EAR)

Let's call it **Trust Context** for better wording. The Trust Context should be represented **ONLY INSIDE** KBS. The returned thing from KBS to the Client after RCAR is still the token generated by the backend AS.

#### TC Usages

TC can be used in two cases
- To call the concrete plugin of KBS, TC should be added to the [function](https://github.com/confidential-containers/trustee/blob/main/kbs/src/plugins/plugin_manager.rs#L29).  Then the concrete plugin can do different operations based on the TC. e.g., hsm can only return the key when TC meets some needs.
- KBS resource policy.

#### How to generate TC?

[...]

PR #1190, opened 2026-02-25
Use Key-Value Storage and Policy Engine for RVPS/AS/KBS Configuration

Xynnn007:
This is following up for https://github.com/confidential-containers/trustee/pull/1156. This patch uses a unified storage config item for each components launch config.

Related to https://github.com/confidential-containers/trustee/pull/1156

### DIFF
--- a/rvps/src/extractors/swid/mod.rs
+++ b/rvps/src/extractors/swid/mod.rs
@@ -13,7 +13,7 @@ use crate::ReferenceValue;
 use super::Extractor;
 
 #[derive(Deserialize, Clone, Debug, PartialEq, Default)]
-pub struct SwidExtractorConfig;
+pub struct SwidExtractorConfig {}
 
 #[derive(Default)]
 pub struct SwidExtractor;


rvps/src/lib.rs
```diff
@@ -9,20 +9,24 @@ pub mod extractors;
...
 
 pub use serde_json::Value;
 
 use anyhow::{bail, Context, Result};
-use log::info;
 use serde::{Deserialize, Serialize};
 
+use crate::extractors::ExtractorsConfig;
+
+pub const REFERENCE_VALUE_STORAGE_INSTANCE: &str = "reference-value";
```
fitzthum: `instance` -> `namespace`.

rvps/src/reference_value.rs
```diff
@@ -120,6 +120,11 @@ impl ReferenceValue {
     pub fn name(&self) -> &String {
         &self.name
     }
+
+    pub fn to_bytes(&self) -> Result<Vec<u8>> {
+        let value_bytes = serde_json::to_vec(&self).context("serialize reference value")?;
+        Ok(value_bytes)
+    }
```
fitzthum, line 127: Might be nicer to implement `Into<Vec<u8>>` here instead, but not a big deal.

rvps/README.md
```diff
@@ -92,17 +92,142 @@ podman run -d -p 50003:50003 --net host rvps
...
+    "storage": {
+        "storage_type": "LocalFs",
+        "backends": {
+            "local_fs": {
+                "dir_path": "/opt/confidential-containers/storage/local_fs"
+            }
+        }
+    },
+    "extractors": {
+        "swid_extractor": {}
```
fitzthum, line 164: No point in having this empty config here.

rvps/README.md
```diff
@@ -92,17 +92,142 @@ podman run -d -p 50003:50003 --net host rvps
...
+
+1. **Sample Extractor** (`sample`): A simple extractor for testing and demos. It directly extracts reference values from input without verifying signatures. See [Sample Extractor README](src/extractors/sample/README.md) for details.
+
+2. **SWID/RIM Extractor** (`swid`): Extracts reference values from SWID tags with RIM (Reference Integrity Manifest) bindings. Currently tested with NVIDIA RIM manifests for GPU vbios. See [SWID RIM Extractor README](src/extractors/swid/README.md) for details.
+   - Currently, `swid_extractor` has no configuration options (empty object `{}`)
+   - The extractor does not verify manifest signatures; authenticity should be verified before providing to RVPS
+   - Default expiration time: 12 months
+   - Default hash algorithm: `sha384`
+
+3. **In-toto Extractor** (`in-toto`): Verifies and extracts reference values from in-toto provenance metadata. Requires the `in-toto` feature flag. See [In-toto Extractor README](src/extractors/in_toto/README.md) for details.
```
fitzthum, line 149: As a side note, we should add a corim extractor sometime and add signature support to the in-toto one.

attestation-service/docs/config.md
```diff
@@ -179,17 +179,43 @@ Available when the `tdx-verifier`, `sgx-verifier`, or `az-tdx-vtpm-verifier` fea
...
+            }
+        }
+    },
     "rvps_config": {
         "type": "BuiltIn",
-        "storage": {
-            "type": "LocalFs"
-            "file_path": "/var/lib/attestation-service/reference-values"
+        "extractors": {
+            "swid_extractor": {}
```
fitzthum, line 218: Again, not really any point in doing this.

attestation-service/docs/config.md
```diff
@@ -284,12 +302,58 @@ Configuration with AMD SEV-SNP verifier using KDS:
 }
 ```
 
+Running with unified storage backend (recommended):
```
fitzthum: It's slightly confusing for this to be the "recommended" option, given that the next example also uses the unified storage backend.

attestation-service/src/ear_token/broker.rs
```diff
@@ -98,6 +104,22 @@ impl EarAttestationTokenBroker {
     }
 }
 
+/// Extensions that will be added to the attestation token.
+#[derive(Debug, Deserialize)]
+pub struct Extension {
```
fitzthum: In hindsight, maybe we should call this `TokenExtension` so it's not confused with the opa extensions.

attestation-service/src/ear_token/broker.rs
```diff
@@ -132,25 +154,101 @@ impl EarAttestationTokenBroker {
...
-
-                appraisal
-                    .trust_vector
-                    .mut_by_name(k)
-                    .unwrap()
-                    .set(claim_value as i8);
+            match policy_results
+                .eval_rules_result
+                .get(TRUST_CLAIMS_RULE)
+                .context("Trust claims rule not found")?
```
fitzthum: Might be nice to say something like "AS policy <policy_id> must contain rule <trust_claims_rule>" with the actual rule, as long as that doesn't leak too much information.

attestation-service/src/ear_token/mod.rs
```diff
@@ -41,86 +32,45 @@ pub struct TokenSignerConfig {
...
-            duration_min: default_duration(),
-            issuer_name: default_issuer_name(),
-            developer_name: default_developer(),
-            build_name: default_build(),
-            profile_name: default_profile(),
+            duration_min: DEFAULT_TOKEN_DURATION,
+            issuer_name: COCO_AS_ISSUER_NAME.to_string(),
+            developer_name: DEFAULT_DEVELOPER_NAME.to_string(),
+            build_name: format!("{} {}", env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION")),
+            profile_name: DEFAULT_PROFILE.to_string(),
```
fitzthum, line 72: This doesn't seem equivalent to the default macros you remove above. Those would fill in the default if a particular field is missing, whereas this will just create a default struct if this function is called.

kbs/docs/config.md
```diff
@@ -402,11 +450,14 @@ type = "Simple"
 id = "admin"
 public_key_path = "/etc/kbs-admin.pub"
 
-[policy_engine]
-policy_path = "/etc/kbs-policy.rego"
+# Legacy per-component storage configuration (deprecated, use storage_backend instead)
```
fitzthum: Is deprecated the right word? We do still support this, right?

kbs/docs/config.md
```diff
@@ -229,15 +216,29 @@ The regular expression must start with `^/kbs` and end with `$` to avoid ambigui
 If no matching role is found (and at least one role is specified), the admin will not be able
 to access any endpoints.
 
-### Policy Engine Configuration
+### Storage Backend Configuration
+
+KBS supports a unified storage backend configuration that allows you to declare a single storage configuration that will be used for all storage needs in KBS, including
+all the persistent storages that KBS relies on.
```
fitzthum: If we have a built-in AS, will this config also apply to that?

fitzthum (commented):
Working my way through this one. Left some comments on the first half. Looks good in general.

rvps/src/reference_value.rs
Xynnn007: Yes. Into trait requires a function returning without a Result. Thus we use this. Btw, TryInto will consume the object (it uses `self` than `&self`), so it's not ideal too

rvps/README.md
Xynnn007: It only shows that the extractors can have a config. coincidencially that swid_extractor now has an empty config struct definition

rvps/README.md
Xynnn007: Agreed.

attestation-service/docs/config.md
Xynnn007: The same as before

attestation-service/docs/config.md
Xynnn007: Let me delete this note

attestation-service/src/ear_token/broker.rs
Xynnn007: done

attestation-service/src/ear_token/mod.rs
Xynnn007: Almost the same. They both handles "how to fill empty fields when get a JSON". The backend logic of the previous one is to fill each field with a function call, while of the new one is to create a default instance and fill with given fields. From blackbox view, they are the same. This change will have fewer code lines to achieve a same goal.

kbs/docs/config.md
Xynnn007: Oh let me change this commit. The final shape does not include this line

kbs/docs/config.md
Xynnn007: Yes. It's for built-in AS and builti-in RVPS

attestation-service/src/ear_token/mod.rs
fitzthum: But doesn't serde default apply to configs that are partly set i.e. if the admin sets all but one of the values in the config file, the empty one will be set with the default. The default function doesn't do that.

Xynnn007: The overall `#[default]` feature will set the parts that is not given with default structure value. Let me add a unit test for this.

attestation-service/src/bin/restful/mod.rs
bpradipt: nit: the docstring must be changed to reflect the changes. Only `policy-id: id` is returned

kbs/src/api_server.rs
```diff
@@ -311,12 +364,34 @@ pub(crate) async fn api(
...
+                        &claim_str,
+                        "resource-policy",
+                        vec!["data.policy.allow"],
+                        vec![],
+                    )
                     .await
                     .inspect_err(|_| KBS_POLICY_ERRORS.inc())?
+                    .eval_rules_result
+                    .get("data.policy.allow")
+                    .expect("`data.policy.allow` rule not put as parameter found")
```
bpradipt, line 387: If `data.policy.allow` is not found, don't we want to return the error ?

kbs/src/attestation/backend.rs
```diff
@@ -144,13 +145,17 @@ pub struct SetPolicyInput {
 }
 
 impl AttestationService {
-    pub async fn new(config: AttestationConfig) -> Result<Self> {
+    pub async fn new(
+        config: AttestationConfig,
+        _storage_backend_config: &StorageBackendConfig,
```
bpradipt, line 147: Why prefixed with `_` ?

kbs/src/api_server.rs
```diff
@@ -311,12 +364,34 @@ pub(crate) async fn api(
...
+                    .as_ref()
+                    .unwrap_or_else(|| {
+                        warn!("The KBS Resource Policy does not define the `data.policy.result` rule, use false as default");
+                        KBS_POLICY_ERRORS.inc();
```
bpradipt, line 391: should this be `data.policy.allow` ?

kbs/src/attestation/backend.rs
Xynnn007: It's only used under the feature flag `#[cfg(any(feature = "coco-as-builtin", feature = "coco-as-builtin-no-verifier"))]`. If we do not build/check/clippy with specified features(s), there will be a lint warning with unused variable.

kbs/src/api_server.rs
Xynnn007: Yes. Let me make a const.

tools/kbs-client/src/main.rs
```diff
@@ -304,7 +304,7 @@ async fn main() -> Result<()> {
                     } else if affirming {
                         include_bytes!("../../../kbs/sample_policies/affirming.rego").into()
                     } else if default {
-                        include_bytes!("../../../kbs/src/policy/default_policy.rego").into()
+                        include_bytes!("../../../kbs/sample_policies/deny_all.rego").into()
```
fitzthum, line 311: I think this change maybe should be discussed separately. Can you elaborate on why this needed here? I don't quite understand the commit message.

kbs/sample_policies/deny_all.rego
```diff
@@ -1,3 +1,78 @@
...
+#      - resource-path format: three items slice
+#      - query: {}
+#
+# Policy Rules
+# ------------
+# The policy evaluates access based on the above inputs. See the allow rules
+# below for specific conditions.
+# 
+# This default policy will only allow accesses to "resource" plugin if the client
+# is a TEE/TPM platform.
```
fitzthum, line 75: I don't think we need to add this big comment to every policy.

deps/key-value-storage/src/local_fs/mod.rs
```diff
@@ -27,6 +27,7 @@ pub struct LocalFs {
 
 #[derive(Deserialize, Clone, PartialEq, Debug)]
 #[serde(default)]
+#[serde(deny_unknown_fields)]
```
fitzthum: Not sure this is really needed in this PR. There are some tradeoffs to main the config more strict. Might be better to discuss separately.

rvps/src/bin/rvps.rs
```diff
@@ -30,8 +33,27 @@ pub struct Cli {
...
+    let cli = Cli::parse();
+
+    if let Some(command) = cli.command {
+        match command {
+            Commands::PrintExampleConfig => {
+                print!("{}", Config::example_config_toml());
+                return Ok(());
+            }
+        }
+    }
```
fitzthum: This also seems unrelated to this PR. I guess it's a somewhat handy, although very few people will run the RVPS from the CLI like this, but maybe make this separate.

attestation-service/src/config.rs
```diff
@@ -31,24 +32,56 @@ pub struct Config {
     pub storage_backend: StorageBackendConfig,
 }
 
+impl Config {
+    /// An example Attestation Service configuration in TOML format with per-field comments.
+    pub fn example_config_toml() -> &'static str {
```
fitzthum: I don't really like hard-coding an example configuration here. This will be annoying to maintain.

kbs/sample_policies/deny_all.rego
Xynnn007: Let me only include this in one sample policy. In future when we get better handling on https://github.com/confidential-containers/trustee/pull/1188 and https://github.com/confidential-containers/trustee/issues/1132, I will put a document for the policy writing.

tools/kbs-client/src/main.rs
Xynnn007: This is because of the moving of policy files. If we do not change the code, the building of trustee-cli will fail. Let me re-write the commit message.

deps/key-value-storage/src/local_fs/mod.rs
Xynnn007: Let me continue this discussion on the public.

rvps/src/bin/rvps.rs
Xynnn007: The same as mentioned. Let me close this and discuss on the public,

attestation-service/src/config.rs
Xynnn007: Actually this is generated by vibe coding. We need some discussion upon the trade-offs. Let me close this and continue on the public channel.

fitzthum:
@Xynnn007 

I don't think we should try to address backwards compatibility in this PR. We should discuss this, though. It's a good subject for a community discussion.

As for the example configs, I would move this to another PR. Personally, I don't really like having the examples hard-coded. It would be one thing if the example were automatically generated (from default values, perhaps) or if it were connected to the tests. Either way, I think we can figure that out later.

For now, I would try to drop whatever is not absolutely necessary so we can merge this soon.

Xynnn007:
@fitzthum good. Let me drop the commits and have a discussion later. Keep in mind that before that, the breakthrough changes will happen in this PR without any notification/helpers.

attestation-service/src/config.rs
```diff
@@ -21,6 +22,13 @@ pub struct Config {
     /// Optional configuration for verifier modules
     #[serde(default)]
     pub verifier_config: Option<VerifierConfig>,
+
+    /// Unified storage backend configuration for all storage needs in CoCo AS.
+    /// When provided, this will be used to create storage instances for:
+    /// - Built-in AS policy storage (instance: "attestation-service-policy")
+    /// - Built-in AS RVPS storage (instance: "reference-value")
+    #[serde(default)]
+    pub storage_backend: StorageBackendConfig,
```
mythi, line 31: Users moving to this version get `Memory` and `Postgres` by default? `Local*` based might be a safer choice as that is what most(?) people have used, WDYT

attestation-service/src/config.rs
```diff
@@ -29,22 +37,34 @@ pub enum ConfigError {
     IO(#[from] std::io::Error),
     #[error("failed to parse AS config file: {0}")]
     FileParse(#[source] std::io::Error),
-    #[error("failed to parse AS config file: {0}")]
+    #[error(
+        "failed to parse AS config file: {0}. \
+        If you are upgrading from an older version, the configuration format may have changed. \
+        Fields removed or replaced in this version: work_dir, policy_engine, attestation_token_broker.policy_dir, rvps_config.storage (for BuiltIn). \
```
mythi: could some fields become just serde skip, like `policy_engine` we could ignore. i

attestation-service/docs/config.md
```diff
@@ -56,25 +54,27 @@ This section is **optional**. When omitted, a new EC key pair is generated and u
...
-| `storage`   | ReferenceValueStorageConfig | Configuration of storage for reference values (`LocalFs` or `LocalJson`)       | No       | `LocalFs`|
+If `type` is set to `BuiltIn`, the following extra properties can be set:
 
-`ReferenceValueStorageConfig` can contain either a `LocalFs` configuration or a `LocalJson` configuration.
+| Property | Type | Description | Required | Default |
+|----------|------|-------------|----------|---------|
+| `extractors` | Object | Optional configuration for provenance extractors | No | None |
 
-For `LocalFs`, the following properties can be set
+**Note:** Storage configuration for BuiltIn RVPS is now managed through the unified `storage_backend` configuration (see [Storage Backend Configuration](#storage-backend-configuration)). The BuiltIn RVPS will use the `reference-value` namespace from the unified storage backend.
```
mythi, line 63: some migration guide could be added in a follow-up PR.

kbs/src/plugins/implementations/resource/backend.rs
```diff
@@ -69,10 +73,12 @@ impl fmt::Display for ResourceDesc {
     }
 }
 
-#[derive(Clone, Debug, Deserialize, PartialEq)]
-#[serde(tag = "type")]
+#[derive(Clone, Debug, Deserialize, PartialEq, Default)]
+#[serde(tag = "backend")]
```
mythi: why not keep `type`?

attestation-service/src/config.rs
Xynnn007: The current PR uses memory by default.

I initially thought that if the local storage path wasn't explicitly specified in the configuration, the default path would be used, which might lead to users being used without being informed.

However, I feel that memory usage isn't actually a big deal and is intuitive, since our do not configuring anything means somehow testing.

I think this choice can be discussed at the weekly meeting. If necessary, it can be fixed in a future PR, but without making it an obstacle to stateless implementation. wdyt?

attestation-service/src/config.rs
Xynnn007: After some thoughts, if we use `#[serde(skip)]` macro, we need to explicitly claim the field in the `Config` struct. So if we are deleting some fields, this does not help imo.

btw, this error information works for the back-ward compability. Let me delete this part and use the old one. Then we can handle back-ward compability altogether.

kbs/src/plugins/implementations/resource/backend.rs
Xynnn007: Good question. I originally thought "backend" would be a more appropriate expression here. However, changing the synonymous field name doesn't seem like a good idea, so I'll change it back.

attestation-service/src/config.rs
mythi: maybe `#[serde(skip)]` is not the suitable macro to use. My thinking was just that if user provides config entries that are harmless to just ignore, maybe we don't need to error on those.

attestation-service/src/config.rs
mythi: Just thinking the case that the new default could be the same as what the default is today without this PR.

attestation-service/src/config.rs
Xynnn007: This involves discussions about backward compatibility, which we can refine in another PR.

In that context, we need to consider the following scenario: a configuration item is removed in a certain version, but the user still configures that field. However, the code logic will discard this user-configured item by "skipping" and it won't take effect as expected. This is why I considered forcibly blocking disallowed values; because an error at startup would make the user aware that "the current configuration file is incorrect, and I need to modify the existing configuration according to the example file."

attestation-service/src/ear_token/mod.rs
fitzthum: btw, when I rebased the PPCIE PR, I saw that this does potentially change the behavior. Setting `#[default]` on the struct itself will use the generic serde default values which might not match what we want.

Xynnn007: Let me take a test, with config
```json
{
    "work_dir": "/var/lib/attestation-service/",
    "rvps_config": {
        "type": "BuiltIn",
        "storage": {
            "type": "LocalFs"
        }
    },
    "attestation_token_broker": {
        "duration_min": 5
    }
}
```
### Test 1: with rev `655d1090` (some version before this PR)

The launch log
```
2026-03-31T01:13:00.045793Z  INFO restful_as: Welcome to Confidential Containers Attestation Service (RESTful version)!


 ________  ________  ________  ________  ________  ________      
|\   ____\|\   __  \|\   ____\|\   __  \|\   __  \|\   ____\     
\ \  \___|\ \  \|\  \ \  \___|\ \  \|\  \ \  \|\  \ \  \___|_    
 \ \  \    \ \  \\\  \ \  \    \ \  \\\  \ \   __  \ \_____  \   
  \ \  \____\ \  \\\  \ \  \____\ \  \\\  \ \  \ \  \|____|\  \  
   \ \_______\ \_______\ \_______\ \_______\ \__\ \__\____\_\  \ 
    \|_______|\|_______|\|_______|\|_______|\|__|\|__|\_________\
                                                     \|_________|
                                                                                    
version: v0.1.0
commit: 655d1090b83bb59a7b57501bae63bb596b7ef2c1
buildtime: 2026-03-31 09:11:11 +08:00
loglevel: debug

2026-03-31T01:13:00.046109Z  INFO restful_as: Using config file attestation-service/config.json
2026-03-31T01:13:00.046362Z DEBUG restful_as: Attestation Service config: Config {
    work_dir: "/var/lib/attestation-service/",
    rvps_config: BuiltIn(
        Config {
            storage: LocalFs(
                Config {
                    file_path: "/opt/confidential-containers/attestation-service/reference_values",
                },
            ),
            extractors: None,
        },
    ),
    attestation_token_broker: EarTokenConfiguration {
        duration_min: 5,
        issuer_name: "CoCo-Attestation-Service",
        developer_name: "https://confidentialcontainers.org",
        build_name: "attestation-service 0.1.0",
        profile_name: "tag:github.com,2024:confidential-containers/Trustee",
        signer: None,
        policy_dir: "/opt/confidential-containers/attestation-service/token/policies",
    },
    verifier_config: None,
}
```

### Test 2: with the latest code

The launch log
```
2026-03-31T01:14:18.387962Z  INFO restful_as: Welcome to Confidential Containers Attestation Service (RESTful version)!


 ________  ________  ________  ________  ________  ________      
|\   ____\|\   __  \|\   ____\|\   __  \|\   __  \|\   ____\     
\ \  \___|\ \  \|\  \ \  \___|\ \  \|\  \ \  \|\  \ \  \___|_    
 \ \  \    \ \  \\\  \ \  \    \ \  \\\  \ \   __  \ \_____  \   
  \ \  \____\ \  \\\  \ \  \____\ \  \\\  \ \  \ \  \|____|\  \  
   \ \_______\ \_______\ \_______\ \_______\ \__\ \__\____\_\  \ 
    \|_______|\|_______|\|_______|\|_______|\|__|\|__|\_________\
                                                     \|_________|
                                                                                    
version: v0.1.0
commit: 8632b0230839b4e8f2b171b820503eb50decffcc
buildtime: 2026-03-31 09:13:51 +08:00
loglevel: debug

2026-03-31T01:14:18.388298Z  INFO restful_as: Using config file attestation-service/config.json
2026-03-31T01:14:18.388504Z DEBUG restful_as: Attestation Service config: Config {
    rvps_config: BuiltIn {
        extractors: None,
    },
    attestation_token_broker: EarTokenConfiguration {
        duration_min: 5,
        issuer_name: "CoCo-Attestation-Service",
        developer_name: "https://confidentialcontainers.org",
        build_name: "attestation-service 0.1.0",
        profile_name: "tag:github.com,2024:confidential-containers/Trustee",
        signer: None,
        verbose_token: true,
    },
    verifier_config: None,
    storage_backend: StorageBackendConfig {
        storage_type: Memory,
        backends: KeyValueStorageStructConfig {
            postgres: None,
            local_json: None,
            local_fs: None,
        },
    },
}
```

They are the same

fitzthum: Yes, but I had to add a custom default function for `verbose_token`. Otherwise, it would default to false.

Xynnn007: Oh. Yes. You should specify the default value for `verbose_token`. But actually you do not need to add [a function](https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/ear_token/mod.rs#L88), just specifying it in the default function can work. i.e.
```rust
impl Default for EarTokenConfiguration {
    fn default() -> Self {
        Self {
            duration_min: DEFAULT_TOKEN_DURATION,
            issuer_name: COCO_AS_ISSUER_NAME.to_string(),
            developer_name: DEFAULT_DEVELOPER_NAME.to_string(),
            build_name: format!("{} {}", env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION")),
            profile_name: DEFAULT_PROFILE.to_string(),
            signer: None,
            verbose_token: true,
        }
    }
}
```

fitzthum: That's what I had, but it didn't work. It's possible something weird happened with the rebase, though. I can take another look later.

PR #1192, opened 2026-02-27
Replace jwt-simple with jsonwebtoken

Cropi:
Replace jwt-simple with jsonwebtoken for Ed25519 JWT signing/verification

The project used jwt-simple solely for Ed25519 JWT signing (client) and verification (server) in the KBS admin API. Since jsonwebtoken v10 is already a workspace dependency (used in kbs token/jwk, attestation-service, and verifier), consolidating on it eliminates ~17 transitive dependencies from the jwt-simple tree without adding any new crates.

Changes:
- Remove jwt-simple from workspace and crate dependencies Add "use_pem" feature to workspace jsonwebtoken (makes from_ed_pem() explicitly available)
- Replace server-side verification in kbs/src/admin/simple.rs with jsonwebtoken::decode using Validation::new(Algorithm::EdDSA)
- Replace client-side signing in tools/kbs-client/src/lib.rs with a shared sign_admin_token() helper using jsonwebtoken::encode
- Update error type in kbs/src/admin/error.rs


tools/kbs-client/src/lib.rs
```diff
@@ -7,16 +7,37 @@
...
+
+fn sign_admin_token(auth_key: &str) -> Result<String> {
+    let encoding_key = EncodingKey::from_ed_pem(auth_key.as_bytes())?;
+    let now = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map_err(|e| anyhow!("System time error: {e}"))?
+        .as_secs();
+    let claims = AdminClaims {
+        iat: now,
+        exp: now + 7200,
```
mkulke: nit: maybe move that number as a const to the top of the file (TOKEN_EXPIRATION_SECONDS or something)

Cropi: I modified this, thanks.

mythi (approved):
LGTM. Can you squash the commits still. 

guest-components uses jwt-simple so we still need to keep an eye on it and its dependencies tho.

PR #1219, opened 2026-03-12
Refactor the CI workflows and scripts to make it clearer

Xynnn007:
Our CI surface had grown organically and it was hard to tell which tests ran when, with some e2e paths even executing twice and docs-only changes still triggering heavy jobs. This change separates KBS e2e build from execution, unifies different flows under a single suite, and makes test workflows consistently skip documentation-only pull requests
so that failures are easier to reason about and CI capacity is spent only where it provides signal.

The same logic also applies for the image building process. See concrete commit message.


Makefile
```diff
@@ -0,0 +1,68 @@
...
+	$(MAKE) -C tools/trustee-cli check
+
+# KBS e2e tests (reuses kbs/test Makefile)
+# Build-only target used by CI to prepare kbs/test binaries and artifacts
+kbs-e2e-build:
+	$(MAKE) -C kbs/test install-dev-dependencies
+	$(MAKE) -C kbs/test bins TEST_FEATURES="$(TEST_FEATURES)"
+
+test-kbs-e2e: kbs-e2e-build
+	$(MAKE) -C kbs/test e2e-test
```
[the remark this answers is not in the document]
Xynnn007: @copilot `e2e-test` target calls two inner `make` commands to do build things. If the target already exists, the build process will be skipped. Does this make sense?

.github/actions/build-single-image/action.yml
```diff
@@ -0,0 +1,74 @@
...
+    required: false
+    default: "."
+  build_args:
+    description: Extra build args string (e.g. --build-arg FOO=bar)
+    required: false
+    default: ""
+  build_option:
+    description: Extra docker build options (e.g. --push)
+    required: false
+    default: ""
```
fitzthum: It might be a little more clear if there were an arg specifically for whether to push the image.

.github/workflows/release-clis-to-ghcr.yml
fitzthum, line 1: This workflow is called `release`, but it looks like it's using the staged image tags. Is that right?

.github/actions/build-single-image/action.yml
```diff
@@ -0,0 +1,74 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+
+    - name: Login to GHCR Container Registry
```
fitzthum, line 37: I guess we only need to do this if we are actually planning on pushing the image. Not a big deal to do it either way though.

.github/workflows/test-e2e-as.yml
```diff
@@ -0,0 +1,50 @@
...
+      with:
+        path: |
+          ~/.cargo/registry/index/
+          ~/.cargo/registry/cache/
+          target/
+        key: rust-${{ hashFiles('./Cargo.lock') }}
+
+    - name: Install dependencies and run CoCo-AS e2e tests
+      run: |
+        go install github.com/fullstorydev/grpcurl/cmd/grpcurl@latest
```
fitzthum: Why do we need this? Could it go inside the Makefile?

.github/actions/build-single-image/action.yml
Xynnn007: Yes. Currently it's only used for `--push`. Let me refactor this

.github/workflows/release-clis-to-ghcr.yml
Xynnn007: Yes. It's following the original logic
https://github.com/confidential-containers/trustee/blob/main/.github/workflows/push-trustee-cli-to-ghcr.yml#L57
https://github.com/confidential-containers/trustee/blob/main/.github/workflows/push-kbs-client-to-ghcr.yml#L57

.github/actions/build-single-image/action.yml
Xynnn007: Yes. There is a condition flag (updated version)
```yaml
if: inputs.push == 'true'
```

.github/workflows/test-e2e-as.yml
Xynnn007: This is used in the CoCoAS e2e test to play the role as a client. Let me find a way to move it into Makefile

.github/workflows/release-clis-to-ghcr.yml
```diff
@@ -0,0 +1,105 @@
...
+            dockerfile: kbs/docker/kbs-client/Dockerfile
+            output_path: kbs-client
+            repo: ghcr.io/confidential-containers/staged-images/kbs-client
+          - component: kbs-client
+            arch: aarch64
+            instance: ubuntu-24.04-arm
+            dockerfile: kbs/docker/kbs-client/Dockerfile
+            output_path: kbs-client
+            repo: ghcr.io/confidential-containers/staged-images/kbs-client
+          # trustee-cli
```
mythi, line 32: not related to your PR but this seems broken in many ways. First, the binary is `trustee`:
```
$ oras pull ghcr.io/confidential-containers/staged-images/trustee-cli:0152a46142255a3e18a6c9925ebaf4c9458f39c9-x86_64
✓ Pulled      usr/local/bin/trustee
...
```

Second, it cannot run: 
```
$ usr/local/bin/trustee 
usr/local/bin/trustee: error while loading shared libraries: libsgx_dcap_quoteverify.so.1: cannot open shared object file: No such file or directory
```

.github/workflows/release-clis-to-ghcr.yml
```diff
@@ -0,0 +1,105 @@
...
+
+jobs:
+  build_and_push:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # kbs-client
+          - component: kbs-client
+            arch: x86_64
```
mythi: I wonder if `arch` is still needed with the native runners? With some Dockerfile location changes, the matrix can be simplified a bit to only include `component`  and `instance`?

.github/workflows/workflow-call-build-staged-images.yml
```diff
@@ -0,0 +1,269 @@
...
+      GHCR_TOKEN:
+        description: 'GitHub token for GHCR'
+        required: false 
+
+permissions:
+  contents: read
+
+jobs:
+  validate-inputs:
+    runs-on: ubuntu-latest
```
mythi:
```suggestion
    runs-on: ubuntu-24.04
```

Makefile
```diff
@@ -0,0 +1,91 @@
...
+test-e2e: test-kbs-e2e test-as-e2e
+
+# KBS: lint, fmt, unit/integration tests
+test-kbs-unit:
+	$(MAKE) -C kbs lint TEST_FEATURES="$(TEST_FEATURES)"
+	$(MAKE) -C kbs format
+	$(MAKE) -C kbs check TEST_FEATURES="$(TEST_FEATURES)"
+
+# Attestation service / RVPS / shared deps: fmt, clippy, tests
+test-as-unit:
```
mythi, line 23: these two targets seem a bit inconsistent especially when AS has its own `Makefile`.

.github/workflows/workflow-call-build-staged-images.yml
```diff
@@ -0,0 +1,269 @@
...
+          if [ -z "${{ secrets.GHCR_TOKEN }}" ]; then
+            echo "Error: GHCR_TOKEN secret must be provided when 'push' is true."
+            exit 1
+          fi
+  kbs:
+    if: inputs.image_group == 'kbs' || inputs.image_group == 'all'
+    needs: validate-inputs
+    strategy:
+      fail-fast: false
+      matrix:
```
mythi, line 40: same question on `arch` and `ARCH` here.

.github/workflows/release-clis-to-ghcr.yml
Xynnn007: Looks like the reason is that the binary is not statically linked, thus need dynamic libs.

btw we once agreed to use docker image of kbs-client / trustee-cli tools because of dynamic libs. We can get rid of binaries totally - unless we find a way to build statically.

.github/workflows/release-clis-to-ghcr.yml
Xynnn007: The target artifact repo tag still needs the arch argument. Like `${repo}:${commit_sha}-${arch},latest-${arch}`

.github/workflows/workflow-call-build-staged-images.yml
Xynnn007: Good. I also fixed other two parts of OS version in kbs-e2e

Makefile
Xynnn007: Yes. AS Makefile does not include cargo fmt definitions and only the building targets. Also, we include other crates into this target.

.github/workflows/workflow-call-build-staged-images.yml
Xynnn007: It's for the tag naming too

.github/workflows/release-clis-to-ghcr.yml
mythi: `export arch=$(uname -m)`

.github/workflows/release-clis-to-ghcr.yml
mythi: `kbs-client` seems OK and we could even enable more attesters for it and keep it statically linked.

I feel `trustee-cli` causes too much maintenance trouble vs how much it's used (probably nowhere?) so perhaps keep things minimal for it (test build in a container).

.github/workflows/release-clis-to-ghcr.yml
Xynnn007: Updated the workflows. Let's see the CI.

.github/workflows/test-rust-ci.yml
```diff
@@ -0,0 +1,139 @@
...
+    name: Key Broker Service (KBS)
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - instance: ubuntu-24.04
+            test_features: "pkcs11"
+          - instance: ubuntu-24.04-arm
+            test_features: "coco-as-builtin,coco-as-grpc,intel-trust-authority-as,cca-attester,pkcs11"
+          - instance: s390x-large
```
mythi, line 28:
```suggestion
          - instance: ubuntu-24.04-s390x
```

@BbolroC can you confirm?

Xynnn007: Originally https://github.com/confidential-containers/trustee/pull/1219/changes/c7ccfca5a839ad1605eb26216515becdd0d6c72d#diff-6e4d92e5bc3362db6d63a292cb90897f141a7265b02ee7875d70e06c3924a5a1L36

Xynnn007: Kindly ping @BbolroC for a double check : )

mythi: I believe he is out. It is fine to keep it unchanged and we can update it later

PR #1224, opened 2026-03-13
as: ear: tdx policy updates

mythi:


attestation-service/src/ear_token/ear_default_policy_cpu.rego
```diff
@@ -174,9 +174,7 @@ hardware := 2 if {
 	input.tdx.quote.header.tee_type == "81000000"
 	input.tdx.quote.header.vendor_id == "939a7233f79c4ca9940a0db3957f0607"
 
-	# Check TDX Module version and its hash. Also check OVMF code hash.
-	input.tdx.quote.body.mr_seam in query_reference_value("mr_seam")
-	input.tdx.quote.body.tcb_svn in query_reference_value("tcb_svn")
```
Xynnn007, line 179: I suggest to comment out the conditions and add the notes to tell the reason.

mythi: not sure why. I want to remove them.

mythi: `mr_seam` could maybe be like that but the other one is not needed.

Xynnn007: Yes. Does it make sense to comment out `input.tdx.quote.body.tcb_svn in query_reference_value("tcb_svn")` and add a note that use `tcb_status` instead as the `tcb_svn` is not easily understood-able.

Xynnn007: My idea is to explicitly show the users what fields can be configured than not telling anything, even if they can be replaced by better fields.

mythi: updated

PR #1226, opened 2026-03-16
WIP: Abstraction layer for different Attestation Services Backends

### LINKED ISSUE #1132: [KBS] Abstraction layer for different Attestation Services Backends

Xynnn007:
Close #1132


kbs/src/trust_context.rs
```diff
@@ -0,0 +1,161 @@
...
+
+/// The overall abstraction of the attestation result.
+/// This is used to hide the different token formats of
+/// backend attestation services and provide a unified interface
+/// to the KBS.
+///
+/// TODO: add more fields to the summary, or change the existing fields.
+#[derive(Default, Serialize, Deserialize, Clone)]
+pub struct AttestationSummary {
+    pub allowed: bool,
```
fitzthum: I don't think this is a very good idea. Whether the attestation token is allowed or not is more nuanced than this, which is exactly why we have the KBS policy.

kbs/src/attestation/coco/grpc.rs
```diff
@@ -142,6 +149,15 @@ impl Attest for GrpcClientPool {
         Ok(token)
     }
 
+    fn claims_to_trust_context(&self, claims: Value) -> Result<TrustContext> {
+        let (tee_pubkey, allowed) = parse_token_claims(claims)?;
+        Ok(TrustContext {
+            attestation_summary: AttestationSummary { allowed },
+            tee_pubkey,
+            custom_claims: Value::default(),
```
fitzthum: Wouldn't we want to extract the claims from the attestation token here?

kbs/src/trust_context.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+#[derive(Debug, Clone, Deserialize, PartialEq, Default)]
+pub struct TrustContextConfig {
+    /// URLs (file:// and https:// schemes accepted) pointing to a public key file
+    #[serde(default)]
+    pub public_keys: Vec<String>,
+
+    /// Path to the private key file.
+    pub private_key: Option<String>,
```
fitzthum: Need better names/comments here. Unclear how a user should populate this config.

fitzthum (commented):
Hmm. Not sure I see the value in adding this abstraction. Yes, different tokens have different properties, but this can be accounted for via the KBS policy.

I do see one potential benefit, which is that we probably do want to pass the token to plugins in the future. It's a little messy if different tokens have different formats. We don't want the plugin to need to account for this. Rather than introducing a new layer, it seems like we could just have the plugins parse out their claims and pass those to the plugins.

Xynnn007:
If the KBS policy were to consider this abstraction, it would mean that the KBS policy would be extremely complex, requiring the configuration of everything related to the tokens returned by different types of backend ASs. This abstraction masks the differences in AS Token formats.

Also, let the plugin to have an common signt to cover different AS tokens helps to fulfill such case:
1. There are multiple KBSs with different AS backends to act as a verifier, e.g. KBS A with CoCoAS + KBS B with Keylime + KBS C with ITA
2. One KBS D works as a relying party, to provide resources and other abilities facilitated by plugin system

In such case, the KBS D only needs one overall policy for different plugins. Similarly, only one code without duplication is needed for different plugin handler.

In engineering design, this decoupling is also very elegant, and can avoid more potential problems caused by coupling.

kbs/src/trust_context.rs
Xynnn007: I get both the overall result and claims and the original claims into it, to keep the corner case work as you mentioned.

kbs/src/attestation/coco/grpc.rs
Xynnn007: Added

PR #1234, opened 2026-03-18
nvidia: add PPCIE support

fitzthum:
PPCIE is a configuration of Hopper devices where 8 GPUs and 4 nvswitches are passed into the guest.

Add a special topology check that identifies this configuration and conditionally adds a new device to represent it. Since PPCIE is not reflected in the individual devices, but in the configuration of all the devices together, it is better to add this additional submod.

Also, add a few little fixes to the NVIDIA verifier. Start to remove the UUID parameter, which is not used, and get rid of the NVML wrapper which is not used.

Also, let's update the docker compose configuration to use the remote verifier by default.

### DIFF
--- a/attestation-service/src/ear_token/broker.rs
+++ b/attestation-service/src/ear_token/broker.rs
@@ -159,6 +159,7 @@ impl EarAttestationTokenBroker {
                 tee_claims.init_data_claims.clone(),
                 tee_claims.runtime_data_claims.clone(),
                 tee_claims.tee,
+                self.config.verbose_token,
             )?;
 
             let tcb_claims_json = serde_json::to_string(&tcb_claims)?;
@@ -442,8 +443,9 @@ fn generate_ec_keys() -> Result<(EcKey<Private>, Vec<u8>, Vec<u8>)> {
 pub fn transform_claims(
     mut input_claims: Value,
     init_data_claims: Value,
-    runtime_data_claims: Value,
+    mut runtime_data_claims: Value,
     tee: Tee,
+    verbose: bool,
 ) -> Result<BTreeMap<String, RawValue>> {
     let mut output_claims = BTreeMap::new();
 
@@ -470,8 +472,20 @@ pub fn transform_claims(
                 "report_data".to_string(),
                 RawValue::String(report_data.as_str().unwrap().to_string()),
             );
+
+            // When attesting environments with lots of devices, the additional evidence
+            // can create a token that is very large. This can overwhelm the header size
+            // limitation of some http servers (including the KBS).
+            // If verbose is false, don't include the additional evidence in the token.
+            if !verbose {
+                runtime_data_claims
+                    .as_object_mut()
+                    .and_then(|map| map.remove("additional-evidence"));
+            }
+
             let transformed_claims: RawValue =
                 serde_json::from_str(&serde_json::to_string(&runtime_data_claims)?)?;
+
             output_claims.insert("runtime_data_claims".to_string(), transformed_claims);
         }
     }
@@ -625,7 +639,7 @@ mod tests {
         let init_data_claims = Value::String("".to_string());
         let runtime_data_claims = Value::String("".to_string());
         let transformed_claims =
-            transform_claims(json, init_data_claims, runtime_data_claims, Tee::Tdx)
+            transform_claims(json, init_data_claims, runtime_data_claims, Tee::Tdx, true)
                 .expect("flatten failed");
 
         let expected_claims = json!({
--- a/attestation-service/docs/config.md
+++ b/attestation-service/docs/config.md
@@ -35,6 +35,7 @@ section:
 | `build_name`  | String                  | The build name to be used as part of the Verifier ID in the EAR         | No       | Automatically generated from Cargo package and AS version|
 | `profile_name`  | String                  | The Profile that describes the EAR token         | No       |tag:github.com,2024:confidential-containers/Trustee`|
 | `signer`       | [TokenSignerConfig][2]  | Signing material of the attestation result token.    | No       | None       |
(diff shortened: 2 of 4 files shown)


deps/verifier/src/nvidia/mod.rs
```diff
@@ -121,6 +128,79 @@ impl NvDeviceReportAndCertClaim {
...
+                .ok_or(anyhow!("switch must specify GPU PDIs"))?
+                .as_array()
+                .ok_or(anyhow!("switch must specify it's GPU PDIs as array"))?
+                .iter();
+            if switch_gpu_pdis.len() != 8 {
+                bail!("Each switch must be connectd to 8 GPUs");
+            }
+
+            for gpu_pdi in switch_gpu_pdis {
+                gpu_pdis.insert(gpu_pdi.as_str().unwrap().to_string());
```
Xynnn007:
```suggestion
                gpu_pdis.insert(gpu_pdi.as_str().ok_or(anyhow!("gpu pdi must be string"))?.to_string());
```

kbs/config/as-config.json
```diff
@@ -11,5 +11,10 @@
             "key_path": "/opt/confidential-containers/kbs/user-keys/token.key",
             "cert_path": "/opt/confidential-containers/kbs/user-keys/token-cert-chain.pem"
         }
+    },
+    "verifier_config" : {
+	"nvidia_verifier": {
+	    "type": "Remote"
+	}
```
Xynnn007: would you help to add an ident for formatting?

attestation-service/src/ear_token/broker.rs
```diff
@@ -61,6 +61,22 @@ impl EarAttestationTokenBroker {
...
             .await?;
 
+        let default_switch_policy = include_str!("ear_default_policy_switch.rego").to_string();
+        let default_switch_policy =
+            base64::engine::general_purpose::URL_SAFE_NO_PAD.encode(default_switch_policy);
+
+        policy_engine
+            .set_policy("default_switch".to_string(), default_switch_policy, false)
+            .await?;
+
```
mythi: was this missed from 1bd9eb2fd? maybe have a separate commit for it.

attestation-service/src/ear_token/ear_default_policy_ppcie.rego
```diff
@@ -0,0 +1,35 @@
...
+
+default hardware := 0
+
+default executables := 0
+
+# Currently, the PPCIE device class is only added
+# when a valid PPCIE topology is found.
+# Thus, the configuration is known to be valid.
+# For information about the switches and gpus used
+# in PPCIE, refer to the individual device submods.
```
mythi, line 13: `ppcie` is still a mode enforced in all devices rather than any "device" to justify a dedicated class, IMO.

Why did you abandon the idea of having a "device belongs to a multi-gpu ppcie cluster"  boolean claim for the individual devices?

fitzthum: I'm not sure I ever really had the idea of adding a ppcie claim to every device. The individual device evidence does not report a PPCIE mode. We rely on the overall topology check to determine if PPCIE is enabled. Arguably it is stretching the device model a little bit to introduce a new submod for PPCIE, but I think this is more intuitive than updating the individual devices and it is certainly easier for a user to check for one field rather than 12.

Anyway I am open to massaging this a little bit in the future, particularly if the device attestation results are updated, but this seems like the best way to do it for now.

kbs/config/as-config.json
fitzthum: The indentation seems correct on my end. Do you mean switching from tabs to spaces?

attestation-service/src/ear_token/broker.rs
fitzthum: Yes, I think I forgot to add that file. Moved it to a second commit.

Xynnn007:
@fitzthum a rebase is neeeded

attestation-service/src/ear_token/ear_default_policy_ppcie.rego
mythi: > The individual device evidence does not report a PPCIE mode. 

the local verifier has that: `config.protected_pcie_status`: Protected PCIe status. Maybe it's also hidden in some NRAS claims values?

fitzthum: NRAS does not report that claim. That claim alone just states that PPCIE mode is on, but it isn't enough to check if it's properly configured. The PDI claims that we use for the topology check are only returned in PPCIE mode, so what we do here is a superset of checking the PPCIE status.

mythi: OK, and I can see `validate_ppcie()` already works with NRAS claims so `protected_pcie_status` is not needed.

How is this going to work for any future topologies? As it is now, the topologies must be checked by the verifier and each topology comes out as a device class. it does not look impossible to have `validate_ppcie()` as a rego check in the resource policies.

fitzthum: PPCIE is specific to Hopper and only supports one configuration. I don't think other MPT platforms/configurations will require a check like this, but we'll see. I did look into implementing this in the KBS policy, but the Rego was unwieldy. Like I said, we can update if this ends up being annoying.

mythi: maybe there's a way to offer some rego libraries for these

fitzthum: yeah, we could add an extension for this as well. the tricky thing is that really this would be better to evaluate as part of the AS policy imo (so that the token can have the ppcie status), but we don't have the full view of all the devices there. hence, doing it in the verifier.

Issue #1112, opened 2025-12-16
Support multiple resource policies


fitzthum:
Today we have one KBS policy covering all of the resources. Of course the policy receives the resource URI as input, so this works fine. You can easily create rules for any particular resource.

That said, when we introduce admin roles, we will have a limitation. Admin roles work at the granularity/level of the request URI. The resource policy only has one URI but it has a very broad scope. If you can set the resource URI, you can release any secrets.

Is there a way that we can partition resources between different admin personas, allowing one admin to set and release one subset of secrets while another admin handles another subset? This already works with setting secrets, since the resource URI is in the path.

To support resource partitioning in the policy, we can introduce scoped resource policies. Basically, an admin could set as many resource policies as they want. Each policy would correspond to a portion of a resource URI. For instance a policy called `a/b/c` would cover once specific resource, while a policy `a/b` would correspond to any resource that match the pattern. Each of these policies would be set via a different request path. For a resource to be released, all applicable policies would have to be valid. The existing policy would remain and would always need to be valid.

Most users would probably just use one policy, but this would allow more advanced organizations to partition resources between different admin, which seems cool. wdyt?

bpradipt:
How about partitioning at the namespace level (ie `a` in a/b/c)? 
This will nicely fit with the K8s world as well where the KBS namespace name maps to K8s namespace. This makes administration and auditing easier for the secrets gated by KBS
User having access to K8s namespace `dev1` has access to resources in KBS under `dev1/` via KBS resource policy.

PR #1239, opened 2026-03-20
Kbs/Admin: modularize admin authentication and authorization

Xynnn007:
Refactor the admin module from simple monolithic implementations to a modular architecture that separates token verification from authorization decisions.

This work will decouple the token provisioning to outer services. And the inner logic inside KBS will be only
- Do nothing
- Deny anything
- verify token & do filter based on the `subject` claim of the token

TODOs (New):
- [ ] Add blogs or documents on coco.org about integrate outside auth services.


kbs/src/admin/mod.rs
```diff
@@ -2,146 +2,110 @@
...
-
-            let re = Regex::new(&role.allowed_endpoints)?;
-
-            if roles.insert(role.id, re).is_some() {
-                return Err(Error::DuplicateAdminRole);
-            }
+        match value {
+            AdminConfig::InsecureAllowAll => Ok(Admin::InsecureAllowAll),
+            AdminConfig::DenyAll => Ok(Admin::DenyAll),
+            AdminConfig::Enforce {
```
fitzthum: Can we come up with a better name here? `CheckToken`, `TokenACL`, `TokenAuthorization`? Something like that.

kbs/src/config.rs
```diff
@@ -473,12 +488,7 @@ mod tests {
             insecure_http: true,
             ..Default::default()
         },
-        admin: AdminConfig {
-            admin_backend: AdminBackendType::Simple(SimpleAdminConfig {
-                personas: Vec::new(),
-            }),
-            roles: Vec::new(),
-        },
+        admin: make_enforce_admin_config_with_defaults(),
```
fitzthum: Regarding our earlier discussion about printing sample configs, I think it's a good idea to pass them through the tests like this.

kbs/config/docker-compose/setup.sh
fitzthum, line 1: Thanks for adding this. It's much nicer than having the huge command in the docker compose file.

kbs/config/docker-compose/as-config.json
fitzthum, line 1: Ahhh thanks for moving this. This has been confusing for a long time.

fitzthum (commented):
Sorry for the delay. I like this PR. 

Providing a private key to the KBS Client has always seemed a bit sketchy (even though it isn't).

That said, there are some downsides to providing a token. First, it's a little more annoying for admin to generate this. Usually this will happen automatically, I guess. Second, there is potential for the token to expire. You set a long expiration in the script, but this will still be a tradeoff in some cases. 

Still, I think this is probably the right way to go because it unifies things with a flow coming rom an external token service.

Anyway, I made a couple comments. Can make some more once it is ready to review.

kbs/src/admin/mod.rs
Xynnn007: Let's use TokenAuthorization

kbs/src/admin/authorization/regex_acl.rs
```diff
@@ -0,0 +1,78 @@
...
+
+use crate::admin::{authorization::AuthorizationTrait, error::*, AuthorizationDecision, Claims};
+
+/// An ACL rule that grants a specific audience access to selected endpoints.
+#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct AdminAclRule {
+    // TODO: add more fields here
+    /// The admin audience that this rule applies to.
+    pub audience: String,
```
fitzthum: Pretty sure this is the wrong field. The audience refers to the application or client that is intended to receive the token. The individual represented by the token is the subject. This should probably be the subject or the role.

kbs/config/docker-compose/setup.sh
```diff
@@ -0,0 +1,41 @@
...
+
+if [ ! -s admin-token ]; then
+  b64url() {
+    openssl base64 -A | tr '+/' '-_' | tr -d '='
+  }
+
+  header='{"alg":"EdDSA","typ":"JWT"}'
+  iat="$(date +%s)"
+  exp="$((iat + 315360000))" # 10 years
+  payload="{\"iss\":\"admin\",\"sub\":\"admin\",\"aud\":[\"admin\"],\"iat\":${iat},\"exp\":${exp}}"
```
fitzthum: It's a bit confusing for the issuer to be `admin`. The issuer is this sectup script. Maybe should say something like `TrusteeInDocker` or just `Trustee` to signify that the token is not coming from some external service.

Also, see previous note regarding the audience, which should probably be `Trustee` or `KBS` afaik.

kbs/src/admin/token_verifier/bearer_jwt.rs
```diff
@@ -0,0 +1,127 @@
...
+use std::collections::BTreeMap;
+use tracing::{error, info};
+
+use crate::admin::{error::*, token_verifier::TokenVerifierTrait, Claims};
+use crate::crypto::jwt::JwtVerifier;
+
+#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+#[serde(default)]
+pub struct BearerJwtConfig {
+    pub idps: Vec<IssuerConfig>,
```
fitzthum: maybe spell out this field. not sure everyone knows what `idp` is.

kbs/src/admin/token_verifier/bearer_jwt.rs
```diff
@@ -0,0 +1,127 @@
...
+        let bearer = Authorization::<Bearer>::parse(request)?.into_scheme();
+        let token = bearer.token();
+
+        for (issuer, verifier) in &self.trusted_issuers {
+            match verifier.verify(token) {
+                Ok(raw_claims) => match claims_from_value(raw_claims) {
+                    Ok(claims) => {
+                        info!("Admin access check for a token from {} succeeded.", issuer);
+                        info!("Admin claims: {:?}", claims);
+                        return Ok(claims);
```
fitzthum: Do we also need to check that the issuer matches the issuer claim?

fitzthum (commented):
NIce. The integration is great.

I have a few comments. Mainly they are related to how we use various claims. I think this can be tightened up across the PR.

kbs/src/admin/token_verifier/bearer_jwt.rs
Xynnn007: It's also a question from my side. Currently we only have `subject` as configurable item in the config file and leave more fields as todo. Do you think we should include `issuer` and `audience` also in this PR?

kbs/src/admin/authorization/regex_acl.rs
Xynnn007: Yes. Fixed, thanks for this

kbs/src/admin/token_verifier/bearer_jwt.rs
Xynnn007: Good. Let me use `identity_providers`

kbs/src/admin/token_verifier/bearer_jwt.rs
fitzthum: The issuer is also in the config file, right? It's the key in the idps map. I think we should probably check that the issuer specified there actually matches the issuer claim.

Probably we should check that audience claim as well. I guess the idea is to make sure that the token that is used is really intended to login to this service and is not just some existing token from the idp. Ofc it makes things a little more complicated to setup. Potentially we could have this be configurable. Generally, though we should probably lean towards having the admin backend be more rigorous.

kbs/src/admin/token_verifier/bearer_jwt.rs
```diff
@@ -0,0 +1,145 @@
...
+                continue;
+            }
+
+            if !audience.is_empty() && !claims.audiences.contains(audience) {
+                error!(
+                    "Token audience mismatch: expected {audience}, got {:?}",
+                    claims.audiences
+                );
+                continue;
+            }
```
Xynnn007: @fitzthum FYI. I add iss / aud checking logic. either of them will be triggered if configured in config.

kbs/src/admin/token_verifier/bearer_jwt.rs
```diff
@@ -0,0 +1,145 @@
...
+        for (issuer, (audience, verifier)) in &self.trusted_issuers {
+            let Ok(raw_claims) = verifier.verify(token) else {
+                debug!("Failed to verify token with issuer {issuer}");
+                continue;
+            };
+            let claims = claims_from_value(raw_claims).map_err(|e| Error::AdminAccessDenied {
+                reason: format!("Failed to parse claims from token: {e:?}"),
+            })?;
+
+            if !issuer.is_empty() && claims.issuer != *issuer {
```
fitzthum: maybe we should be using options for issuer and audience? also might want to print a warning if they are not set.

fitzthum (commented):
Ok. I think we are close here. One more nit.

@confidential-containers/trustee-maintainers PTAL this is an important PR

kbs/src/admin/token_verifier/bearer_jwt.rs
fitzthum: Also it's slightly odd that we're using the issuer as the key in a map when it can apparently be empty? It seems like maybe that should not be optional.

Xynnn007: > Also it's slightly odd that we're using the issuer as the key in a map when it can apparently be empty? It seems like maybe that should not be optional.

This can be optional. Actually for a user, he can either set the issuer name or not. But what's required is the public key. If it's given, the verification happens; if not, only check the endorsement against the public key,

kbs/src/admin/token_verifier/bearer_jwt.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use actix_web::{http::header::Header, HttpRequest};
+use actix_web_httpauth::headers::authorization::{Authorization, Bearer};
+
+use serde::Deserialize;
+use serde_json::Value;
+use tracing::{debug, error, info, warn};
+
+use crate::admin::{error::*, token_verifier::TokenVerifierTrait, Claims};
+use crate::crypto::jwt::JwtVerifier;
```
mythi, line 15: not available for this commit. do we need 3bd16e01c5c531a79c0fec2d8b63775f67e8e84a in this PR or move it to a separate one?

kbs/src/admin/token_verifier/mod.rs
```diff
@@ -0,0 +1,47 @@
+// Copyright (C) Copyright Confidential Containers Project Authors.
```
mythi: nit: the copyright claims are used somewhat inconsistently in this PR

kbs/docs/admin.md
```diff
@@ -0,0 +1,209 @@
...
+- KBS verifies JWT signature using configured JWKS and/or PEM public keys
+- KBS checks JWT `iss` an `aud` in `identity_providers` if configured
+- KBS maps JWT `sub` to local admin subjects
+- KBS authorizes each admin API path via regex ACL rules
+
+This lets you keep identity issuance external while keeping KBS-side authorization explicit and auditable.
+
+## Quick Demo: Authentik + authn Service + kbs-client
+
+[Authentik](https://docs.goauthentik.io/) is an IdP (Identity Provider) and SSO (Single Sign On) platform that is built with security at the forefront of every piece of code, every feature, with an emphasis on flexibility and versatility.
```
mythi: I'd keep any 3rd party services out from the main docs. Maybe a blog or something once this feature settles.

kbs/docs/admin.md
```diff
@@ -0,0 +1,209 @@
...
+  --resource-file /tmp/test_resource
+```
+
+## `BearerJwt` Config Reference
+
+`[admin.token_verifier]` with `type = "BearerJwt"` accepts:
+
+- `identity_providers` (array): list of trusted identity providers
+- `insecure_public_key_from_uri` (optional bool, default `false`): allow fetching
+  `public_key_uri` and `jwk_set_uri` over plaintext `http://` for compatibility
```
mythi: I don't think we ever supported `http://`. What compatibility this is referring to?

Makefile
```diff
@@ -65,18 +65,19 @@ test-kbs-docker-e2e:
...
-	! ./kbs-client --url http://127.0.0.1:8080 get-resource --path "test-org/test-repo/test-secret" && \
-	./kbs-client --url http://127.0.0.1:8080 config --auth-private-key $$E2E_DIR/kbs/config/private.key set-resource-policy --policy-file "$(CURDIR)/deps/policy-engine/test_data/policy_2.rego" && \
-	./kbs-client --url http://127.0.0.1:8080 get-resource --path "test-org/test-repo/test-secret"
+	sudo ./kbs-client --url http://127.0.0.1:8080 config --admin-token-file $$E2E_DIR/kbs/config/docker-compose/admin-token set-resource --path "test-org/test-repo/test-secret" --resource-file test-secret && \
+	! sudo ./kbs-client --url http://127.0.0.1:8080 get-resource --path "test-org/test-repo/test-secret" && \
+	sudo ./kbs-client --url http://127.0.0.1:8080 config --admin-token-file $$E2E_DIR/kbs/config/docker-compose/admin-token set-resource-policy --policy-file "$(CURDIR)/deps/policy-engine/test_data/policy_2.rego" && \
+	sudo ./kbs-client --url http://127.0.0.1:8080 get-resource --path "test-org/test-repo/test-secret"
```
mythi: sudoless e2e worked with the private key method so it's better to keep the same approach for the token. if this is needed, then all docs in 8538cac86c56dede5915115925350a1b3aaf42bb are also broken, right?

kbs/docs/config.md
```diff
@@ -173,48 +173,52 @@ Detailed [documentation](https://docs.trustauthority.intel.com).
...
-If a matching role is found, the `allowed_endpoints` regular expression is used to determine
-which endpoints the admin can access.
-The regular expression must start with `^/kbs` and end with `$` to avoid ambiguity.
-If no matching role is found (and at least one role is specified), the admin will not be able
-to access any endpoints.
+Each ACL entry:
+
+| Property | Type | Description | Required |
+|----------|------|-------------|----------|
+| `subject` | String | JWT `sub` value to match | Yes |
```
mythi: `subject` seems specific to JWT but what if we wanted to add some api-key based flow later. how to map users with rules in that cases? (maybe there is no answer right now but just thinking how to keep this future proof)

mythi (commented):
1st round

kbs/src/admin/token_verifier/bearer_jwt.rs
Xynnn007: In this PR I think. As both Attestation token verifier and Admin token verifier use JWK verification logic. It's good to bring them together in one place, than inventing another

kbs/docs/admin.md
Xynnn007: That means if `http://` is used, it will try to fetch with HTTP. Both HTTPS and HTTP are using the same field, thus "for compatibility". Let me use another expression

Makefile
Xynnn007: Let me do a fixup for the Makefile change into https://github.com/confidential-containers/trustee/commit/8538cac86c56dede5915115925350a1b3aaf42bb

The reason we need sudo here is that we need to read files under `tmp` that's created by `sudo docker-compose`. In previous version, the admin keys are generated outside the docker compose and using `openssl` command. In this new version, the admin tokens are generated by the script in `docker-compose` init container.

kbs/docs/config.md
Xynnn007: You can see that we now have three modes for admin module
1. InsecureAllowAll
2. DenyAll
3. TokenAuthorization: means uses JWT to do auth. It includes two sub-parts. One for JWT verification, one for authZ.

If needs, we can extend another mode named API key. But as we stated before, it's more recommended to have the authN/authZ as network gateway or something else in front of KBS, and use `InsecureAllowAll` mode for KBS to only work as a black box - this would help us avoid re-inventing wheels.

kbs/docs/admin.md
Xynnn007: tbo I am ok with either. so kindly ping @fitzthum for a consensus

Makefile
mythi: I understand why it is added. My comment is that the new approach is not good if we need add sudo here as a workaround.

kbs/src/admin/token_verifier/bearer_jwt.rs
mythi: OK, but I believe then you need to bring that commit as the first one in this PR.

Makefile
Xynnn007: Oh. so the root logic is using docker-compose to deploy the trustee. and during launch the docker-compose generates the logic token. and the token natually has "sudo" permission. One way is to change the script in docker-compose of generating token, to make the token "chmod 0644"?

kbs/docs/config.md
mythi: my comment was just that if `subject` is the best entry to describe the role because it seems specific to what is in a JWT.

Xynnn007: Yes. Continue my comment - 

Coincidently in `TokenAuthorization`, we could use `subject` to do ACLs. Certainly for API Key mode we can reuse some of code, to mapping a API key to also a "subject" abstraction and use the same ACL module. But tbo if we do not have code, it's more like just talking and probably get nothing. Let me treat your original comment as an open, and I am sure the answer is - yes! The architecture is compatible.

mythi: I don't quite agree `subject` is a good choice. Just hoping we don't need to change it later.

This triggers another comment: we seem to drop the idea of roles and ask users to maintain complex regular expressions. Did you consider having the regular expression rules defined as some roles?

fitzthum: personally I think role is a better term than subject. Technically tho I think these terms don't actually mean the same thing. The subject refers to a particular instance whereas a role is more like a class. So are we trying to express our admin ACLs in terms of specific users or are we talking about classes of users? i suspect we want the second one (a role). i think subjects are supposed to be unique. if we express our ACLs in terms of subject, we're going to need an entry for every user, which is not ideal. JWTs can have both a subject and a role btw. possibly we should expect our tokens to have a role set? this does make things more complex tho.

kbs/docs/admin.md
fitzthum: Personally I don't care. Maybe a compromise could be to have an example on the website but not in the repo.

kbs/src/admin/token_verifier/bearer_jwt.rs
fitzthum: I mostly agree, but it does seem like it would be safer to also check that the issuer matches what is actually in the token. And why not be safer in this part of the project?

Xynnn007: Ah yes. Actually it depends on users. The user can config this - my actual deployment with authentik told me that the issuer is a random string for each authentik lifetime, and I frequently restarts authentik to do tests without changing the kbs config.

So, my idea is not to require this field to be configured, but once it's configured, we will do the check.

kbs/src/admin/token_verifier/bearer_jwt.rs
Xynnn007: Ok. let me do a rebase.

kbs/docs/config.md
Xynnn007: Let me try to explain in more detail, ensuring we're all on the same thread.

In a PR, when we enable `TokenAuthorization` mode, the following two items are also pluggable:

1. Token verifier: Verifies the token's endorsement and returns a claim.

2. authZ: now we only have an ACL based on the claim.

When we select JWT, step 1 verifies the endorsement, checking if the issuer and audience match the configuration, and then returns the JWT's claim to step 2. This is the current shape in the PR.

So to avoid too specific to JWT, we need to change the "claim" thing, making it a generic entity used to select an ACL rule set or mark as a role for authZ plugin. 

So the question would be, for JWT, let's return `sub` field as the role?
Or, as @fitzthum said, "JWTs can have both a subject and a role btw". I noticed in JWT spec role is not a fixed field but can be embedded in payloads by generator. Do you mean that we require the generator of the JWT to embed a "role" field in claims? Or - do we have other mechanism to map a JWT's claims into a role?

Xynnn007: btw, still I want to state again. For complex systems, it's strongly recommended to build an authN and authZ service in front of KBS, and use insecureAllowAll for KBS.

mythi: > The subject refers to a particular instance whereas a role is more like a class. So are we trying to express our admin ACLs in terms of specific users or are we talking about classes of users?

I side-tracked a bit from that `subject` topic asking if we should do something like

```json
{ user = "kbs-admin", roles = [resources:read, policies:write, foo:bar ] },
```

instead of

```json
{ subject = "kbs-admin", allowed_endpoints = "^/kbs/v0/resource/.+$" },
```

Xynnn007: looks nice. This can be achieve in this PR. Still the question. How to map JWT claims into a `user/role`? By requiring that `role/user` should be included into the claims?

fitzthum: Unfortunately I think the world is a little bit more complicated than we first realized. JWT does not have a role or a scope as a public claim. Schemes like 0Auth, OID, Auth0, etc all have private claims that address this sort of thing.

So it could be that trying to implement this at the JWT level doesn't quite add up. Mapping a JWT subject to a role doesn't seem correct. To support JWTs that don't have any private claims, we might need some internal logic to map subjects to roles. This doesn't seem very good either. Might be better to just require that JWTs have a role claim and make sure the ones that we generate do.

It's possible there should be another step in here. 1) Validate JWT 2) Map JWT claims (this depends on the authentication scheme being used) 3) ACLs. 

but on the other hand I don't want things to get too complicated in this PR.

fitzthum: > `{ user = "kbs-admin", roles = [resources:read, policies:write, foo:bar ] },`

This is much less powerful than the regex, which allows you to partition resources into groups that can be modified by particular admins.

Xynnn007: > Might be better to just require that JWTs have a role claim and make sure the ones that we generate do.

This seems a practical way. But we need to notice that in this way the `sub` field will be "ignored", and only printed in the log. I am in favor of this way. Let me get this change. And for more complex ways to configure, I'd prefer a blog or document to leverage outside authN/authZ services + trustee.

Xynnn007: Let's not make the core trustee too complex, but keep the plugin mechanism enough.

mythi: > > `{ user = "kbs-admin", roles = [resources:read, policies:write, foo:bar ] },`
> 
> This is much less powerful than the regex, which allows you to partition resources into groups that can be modified by particular admins.

True. This can be useful indeed but then we might also need to update the resource policy model to have a policy attached to these groups to avoid one global policy admin that controls read access to everything, right?

fitzthum: > This can be useful indeed but then we might also need to update the resource policy model to have a policy attached to these groups to avoid one global policy admin that controls read access to everything, right?

See https://github.com/confidential-containers/trustee/issues/1112, but even without this you could still have one higher-privilege admin who can set the policy and then different lower-privilege admin who can set different resources. It's also handy for plugins that could have arbitrary request structures. The regex is kind of annoying in some ways tho. I am open to changing it down the road, but let's see how it plays out.

fitzthum (approved):
LGTM

Xynnn007:
Let's wait for @mythi before getting merged

kbs/docs/admin.md
```diff
@@ -0,0 +1,93 @@
+# KBS Admin Module
+
+The `kbs/src/admin` module protects KBS admin APIs. It supports three modes:
+
+- `InsecureAllowAll`
+- `DenyAll`
+- `TokenAuthorization(TokenVerifier, Authorization)`
+
+In `TokenAuthorization` mode, KBS validates an admin token and then applies endpoint authorization rules.
```
mythi, line 9: if new authentication mechanism is added, is it going to be a `mode` here or should it be a (sub-)type? At the moment both `mode` and `type` seem to be JWT specific and partly overlapping.

Thinking out loud: maybe `mode` should be `authn` to better describe the intended functionality?

Xynnn007: It makes sense to use `mode` here. There are three "modes"
- `DenyAll`: all admin endpoints are abandoned.
- `InsecureAllowAll`: all admin endpoints are accessible.
- `TokenAuthorization `: verify a token, then do authZ upon the token
  - `token_verifier.type`: what token is checked, now we supports `BearerJwt`
  - `authorizer.type`: how to do authZ, now we supports `RegexAcl`

Where a new plugin will ly depends on In which level you will implement a plugin

Xynnn007: It makes sense to use `mode` here. There are three "modes"
- `DenyAll`: all admin endpoints are abandoned.
- `InsecureAllowAll`: all admin endpoints are accessible.
- `TokenAuthorization `: verify a token, then do authZ upon the token
  - `token_verifier.type`: what token is checked, now we supports `BearerJwt`
  - `authorizer.type`: how to do authZ, now we supports `RegexAcl`

Where a new plugin will ly depends on In which level you will implement a plugin

mythi: > Where a new plugin will ly depends on In which level you will implement a plugin

This does not help to answer my question. Which level #1025 or some api-key authentication should go?

Xynnn007: Emmmm actually the "login" operation is part of authN, and which we decided to implement outside the trustee. What leave inside the KBS is a simple authZ mechanism. For example "how to provision an admin token" is a authentication problem - which defines who you are. Authorization - which can be simply done inside KBS - cares about "what you can do".  The reason is that authentiation logic is mature in software system and we can easily find projects/docker-images that fullfill this need and we can integrate them with current PR seamlessly.

So go back to the question about the password/username mechanism, I do not think it will get into the system.  You can see the deprecated "Authentik" component is a out-of-band authentication service that supports username/password login.

Xynnn007: About the API key. We can implement a `token_verifier` named `BearerApiKey` (means that apikey is carried in HTTP bearer head) or something else.

mythi: > What leave inside the KBS is a simple authZ mechanism.

right, maybe the PR title made me confused on 'authn' or something. thanks for the clarification

> About the API key. We can implement a `token_verifier` named `BearerApiKey` (means that apikey is carried in HTTP bearer head) or something else.

or `mode = ApiKeyAuthorization` :-) I don't want to be too pedantic here but I was hoping to get the entries in the config files self-explanatory so that they don't need to change if new functionality is added (a use case where an admin can generate api-keys that control access to some endpoints seems reasonable, no?).

Xynnn007: > or `mode = ApiKeyAuthorization` :-) I don't want to be too pedantic here but I was hoping to get the entries in the config files self-explanatory so that they don't need to change if new functionality is added (a use case where an admin can generate api-keys that control access to some endpoints seems reasonable, no?).

Yes.

TBO, as the plugin design is flexible enough thus both way (`mode` level, or `token_verifier` level) might be ok. But if we use `token_verifier`, we should have a proper way to map one api key to a `role` abstraction. The `role` is used to do ACLs or other `authorizer`. If we cannot define a proper way to map an api key to a `role` abstraction, we should consider the `mode`.

Personally, I tend to prioritize classifying abstractions into small-grained forms (`token_verifier + authorizer` level), and only use larger-grained abstractions (`mode`) if that fails.

kbs/test/config/external-plugin-tls.toml
```diff
@@ -2,8 +2,11 @@
 sockets = ["127.0.0.1:8086"]
 insecure_http = true
 
+[attestation_token]
+insecure_key = true
```
Xynnn007, line 6: Needs to add some notes to this line:

Previously, if we do not set the KBS `attestation_token` claims (default `insecure_key` is `false`), the token verification will still succeed. This is a hidden bug! Now it's fixed by the [commit](https://github.com/confidential-containers/trustee/pull/1239/changes/a4e53c144e66ff5acf510db2079ebe14bb01eac8) in a way that requires to set the trusted pubkey/anchor.

Also, it will get more test cases in the [commit](https://github.com/confidential-containers/trustee/pull/1150/changes/59d2d9d489c216f5d5e908e52e282eb64bf536ee)

fitzthum:
Are we ready to merge this? cc @mythi

mythi:
> Are we ready to merge this? cc @mythi

Sorry, I missed this. My 2 cents is that the config entries are not very logical as I commented earlier.  "admin mode" does not tell much that it's about how authorization is configured ("admin auth mode" would already help a lot). The use of "token" matches with the current functionality well but is it a good fit for other ways to authorize admins.

kbs/docs/admin.md
fitzthum: I guess I might be missing the issue here, but to me `TokenAuthorization` makes sense. This mode would support anything that uses a token, which is basically everything.

Xynnn007:
> Sorry, I missed this. My 2 cents is that the config entries are not very logical as I commented earlier. "admin mode" does not tell much that it's about how authorization is configured ("admin auth mode" would already help a lot). The use of "token" matches with the current functionality well but is it a good fit for other ways to authorize admins.

So the suggestion is to change
```toml
[admin]
mode = xxx
```

to 

```toml
[admin]
authrozation_mode = xxx
```
- or something else?

kbs/docs/admin.md
mythi: Only trying think how to make the config entries intuitive for the users. And keep them so if new functionality is added.

I believe the top-level functionality only about enabling authorization:

```
[admin]
endpoint_authorization(_mode)  = InsecureAcceptAll, DenyAll, Authorize(...)
```

and then, can we say `Authorize` takes `Config` and `Rules`.

mythi:
> So the suggestion is to change

brainstorming options here https://github.com/confidential-containers/trustee/pull/1239#discussion_r3128631574

kbs/docs/admin.md
Xynnn007: Hey @fitzthum could you help to get a better name consistently with @mythi ? In a way I know what the entries mean as I designed it : / 
 
Therefore, I cannot prejudge what the configuration items mean like someone configuring for the first time.

fitzthum: `authorization_mode` seems fine to me. `endpoint` seems redundant. I think we use the plain `mode` field in a number of other config blocks, but maybe those are worth revisiting also.

Xynnn007: I changed to `authorization_mode`.

mythi: what about the others, like making the configs all verbs.

Xynnn007: So `AuthorizeWithToken`?

mythi: If everything is a token, then that part sounds redundant too. However, based on the current implementation it feels to me that adding new auth is simpler at `AdminConfig` level by using `AuthorizeWithXyz` (e.g., `Jwt` for now) because then each mode can have their own config structs defined that suits the mode language and helps to make `token_verifier` simpler (`type` becomes redundant). Thougths on this?

kbs/docs/admin.md
mythi: Checking my old feedback and I had missed your response here. My question was why do we need `http`?

kbs/docs/admin.md
Xynnn007: Sry @mythi I did not get the point quite well. Could you give an expected shape in toml - or some examples?

kbs/docs/admin.md
Xynnn007: Let me take an example - when using self-deployed authentik-like 3rd authN services, it provides a way to get the token verification keys.

kbs/docs/admin.md
mythi: The example I had in mind was: `authorization_mode = AuthJwt`, can use `[admin.identity_providers]`and `AuthFoo` can have its own. But I realized this can be problematic when we need both enabled. How the current implementation scales to this use-case?

kbs/docs/admin.md
mythi: would it make more sense to not allow `http` but keep the option allow untrusted certs

Xynnn007:
@fitzthum @mythi changed the config item like this

```json
{
            "authorization_mode": "AuthenticatedAuthorization",
            "authorizer": {
                "type": "RegexAcl",
                "acls": [{
                    "role": "Alice",
                    "allowed_endpoints": "^/kbs/v0/admin/allowed/.+$"
                }]
            },
            "authentication": {
                "type": "BearerJwt",
                "identity_providers": [{
                    "issuer": expected_issuer,
                    "audience": expected_audience,
                    "public_key_uri": public_key_path
                }]
            }
        }
```

fitzthum:
> changed the config item like this

I think I liked the old one a little bit more, but mainly I just want to get this merged. 

As nits `AuthenticatedAuthorization` seems too general.

Also `authorizer` and `authentication` are different parts of speech. I would maybe call it `authorizer` and `authenticator` or `authorization` and `authentication`.

mythi:
> > changed the config item like this
> 
> I think I liked the old one a little bit more, but mainly I just want to get this merged.
>
> As nits `AuthenticatedAuthorization` seems too general.
>

Any suggestions? My feedback has been to avoid making it too tied to Jwt if other modes or types are added later without breaking users again but yesterday I learnt that these won't be implemented in Trustee to keep things simple. With than, something like `AuthorizeJwt` (verb) would work too I guess. Perhaps the `type` carries redundant information too that we don't need to ask users to provide. :thinking:

fitzthum:
> Any suggestions? 

I liked `TokenAuthorization`

Xynnn007:
Ok folks.  I asked AI to analyze the design structure. The feed back is that
1. Now we have three auth modes: (block nothing) allowAll, (block everything) denyAll, (block something) Now it's called `AuthenticatedAuthorization/TokenAuthorization`. The three modes can cover almost all scenarios we (will) meet.
2. The hardest part is define "block something" part. My idea is that this part an be abstractly divided into two parts
    a. identify who you are
    b. judge what you can do
    For A, the term is authentication (authN). We did discuss that KBS does not need to implement too complex authN & authZ parts. But if we need a simple auth system, we need to at least know who you are. This includes things like a JWT inside HTTP header (in this PR), a KEY/TOKEN inside HTTP header, a sub-field in the request body, etc. Note that we still did not care about how to provision the token/identifier to the client in KBS, this is also authN but out of KBS's responsibility. The "who you are" question of KBS cares about where the "identifier" can be found and what kind of the "identifier", thus we might have different enums, like `BearerJwt` (means from Bearer header, and it's a JWT)
    For B, the term is authorization (authZ). The workflow is to judge whether the access is acceptable based on the "who you are". So to keep the ability we have before, I added `RegexAcl` in this PR.

So I choose to rename to `AuthenticatedAuthorization` to mean that this mode is to "block something". Maybe `Enforcement` can be a good name? My suggestion is that let's just find the names for current
- `AuthenticatedAuthorization`: Maybe `Enforcement` vs `AuthenticatedAuthorization` vs something else?
- `authorizer` and `authentication`: I'd like to follow Tobin's `authorization` and `authentication` idea.

kbs/docs/admin.md
mythi: @Xynnn007 ^

Xynnn007: Still think that `http://` is good for dev cases. I can highlight the risk for `http`, but no need to delete it

fitzthum:
Idk if I agree entirely. The three modes are really allow all, deny all, or allow with an admin token with a particular role/acls. In theory we could add more modes in the future that would support other types of fine-grained access. That's why I like having token in the mode name i.e. TokenAuthorization. 

But either way I think we should just merge the PR. Let's not get stalled on the last tiny detail.

Xynnn007:
Ok. Adopted `authorization` and `authentication` idea. 

@mythi please, check if the current shape is good to you.

mythi:
I guess nobody is quite happy. :-) My main comment was that this setup can be extended without breaking users (with this PR included we get a hat trick of third consecutive release with user breaking config). If that can be achieved, then it's good.

Xynnn007:
When discussing changes in user experience, the most crucial aspect is configuring the `[admin]` section in the KBS configuration file.

before this PR, we configure
```toml
[admin]
# Do not use this in production.
type = "InsecureAllowAll"
```

after this, we have
```toml
[admin]
authorization_mode = "InsecureAllowAll"
```

The **ONLY** change I see is `type` -> `authorization_mode`. So do you mean we'd better change back to `type`?

About the previous `Simple` auth way. I think we've been discussing this for a long time regarding how to implement auth. We've been using public and private keys for admin auth for a long time, but frankly, that's not ideal; using tokens is much better and this is what we have discussed. And the current plugin design that's compatible with tokens is a better approach - if admin auth is still needed for KBS.

As mentioned several times before, it's best to avoid having KBS handle authentication-related tasks and instead outsource these to external gateway services. This approach remains recommended; use InsecureAllowAll and then add an out-of-band authentication service.

mythi:
> The **ONLY** change I see is `type` -> `authorization_mode`. So do you mean we'd better change back to `type`?

`authorization_mode` is good. I meant that hopefully the future changes are purely additive and the structure serves new types without breakages.

tools/kbs-client/src/main.rs
```diff
@@ -89,11 +90,10 @@ struct Config {
...
     command: ConfigCommands,
 
-    /// PEM file path of private key used to authenticate the resource registration endpoint token (JWT)
-    /// to Key Broker Service. This key can sign legal JWTs.
-    /// This client tool only support ED22519 key now.
+    /// Optional admin bearer token file path.
+    /// If omitted, client will try ~/.trustee/admin-token, then fallback to anonymous requests.
     #[clap(long, value_parser)]
-    auth_private_key: PathBuf,
+    admin_token_file: Option<PathBuf>,
```
fitzthum, line 97: btw can we add some kind of deprecation warning for the old arg? maybe keep the old arg but print a warning if anyone uses it.

Xynnn007: Good point. Let me go furthur.

The goal is to inform the user that the current configuration will not take effect. Could I do this by mandating that the input configuration format must conform to certain fields, otherwise throwing an error and referencing a configuration file format documentation?

Because if we set a precedent here—meaning that every time the configuration file is modified, we need to record historical configuration items in the code—we'll need to do the same elsewhere in the future. This will result in a lot of redundant code. My suggestion is to avoid saying "which are no longer supported" and instead directly say "the current format is as follows, there are configuration errors, startup failed, and modifications are needed."

mythi:
Per request, I'm summarizing my feedback that I've tried to ask/comment on and I believe are still open:

- `http://` seems useless and exposes complexity (insecure boolean). An argument was made that it's nice for development purposes but that is possible by doing to manual download and to use`file://`. We haven't supported `http://` earlier either: b6fa7ca879fbd8f2c8f7c237e840c2f763dfd15e says it's a extraction of old functionality but that is also not true, it adds new logic (`http://`) that gets hidden in the history. These should be separated at minimum.
- the config entries users need to set: users have already opted-in to do (JWT) authn/authz by selecting the mode. Since `type = "BearerJwt"` and `type = "RegexAcl"` have no alternatives, it seems strange these need to be set at all. Can these be hidden from users? Do you see any benefit/use-case for adding a separate type entry since it looks the required config entry is type specific. How about letting users to simply specify:

```toml
[[admin.authorization.regex_acls]]
role = "admin1"
allowed_endpoints = "^/kbs/v0/resource/.+$"
...
```

Looks like 612acdcc97b7c0fd8387636f53c797fa43ffca7c does not belong to this PR

Xynnn007:
> * `http://` seems useless and exposes complexity (insecure boolean). An argument was made that it's nice for development purposes but that is possible by doing to manual download and to use`file://`. We haven't supported `http://` earlier either: [b6fa7ca](https://github.com/confidential-containers/trustee/commit/b6fa7ca879fbd8f2c8f7c237e840c2f763dfd15e) says it's a extraction of old functionality but that is also not true, it adds new logic (`http://`) that gets hidden in the history. These should be separated at minimum.

Some more context for https://github.com/confidential-containers/trustee/pull/1239#discussion_r3145222329. For authentik like services, they will expose a `.well-known` http(s) api for clients to get the public key which is used to verify the token provisioned themselves. Thus, a "warning" supporting would be help for such test case. Note that KBS server now supports `insecure_http` option for non-HTTPS server also.




> ```toml
> [[admin.authorization.regex_acls]]
> role = "admin1"
> allowed_endpoints = "^/kbs/v0/resource/.+$"
> ...
> ```

Note that if we use this kind of configuration, it hints that the only supporting mode for authorization is `BearerJwt` and `RegexAcl` as there is no field to specify "where the auth token is got" and "how to do filter logic based on auth token". Please ensure this is the expectation.



> Looks like [612acdc](https://github.com/confidential-containers/trustee/commit/612acdcc97b7c0fd8387636f53c797fa43ffca7c) does not belong to this PR

I once met CI terbulance and this commit fixes. I can bring it out to a separate PR.

mythi:
> Note that if we use this kind of configuration, it hints that the only supporting mode for authorization is `BearerJwt` and
> `RegexAcl` as there is no field to specify "where the auth token is got" and "how to do filter logic based on auth token".

Ah, it was just an example for authorization. I left authentication out but without type for authentication, it could be

```toml
[[admin.authentication.jwt_identity_providers]]
issuer = "authentik"
audience = "kbs-admin"
jwk_set_uri = "https://auth.example.com/application/o/kbs/jwks/" 
```
or something like that. Unless you can think of some use for `type`.

Xynnn007:
> ```toml
> [[admin.authentication.jwt_identity_providers]]
> issuer = "authentik"
> audience = "kbs-admin"
> jwk_set_uri = "https://auth.example.com/application/o/kbs/jwks/" 
> ```

Does this example hint that `Bearer` is the only way to get JWT and not configurable?

Note that the example uses `.jwt_identity_providers` suffix. This means that all authentication materials are JWT - if we use a token than JWT, how can the config be set?

mythi:
> Does this example hint that `Bearer` is the only way to get JWT and not configurable?

Not sure I understand the question. There can be multiple table entries just like in the implementation currently.

> This means that all authentication materials are JWT - if we use a token than JWT, how can the config be set?

Answering with a question: do you think these other tokens would be able to use `identity_providers` config structure.

Xynnn007:
> Not sure I understand the question. There can be multiple table entries just like in the implementation currently.

This is similar to unified storage configuration - we could enumerate different authentication's configs, but we still need a place to tell "which one to use". 
Or - do you mean that if `[admin.authorization.regex_acls]` given, means that `regex_acls` is used. If both `[admin.authorization.regex_acls]` and `[admin.authorization.B]` is given, there will raise an error, or have some 
priority order?






> > This means that all authentication materials are JWT - if we use a token than JWT, how can the config be set?
> 
> Answering with a question: do you think these other tokens would be able to use `identity_providers` config structure.

Let me explain my question. Suppose we have two authentication types, like "Jwt in BearerHeader", and "Token in AnotherHeader". How can we choose which one to use? Note that `jwt_identity_providers` could still needs context about "where the token is got"

mythi:
OK got it. Well, my original comment tried to say we could add `type` later if there are multiple options to choose from (or have another mechanism if multiple options can be enabled simultaneously).

Xynnn007:
> OK got it. Well, my original comment tried to say we could add `type` later if there are multiple options to choose from (or have another mechanism if multiple options can be enabled simultaneously).

The current plug-in design uses a `type` field to select plugins. The advantage of this is that even if new plugins are added later, existing user configurations don't need to be changed. 

If we implemented the way you proposed first and then added plugins design as more modes are added later, the risk would be that the user interface would need to be changed in future.

mythi:
As I see it, the absence of `type` if new plugins are added can be handled via defaults. Do you see other reasons why this could be a breaking change? `type` has also the downside that only one can be enabled at a time but I've shared use-cases were multiple types could be enabled.

tools/kbs-client/src/main.rs
fitzthum: Yeah, I think that's fine. I was just thinking about this after the recent conversation regarding the version pinning. Once we merge this, there is potential for the website to have the wrong directions.

Xynnn007: Nice catch. There might be two ways 

1. We are in the cycle between 0.20.0~0.21.0, so code could have the link to tag 0.21.0
2. We have the first commit to change the document, and another commit to link with the ref of the previous commit.

Xynnn007:
> As I see it, the absence of `type` if new plugins are added can be handled via defaults. Do you see other reasons why this could be a breaking change? `type` has also the downside that only one can be enabled at a time but I've shared use-cases were multiple types could be enabled.

Good. Let me drop the type field of the user interface and by default only support `[admin.authentication.bearer_jwt]` and `[admin.authorization.regex_acl]` sections. I think in future if we bring `type` back, there would be some logic change inside the configuration blackbox but the modifications upon the user configuration file can be very little.

tools/kbs-client/src/main.rs
Xynnn007: Let me draft a PR different from this one.

Xynnn007:
@mythi 
> * `http://` seems useless and exposes complexity (insecure boolean). An argument was made that it's nice for development purposes but that is possible by doing to manual download and to use`file://`. We haven't supported `http://` earlier either: [b6fa7ca](https://github.com/confidential-containers/trustee/commit/b6fa7ca879fbd8f2c8f7c237e840c2f763dfd15e) says it's a extraction of old functionality but that is also not true, it adds new logic (`http://`) that gets hidden in the history. These should be separated at minimum.

I added the part to tell users in the config README.md and raise warning in the terminal when users actually uses this.



> * the config entries users need to set: users have already opted-in to do (JWT) authn/authz by selecting the mode. Since `type = "BearerJwt"` and `type = "RegexAcl"` have no alternatives, it seems strange these need to be set at all. Can these be hidden from users? Do you see any benefit/use-case for adding a separate type entry since it looks the required config entry is type specific. How about letting users to simply specify:

Get to the shape like

```toml
[admin]
authorization_mode = "AuthenticatedAuthorization"

[admin.authentication.bearer_jwt]
identity_providers = [
    { issuer = "TrusteeInDocker", public_key_uri = "/opt/confidential-containers/kbs/user-keys/public.pub" },
]

[admin.authorization.regex_acl]
acls = [{ role = "admin", allowed_endpoints = "^/kbs/.+$" }]
```

and we can extend it later if new things come.


I think this will cover all the worries you raised. If there is no blockers until this weekend, I'd like to get this merged as it took a long time to bring in this feature.

mythi:
> @mythi
> 
> > * `http://` seems useless and exposes complexity (insecure boolean). An argument was made that it's nice for development purposes but that is possible by doing to manual download and to use`file://`. We haven't supported `http://` earlier either: [b6fa7ca](https://github.com/confidential-containers/trustee/commit/b6fa7ca879fbd8f2c8f7c237e840c2f763dfd15e) says it's a extraction of old functionality but that is also not true, it adds new logic (`http://`) that gets hidden in the history. These should be separated at minimum.
> 
> I added the part to tell users in the config README.md and raise warning in the terminal when users actually uses this.

My suggestion is to drop it from this PR and moved to a PR of its own. It needs to be moved out from https://github.com/confidential-containers/trustee/pull/1239/commits/a19e96b0e2cd6972c41997586f047e14b641e77f anyway (that commit is just moving existing functionality). It also helps to unblock this PR.

Xynnn007:
@mythi removed and I will put another PR for it.

mythi (approved):
I don't have anything to add so adding LGTM. Maybe @fitzthum wants to take another look before merging since things got changed since the previous approval

kbs/docs/cluster.md
```diff
@@ -16,19 +16,32 @@ Users can use very simple command to:
...
+
 ```bash
-docker compose up -d
+kbs-client config \
+  --url http://127.0.0.1:8080 \
+  --admin-token-file kbs/config/docker-compose/admin-token \
+  set-resource-policy --allow-all
 ```
 
+If `--admin-token-file` is omitted, `kbs-client` tries `~/.trustee/admin-token`,
```
fitzthum: Maybe we should make the default path match what is setup via docker compose?

deps/verifier/src/se/README.md
```diff
@@ -100,11 +100,7 @@ sockets = ["0.0.0.0:8080"]
 insecure_http = true
 
 [admin]
-type = "Simple"
-
-[[admin.personas]]
-id = "admin"
-public_key_path = "/kbs/kbs.pem"
+authorization_mode = "InsecureAllowAll"
```
fitzthum: Seems like this is not equivalent to what is currently in the docs. We should probably be using token-based auth here.

fitzthum (approved):
LGTM

A couple little nits, but I think we are ready here

kbs/docs/cluster.md
Xynnn007: let's have multiple tries. first (relative path) for docker-compose. then this absolute path.

PR #1245, opened 2026-03-24
ci: preserve package-write scope in reusable image workflows

Xynnn007:
Reusable workflows were forcing contents-read at the top level, which can narrow inherited GITHUB_TOKEN scopes and block GHCR organization package pushes on push-enabled paths. Relying on caller/job permissions keeps least privilege for PRs while allowing package publication only where explicitly granted.

Fixes errors like

https://github.com/confidential-containers/trustee/actions/runs/23468514542/job/68285934012

### DIFF
--- a/.github/workflows/workflow-call-build-staged-images.yml
+++ b/.github/workflows/workflow-call-build-staged-images.yml
@@ -18,9 +18,6 @@ on:
         description: 'GitHub token for GHCR'
         required: false 
 
-permissions:
-  contents: read
-
 jobs:
   validate-inputs:
     runs-on: ubuntu-24.04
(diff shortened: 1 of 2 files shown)


mkulke (commented):
I don't think it's a good idea to remove explicit permissions, we don't use zizmor or other static analysis tools for workflows in trustee yet, but they will probably flag this. If those reusable workflows require permissions we should enumerate them, ideally on the job level:

Image

fitzthum (approved):
LGTM

mythi:
I checked the scorecard results yesterday and now with the latest. This seems to give regression and drops the score by 0.5 (zero points from "Token Permissions").

The check:
> The highest score is awarded when the permissions definitions in each workflow's yaml file are set as read-only at the top level and the required write permissions are declared at the run-level. 

is no longer met with:
```
Warn: no topLevel permission defined: .github/workflows/workflow-call-build-staged-images.yml:1
Info: topLevel 'contents' permission set to 'read': .github/workflows/workflow-call-kbs-e2e.yml:30
Warn: no topLevel permission defined: .github/workflows/workflow-call-publish-staged-manifests.yml:1
```

PR #1248, opened 2026-03-25
workflows: Set timeout for KBS lint/format/test

BbolroC:
The `Run KBS/lint/format/test` step was observed running longer than 1 hour. The default timeout for a single workflow step is 6 hours, while the expected runtime for this step is under 5 minutes. Set an explicit timeout accordingly.

Example run: https://github.com/confidential-containers/trustee/actions/runs/23529279959/job/68489297233


Xynnn007:
cc @fitzthum looks like the integration test has fallen into an endless dream

PR #1251, opened 2026-03-26
Trustee Stateless work 3: Change Attestation Session to storage backend.

Xynnn007:
The last code part of https://github.com/confidential-containers/trustee/issues/1092

This patch mainly decouple the kbs attestation session status to storages. This is the last part (let's ignore admin configs for a while) of the KBS's state inside an instance to be decoupled.

Usually, the session state are of short liveness, so we define another overwriting item for it to choose a different storage backend. 

In future, we could add support for redis or so. This will help multiple have a common redis instance for HA.

After this PR, I am planning to perfect the helm chart deployment one.

Vibe coding with help of cursor.

### DIFF
--- a/kbs/src/attestation/backend.rs
+++ b/kbs/src/attestation/backend.rs
@@ -15,7 +15,7 @@ use semver::{BuildMetadata, Prerelease, Version, VersionReq};
 use serde::Deserialize;
 use serde_json::json;
 use std::collections::HashMap;
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 
 use crate::attestation::session::{SessionMap, KBS_SESSION_ID};
 use crate::prometheus::{
@@ -180,9 +180,38 @@ impl AttestationService {
             .await
             .map_err(|e| Error::SessionStorageInitialization { source: e })?;
 
-        let session_map = SessionMap::new(session_storage_backend);
-
-        // TODO: implement session storage cleanup
+        let session_map = SessionMap::new(session_storage_backend.clone());
+        // Start background cleanup of expired session records in the `kbs-session` namespace.
+        {
+            let cleanup_session_map = session_map.clone();
+            // Run periodic cleanup every minute when healthy.
+            let cleanup_interval = ::tokio::time::Duration::from_secs(60);
+            // Use exponential backoff on failures to avoid noisy retry loops during outages.
+            let cleanup_retry_initial_interval = ::tokio::time::Duration::from_secs(5);
+            let cleanup_retry_max_interval = ::tokio::time::Duration::from_secs(300);
+            ::tokio::spawn(async move {
+                let mut retry_interval = cleanup_retry_initial_interval;
+                loop {
+                    match cleanup_session_map.cleanup_expired().await {
+                        Ok(()) => {
+                            retry_interval = cleanup_retry_initial_interval;
+                            ::tokio::time::sleep(cleanup_interval).await;
+                        }
+                        Err(err) => {
+                            warn!(
+                                ?err,
+                                retry_secs = retry_interval.as_secs(),
+                                "failed to clean up expired session records"
+                            );
+                            ::tokio::time::sleep(retry_interval).await;
+                            retry_interval = retry_interval
+                                .saturating_mul(2)
+                                .min(cleanup_retry_max_interval);
+                        }
+                    }
+                }
+            });
+        }
         Ok(Self {
             inner,
             timeout: config.timeout,
@@ -282,6 +311,7 @@ impl AttestationService {
             debug!("Session ID {}", session.id());
 
             if session.is_expired() {
+                let _ = self.session_map.storage.delete(session_id).await?;
                 bail!("session expired.");
             }
 
--- a/kbs/src/attestation/session.rs
+++ b/kbs/src/attestation/session.rs
@@ -143,6 +143,25 @@ impl SessionMap {
         }
         Ok(Some(session))
     }
+
+    pub async fn cleanup_expired(&self) -> Result<()> {
+        let keys = self.storage.list().await?;
+        for key in keys {
+            let Some(value) = self.storage.get(&key).await? else {
+                continue;
+            };
+
+            let Ok(session) = serde_json::from_slice::<SessionStatus>(&value) else {
+                continue;
+            };
+
+            if session.is_expired() {
+                let _ = self.storage.delete(&key).await?;
+            }
+        }
+
+        Ok(())
+    }
 }
 
 #[cfg(test)]


kbs/src/attestation/session.rs
```diff
@@ -99,18 +107,90 @@ impl SessionStatus {
...
+            };
+
+            let Ok(session) = serde_json::from_slice::<SessionStatus>(&value) else {
+                continue;
+            };
+
+            if session.is_expired() {
+                let _ = self.storage.delete(&key).await?;
+            }
         }
```
Xynnn007, line 161: Note that we use a O(n) complexity algorithm to clean up the expired sessions. I want to have a following PR to let the KeyValueStorage have a TTL configuration. It will help to decrease the algorithm time-space complexity.

kbs/src/api_server.rs
```diff
@@ -106,6 +110,13 @@ impl ApiServer {
         #[cfg(feature = "as")]
         let attestation_service = crate::attestation::AttestationService::new(
             config.attestation_service.clone(),
+            &config.session_storage_type.unwrap_or_else(|| {
+                info!(
+                    "Session storage type not configured, using storage backend type: {:?}",
+                    config.storage_backend.storage_type
```
Xynnn007, line 116: Note that we can configure the session storage different from the unified one here. And if it is not configured, it will align with the unified one. I am not sure if this is the best.
For example, users want to use localfs/localjson storage, and the sessions will also be stored to it, which might cause performance decrease and lock issue. I wonder if we should set this to `Memory` by default if not set?

kbs/src/attestation/backend.rs
```diff
@@ -9,26 +9,28 @@ use anyhow::{anyhow, bail, Context};
...
 };
 
 use super::{
     config::{AttestationConfig, AttestationServiceConfig},
-    session::{SessionMap, SessionStatus},
+    session::SessionStatus,
     Error, Result,
 };
 
+const KBS_SESSION_STORAGE_NAMESPACE: &str = "kbs-session";
```
fitzthum: nit: maybe `kbs-protocol-session` is slightly clearer.

kbs/src/config.rs
```diff
@@ -81,6 +81,13 @@ pub struct KbsConfig {
     #[serde(default)]
     pub storage_backend: StorageBackendConfig,
 
+    /// Optional session storage type used for attestation session state.
+    ///
```
fitzthum, line 85: Maybe can make this a little more clear by saying something like 

"KBS protocol state will be stored in system-wide key value storage by default. Use this field to specify an additional storage backend just for KBS protocol session state."

kbs/src/attestation/backend.rs
```diff
@@ -180,9 +180,38 @@ impl AttestationService {
...
 
-        let session_map = SessionMap::new(session_storage_backend);
-
-        // TODO: implement session storage cleanup
+        let session_map = SessionMap::new(session_storage_backend.clone());
+        // Start background cleanup of expired session records in the `kbs-session` namespace.
+        {
+            let cleanup_session_map = session_map.clone();
+            // Run periodic cleanup every minute when healthy.
+            let cleanup_interval = ::tokio::time::Duration::from_secs(60);
```
fitzthum, line 188: I wonder if this should be configurable. 60s might be too frequent for simple setups.

Xynnn007: The original timeout is 60s and has been working fine. https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/backend.rs#L176-L187

For now it is a workaround to clean up using the loop here. I prefer to provide expiration as part of the kv store and then we can delete this part in future. So probably we can still leave it as-is?

Xynnn007:
@fitzthum Thanks for the comments. Not sure I understand correctly - the two issues may be resolved altogether? e.g. we can bring a `ttl` option to key-value storage's `SetParameters`. Then it can be deleted background.

kbs/src/attestation/backend.rs
fitzthum: That's fine.

kbs/src/attestation/backend.rs
```diff
@@ -171,20 +174,44 @@ impl AttestationService {
...
-                    session_map_clone
-                        .sessions
-                        .retain_async(|_, v| !v.is_expired())
-                        .await;
+                    match cleanup_session_map.cleanup_expired().await {
+                        Ok(()) => {
+                            retry_interval = cleanup_retry_initial_interval;
+                            ::tokio::time::sleep(cleanup_interval).await;
+                        }
+                        Err(err) => {
```
mkulke, line 200: when can this happen? and if it's entirely unexpected should we panic?

kbs/src/attestation/session.rs
```diff
@@ -99,18 +107,90 @@ impl SessionStatus {
...
+
+    pub async fn insert(&self, session: SessionStatus) -> Result<()> {
+        let session_bytes = serde_json::to_vec(&session)?;
+        let session_id = session.id();
+        let _ = self
+            .storage
+            .set(
+                session_id,
+                &session_bytes,
+                SetParameters { overwrite: true },
```
mkulke, line 128: is there a legitimate reason why insert for a session would be called twice?

kbs/src/attestation/session.rs
```diff
@@ -99,18 +107,90 @@ impl SessionStatus {
...
+    }
+
+    pub async fn get(&self, session_id: &str) -> Result<Option<SessionStatus>> {
+        let session_bytes = self.storage.get(session_id).await?;
+        let Some(session_bytes) = session_bytes else {
+            return Ok(None);
+        };
+        let session: SessionStatus = serde_json::from_slice(&session_bytes)?;
+        if session.is_expired() {
+            let _ = self.storage.delete(session_id).await?;
```
mkulke, line 141: should we avoid performing a side-effect like this and maybe have a discrete cleanup look that will purge expired objects?

as a caller of this function I wouldn't necessarily expect a get() function like this to perform a destructive option

kbs/src/attestation/backend.rs
Xynnn007: It will happen when any call in `list_all`, `get`, `delete` fails. My advice is not to panic. Even if expired sessions are not cleaned up, it will not disrupt the normal logic of KBS. Retries and alerts are sufficient. If the core logic—which relies on interactions with the KV store—were to fail, it would result in a panic or an error-driven exit at that specific point.

kbs/src/attestation/session.rs
Xynnn007: Oh. This will only occur when a collision takes place. This is the same as current [code in tree](https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/session.rs#L114)

kbs/src/attestation/session.rs
Xynnn007: Yes, this is a workaround. We are essentially performing a scan on one hand, while simultaneously removing expired sessions whenever a session is retrieved. I plan to add a TTL option for key-value objects later on; once implemented, both the dedicated thread currently responsible for session cleanup and the deletion logic implemented here can be removed.

mkulke: it would be good to add an issue for this so, so we don't forget this

Issue #1206, opened 2026-03-05
Docker compose doesn't work on ARM


fitzthum:
It looks like the coco-keyprovider binary is not supported on ARM, so docker compose explodes on startup on that platform. We don't really even use the cooc-keyprovider. Maybe we should just delete that part, or fix the build (which is actually in guest-components).

Xynnn007:
We can close this as now coco keyprovider is not part of trustee

mythi:
If I remember correctly we agreed to keep this open until the image build is fixed. Is it working now?

Xynnn007:
@mythi Oh. Time flies. I forgot the thread and here no context in the dialogue :sweat_smile:. Let me create a new issue under GC to explicit show the issue.

PR #1256, opened 2026-03-27
docker-compose: remove keyprovider container

Xynnn007:
Now the image encryption follows the guide
https://confidentialcontainers.org/docs/features/encrypted-images/ and the key registration to KBS relies on kbs-client, than directly the keyprovider.

Let's delete this part if we ensure that no one uses this.

cc @fitzthum @bpradipt

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -70,30 +70,6 @@ services:
       "0.0.0.0:50003"
     ]
 
-  keyprovider:
-    image: ghcr.io/confidential-containers/coco-keyprovider:latest
-    environment:
-      - RUST_LOG
-    restart: always
-    ports:
-      - "50000:50000"
-    volumes:
-      - ./kbs/config:/opt/confidential-containers/kbs/user-keys
-    command: [
-      "coco_keyprovider",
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -70,30 +70,6 @@ services:
       "0.0.0.0:50003"
     ]
 
-  keyprovider:
-    image: ghcr.io/confidential-containers/coco-keyprovider:latest
```
mythi, line 74: removing `keyprovider` just hides the problem, right? The way I read #1206 is that the image is broken on ARM. Should we just remove the image also?

Xynnn007: Yes. For docker-compose of trustee, it's enough to remove the image.

But whether we need an arm version cocokeyprovider image still depends. No one up to now requires this and I think we can lazily impl it.

fitzthum: i would just keep the image for now. we can fix arm in the future or delete depending on requirements.

mythi: Works for me but maybe keep the issue open so that users know the image is broken.

PR #1259, opened 2026-03-31
Enable TLS 1.3 and add PQC TLS support

yafu-1:
Previously, the KBS server was configured for TLS 1.2 only. This change switches to mozilla_intermediate_v5, enabling support for both TLS 1.2 and TLS 1.3.

With TLS 1.3 available, PQC (post-quantum cryptography) hybrid key exchange groups can be enabled. The KBS server detects OpenSSL capabilities at runtime and selects the best supported PQC group.

If no PQC groups are available, the KBS server falls back to classical ECDHE groups.

---

## Details

- Switch to `mozilla_intermediate_v5` to allow TLS 1.3
- Add runtime detection of PQC hybrid groups
- Select the first supported group from a prioritized list
- Fallback to classical groups if PQC is unavailable
- Cache detection result using `OnceLock`

## Testing

Tested with trustee-attester and OpenSSL 3.5.1:

- TLS 1.3 + PQC negotiation successful (X25519MLKEM768)
- TLS 1.3 uses classical key exchange algorithms as fallback when PQC groups are unavailable due to system/OpenSSL configuration
- TLS 1.2 fallback works as expected

### DIFF
--- a/attestation-service/src/bin/restful-as.rs
+++ b/attestation-service/src/bin/restful-as.rs
@@ -177,7 +177,7 @@ loglevel: {env_filter}
 
     let server = match (cli.https_prikey, cli.https_pubkey_cert) {
         (Some(prikey), Some(pubkey_cert)) => {
-            let mut builder = SslAcceptor::mozilla_modern(SslMethod::tls())?;
+            let mut builder = SslAcceptor::mozilla_intermediate_v5(SslMethod::tls())?;
 
             let prikey = tokio::fs::read(prikey)
                 .await


kbs/src/http.rs
```diff
@@ -19,9 +69,25 @@ pub fn tls_config(config: &HttpServerConfig) -> Result<openssl::ssl::SslAcceptor
...
+    let mut builder = SslAcceptor::mozilla_intermediate_v5(SslMethod::tls())?;
     builder.set_private_key_file(key_file, SslFiletype::PEM)?;
     builder.set_certificate_chain_file(cert_file)?;
 
+    // Auto-detect and configure TLS groups based on OpenSSL capabilities
+    let groups = if let Some(pqc_group) = detect_best_pqc_group() {
+        tracing::info!("PQC TLS enabled");
+        format!("{}:{}", pqc_group, CLASSICAL_GROUPS)
+    } else {
+        tracing::info!("PQC TLS not available, using classical algorithms");
```
Xynnn007: Do we need to mention the algorithm name?

kbs/src/http.rs
```diff
@@ -19,9 +69,25 @@ pub fn tls_config(config: &HttpServerConfig) -> Result<openssl::ssl::SslAcceptor
...
+    // Use mozilla_intermediate_v5 which supports both TLS 1.2 and TLS 1.3
+    // Note: mozilla_intermediate (v4) explicitly disables TLS 1.3 via NO_TLSV1_3 flag
+    // Version 5 is required for TLS 1.3 and PQC support
+    let mut builder = SslAcceptor::mozilla_intermediate_v5(SslMethod::tls())?;
     builder.set_private_key_file(key_file, SslFiletype::PEM)?;
     builder.set_certificate_chain_file(cert_file)?;
 
+    // Auto-detect and configure TLS groups based on OpenSSL capabilities
+    let groups = if let Some(pqc_group) = detect_best_pqc_group() {
+        tracing::info!("PQC TLS enabled");
```
Xynnn007: Let's `use tracing::{info, warn};` in the front of the file and directly use `info!` here. So do other parts.

Xynnn007:
btw you might need to fight with the CI error for a while ; )

kbs/src/http.rs
yafu-1: Good point!  Updated to include the algorithm name.

kbs/src/http.rs
yafu-1: Done, thanks for the review!

PR #1280, opened 2026-04-15
rvps: add corim extractor

fitzthum:
CoRIM is one of the most prominent formats for reference values. Add an extractor that parses reference values from CoRIMs.

CoRIM can do many different things. This is a first iteration that does not support the entire spec. Specifically, the extractor only cares about reference triples, which probably makes sense for the RVPS.

Within those reference triples, the extractor only knows how to handle certain types of claims. This is enough to parse the various CoRIMs that I used for testing and can easily be extended in the future.

Since CoRIMs are quite flexible, it's hard to create a general rule for how reference values should be named. This PR creates a URI-style name from whatever identifiers are found. In the future we may want to create a more explicit/configurable mapping from CoRIM to RV name, but let's start with something simple.

Finally, since the CoRIM spec is still in flux, there can be mismatches between different versions and different implementations. I have added one manual workaround to handle CoRIMs that start with unexpected ASN tags (which are fairly common). Some older CoRIMs might not be supported, but we can potentially add a workaround for this in the future.

oh and no support for checking the corim signature yet


fitzthum:
Hmm looks like we are running into this interesting compiler overflow https://github.com/bootc-dev/canon-json-rs/issues/16. Might try switching to another canonicalization crate. This one doesn't seem that great.

corim-rs needs the `arbitrary-precision` feature, which causes `canon-json` to explode.

We need to pickup http://github.com/confidential-containers/guest-components/pull/1416 from gc before this will pass all the tests. Otherwise we will hit problems when trying to build the whole workspace which includes kbs-client and rvps.

Xynnn007:
The gc side PR seems red. Please fix that and I will take a look on this after that one gets merged.

fitzthum:
Ok, dependency issues finally sorted out.

rvps/Cargo.toml
```diff
@@ -25,13 +25,16 @@ required-features = ["bin"]
...
 base64.workspace = true
+hex.workspace = true
 chrono = { workspace = true, features = ["serde"] }
 clap = { workspace = true, optional = true }
 config.workspace = true
 key-value-storage.path = "../deps/key-value-storage"
 path-clean = { version = "1.0.1", optional = true }
 prost.workspace = true
 roxmltree = "0.21.1"
+ciborium = "0.2"
```
Xynnn007, line 36: This crate has not been updated for ~ 2 years. Do you have any other options?

rvps/src/extractors/corim/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+            // We will join these all together to make the rv path.
+            // For now, ignore the instance and group maps.
+            let mut base_rv_path_components = vec![];
+
+            // Collect environment information if it exists.
+            let env = &triple.ref_env;
+
+            if let Some(class) = &env.class {
+                // Only add stringlike fields to the path for now.
+                if let Some(v) = class.vendor.clone() {
```
Xynnn007:
```suggestion
                if let Some(v) = &class.vendor {
```

rvps/src/extractors/corim/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+                    base_rv_path_components.push(v.to_string())
+                }
+
+                if let Some(m) = class.model.clone() {
```
Xynnn007:
```suggestion
                if let Some(m) = &class.model {
```

rvps/src/extractors/corim/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+    Ok(rvs)
+}
+
+// Some CoRIMs start with unexpected CBOR tags.
+// If we find CBOR tags 500 or 502 at the start of the CoRIM,
+// remove them.
+//
+// Specifically, this might look like d901 f4d9 01f6
+// where 0xd9 denotes a CBOR tag and 0x01f4 are 0x01f6
+// the problematic tag numbers.
```
Xynnn007, line 130: Where can we get such "ill" CoRIMs? can we get some unit test cases.

rvps/src/extractors/corim/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+
+            if let Some(class) = &env.class {
+                // Only add stringlike fields to the path for now.
+                if let Some(v) = class.vendor.clone() {
+                    base_rv_path_components.push(v.to_string())
+                }
+
+                if let Some(m) = class.model.clone() {
+                    base_rv_path_components.push(m.to_string())
+                }
```
Xynnn007, line 49: Also we have `class_id`, `layer`, `index` field. How to let them join the game here to generate a rv path?

rvps/src/extractors/corim/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+                            let uuid = &val.0 .0.as_slice();
+                            let uuid = uuid::Uuid::from_slice(uuid)?.to_string();
+                            claim_rv_path_components.push(uuid);
+                        }
+                        _ => (), // todo: support other key types
+                    }
+                }
+
+                // todo: add non-string types
+                if let Some(n) = claim.mval.name.clone() {
```
Xynnn007:
```suggestion
                if let Some(n) = &claim.mval.name {
```

rvps/src/extractors/corim/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+                    claim_rv_path_components.push(n.to_string())
+                }
+
+                if let Some(s) = claim.mval.serial_number.clone() {
```
Xynnn007:
```suggestion
                if let Some(s) = &claim.mval.serial_number {
```

rvps/src/extractors/corim/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+                    let mut digest_rv_path_components = claim_rv_path_components.clone();
+                    digest_rv_path_components.push("digests".to_string());
+
+                    for digest in digests {
+                        let hash_bytes = digest.val.as_ref();
+                        let hash_hex = hex::encode(hash_bytes);
+                        rv_digests.push(hash_hex);
+                    }
+
+                    let name = digest_rv_path_components.join("/");
```
Xynnn007, line 93: Note that the CoMID structure definition is more complex than what we have as reference value URI. Although reference value URI s.t. `rvps://` is not used in this PR, the `/`-splited path querying keys share a same topology with RVPS URI.

So there must be a mechanism that we defined by ourselves here - how different fields of a CoMID can map to a reference value key/value in RVPS. Let's make a README.md under `corim` path. Probably we need another round review upon the design.

rvps/src/extractors/corim/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+                continue;
+            }
+        }
+        break;
+    }
+    slice
+}
+
+impl Extractor for CorimExtractor {
+    fn verify_and_extract(&self, provenance_base64: &str) -> Result<Vec<ReferenceValue>> {
```
bpradipt, line 152: I think a brief comment mentioning that currently unsigned CBORs are used will be useful.

rvps/src/extractors/corim/mod.rs
fitzthum: Most of the NVIDIA ones use this pattern. I can't add them as test cases because of licensing stuff, but I tested them out locally. I don't really like handling these tags, but note that the veraison cocli project does the same thing.

rvps/src/extractors/corim/mod.rs
fitzthum: I'm thinking we'll just add additional conditions for the other fields an add whatever is present to the id. For this PR I am only handling the string types, so I skipped it. Can add in the future or in this one if it's high priority.

rvps/src/extractors/corim/mod.rs
fitzthum: Yeah I don't think there is one general strategy for mapping CoRIMs to URIs. The CoRIM spec is quite big/flexible. Thus, I am taking the approach of collecting any identifiers that we find while traversing the CoRIM, and concatenating them all together to make the URI. I think this is a pretty good approach, especially for the first iteration, but if you have any other ideas lmk.

rvps/src/extractors/corim/mod.rs
fitzthum: Sure. I can add. This will do signed corims as well it just doesn't check the signature. ofc hope to add that in the future

rvps/Cargo.toml
fitzthum: Hmm yeah this crate was part of enarx. Looks like there are still some updates to the project but no releases. Let me ask around. I'm not sure there is a better option and the corim-rs crate relies on ciborium anyway.

rvps/src/extractors/corim/mod.rs
Xynnn007: In a way this looks not that "transparent". If we cannot handle it in this PR, it would be a good choice to raise an issue to record this and add the issue link note inside code.

rvps/src/extractors/corim/mod.rs
Xynnn007: Oh. I am not against your design. On the contrary, I like it.

 TBO in history when I read the CoRIM specification together with other RATS specifications I found it hard to map the CoRIM fields into a specific URI, and at least we need a strategy but we did not have that in the ecosystem. The same situation applied for the [TE](https://github.com/confidential-containers/trustee/issues/350). So I chose to delay and wait.

Now I am very happy that we have met the same situation. So a practical way is to
1. write the way of mapping to a `README.md` under CoRIM directory in this PR - we do not need it to be perfect and final in this PR - we cannot achieve the destination without process. But I must say that I love your concatenating way in this PR.
2. As you said kindly, we can make it better during our using it. Let's see

rvps/Cargo.toml
Xynnn007: Good. I also see that our old friend @mingweishih from Azure has built a crate named [corim](https://crates.io/crates/corim) very recently, which provides ser/de-ser functionalities for CoRIMs. Not sure if this will help you

rvps/src/extractors/corim/mod.rs
fitzthum: Yeah let me add a README for this extractor.

rvps/src/extractors/corim/mod.rs
fitzthum: Let me find a good reference for this. The cocli issue is [here](https://github.com/veraison/corim/pull/133), but there's not much information about the root cause.

Btw there are a few other issues with corim that I did not patch in this PR. For instance the spec gets updated with breaking changes and older corims might not be compatible. cocli has some other fixes for this (in a compatibility mode), but I skipped it for this one.

rvps/Cargo.toml
fitzthum: oh yes great to bump into @mingweishih again. Let me take a look at this. I think first iteration we may want to stick with corim-rs, which seems a little more mature, but I am open to switching.

The ciborium dep comes from supporting corims that have comids tagged as bytes rather than as comids. Not sure how this crate addresses that.

rvps/src/extractors/corim/mod.rs
fitzthum: Updated the comment. lmk if you had something else in mind.

rvps/src/extractors/corim/mod.rs
fitzthum: Added a readme.

rvps/src/extractors/corim/README.md
```diff
@@ -0,0 +1,34 @@
...
+* Environment map model field
+* MKey String, integer, or Uuid
+* Mval name
+* Mval serial number
+
+If the mval is a list of digests, the string "digests" will be added to
+the end of the URI, and the value will be a list of digests.
+
+If the mval is raw bytes, the string "raw_bytes" will be added to the end
+of the URI and the value will be the bytes converted to hex.
```
Xynnn007, line 34: Let's also mention that the parts are concat-ed with `/`, and provide some examples

Xynnn007 (approved):
LGTM! Thanks @fitzthum only one little thing left

Xynnn007:
you might need a rebase

rvps/src/extractors/corim/README.md
fitzthum: Added

fitzthum:
@mythi ptal

rvps/Cargo.toml
mingweishih: Same here! Glad we come across again. We're actively working on this new corim project because there are lots of demands in CC space but there is no crate from crate.io (we're aware of corim-rs but no idea about its release plan). Avoiding the dependency of ciborium is another goal (for the same reason of unknown maintainence status @Xynnn007 pointed out). Given corim only needs a subset of CBOR, we decided to implement an in-house se/deser. There is other goal like `no_std` support.

We do handle the comid tags in bstr that follows corim spec. Feel free to give it a try and let me know any feedback or issue.

rvps/src/extractors/corim/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    #[test]
+    fn extract_test_corim() {
+        tracing_subscriber::fmt()
+            .with_max_level(tracing::Level::DEBUG)
+            .with_test_writer()
+            .try_init()
+            .ok();
+
+        let cbor_bytes = include_bytes!("unsigned-example-corim.cbor");
```
mythi, line 202: you probably have the code that encoded this. would it make sense to make the test cases do that instead? Maybe this works for the broken ones too (requiring `strip_outer_tags()`).

mythi (approved):
One minor suggestion but not a blocker. Needs a rebase too.

This provides a good foundation. I'll need to see how intel POE can be supported once the spec matures. I can use this baseline to test things.

FWIW, on the corim crate topic. I'm currently using the "Azure" version to get POEs generated.

rvps/src/extractors/corim/mod.rs
fitzthum: I didn't generate this case. I borrowed it from one of the veraison crates. Once we switch to the other corim crate, we can add some cases that construct all the various things we support.

PR #1284, opened 2026-04-17
se-verifier: Handle report data properly for IBM SEL

BbolroC:
IBM SEL hasn’t handled the report data properly and does not include initdata in the parsed claim. 
Plus, the attestation outcome is `Contraindicated` because there is no appraisal section in the attestation policy for the TEE. 
This PR resolves the issue. Please refer to the individual commit messages for details.

Depends-on: https://github.com/confidential-containers/guest-components/pull/1417

Assisted-by: IBM Bob

### DIFF
--- a/deps/verifier/src/se/ibmse.rs
+++ b/deps/verifier/src/se/ibmse.rs
@@ -3,8 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use crate::{TeeEvidence, TeeEvidenceParsedClaim};
-use anyhow::{anyhow, bail, Context, Result};
+use crate::{ReportData, TeeEvidence, TeeEvidenceParsedClaim, ToHex};
+use anyhow::{anyhow, Context, Result};
 use core::result::Result::Ok;
 use openssl::encrypt::{Decrypter, Encrypter};
 use openssl::pkey::{PKey, Private, Public};
@@ -19,6 +19,7 @@ use pv::uv::ConfigUid;
 use serde::{Deserialize, Serialize};
 use serde_with::{base64::Base64, hex::Hex, serde_as};
 use std::{env, fs};
+use thiserror::Error;
 use tracing::{debug, info, warn};
 
 const DEFAULT_CERTS_OFFLINE_VERIFICATION: &str = "false";
@@ -62,6 +63,42 @@ fn list_files_in_folder(dir: &str) -> Result<Vec<String>> {
     }
 
     Ok(file_paths)
+}
+
+/// Error types for SE verifier operations
+#[derive(Error, Debug)]
+pub enum SeError {
+    #[error(
+        "USER_DATA content mismatch in IBM SEL evidence, expected: {expected:?}, got: {actual:?}"
+    )]
+    UserDataMismatch { expected: Vec<u8>, actual: Vec<u8> },
+
+    #[error("Failed to verify the measurement")]
+    MeasurementVerificationFailed,
+
+    #[error("Failed to decrypt measurement key")]
+    DecryptMeasurementKey(#[source] anyhow::Error),
+
+    #[error("Failed to decrypt request nonce")]
+    DecryptRequestNonce(#[source] anyhow::Error),
+
+    #[error("Failed to convert nonce from Vec<u8> to [u8; 16], must have exactly 16 elements")]
+    InvalidNonceLength,
+
+    #[error("Failed to get image_public_host_key_hash")]
+    MissingImagePublicHostKeyHash,
+
+    #[error("Failed to get attestation_public_host_key_hash")]
+    MissingAttestationPublicHostKeyHash,
+
+    #[error("Failed to deserialize evidence")]
+    DeserializeEvidence(#[source] serde_json::Error),
+
+    #[error("Failed to build json value from SE claims")]
+    BuildJsonClaims(#[source] serde_json::Error),
+
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
 }
 
 #[repr(C)]
@@ -114,6 +151,9 @@ pub struct SeAttestationRequest {
     encr_request_nonce: Vec<u8>,
     #[serde_as(as = "Base64")]
     image_hdr_tags: BootHdrTags,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde_as(as = "Option<Base64>")]
+    runtime_data_digest: Option<Vec<u8>>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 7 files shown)


BbolroC:
For reviewers, the change is verified by the current Kata nightly test for IBM SEL. (with https://github.com/confidential-containers/guest-components/pull/1417)

stevenhorsman (approved):
Looks okay to me. Thanks @BbolroC

BbolroC:
Thanks @stevenhorsman for the review! I've made this to a draft to get feedback from @Xynnn007.

deps/verifier/src/se/ibmse.rs
```diff
@@ -185,6 +226,40 @@ impl SeVerifierImpl {
             .try_into()
             .map_err(|_| anyhow!("Failed to convert nonce from Vec<u8> to [u8; 16], It must have exactly 16 elements."))?;
 
+        // Validate user_data - expected_report_data is required
+        let report_data = match expected_report_data {
+            ReportData::Value(data) => data,
+            ReportData::NotProvided => {
+                bail!("expected_report_data must be provided for IBM SEL verification");
+            }
+        };
```
Xynnn007: Looks like that the report-data/initdata here are used to check the UserData. For other verifiers, we usually mark the report-data and init-data optional. That means if given, we check; if not, we do not check. Probably we could have similar logic here but a little different: if both given, we check; if not, we do not check; if only one given, we raise error

deps/verifier/src/se/ibmse.rs
```diff
@@ -316,9 +392,452 @@ impl SeVerifierImpl {
...
+        // Should fail at measurement verification (we don't have valid measurement),
+        // but NOT at user_data validation
+        if let Err(e) = result {
+            let error_msg = e.to_string();
+            assert!(
+                !error_msg.contains("USER_DATA content mismatch"),
+                "Should not fail at USER_DATA validation, got: {}",
+                error_msg
+            );
+        }
```
Xynnn007, line 485: There might be a maintaining risk here that the check logic is "if not containing string". If we change the error message inside the check logic and breaks check logic togetherly, the test still passes. So a suggestion might be to have `thiserror` error enum definition in function `fn evaluate`, and here we compare the error type is not e.g. `UserDataMisMatch` explicitly.

deps/verifier/src/se/ibmse.rs
```diff
@@ -316,9 +392,452 @@ impl SeVerifierImpl {
...
+            image_hdr_tags: create_dummy_boot_hdr_tags(),
+        };
+
+        let evidence = serde_json::to_value(&response).expect("Failed to serialize");
+        let expected_report_data = ReportData::Value(&report_data);
+        let expected_init_data = InitDataHash::Value(&init_hash);
+
+        let result = verifier.evaluate(evidence, &expected_report_data, &expected_init_data);
+
+        assert!(result.is_err(), "Should fail with mismatched user_data");
```
Xynnn007, line 522: See the previous comment

Xynnn007 (commented):
Thanks @BbolroC for the PR! The code looks very good to me. There are mainly two comments. btw about the error comparation comment is mainly for maintaince, and it's acceptable to me to open an issue to record this and make it better in another PR.

deps/verifier/src/se/ibmse.rs
BbolroC: Gotcha! Thanks! I will update the PR.

deps/verifier/src/se/ibmse.rs
BbolroC: Yep!

deps/verifier/src/se/ibmse.rs
BbolroC: Looking at the attester side, I was thinking that the report data shouldn't be optional. That is why I ended up with this logic. What do you think about this?

deps/verifier/src/se/ibmse.rs
```diff
@@ -316,9 +392,452 @@ impl SeVerifierImpl {
...
+        };
+
+        let evidence = serde_json::to_value(&response).expect("Failed to serialize");
+        let expected_report_data = ReportData::Value(&report_data);
+        let expected_init_data = InitDataHash::NotProvided;
+
+        let result = verifier.evaluate(evidence, &expected_report_data, &expected_init_data);
+
+        // Should fail at measurement verification, but NOT at user_data validation
+        if let Err(e) = result {
```
mkulke: I think you want to use unwrap_err() in a test to assert that there is an error

deps/verifier/src/se/ibmse.rs
```diff
@@ -316,9 +392,452 @@ impl SeVerifierImpl {
...
+        let expected_init_data = InitDataHash::NotProvided;
+
+        let result = verifier.evaluate(evidence, &expected_report_data, &expected_init_data);
+
+        // Should fail because report_data is required
+        assert!(
+            result.is_err(),
+            "Should fail when report_data is not provided"
+        );
+        if let Err(e) = result {
```
mkulke: see above

deps/verifier/src/se/ibmse.rs
```diff
@@ -316,9 +392,452 @@ impl SeVerifierImpl {
...
+        };
+
+        let evidence = serde_json::to_value(&response).expect("Failed to serialize");
+        let expected_report_data = ReportData::Value(&report_data);
+        let expected_init_data = InitDataHash::Value(&init_hash);
+
+        let result = verifier.evaluate(evidence, &expected_report_data, &expected_init_data);
+
+        // Should fail at measurement verification, but NOT at user_data validation
+        if let Err(e) = result {
```
mkulke: see above

deps/verifier/src/se/ibmse.rs
```diff
@@ -316,9 +392,452 @@ impl SeVerifierImpl {
...
+
+        let evidence = serde_json::to_value(&response).expect("Failed to serialize");
+        let expected_report_data = ReportData::Value(&report_data);
+        let expected_init_data = InitDataHash::Value(&init_hash);
+
+        let result = verifier.evaluate(evidence, &expected_report_data, &expected_init_data);
+
+        assert!(result.is_err(), "Should fail with wrong order of user_data");
+
+        if let Err(e) = result {
```
mkulke: see above

deps/verifier/src/se/ibmse.rs
BbolroC: Yep, it can simplify things a bit more. Thanks for the suggestion!

BbolroC:
@fitzthum @Xynnn007 Looking forward to your feedback. Thanks a lot!

deps/verifier/src/se/ibmse.rs
```diff
@@ -168,22 +209,44 @@ impl SeVerifierImpl {
...
-            .map_err(|_| anyhow!("Failed to convert nonce from Vec<u8> to [u8; 16], It must have exactly 16 elements."))?;
+            .map_err(SeError::DecryptRequestNonce)?;
+
+        let nonce_array: [u8; 16] = nonce.try_into().map_err(|_| SeError::InvalidNonceLength)?;
+
+        // Validate runtime_data_digest if provided
+        if let ReportData::Value(expected_report_data) = expected_report_data {
+            let user_data = general_purpose::STANDARD
+                .decode(&se_response.user_data)
+                .context("Failed to decode user_data")?;
```
Xynnn007: The same UserData base64 encoding question as GC side.

Xynnn007 (approved):
LGTM. Thanks! @BbolroC

deps/verifier/src/se/ibmse.rs
BbolroC: This will be updated as https://github.com/confidential-containers/guest-components/pull/1417#discussion_r3122060025. Thanks a lot!

fitzthum (approved):
LGTM

PR #1286, opened 2026-04-20
kbs: pass VAULT and EXTERNAL_PLUGIN build args to Dockerfiles

bpradipt:
Helps to build them as part of docker-compose and other docker workflows


Xynnn007 (approved):
LGTM. Let's see the CI color

fitzthum (approved):
LGTM. No external plugins enabled by default so should be fine.

PR #1288, opened 2026-04-21
kbs: enable vault backend by default

### LINKED ISSUE #1272: Enable OpenBao/Vault built by default
Starting this issue to discuss if it's fine to enable openbao/vault support as resource backend by default in KBS?

Today vault/openbao is tested in e2e CI. If additional test cases are required those can be added as well.

bpradipt:
Set VAULT=true as the default in the Makefile and both Dockerfiles so that vault/OpenBao KV backend support is compiled in without requiring an explicit flag. Update vault_kv.md to reflect the new default.

Fixes: #1272


mkulke (approved):

Xynnn007 (approved):
Retrigger the CI. Let's see the color

fitzthum (approved):
LGTM

aramirez-capacity:
Hi. I think there is a bug in this MR. New builds still don't seem to include vault support by default.

If I dry-run the `make` command the Dockerfile runs, we can see `vault` is not in `--features` despite setting `VAULT=true`:
```
0 15:08:16 /dev/shm/trustee/kbs
$ make -n AS_FEATURE=coco-as-builtin FEATURES=pkcs11 VAULT=true 2>&1 | grep "cargo build"

cargo build -p kbs --locked --release --no-default-features --features coco-as-builtin,pkcs11
```

I believe the problem stems from this behaviour from GNU `make`: https://www.gnu.org/software/make/manual/make.html#Override-Directive
> If a variable has been set with a command argument (see [Overriding Variables](https://www.gnu.org/software/make/manual/make.html#Overriding)), then ordinary assignments in the makefile are ignored.

In other words: [this line](https://github.com/confidential-containers/trustee/blob/main/kbs/Makefile#L61) is being silently ignored due to `FEATURES` being set as a command line argument.

If that's really the issue, it might be worth reviewing the Makefile in general (and maybe the CI tests?).

fitzthum:
@bpradipt did you see ^^

bpradipt:
Ah no, I missed the update. I'll take a look and revert.

fitzthum:
I don't think it's your fault actually. Seems like the issue would be that we're setting `FEATURES=pkcs11` in the Makefile. Needs to be confirmed tho.

bpradipt:
> Hi. I think there is a bug in this MR. New builds still don't seem to include vault support by default.
> 
> If I dry-run the `make` command the Dockerfile runs, we can see `vault` is not in `--features` despite setting `VAULT=true`:
> 
> ```
> 0 15:08:16 /dev/shm/trustee/kbs
> $ make -n AS_FEATURE=coco-as-builtin FEATURES=pkcs11 VAULT=true 2>&1 | grep "cargo build"
> 
> cargo build -p kbs --locked --release --no-default-features --features coco-as-builtin,pkcs11
> ```
> 
> I believe the problem stems from this behaviour from GNU `make`: https://www.gnu.org/software/make/manual/make.html#Override-Directive
> 
> > If a variable has been set with a command argument (see [Overriding Variables](https://www.gnu.org/software/make/manual/make.html#Overriding)), then ordinary assignments in the makefile are ignored.
> 
> In other words: [this line](https://github.com/confidential-containers/trustee/blob/main/kbs/Makefile#L61) is being silently ignored due to `FEATURES` being set as a command line argument.
> 
> If that's really the issue, it might be worth reviewing the Makefile in general (and maybe the CI tests?).

The issue is confirmed. I'll create a PR with the fix.

PR #1292, opened 2026-04-23
Verifier: add PCCS to DCAP configuration and tidy the codes

Xynnn007:
cc @mythi

### DIFF
--- a/deps/verifier/src/tdx/quote.rs
+++ b/deps/verifier/src/tdx/quote.rs
@@ -420,7 +420,7 @@ mod tests {
     /// Also, you need to configure DCAP to work with alibaba cloud's PCCS.
     /// create `/tmp/sgx_test_qcnl.conf` with content
     /// ```json
-    /// {"pccs_url" :"https://sgx-dcap-server.cn-beijing.aliyuncs.com/sgx/certification/v4/"}
+    /// {"collateral_service" :"https://sgx-dcap-server.cn-beijing.aliyuncs.com/sgx/certification/v4/"}
     /// ```
     /// Test can be run using exported environment variable:
     /// ```QCNL_CONF_PATH=/tmp/sgx_test_qcnl.conf```


mythi:
QCNL logic is cryptic but you should be able to use local PCCS as `collateral_service` in Trustee since we only need verification collateral:

```
 // You can use the Intel PCS or another PCCS to get quote verification collateral.  Retrieval of PCK 
  // Certificates will always use the PCCS described in pccs_url.  When collateral_service is not defined, both 
  // PCK Certs and verification collateral will be retrieved using pccs_url  
  //,"collateral_service": "https://api.trustedservices.intel.com/sgx/certification/v4/"
```

Xynnn007:
@mythi Yes I had read this comment. It means that collateral service can be replaced by pccs_url, but pccs_url cannot be replaced by collateral service.

mythi:
> @mythi Yes I had read this comment. It means that collateral service can be replaced by pccs_url, but pccs_url cannot be replaced by collateral service.

What is the error you get when you add your local PCCS to Trustee's config. In my local testing only `collateral_service` was used for verification collateral

Xynnn007:
> What is the error you get when you add your local PCCS to Trustee's config. In my local testing only `collateral_service` was used for verification collateral

Let me take a try and reply to you then

mythi:
> 2. If the specified file of QCNL_CONF_PATH exists, skip overwriting

From what I've understood, `File::create_new` handles the case atomically

Xynnn007:
> From what I've understood, `File::create_new` handles the case atomically

The original code submit by you use a error kind condition to check if it's AlreadyExists also. This PR does not change the logic

Xynnn007:
> > What is the error you get when you add your local PCCS to Trustee's config. In my local testing only `collateral_service` was used for verification collateral
> 
> Let me take a try and reply to you then

I cannot produce some issue here. In history there was a case where a quote generated on alibaba cloud cannot be verified by intel PCS.

So I dig into DCAP source code, and find that the `pccs_url` is used to [get PCK](https://github.com/intel/confidential-computing.tee.dcap/blob/main/QuoteGeneration/qcnl/certification_service.cpp#L189). But if we only configure `collateral_server`, the PCK will be looked up via default value [https://localhost:8081/sgx/certification/v4/](https://github.com/intel/confidential-computing.tee.dcap/blob/2d483e35e7ebd70d91d88cd4abfd4c097f2e634c/QuoteGeneration/qcnl/inc/qcnl_config.h#L85) thus error happens.

I noticed that there are [examples](https://github.com/intel/confidential-computing.tee.dcap/blob/main/QuoteGeneration/qcnl/linux/sgx_default_qcnl_azure.conf) that **both collateral_service and pccs_url are configured but they are different** for QCNL config.

If collateral_service is enough, why we need to configure a different pccs_url?

Xynnn007:
@mythi Temporarily it's hard for me to create a scenario that PCKs are not retrievable on Intel PCS but only self-hosted PCCS, while collaterals can be get from Intel PCS.

mythi:
> If collateral_service is enough, why we need to configure a different pccs_url?

QCNL serves both quote generation and quote verification. In Trustee we don't have a scenario where quote signing and PCK certs retrieval would be needed.

QCNL with `pccs_url` is needed for things like TDX QGS or SGX AESMD to work to get PCK certs from a local cache but if you only need Trustee to use Alibaba's PCCS, you can use the existing `collateral_service` DCAP verifier config entry.

Xynnn007:
@mythi Got it. Thanks for the explanation! Let me change the code.

mythi:
Also related: I'm planning to change Trustee a bit to *not* use QCNL at all because of some new requirements I have and because limitations like `QCNL_CONF_PATH` writing. So with that, can you live without 4403330 for the time being since I think that writing function can be dropped later.

Xynnn007:
> Also related: I'm planning to change Trustee a bit to _not_ use QCNL at all because of some new requirements I have and because limitations like `QCNL_CONF_PATH` writing. So with that, can you live without [4403330](https://github.com/confidential-containers/trustee/commit/4403330b658a92c132eae88c3e7e138093527b5d) for the time being since I think that writing function can be dropped later.

What's the requirements you are facing? Does DCAP will support to get the parameters directly from API or something?

I can close this PR as PCCS is not needed anymore.

mythi:
> > Also related: I'm planning to change Trustee a bit to _not_ use QCNL at all because of some new requirements I have and because limitations like `QCNL_CONF_PATH` writing. So with that, can you live without [4403330](https://github.com/confidential-containers/trustee/commit/4403330b658a92c132eae88c3e7e138093527b5d) for the time being since I think that writing function can be dropped later.
> 
> What's the requirements you are facing? Does DCAP will support to get the parameters directly from API or something?

`tee_verify_quote()` takes user-provided collateral. I'm planning to change
 https://github.com/confidential-containers/trustee/blob/ec63e59123349a9e4bdabb86d57943e898064c8f/deps/verifier/src/intel_dcap/mod.rs#L124-L134 

and handle collateral pulling in Rust (P(C)CS compatible) and to support air-gapped use case alternative without having to have PCCS running.

Xynnn007:
> and handle collateral pulling in Rust (P(C)CS compatible) and to support air-gapped use case alternative without having to have PCCS running.

Good. This would be a right way to go. btw, do you have any plan to do quote verification in pure Rust also?

mythi:
> > and handle collateral pulling in Rust (P(C)CS compatible) and to support air-gapped use case alternative without having to have PCCS running.
> 
> Good. This would be a right way to go. btw, do you have any plan to do quote verification in pure Rust also?

Let's see. I have a POC but it needs more work. Moving step by step closer to it anyway.

mythi (approved):
LGTM

PR #1299, opened 2026-04-28
kbs: get tid of rand dependency

Xynnn007:
The rand is used in AES algorithm in KBS. AES crates provide crate-compatible functions thus we do not need rand crate explicitly.

This commit removes rand dependency.


mythi (approved):
Looks good, thanks

PR #1302, opened 2026-04-28
ci: add white list to link check

Xynnn007:
Errors happen very frequently like in https://github.com/confidential-containers/trustee/actions/runs/25042336665/job/73348459915?pr=1239 Thus we add them to white list.


fitzthum (approved):
LGTM

hopefully we don't move that docs page in the future

PR #1303, opened 2026-04-28
kbs: align external plugin TLS config schema with runtime behavior

Xynnn007:
~~Require explicit `tls_mode` and model TLS-only fields under the `tls` variant so config parsing, validation, and startup errors share one consistent contract. This avoids ambiguous defaults and makes misconfiguration fail fast with clearer diagnostics.~~

Automatically detect `tls_mode` based on `endpoint`, and require ca-cert if tls-mode is used.

cc @bpradipt


bpradipt:
e2e ext plugin CI is failing

kbs/docs/config.md
```diff
@@ -354,8 +354,8 @@ Each backend is reachable at `/kbs/v0/external/<name>/...`.
 |---|---|---|---|---|
 | `name` | string | Yes | — | Sub-plugin name used in URL routing |
 | `endpoint` | string | Yes | — | gRPC endpoint (`http://` for insecure, `https://` for TLS) |
-| `tls_mode` | string | No | `"tls"` | `"insecure"` or `"tls"` |
-| `ca_cert_path` | string | No | — | CA certificate path (required for `tls`) |
+| `tls_mode` | string | Yes | — | `"insecure"` or `"tls"` |
```
fitzthum: This is marked as a required field, but it seems like you're actually removing this from the config?

fitzthum (commented):
One question but seems good

kbs/docs/config.md
bpradipt: We should call it as "auto" may be with a comment that the mode is auto detected based on the url

fitzthum (approved):
LGTM

bpradipt (approved):
/lgtm
Thanks @Xynnn007

PR #1305, opened 2026-04-29
kbs/plugins: add provisioner plugin

MatiasVara:
Add a provisioner plugin that automates LUKS key generation and resource creation for confidential VMs. When a POST request is received at `kbs/v0/provisioner/provision`, the plugin:

1. Generates a UUID and a random LUKS encryption key. The UUID is generated from the VM's namespace and name. 
2. Builds an initdata.toml pointing the guest to the KBS resource path `default/{UUID}/root` where the key is stored.
3. Stores the key in the KBS storage backend so the guest can retrieve it after attestation via the existing `resource` plugin.
4. Returns a JSON response:

   struct ProvisionResponse { uuid: String, resource_path: String, }

The plugin also supports DELETE requests at
`kbs/v0/provisioner/provision/{uuid}` to remove provisioned resources.

Authentication is routed through the KBS admin auth path (validate_auth returns true) instead of requiring a TEE attestation token, since the caller is infrastructure (hook sidecar), not a TEE guest.

Note that I already did a first review using Claude.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ regorus = { version = "0.11.0", default-features = false, features = [
     "time",
     "std",
 ] }
+rand = "0.10.2"
 reqwest = { version = "0.13", default-features = false, features = ["json"] }
 rstest = "0.26.1"
 serde = { version = "1.0", features = ["derive"] }
@@ -74,7 +75,7 @@ rsa = { version = "0.9.10", features = ["sha2"] }
 time = { version = "0.3.47", features = ["std"] }
 tracing = "0.1.44"
 tracing-subscriber = { version = "0.3.23", features = ["env-filter"] }
-uuid = { version = "1.23.3", features = ["v4"] }
+uuid = { version = "1.23.3", features = ["v4", "v5"] }
 zeroize = "1"
 
 [patch.crates-io]
--- a/kbs/Cargo.toml
+++ b/kbs/Cargo.toml
@@ -85,6 +85,7 @@ zeroize.workspace = true
 policy-engine.path = "../deps/policy-engine"
 prometheus = "0.14.0"
 prost = { workspace = true, optional = true }
+rand.workspace = true
 regex = "1.12.4"
 regorus.workspace = true
 reqwest = { workspace = true, features = ["default-tls"] }
--- a/kbs/Cargo.toml
+++ b/kbs/Cargo.toml
@@ -91,6 +91,7 @@ zeroize.workspace = true
 policy-engine.path = "../deps/policy-engine"
 prometheus = "0.14.0"
 prost = { workspace = true, optional = true }
+rand.workspace = true
 regex = "1.12.4"
 regorus.workspace = true
 reqwest = { workspace = true, features = ["default-tls"] }
--- a/kbs/src/api_server.rs
+++ b/kbs/src/api_server.rs
@@ -396,7 +396,7 @@ pub(crate) async fn api(
                 // Plugin calls need to be authorized by the admin auth
                 core.admin.check_admin_access(&request)?;
                 let response = plugin
-                    .handle(&body, &query, resource_path, request.method())
+                    .handle(&body, &query, resource_path, request.method(), None)
                     .await
                     .map_err(|e| Error::PluginInternalError { source: e })?;
 
@@ -446,8 +446,11 @@ pub(crate) async fn api(
                 }
                 KBS_POLICY_APPROVALS.inc();
 
+                let init_data = claims
+                    .pointer("/submods/cpu0/ear.veraison.annotated-evidence/init_data_claims");
+
                 let response = plugin
-                    .handle(&body, &query, resource_path, request.method())
+                    .handle(&body, &query, resource_path, request.method(), init_data)
                     .await
                     .map_err(|e| Error::PluginInternalError { source: e })?;
 
--- a/kbs/src/plugins/implementations/provisioner/mod.rs
+++ b/kbs/src/plugins/implementations/provisioner/mod.rs
@@ -1,0 +1,228 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Provisioner plugin for KBS.
+//!
+//! Generates per-VM LUKS encryption keys and stores them in a dedicated
+//! storage namespace so that attested guests can retrieve them via
+//! `GET /kbs/v0/provisioner/default/{uuid}/root`.
+//!
+//! The hook sidecar calls `POST /kbs/v0/provisioner/provision` before the VM
+//! boots and receives `{uuid, resource_path}` to inject into the VM.
+//! On boot the guest attests and fetches the key through the provisioner's
+//! own serving endpoint.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use actix_web::http::Method;
+use anyhow::{anyhow, bail, Context, Result};
+use key_value_storage::{KeyValueStorageInstance, SetParameters, StorageProvider};
+use rand::distr::{Alphanumeric, SampleString};
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+const PROVISIONER_STORAGE_NAMESPACE: &str = "provisioner";
+
+// UUID v5 namespace for the provisioner plugin.
+// Derived from: Uuid::new_v5(&Uuid::NAMESPACE_URL, b"provisioner_plugin")
+const PROVISIONER_NAMESPACE: Uuid = uuid::uuid!("feab5599-6854-52bc-95f0-0241c03e620a");
+
+/// Config (deserialized from kbs-config.toml)
+#[derive(Deserialize, Clone, Debug, PartialEq)]
+pub struct ProvisionerConfig {
+    /// Length of the random key in bytes (default 32).
+    #[serde(default = "default_key_length")]
+    pub key_length: usize,
+}
+
+fn default_key_length() -> usize {
+    32
+}
+
+pub struct Provisioner {
+    storage: KeyValueStorageInstance,
+    key_length: usize,
+}
+
+#[derive(Serialize, Deserialize, Clone)]
+struct ProvisionRequest {
+    vm_name: String,
+    #[serde(default = "default_namespace")]
+    namespace: String,
+}
+
+fn default_namespace() -> String {
+    "default".into()
+}
+
+#[derive(Serialize, Clone)]
+struct ProvisionResponse {
+    uuid: String,
+    resource_path: String,
+}
+
+#[derive(Serialize)]
+struct StatusResponse {
+    status: String,
+}
+
+impl Provisioner {
+    pub async fn new(
+        config: ProvisionerConfig,
+        storage_provider: Arc<dyn StorageProvider>,
+    ) -> Result<Self> {
+        let storage = storage_provider
+            .get_or_register(PROVISIONER_STORAGE_NAMESPACE)
+            .await
+            .context("Provisioner: failed to init storage backend")?;
+
+        Ok(Self {
+            storage,
+            key_length: config.key_length,
+        })
+    }
+}
+
+impl Provisioner {
+    fn generate_random_key(&self) -> String {
+        Alphanumeric.sample_string(&mut rand::rng(), self.key_length)
+    }
+
+    fn generate_confdata_toml(luks_key: &str) -> String {
+        format!(
+            "version = \"0.1.0\"\n\
+             \n\
+             [data]\n\
+             \"io.cryptsetup.key.text.root\" = \"{}\"\n",
+            luks_key
+        )
+    }
+
+    async fn handle_get_resource(
+        &self,
+        path: &[&str],
+        init_data: Option<&serde_json::Value>,
+    ) -> Result<Vec<u8>> {
+        let resource_key = path.join("/");
+
+        let Some(claims) = init_data else {
+            bail!("provisioner requires init_data for resource access");
+        };
+
+        let bound_resource = claims
+            .get("trustee.kbs.resource")
+            .and_then(|v| v.as_str())
+            .ok_or_else(|| anyhow!("init_data missing trustee.kbs.resource"))?;
+        let expected = format!("kbs+provisioner:///{resource_key}");
+        if bound_resource != expected {
+            bail!("init_data resource mismatch: expected {expected}, got {bound_resource}");
+        }
+
+        let data = self
+            .storage
+            .get(&resource_key)
+            .await?
+            .ok_or_else(|| anyhow!("provisioner resource not found: {resource_key}"))?;
+        Ok(data)
+    }
+
+    async fn handle_provision(&self, body: &[u8]) -> Result<Vec<u8>> {
+        let req: ProvisionRequest = serde_json::from_slice(body).context("invalid JSON body")?;
+
+        let vm_identity = format!("{}/{}", req.namespace, req.vm_name);
+
+        // Generate deterministic UUID from VM identity
+        let trustee_uuid = Uuid::new_v5(&PROVISIONER_NAMESPACE, vm_identity.as_bytes()).to_string();
+
+        let resource_path = format!("default/{trustee_uuid}/root");
+
+        if self.storage.get(&resource_path).await?.is_some() {
+            let response = ProvisionResponse {
+                uuid: trustee_uuid,
+                resource_path,
+            };
+            return Ok(serde_json::to_vec(&response)?);
+        };
+
+        let luks_key = self.generate_random_key();
+
+        let confdata_toml = Self::generate_confdata_toml(&luks_key);
+
+        self.storage
+            .set(
+                &resource_path,
+                confdata_toml.as_bytes(),
+                SetParameters { overwrite: false },
+            )
+            .await
+            .context("failed to write resource")?;
+
+        let response = ProvisionResponse {
+            uuid: trustee_uuid,
+            resource_path,
+        };
+
+        Ok(serde_json::to_vec(&response)?)
+    }
+
+    async fn handle_deprovision(&self, path: &[&str]) -> Result<Vec<u8>> {
+        let trustee_uuid = path
+            .first()
+            .ok_or_else(|| anyhow!("missing uuid in path"))?;
+        let resource_path = format!("default/{trustee_uuid}/root");
+
+        let _ = self.storage.delete(&resource_path).await;
+
+        Ok(serde_json::to_vec(&StatusResponse {
+            status: "deleted".into(),
+        })?)
+    }
+}
+
+#[async_trait::async_trait]
+impl super::super::plugin_manager::ClientPlugin for Provisioner {
+    async fn handle(
+        &self,
+        body: &[u8],
+        _query: &HashMap<String, String>,
+        path: &[&str],
+        method: &Method,
+        init_data: Option<&serde_json::Value>,
+    ) -> Result<Vec<u8>> {
+        match (method.as_str(), path.first().copied()) {
+            ("POST", Some("provision")) => self.handle_provision(body).await,
+            ("DELETE", Some("provision")) => self.handle_deprovision(&path[1..]).await,
+            ("GET", _) => self.handle_get_resource(path, init_data).await,
+            _ => bail!(
+                "unsupported: {} /kbs/v0/provisioner/{}",
+                method,
+                path.join("/")
+            ),
+        }
+    }
+
+    async fn validate_auth(
+        &self,
+        _body: &[u8],
+        _query: &HashMap<String, String>,
+        path: &[&str],
+        method: &Method,
+    ) -> Result<bool> {
+        // GET: guest fetches the resource after TEE attestation (validate_auth = false).
+        // POST/DELETE: sidecar provisions/deprovisions via admin auth (validate_auth = true).
+        match (method.as_str(), path.first().copied()) {
+            ("GET", _) => Ok(false),
+            _ => Ok(true),
+        }
+    }
+
+    async fn encrypted(
+        &self,
+        _body: &[u8],
+        _query: &HashMap<String, String>,
+        _path: &[&str],
+        method: &Method,
+    ) -> Result<bool> {
+        Ok(method.as_str() == "GET")
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,7 +90,7 @@ dependencies = [
  "mime",
  "percent-encoding",
  "pin-project-lite",
- "rand 0.10.1",
+ "rand 0.10.2",
  "sha1 0.11.0",
  "smallvec",
  "tokio",
@@ -2127,7 +2127,7 @@ dependencies = [
  "openssl",
  "p256",
  "p521",
- "rand 0.10.1",
+ "rand 0.10.2",
  "rand 0.8.6",
  "rsa",
  "serde",
@@ -3179,7 +3179,7 @@ dependencies = [
  "google-cloud-wkt",
  "http 1.4.0",
  "pin-project",
- "rand 0.10.1",
+ "rand 0.10.2",
  "serde",
  "serde_json",
  "thiserror 2.0.19",
@@ -4346,6 +4346,7 @@ dependencies = [
  "polyval 0.6.2",
  "prometheus",
  "prost",
+ "rand 0.10.2",
  "reference-value-provider-service",
  "regex",
  "regorus",
@@ -4517,7 +4518,7 @@ dependencies = [
  "openssl",
  "prost",
  "protos",
- "rand 0.10.1",
+ "rand 0.10.2",
  "reqwest 0.13.4",
  "resource_uri",
  "serde",
@@ -5935,9 +5936,9 @@ dependencies = [
 
 [[package]]
 name = "rand"
-version = "0.10.1"
+version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d2e8e8bcc7961af1fdac401278c6a831614941f6164ee3bf4ce61b7edb162207"
+checksum = "c7f5fa3a058cd35567ef9bfa5e75732bee0f9e4c55fa90477bef2dfcdbc4be80"
 dependencies = [
  "chacha20",
  "getrandom 0.4.2",
@@ -6187,7 +6188,7 @@ dependencies = [
  "num-bigint 0.5.1",
  "num-traits",
  "parking_lot 0.12.5",
- "rand 0.10.1",
+ "rand 0.10.2",
  "regex",
  "serde",
  "serde_json",
@@ -8461,6 +8462,7 @@ dependencies = [
  "getrandom 0.4.2",
  "js-sys",
  "serde_core",
+ "sha1_smol",
  "wasm-bindgen",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,7 +90,7 @@ dependencies = [
  "mime",
  "percent-encoding",
  "pin-project-lite",
- "rand 0.10.1",
+ "rand 0.10.2",
  "sha1 0.11.0",
  "smallvec",
  "tokio",
@@ -2124,7 +2124,7 @@ dependencies = [
  "openssl",
  "p256",
  "p521",
- "rand 0.10.1",
+ "rand 0.10.2",
  "rand 0.8.6",
  "rsa",
  "serde",
@@ -3168,7 +3168,7 @@ dependencies = [
  "google-cloud-wkt",
  "http 1.4.0",
  "pin-project",
- "rand 0.10.1",
+ "rand 0.10.2",
  "serde",
  "serde_json",
  "thiserror 2.0.19",
@@ -4333,6 +4333,7 @@ dependencies = [
  "polyval 0.6.2",
  "prometheus",
  "prost",
+ "rand 0.10.2",
  "reference-value-provider-service",
  "regex",
  "regorus",
@@ -4468,7 +4469,7 @@ dependencies = [
  "openssl",
  "prost",
  "protos",
- "rand 0.10.1",
+ "rand 0.10.2",
  "reqwest 0.13.4",
  "resource_uri",
  "serde",
@@ -5889,9 +5890,9 @@ dependencies = [
 
 [[package]]
 name = "rand"
-version = "0.10.1"
+version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d2e8e8bcc7961af1fdac401278c6a831614941f6164ee3bf4ce61b7edb162207"
+checksum = "c7f5fa3a058cd35567ef9bfa5e75732bee0f9e4c55fa90477bef2dfcdbc4be80"
 dependencies = [
  "chacha20",
  "getrandom 0.4.2",
@@ -6132,7 +6133,7 @@ dependencies = [
  "num-bigint 0.5.1",
  "num-traits",
  "parking_lot 0.12.5",
- "rand 0.10.1",
+ "rand 0.10.2",
  "regex",
  "serde",
  "serde_json",
@@ -8389,6 +8390,7 @@ dependencies = [
  "getrandom 0.4.2",
  "js-sys",
  "serde_core",
+ "sha1_smol",
  "wasm-bindgen",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5865,7 +5865,7 @@ dependencies = [
  "bytes",
  "getrandom 0.4.2",
  "lru-slab",
- "rand 0.10.1",
+ "rand 0.10.2",
  "rand_pcg",
  "ring",
  "rustc-hash 2.1.2",
--- a/kbs/src/plugins/implementations/external_plugin.rs
+++ b/kbs/src/plugins/implementations/external_plugin.rs
@@ -387,6 +387,7 @@ impl ClientPlugin for ExternalPlugin {
         query: &HashMap<String, String>,
         path: &[&str],
         method: &Method,
+        _init_data: Option<&serde_json::Value>,
     ) -> Result<Vec<u8>> {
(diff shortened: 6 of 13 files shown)


kbs/src/plugins/implementations/provisioner.rs
```diff
@@ -0,0 +1,310 @@
...
+        method: &Method,
+    ) -> Result<Vec<u8>> {
+        self.check_bearer_token(query)?;
+
+        match (method.as_str(), path.first().copied()) {
+            ("POST", Some("provision")) => self.handle_provision(body).await,
+            ("DELETE", Some("provision")) => {
+                self.handle_deprovision(&path[1..]).await
+            }
+            ("GET", Some("healthz")) => {
```
fitzthum: Not sure if we want to add health to this plugin in particular. I think we will be adding an overall trustee health endpoint soon. I guess that won't tell you that a particular plugin is enabled, but I would still lean away from having this here.

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+        let mrconfigid = {
+            let digest = Sha384::digest(initdata_toml.as_bytes());
+            B64.encode(digest)
+        };
+
+        // Write the confdata (LUKS key) to storage via the same backend
+        // that the `resource` plugin reads from.
+        // TODO: Using overwrite: true could silently replace another VM's key
+        // on UUID collision. Use overwrite: false and handle the conflict error,
+        // or check existence before writing.
```
fitzthum: Yeah probably should not do this. Instead, you can introduce another endpoint to this plugin for serving resources.

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    fn generate_initdata_toml(&self, resource_path: &str) -> String {
+        format!(
+            "algorithm = \"sha384\"\n\
+             version = \"0.1.0\"\n\
+             \n\
+             [data]\n\
+             \"trustee.kbs.url\" = \"{}\"\n\
+             \"trustee.kbs.resource\" = \"{}\"\n",
+            self.kbs_url, resource_path
```
fitzthum: You may want to add a comment somewhere that this is not for CoCo and that this init-data is not meant to be compatible with coco.

kbs/src/plugins/implementations/provisioner.rs
```diff
@@ -0,0 +1,310 @@
...
+    ) -> Result<bool> {
+        // Return true so KBS routes through the admin auth path
+        // (instead of the attestation token path which requires a TEE session).
+        //
+        // NOTE: Currently relies on InsecureAllowAll admin backend for dev.
+        // For production, switch to Simple admin with JWT-signed requests
+        // and scoped roles, e.g.:
+        //   [admin] type = "Simple"
+        //   [[admin.personas]] id = "provisioner" public_key_path = "..."
+        //   [[admin.roles]] id = "provisioner" allowed_endpoints = "^/kbs/v0/provisioner/.*$"
```
fitzthum: Note that this config is changing a bit, but I think your approach is generally sound. You will have an admin token just for this plugin that will be given to the hook sidecar thing, right?

kbs/src/plugins/implementations/provisioner.rs
```diff
@@ -0,0 +1,310 @@
...
+            return Ok(());
+        };
+        if expected.is_empty() {
+            return Ok(());
+        }
+        let provided = query.get("authorization").or_else(|| query.get("token"));
+        match provided {
+            Some(tok) if tok == expected => Ok(()),
+            _ => bail!("Unauthorized"),
+        }
```
fitzthum: Why is this necessary? Is there something missing from the admin token checks?

fitzthum (commented):
Nice. A couple high-level comments, but this seems generally workable.

kbs/src/plugins/implementations/provisioner.rs
MatiasVara: It is not necessary. I forgot to remove check_bearer_token().

kbs/Cargo.toml
```diff
@@ -40,6 +37,9 @@ pkcs11 = ["cryptoki"]
 # that want to join a Nebula overlay network
 nebula-ca-plugin = []
 
+# Use provisioner plugin to automate LUKS key generation for confidential VMs
+provisioner = []
```
Xynnn007: If we do not have any new dependencies for plugin `provisioner`. I'd suggest to not to have this feature and other building flags. Underlying, make `provisioner.rs` by default built.

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+use key_value_storage::{KeyValueStorageInstance, SetParameters, StorageBackendConfig};
+use rand::Rng;
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha384};
+use uuid::Uuid;
+
+use super::resource::RESOURCE_STORAGE_NAMESPACE;
+
+// ---------------------------------------------------------------------------
+// Config (deserialized from kbs-config.toml)
```
Xynnn007: Let's use rust style comments. like
```rust
/// Config (deserialized from kbs-config.toml)
#[derive(Deserialize, Clone, Debug, PartialEq)]
pub struct ProvisionerConfig {
...
```

kbs/src/plugins/implementations/provisioner.rs
```diff
@@ -0,0 +1,288 @@
...
+    #[serde(default = "default_key_length")]
+    pub key_length: usize,
+}
+
+fn default_key_length() -> usize {
+    32
+}
+
+// ---------------------------------------------------------------------------
+// Plugin state
```
Xynnn007: ditto

kbs/src/plugins/implementations/provisioner.rs
```diff
@@ -0,0 +1,288 @@
...
+        self.cache.lock().unwrap().retain(|_, v| v.uuid != *trustee_uuid);
+
+        Ok(serde_json::to_vec(&StatusResponse {
+            status: "deleted".into(),
+        })?)
+    }
+}
+
+// ---------------------------------------------------------------------------
+// ClientPlugin trait
```
Xynnn007: ditto

kbs/src/plugins/implementations/provisioner.rs
```diff
@@ -0,0 +1,288 @@
...
+    // TODO: This in-memory cache grows unboundedly and uses std::sync::Mutex
+    // which can block the async runtime. More critically, the cache does
+    // not survive KBS restarts: if the sidecar re-provisions the same VM
+    // after a restart, a new UUID/key pair is generated, replacing the
+    // original resource. The VM's LUKS volume would then fail to unlock
+    // because the key no longer matches. The cache (or the vm->resource
+    // mapping) must be persisted to the storage backend so it can be
+    // restored on startup. Also consider a TTL/LRU eviction policy and
+    // switching to tokio::sync::Mutex.
+    cache: Mutex<HashMap<String, ProvisionResponse>>,
```
Xynnn007: Why not we use `tokio::sync::Mutex` or `tokio::sync::RwLock` here?

Xynnn007 (commented):
Hi @MatiasVara welcome! I left some ideas although it's still draft. I have a question before codes:

KBS now has a mechanism named "external-plugin". See [docs](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/ext_plugin.md). To make it simple, it provides a way for KBS to connect to a gRPC server aside of KBS. In this way, we do not need to make any changes to KBS code. What we need is just changing the KBS config a little. In this way you can have a self-and-easy-maintained plugin server, do you consider to use the mechanism?

MatiasVara:
> Hi @MatiasVara welcome! I left some ideas although it's still draft. I have a question before codes:
> 
> KBS now has a mechanism named "external-plugin". See [docs](https://github.com/confidential-containers/trustee/blob/main/kbs/docs/ext_plugin.md). To make it simple, it provides a way for KBS to connect to a gRPC server aside of KBS. In this way, we do not need to make any changes to KBS code. What we need is just changing the KBS config a little. In this way you can have a self-and-easy-maintained plugin server, do you consider to use the mechanism?

Oh cool, I did not know that there is that mechanism. I'll investigate it! Do you think it is better that way that having a plugin in-tree?

Xynnn007:
> Do you think it is better that way that having a plugin in-tree?

If it's not a concrete scenario for CoCo, I'd like to say yes. As it's defined by you and can be modified anytime. We will also be benefit of less maintainance effort ; )

MatiasVara:
> > Do you think it is better that way that having a plugin in-tree?
> 
> If it's not a concrete scenario for CoCo, I'd like to say yes. As it's defined by you and can be modified anytime. We will also be benefit of less maintainance effort ; )

I think the plan is to use it in together with kubevirt to provide resources to encrypt the disk for confidential VMs. I completely got you point. It's just that for me it would be easier if it is already in trustee and someone can just enable/disable it but I do not have an strong opinion about it.

fitzthum:
Personally, I think it's fine to have this be an internal plugin. Not all users will use every plugin.

kbs/src/plugins/implementations/provisioner.rs
MatiasVara: I removed it.

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+            return Ok(serde_json::to_vec(cached)?);
+        }
+
+        // TODO: UUIDv4 is random but not tied to VM identity. A more robust
+        // approach would derive the UUID deterministically from the VM's
+        // attributes (e.g. name + namespace + cluster ID) to ensure
+        // idempotency and traceability across re-provisions.
+        // A uniqueness check against existing storage entries should also
+        // be added to avoid collisions before writing the resource.
+        let trustee_uuid = Uuid::new_v4().to_string();
```
MatiasVara: @fitzthum  I was starting to investigate SPIFFE to generate the UUID but it may be overkill for this plugin. WDYT?

kbs/src/plugins/implementations/provisioner/mod.rs
MatiasVara: Can you elaborate?

kbs/src/plugins/implementations/provisioner/mod.rs
fitzthum: yeah don't bother with SPIFFE yet. we are looking into various integrations there.

kbs/src/plugins/implementations/provisioner.rs
```diff
@@ -0,0 +1,265 @@
...
+    kbs_url: String,
+    key_length: usize,
+    // TODO: This in-memory cache grows unboundedly and uses std::sync::Mutex
+    // which can block the async runtime. More critically, the cache does
+    // not survive KBS restarts: if the sidecar re-provisions the same VM
+    // after a restart, a new UUID/key pair is generated, replacing the
+    // original resource. The VM's LUKS volume would then fail to unlock
+    // because the key no longer matches. The cache (or the vm->resource
+    // mapping) must be persisted to the storage backend so it can be
+    // restored on startup.
```
Xynnn007: We might need to update the comments here because now `tokio::sync::RwLock` is used.

kbs/src/plugins/implementations/provisioner.rs
```diff
@@ -0,0 +1,265 @@
...
+    // encoding to get a full 256-bit key.
+    fn generate_random_key(&self) -> String {
+        const CHARSET: &[u8] = b"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+        let mut rng = rand::thread_rng();
+        (0..self.key_length)
+            .map(|_| {
+                let idx = rng.gen_range(0..CHARSET.len());
+                CHARSET[idx] as char
+            })
+            .collect()
```
Xynnn007: A shorter way with `Alphanumeric`  distribution can also implement this
```rust
let key = rand::distr::Alphanumeric.sample_string(&mut rand::rng(), self.key_length);
```

Not sure you like the taste

kbs/src/plugins/implementations/provisioner.rs
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    // This init-data is not meant to be compatible with CoCo.
+    // See https://gitlab.com/berrange/cvminjector#initial-data-format
```
Xynnn007: Looks like there are some inner mechanism for this plugin. My preference is that to have a code structure like 
```
- implementations
  - ...
  - provisioner
    - mod.rs (original provisioner.rs)
    - README.md (simply introduce the case, and with a guide or a guide link to help others to use this ability)
```

Xynnn007 (commented):
Thanks for adopting previous comments! 

If we want to get this into the repo, let me take a closer look at the design.

Based on the concrete logic and dataflow of the plugin, I’m wondering what was the reason for choosing a plugin approach instead of `kbs-client`+script inside KubeVirt hook sidecar to generate the UUID and set KBS resource?

This question came up when I noticed the `storage` member inside `Provisioner` struct. KBS resource plugin is a plugin, and if there is another `KeyValueStorageInstance` as introduced here with write access, it could potentially race with KBS resource writes.

But yes, if the original idea is that key generation can be triggered by an untrusted entity while keeping the key hidden, then this approach makes sense.

MatiasVara:
> Thanks for adopting previous comments!
> 
> If we want to get this into the repo, let me take a closer look at the design.
> 
> Based on the concrete logic and dataflow of the plugin, I’m wondering what was the reason for choosing a plugin approach instead of `kbs-client`+script inside KubeVirt hook sidecar to generate the UUID and set KBS resource?
> 
My first PoC was something called Secret-Operator that would do exactly that but I had the problem that I can't trust k8s infrastructure so I decided to deploy it in a different infrastructure that I can trust. Then, I though that, to reduce the number of components, I would prefer to have it in trustee. 

> This question came up when I noticed the `storage` member inside `Provisioner` struct. KBS resource plugin is a plugin, and if there is another `KeyValueStorageInstance` as introduced here with write access, it could potentially race with KBS resource writes.
> 
yes, It could. What would be the problem of that? I am trying to understand the implications of this design.

> But yes, if the original idea is that key generation can be triggered by an untrusted entity while keeping the key hidden, then this approach makes sense.

The whole idea is:
image

fitzthum:
> yes, It could. What would be the problem of that? I am trying to understand the implications of this design.

I don't think you should be accessing the resource storage from the plugin. The plugin should have its own storage namespace. This means adding an endpoint to get the resource via the plugin. This may run into issues with the lack of plugin support in the client, but it's still the way to go I think.

MatiasVara:
> > yes, It could. What would be the problem of that? I am trying to understand the implications of this design.
> 
> I don't think you should be accessing the resource storage from the plugin. The plugin should have its own storage namespace. This means adding an endpoint to get the resource via the plugin. This may run into issues with the lack of plugin support in the client, but it's still the way to go I think.

I can understand that would require changes on the client. Should it also require a PR to somehow give support for the plugin?

fitzthum:
> I can understand that would require changes on the client. Should it also require a PR to somehow give support for the plugin?

See https://github.com/confidential-containers/guest-components/pull/1468

kbs/src/plugins/implementations/provisioner.rs
MatiasVara: Yes.

MatiasVara: Rethinking about this, if the hook side car is not trusted then the private key would be compromised too and in that case using `insecureAll` would have the same effect.

MatiasVara:
> > yes, It could. What would be the problem of that? I am trying to understand the implications of this design.
> 
> I don't think you should be accessing the resource storage from the plugin. The plugin should have its own storage namespace. This means adding an endpoint to get the resource via the plugin. This may run into issues with the lack of plugin support in the client, but it's still the way to go I think.

I addressed this comment in last version.

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,274 @@
+// SPDX-License-Identifier: Apache-2.0
+
+//! Provisioner plugin for KBS.
+//!
+//! Generates per-VM LUKS encryption keys and stores them in a dedicated
+//! storage namespace so that attested guests can retrieve them via
+//! `GET /kbs/v0/provisioner/default/{uuid}/root`.
+//!
+//! The hook sidecar calls `POST /kbs/v0/provisioner/provision` before the VM
+//! boots and receives `{uuid, oemstring, mrconfigid}` to inject into SMBIOS.
```
MatiasVara: I need to remove `mrconfigid` word since that is for TDX. In case of sev-snp, the register is `hostData`. At the level the provisioner executes, the name of the register does not mind.

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+use rand::distributions::{Alphanumeric, DistString};
+use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha384};
+use uuid::Uuid;
+
+const PROVISIONER_STORAGE_NAMESPACE: &str = "provisioner";
+
+// Config (deserialized from kbs-config.toml)
+#[derive(Deserialize, Clone, Debug, PartialEq)]
+pub struct ProvisionerConfig {
```
MatiasVara, line 32: Would it make sense to use the Pkcs11Config here in case it is available?

MatiasVara:
@Xynnn007 @fitzthum Would it be OK to move out this PR from draft? Is there something else you want me to review? Thanks.

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+
+// UUID v5 namespace for the provisioner plugin.
+// Derived from: Uuid::new_v5(&Uuid::NAMESPACE_URL, b"provisioner")
+const PROVISIONER_NAMESPACE: Uuid = uuid::uuid!("8e6b2f7a-4c0e-5f02-bf93-2a83a27e4e67");
+
+// Config (deserialized from kbs-config.toml)
+#[derive(Deserialize, Clone, Debug, PartialEq)]
+pub struct ProvisionerConfig {
+    /// URL that will be embedded in `initdata.toml` for the guest.
+    pub kbs_url: String,
```
Xynnn007: Is this a redundant design? Note that the webhook will access the KBS, hinting that it knows the address of KBS. Then the returned initdata also has kbs address inside, meaning  "getting address from a known kbs"?

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+struct StatusResponse {
+    status: String,
+}
+
+impl Provisioner {
+    pub async fn new(
+        config: ProvisionerConfig,
+        storage_backend_config: &StorageBackendConfig,
+    ) -> Result<Self> {
+        let storage = storage_backend_config
```
Xynnn007: We have recently updated the storage API. See https://github.com/confidential-containers/trustee/pull/1417/changes#diff-d058a558543efaf44420a7019071f2afbeb2cd70170f9c8273e7e9e21642e456R105

A rebase would be needed and some apply work.

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,278 @@
...
+}
+
+pub struct Provisioner {
+    storage: KeyValueStorageInstance,
+    kbs_url: String,
+    key_length: usize,
+    // The in-memory cache is an optimization to avoid storage reads on
+    // repeated provisions. It grows unboundedly for the lifetime of the
+    // process but entries are small and bounded by the number of VMs.
+    cache: tokio::sync::RwLock<HashMap<String, ProvisionResponse>>,
```
Xynnn007: We are trying to decouple the storage and computation in architecture. This is for deployment - with multiple same kbs containers in front of a same backend storage instance.

So an idea would be: let's remove the in-process cache design here, as this would make different kbs instances non-idempotent.

kbs/src/plugins/implementations/provisioner/mod.rs
MatiasVara: The KBS is unknown by the guest. The webhook knows where the KBS is and it could add that into the plugin's response. That will require to recalculate mrconfig or just move it to the webhook too. I think this design makes the webhook simpler but I am not sure what is the best approach.

Xynnn007: The assembling logic can lie on the webhook side. In KBS logic, it's strange to have "i know you but you tell me who you are" logic

Xynnn007:
> @Xynnn007 @fitzthum Would it be OK to move out this PR from draft? Is there something else you want me to review? Thanks.

It's ok. And we can keep discussing with the PR!

kbs/src/plugins/implementations/provisioner/mod.rs
MatiasVara: By following this comment, the plugin would return only something like:
```json
  "uuid": "f05c8012-e339-5162-8c06-e08bb2d94fc7",
  "resource_path": "default/f05c8012-e339-5162-8c06-e08bb2d94fc7/root"
```
all the other values like mrconfig and kbs_url would be filled by the webhook or the operator, is that ok?

MatiasVara:
This has been tested using the provisioner-operator at https://github.com/MatiasVara/provisioner-operator

MatiasVara:
@Xynnn007 @fitzthum feel free to let me know if anything is missing. Thanks.

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+    ) -> Result<bool> {
+        // GET: guest fetches the resource after TEE attestation (validate_auth = false).
+        // POST/DELETE: sidecar provisions/deprovisions via admin auth (validate_auth = true).
+        //
+        // NOTE: Currently relies on InsecureAllowAll admin backend for dev.
+        // For production, switch to Simple admin with JWT-signed requests
+        // and scoped roles, e.g.:
+        //   [admin] type = "Simple"
+        //   [[admin.personas]] id = "provisioner" public_key_path = "..."
+        //   [[admin.roles]] id = "provisioner" allowed_endpoints = "^/kbs/v0/provisioner/.*$"
```
Xynnn007: The comments here are out of date. The admin auth now is different, please check https://github.com/confidential-containers/trustee/blob/main/kbs/docs/admin.md if needed

MatiasVara: Can I just remove the comment? Would that work?

Xynnn007: it works

Xynnn007 (approved):
LGTM. Thanks @MatiasVara

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+             \n\
+             [data]\n\
+             \"io.cryptsetup.key.text.root\" = \"{}\"\n",
+            luks_key
+        )
+    }
+
+    async fn handle_get_resource(&self, path: &[&str]) -> Result<Vec<u8>> {
+        let resource_key = path.join("/");
+        let data = self
```
fitzthum, line 121: So to be clear, this plugin doesn't actually check the init-data (which it can't access atm) to make sure that the connection is coming from a particular guest. Does that matter?

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+            .storage
+            .get(&resource_key)
+            .await?
+            .ok_or_else(|| anyhow!("provisioner resource not found: {resource_key}"))?;
+        Ok(data)
```
fitzthum, line 126: So the guest can request the LUKS key as many times as it wants including at runtime. Is that a potential issue?

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+use key_value_storage::{KeyValueStorageInstance, SetParameters, StorageProvider};
+use rand::distributions::{Alphanumeric, DistString};
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+const PROVISIONER_STORAGE_NAMESPACE: &str = "provisioner";
+
+// UUID v5 namespace for the provisioner plugin.
+// Derived from: Uuid::new_v5(&Uuid::NAMESPACE_URL, b"provisioner")
+const PROVISIONER_NAMESPACE: Uuid = uuid::uuid!("8e6b2f7a-4c0e-5f02-bf93-2a83a27e4e67");
```
fitzthum: I guess we haven't really established a convention for naming namespaces. I guess this is fine, but I might prefer something more descriptive. The namespaces aren't particularly dynamic so not too much risk of collision.

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+        })
+    }
+}
+
+impl Provisioner {
+    fn generate_random_key(&self) -> String {
+        Alphanumeric.sample_string(&mut rand::thread_rng(), self.key_length)
+    }
+
+    fn generate_confdata_toml(luks_key: &str) -> String {
```
fitzthum, line 91: nit: maybe call this `luks_key_path` so it doesn't look like it could be the actual key

fitzthum (commented):
Looks pretty good. A couple nits and two high-level questions.

kbs/src/plugins/implementations/provisioner/mod.rs
MatiasVara: Thanks for this question. It made me re think about it. The idea is that trustee would check that the particular guest and the init-data match. I investigated a bit that and this could be achieve using the operator to add the vm-name and the namespace in the init-data which is injected into the guest. Then I need to add the corresponding AS policy to compare. WDYT?

kbs/src/plugins/implementations/provisioner/mod.rs
MatiasVara: It is not an issue. Why would it be?

kbs/src/plugins/implementations/provisioner/mod.rs
MatiasVara: What do you suggeste?

kbs/src/plugins/implementations/provisioner/mod.rs
MatiasVara: `luks_key` is the actual key.

kbs/src/plugins/implementations/provisioner/mod.rs
fitzthum: I think it's probably fine, but it's worth noting that providing the key early in the boot process (i.e. to the initrd) and providing it at runtime have different properties. You can't really control this from Trustee, though without some more advanced stuff.

kbs/src/plugins/implementations/provisioner/mod.rs
fitzthum: Ideally the plugin would do it all. Ding and I are at a stalemate about how to provide the init-data to a plugin tho. I guess this can be handled in a follow-up although it is a fairly significant aspect.

kbs/src/plugins/implementations/provisioner/mod.rs
fitzthum: I would just call it `provisioner_plugin` perhaps. If someone is looking at the database, ideally they will be able to tell what the various namespaces represent.

kbs/src/plugins/implementations/provisioner/mod.rs
fitzthum: ah I thought this was the init-data but it's the secret resource itself. That's fine.

kbs/src/plugins/implementations/provisioner/mod.rs
MatiasVara: I am OK to try to implement it here although I do not know how to do it.

fitzthum: The basic idea would be to change the plugin interface so that it takes the init-data as input. @Xynnn007 has some concerns about this not being generic enough (i.e. some attestation tokens don't have init-data at all), but in this case, the plugin clearly depends on init-data. If init-data is not available it can fail.

MatiasVara: I will propose something and we can discuss it.

MatiasVara: Would it more generic if handle() is defined like that:
```rust
pub struct PluginContext {
    pub attestation_claims: Option<serde_json::Value>,
}
async fn handle(
    &self,
    body: &[u8],
    query: &HashMap<String, String>,
    path: &[&str],
    method: &Method,
    ctx: &PluginContext,
) -> Result<Vec<u8>>;
```

fitzthum: Something like that seems fine. You might also consider just passing in the init-data rather than all of the claims, since this is a bit more well defined. @Xynnn007 has a broader proposal to introduce a "trust context." We don't really agree on that. Maybe we can just start with something specific like you suggest.

MatiasVara: is it OK if I add it to this commit? Or a different commit would be better?

fitzthum: Fine to keep it in this PR but probably a separate commit for changing the interface.

kbs/src/plugins/implementations/provisioner/mod.rs
Xynnn007: This is not resolved btw

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+
+impl Provisioner {
+    pub async fn new(
+        config: ProvisionerConfig,
+        storage_provider: Arc<dyn StorageProvider>,
+    ) -> Result<Self> {
+        let storage = storage_provider
+            .get_or_register(PROVISIONER_STORAGE_NAMESPACE)
+            .await
+            .map_err(|e| anyhow!("Provisioner: failed to init storage backend: {e}"))?;
```
Xynnn007: We can use `.context("Provisioner: failed to init storage backend")?;` here instead of `map_err`

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+                .get("trustee.kbs.resource")
+                .and_then(|v| v.as_str())
+                .ok_or_else(|| anyhow!("init_data missing trustee.kbs.resource"))?;
+            let expected = format!("kbs+provisioner:///{resource_key}");
+            if bound_resource != expected {
+                bail!("init_data resource mismatch: expected {expected}, got {bound_resource}");
+            }
+        } else {
+            bail!("provisioner requires init_data for resource access");
+        }
```
Xynnn007: The code structure here can be changed to 
```
let Some(claims) = init_data else {
    bail!("provisioner requires init_data for resource access");
};
```

This will bring the "if-not-fail" logic early for better reading.

kbs/src/api_server.rs
```diff
@@ -396,7 +396,7 @@ pub(crate) async fn api(
                 // Plugin calls need to be authorized by the admin auth
                 core.admin.check_admin_access(&request)?;
                 let response = plugin
-                    .handle(&body, &query, resource_path, request.method())
+                    .handle(&body, &query, resource_path, request.method(), None)
```
Xynnn007, line 399: the code means "if the plugin does not need auth, it does not need initdata". This might not be expected? We can add it here also.

kbs/src/api_server.rs
```diff
@@ -446,8 +446,11 @@ pub(crate) async fn api(
                 }
                 KBS_POLICY_APPROVALS.inc();
 
+                let init_data = claims
```
Xynnn007, line 449: @fitzthum mentioned my idea of add Truct Context here to support condition like here to deliver attestation contexts. Note that the code here assumes the backend attestation service is CoCo AS, which is coupled. This is not the issue of the PR, but the one of current kbs architecture design.

this is ok for the PR.

I will take a look on the design later

Cargo.toml
```diff
@@ -53,6 +53,7 @@ regorus = { version = "0.11.0", default-features = false, features = [
     "time",
     "std",
 ] }
+rand = "0.8"
```
Xynnn007: rand has newer version `v0.10.2`

Xynnn007 (commented):
More comments. Also please check some previous "resolved comments" which are not resolved

MatiasVara:
> More comments. Also please check some previous "resolved comments" which are not resolved

Thanks for the review. I will address the comments. Apologies for those that I resolved without actually resolving them. I do not know how I missed them.

kbs/src/plugins/implementations/provisioner/mod.rs
MatiasVara: Should I use .context() instead of map_err() in all the cases that is posible?

Xynnn007: That will be good!

kbs/src/api_server.rs
MatiasVara: I am not sure. Are there claims in that path?

fitzthum (approved):
LGTM

fitzthum:
@Xynnn007 are you good with this one? 

@MatiasVara CI broken. looks like the lockfile needs to be fixed.

Xynnn007 (approved):

PR #1306, opened 2026-04-30
KBS API: Add the option to register trusted TPM Attestation Key

iroykaufman:
This pull request introduces support for registering trusted TPM Attestation Key (AK) public keys, exposing new API and CLI commands for registering these keys. 
The implementation follows a similar pattern to register reference value.

### DIFF
--- a/attestation-service/src/lib.rs
+++ b/attestation-service/src/lib.rs
@@ -279,6 +279,43 @@ impl AttestationService {
             .context("query reference values")
     }
 
+    /// Register a trusted TPM Attestation Key (AK) public key.
+    /// The key is stored as a PEM file in the trusted AK keys directory.
+    pub async fn register_attestation_key(&self, key_pem: &str) -> Result<()> {
+        use std::collections::hash_map::DefaultHasher;
+        use std::hash::{Hash, Hasher};
+        use std::io::Write;
+
+        let trusted_ak_keys_dir = self
+            .config
+            .verifier_config
+            .as_ref()
+            .and_then(|c| c.trusted_ak_keys_dir())
+            .unwrap_or_else(|| std::path::PathBuf::from("/etc/tpm/trusted_ak_keys"));
+
+        std::fs::create_dir_all(&trusted_ak_keys_dir).with_context(|| {
+            format!(
+                "Failed to create trusted AK keys directory: {:?}",
+                trusted_ak_keys_dir
+            )
+        })?;
+
+        // Generate a deterministic filename from the key content for idempotency
+        let mut hasher = DefaultHasher::new();
+        key_pem.hash(&mut hasher);
+        let hash = hasher.finish();
+        let filename = format!("{:016x}.pub", hash);
+        let key_path = trusted_ak_keys_dir.join(&filename);
+
+        let mut file = std::fs::File::create(&key_path)
+            .with_context(|| format!("Failed to create AK key file: {:?}", key_path))?;
+        file.write_all(key_pem.as_bytes())
+            .with_context(|| format!("Failed to write AK key file: {:?}", key_path))?;
+
+        info!("Registered TPM attestation key: {}", filename);
+        Ok(())
+    }
+
     pub async fn generate_supplemental_challenge(
         &self,
         tee: Tee,
--- a/kbs/src/api_server.rs
+++ b/kbs/src/api_server.rs
@@ -300,6 +300,21 @@ pub(crate) async fn api(
             Ok(HttpResponse::Ok().content_type("application/json").finish())
         }
 
+        #[cfg(feature = "as")]
+        "attestation-key" if request.method() == Method::POST => {
+            core.admin.check_admin_access(&request)?;
+            let key_pem = std::str::from_utf8(&body).map_err(|_| Error::AttestationKeyError {
+                message: "Failed to parse attestation key: invalid UTF-8".to_string(),
+            })?;
+            core.attestation_service
+                .register_attestation_key(key_pem)
+                .await
+                .map_err(|e| Error::AttestationKeyError {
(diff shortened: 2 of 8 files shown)


attestation-service/src/lib.rs
```diff
@@ -279,6 +279,43 @@ impl AttestationService {
             .context("query reference values")
     }
 
+    /// Register a trusted TPM Attestation Key (AK) public key.
+    /// The key is stored as a PEM file in the trusted AK keys directory.
+    pub async fn register_attestation_key(&self, key_pem: &str) -> Result<()> {
```
mkulke, line 284: I don't this is needs to be async. If we want this to be an async function we have to use async IO methods, not the one from std (I wouldn't recommend it though and just drop async)

attestation-service/src/lib.rs
```diff
@@ -279,6 +279,43 @@ impl AttestationService {
...
+        use std::collections::hash_map::DefaultHasher;
```
mkulke, line 285: can we put those on the top of the file?

attestation-service/src/lib.rs
```diff
@@ -279,6 +279,43 @@ impl AttestationService {
...
+        use std::hash::{Hash, Hasher};
+        use std::io::Write;
+
+        let trusted_ak_keys_dir = self
+            .config
+            .verifier_config
+            .as_ref()
+            .and_then(|c| c.trusted_ak_keys_dir())
+            .unwrap_or_else(|| std::path::PathBuf::from("/etc/tpm/trusted_ak_keys"));
```
mkulke, line 294: let's use a constant DEFAULT_TPM_AK_PATH

mkulke, line 294: nit: std:path:Pathbuf and other fns that are called with their full module path should be imported via use (on top of the file)

kbs/src/api_server.rs
```diff
@@ -300,6 +300,21 @@ pub(crate) async fn api(
             Ok(HttpResponse::Ok().content_type("application/json").finish())
         }
 
+        #[cfg(feature = "as")]
```
mkulke, line 303: should be guarded by a tpm-verifier feature toggle?

I think we are introducing a verifier-specific endpoint as a global primitive. attestation-key is a very generic term, that could apply to a lot of things.

we could either add a clarifying prefix (`.../tpm-attestation-key`) or, which is probably more consistent, introduce a new hierarchy for endpoints like this (`.../verifiers/tpm/attestation-key`, `.../verifiers/snp/ark`, ...)

iroykaufman: Agreed, it'll be better to use tpm-verifier, although it can work without the `tpm-verifier`, it will be useless.
Regarding the endpoint, I understand that `attestation-key` is not the best naming, and I think going with the hierarchy endpoints is better.

mkulke: let's wait for some more input

mkulke (commented):
note: the current proposal introduces statefulness based purely on the filesystem ops, which has negative implications for deployments in kubernetes and is probably not reliable in an API service in its current form. e,g, a register op is not atomic and subject to races.

If we want to manage tpm attestation keys as resources in KBS, then we will have to use the storage subsystem that has been introduced to manage state in a reliable fashion:

also, I think the whole CRUD surface should be covered for such a resource, similar to policies

fitzthum:
Adding a new KBS API here seems a little fishy, especially when I don't really see how these keys are being consumed. Why can't you use the existing reference value flow for this?

iroykaufman:
> Adding a new KBS API here seems a little fishy, especially when I don't really see how these keys are being consumed. Why can't you use the existing reference value flow for this?

In my case I want to add trusted TPM AK, currently the only way it seems possible to do that is to manually add the file with the key to the correct directory, as explained in this [readme](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/tpm/README.md). I want to automate this using the API.

mkulke:
> > Adding a new KBS API here seems a little fishy, especially when I don't really see how these keys are being consumed. Why can't you use the existing reference value flow for this?
> 
> In my case I want to add trusted TPM AK, currently the only way it seems possible to do that is to manually add the file with the key to the correct directory, as explained in this [readme](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/tpm/README.md). I want to automate this using the API.

As an altarnative to an API endpoint for the purpose of automation you can maybe use a kubernetes configmap as backing store for the attestation-keys in Trustee and use the k8s api/tooling to manage the keys.

As mentioned, doing an API integration properly for the AKs requires more work and there are open questions.

fitzthum:
The vTPM code is pretty new. Rather than trying to bolt new features onto what we currently have, let's take a step back and think about the high-level design.

Rather than storing allowed AK pub hashes in some separate file, let's have the verifier report the ak public key (already part of the evidence) and/or the public key hash as a tcb claim. Then you can write an attestation policy that compares this claim to a reference value. To set the allowed public keys, you just set the reference value.

This will stick closer to the flow of other verifiers and will come in handy later when we want to check the binding between this public key and some other device.

iroykaufman:
> Rather than storing allowed AK pub hashes in some separate file, let's have the verifier report the ak public key (already part of the evidence) and/or the public key hash as a tcb claim. Then you can write an attestation policy that compares this claim to a reference value. To set the allowed public keys, you just set the reference value.

This solution sounds good. This involves removing the mechanism for checking the trusted AK in the vTPM verifier and instead checking this in the attestation policy. I'll implement this solution and update the PR.

iroykaufman:
I'm closing this PR in favor of [PR#1328](https://github.com/confidential-containers/trustee/pull/1328) following @fitzthum's suggestion.

PR #1307, opened 2026-04-30
kbs: fix boolean feature flags being ignored when FEATURES is passed on command line

bpradipt:
GNU Make treats command-line variables as overrides, so assignments like 'FEATURES := $(FEATURES),vault' inside the Makefile are silently ignored when FEATURES is already set via the command line (e.g. FEATURES=pkcs11). Use the 'override' directive so that VAULT, ALIYUN, NEBULA_CA_PLUGIN, and EXTERNAL_PLUGIN conditionals always append to FEATURES regardless of how FEATURES was initially set.

Found by @aramirez-capacity
https://github.com/confidential-containers/trustee/pull/1288#issuecomment-4344076591

### DIFF
--- a/kbs/Makefile
+++ b/kbs/Makefile
@@ -50,19 +50,19 @@ else ifneq ($(AS_TYPE), )
 endif
 
 ifeq ($(ALIYUN), true)
-  FEATURES := $(FEATURES),aliyun
+  override FEATURES := $(FEATURES),aliyun
 endif
 
 ifeq ($(NEBULA_CA_PLUGIN), true)
-  FEATURES := $(FEATURES),nebula-ca-plugin
+  override FEATURES := $(FEATURES),nebula-ca-plugin
 endif
 
 ifeq ($(VAULT), true)
-  FEATURES := $(FEATURES),vault
+  override FEATURES := $(FEATURES),vault
 endif
 
 ifeq ($(EXTERNAL_PLUGIN), true)
-  FEATURES := $(FEATURES),external-plugin
+  override FEATURES := $(FEATURES),external-plugin
 endif
 
 ifeq ($(CLI_FEATURES),sample_only)


fitzthum (approved):
LGTM

@confidential-containers/trustee-maintainers any concerns about tweaking the Makefile here. seems probably ok to me

mythi:
I also think it's OK but the whole features construct seems complex and the alternative would be to drop these =true vars and ask users to specify the FEATURES they need?

bpradipt:
> I also think it's OK but the whole features construct seems complex and the alternative would be to drop these =true vars and ask users to specify the FEATURES they need?

Something like having a default FEATURES set and let users override it ?

mythi:
> > I also think it's OK but the whole features construct seems complex and the alternative would be to drop these =true vars and ask users to specify the FEATURES they need?
> 
> Something like having a default FEATURES set and let users override it ?

Yes that's what I had in mind.

bpradipt:
> > > I also think it's OK but the whole features construct seems complex and the alternative would be to drop these =true vars and ask users to specify the FEATURES they need?
> > 
> > 
> > Something like having a default FEATURES set and let users override it ?
> 
> Yes that's what I had in mind.

What do you say @fitzthum ?

mythi:
Should have clarified that it could go to a later PR.

fitzthum:
Yeah, I've always found the Makefile to be a bit convoluted, but let's fix the bug first.

bpradipt:
@mythi @fitzthum based on the discussions so far are we in agreement to proceed with this PR as-is?
Any other changes required for now ?

mythi (approved):

Xynnn007:
> Yeah, I've always found the Makefile to be a bit convoluted, but let's fix the bug first.

we can consider using `.env` or something similar, to declare the flags in a single "configuration" file, than providing them in cmdline

Issue #1273, opened 2026-04-13
Trustee Storage | Add per component storage override support


Xynnn007:
Now in one-in-all KBS, we have a unified storage configuration item. But this will set all the underlying storage plugin with a same type. We might need to have a mechanism to override the "global" config for each concrete component.

Xynnn007:
https://github.com/confidential-containers/trustee/pull/1327

PR #1312, opened 2026-05-01
Add RVPS storage override support

lmilleri:
It partially addresses issue #1273, by providing the option to override the RVPS storage. If RVPS storage not provided, it fallbacks to the general storage configuration.

### DIFF
--- a/attestation-service/docs/config.md
+++ b/attestation-service/docs/config.md
@@ -60,8 +60,11 @@ If `type` is set to `BuiltIn`, the following extra properties can be set:
 | Property | Type | Description | Required | Default |
 |----------|------|-------------|----------|---------|
 | `extractors` | Object | Optional configuration for provenance extractors | No | None |
+| `storage` | [StorageBackendConfig][4] | Optional RVPS-specific storage configuration. If provided, overrides the unified `storage_backend` for RVPS only. If omitted, falls back to the unified `storage_backend`. | No | None |
 
-**Note:** Storage configuration for BuiltIn RVPS is now managed through the unified `storage_backend` configuration (see [Storage Backend Configuration](#storage-backend-configuration)). The BuiltIn RVPS will use the `reference_value` namespace from the unified storage backend.
+> [!NOTE]
+> **Storage Configuration:** By default, BuiltIn RVPS uses the unified `storage_backend` configuration (see [Storage Backend Configuration](#storage-backend-configuration)) with the `reference_value` namespace. However, you can optionally provide a `storage` field to configure RVPS-specific storage that differs from other components.
+> This allows you to use different storage backends for different components (e.g., LocalFs for KBS and AS policies, but LocalJson for RVPS reference values).
 
 For detailed information about extractors configuration, including available extractors and their options, see the [RVPS README](../../rvps/README.md#extractors-configuration).
 
@@ -217,6 +220,35 @@ Running with a built-in RVPS with extractor configuration:
         "type": "BuiltIn",
         "extractors": {
             "swid_extractor": {}
+        }
+    },
+    "attestation_token_broker": {
+        "duration_min": 5
+    }
+}
+```
+
+Running with a built-in RVPS with RVPS-specific storage configuration:
+
+```json
+{
+    "storage_backend": {
+        "storage_type": "LocalFs",
+        "backends": {
+            "local_fs": {
+                "dir_path": "/var/lib/attestation-service/storage"
+            }
+        }
+    },
+    "rvps_config": {
+        "type": "BuiltIn",
+        "storage": {
+            "storage_type": "LocalJson",
+            "backends": {
+                "local_json": {
+                    "file_dir_path": "/var/lib/rvps/references"
+                }
+            }
         }
     },
     "attestation_token_broker": {
--- a/attestation-service/src/rvps/mod.rs
+++ b/attestation-service/src/rvps/mod.rs
@@ -5,7 +5,7 @@
 
 use std::sync::Arc;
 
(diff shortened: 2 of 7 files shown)


attestation-service/src/rvps/mod.rs
```diff
@@ -80,10 +89,34 @@ pub async fn initialize_rvps_client(
...
     match config {
-        RvpsConfig::BuiltIn { extractors } => {
+        RvpsConfig::BuiltIn {
+            extractors,
+            storage,
+        } => {
             info!("launch a built-in RVPS.");
+
+            // Use RVPS-specific storage if provided, otherwise fall back to defaults
+            let (actual_storage_type, actual_storage_config) = if let Some(rvps_storage) = storage {
```
Xynnn007: can we use the following structure to avoid a `if let` to be on the right of the `=`?

```suggestion
            let (actual_storage_type, actual_storage_config) = match storage {
                 Some(rvps_storage) => {...}
                 None => {...}
```

rvps/README.md
```diff
@@ -231,6 +231,62 @@ Using in-memory storage (default):
 
 ## Integrate RVPS into the Attestation Service
 
+### Built-In Mode (Recommended for most deployments)
```
Xynnn007: Note that this new section is the same as `Native Mode (Not Recommend)` as both uses rvps as a crate. We'd better merge the required contents into the `Native Mode (Not Recommend)` section

kbs/docs/config.md
```diff
@@ -127,8 +127,9 @@ If `type` is set to `BuiltIn`, the following extra properties can be set:
...
 |----------|------|-------------|----------|---------|
 | `extractors` | Object | Optional configuration for provenance extractors | No | None |
+| `storage` | Object | Optional RVPS-specific storage configuration (same format as `storage_backend`). If provided, overrides the unified `storage_backend` for RVPS only. If omitted, falls back to the unified `storage_backend`. | No | None |
 
-**Note:** Storage configuration for BuiltIn RVPS is now managed through the unified `storage_backend` configuration (see [Storage Backend Configuration](#storage-backend-configuration)). The BuiltIn RVPS will use the `reference_value` namespace from the unified storage backend.
+**Storage Configuration:** By default, BuiltIn RVPS uses the unified `storage_backend` configuration (see [Storage Backend Configuration](#storage-backend-configuration)) with the `reference_value` namespace. However, you can optionally provide a `storage` field to configure RVPS-specific storage that differs from other components. This allows you to use different storage backends for different components (e.g., LocalFs for KBS resources, but LocalJson for RVPS reference values).
```
Xynnn007: Could you help to turn this to [a markdown highlight](https://github.com/orgs/community/discussions/16925) Note here?

Xynnn007 (commented):
Thanks @lmilleri ! As I promised, I will give a migration guide from old version to the new version if you still need them. Also some questions:
1. Do you folks agree to remove localfs/localjson part from trustee system? The reason is that we can use postgres as storage, which looks more "production". To provision reference values, the deployment could use a job to call `kbs-client` to set the reference values.
2. How trustee-operator considers the HA? For the helm PR, we are seeking to mark all micro services as `deployment` than `statefulset`.

lmilleri:
Thank you @Xynnn007, please have another review

lmilleri:
@Xynnn007 I'd rather keep the localfs/localjson backends in trustee until we have a clear migration plan in the trustee-operator. It’s definitely nice using Postgres, but I wonder if it’s a bit too much of a commitment for production?
Re HA, the trustee operator uses a deployment and the etcd database for data replication

attestation-service/docs/config.md
```diff
@@ -60,8 +60,9 @@ If `type` is set to `BuiltIn`, the following extra properties can be set:
...
 |----------|------|-------------|----------|---------|
 | `extractors` | Object | Optional configuration for provenance extractors | No | None |
+| `storage` | [StorageBackendConfig][4] | Optional RVPS-specific storage configuration. If provided, overrides the unified `storage_backend` for RVPS only. If omitted, falls back to the unified `storage_backend`. | No | None |
 
-**Note:** Storage configuration for BuiltIn RVPS is now managed through the unified `storage_backend` configuration (see [Storage Backend Configuration](#storage-backend-configuration)). The BuiltIn RVPS will use the `reference_value` namespace from the unified storage backend.
+**Storage Configuration:** By default, BuiltIn RVPS uses the unified `storage_backend` configuration (see [Storage Backend Configuration](#storage-backend-configuration)) with the `reference_value` namespace. However, you can optionally provide a `storage` field to configure RVPS-specific storage that differs from other components. This allows you to use different storage backends for different components (e.g., LocalFs for KBS and AS policies, but LocalJson for RVPS reference values).
```
Xynnn007: Could you help to add this to highlight Note also?

Xynnn007:
> @Xynnn007 I'd rather keep the localfs/localjson backends in trustee until we have a clear migration plan in the trustee-operator. It’s definitely nice using Postgres, but I wonder if it’s a bit too much of a commitment for production? Re HA, the trustee operator uses a deployment and the etcd database for data replication

For production I believe postgres is better than localfs/localjson. But I'd respect your ideas to keep them in code before you get plans!

Xynnn007 (approved):

fitzthum (approved):
LGTM

The storage config has turned out to be a little messier than I had hoped, but this PR seems fine.

It seems like if we're using built-in mode, the RVPS should maybe share the top-level storage backend (i.e. define two backends at the top level; one for localfs and one for local json). Let's leave that aside for now, though.

PR #1319, opened 2026-05-06
tests: adjust timeout in integration test

fitzthum:
Last time we had intermittent issues with the integration tests it was related to this timeout. Let's try increasing it. Possibly the recent changes to the config and storage backend have increased startup time.

If this doesn't change anything I will investigate more closely.

### DIFF
--- a/integration-tests/src/common.rs
+++ b/integration-tests/src/common.rs
@@ -41,7 +41,7 @@ use tracing_subscriber::fmt;
 
 const KBS_URL: &str = "http://127.0.0.1:8081";
 const RVPS_URL: &str = "http://127.0.0.1:51003";
-const WAIT_TIME: u64 = 5000;
+const WAIT_TIME: u64 = 10000;
 
 const ALLOW_ALL_POLICY: &str = "
     package policy


Xynnn007:
My suggestion is to address this problem at its root. Since a waiting way often means we don't know if a specific process or procedure has finished executing. Could we use a mechanism to check for completion, or a message to notify us when the corresponding process has finished?

mkulke:
copilot [session](https://github.com/confidential-containers/trustee/tasks/b3475631-81db-4276-8dbc-490bfbb0caba?session_id=1d0358a2-f2f4-497a-9da0-d07b29b7d5f6) concluded this:

<blockquote>
Yes — the root race is startup readiness vs. first use (not test logic itself).

In /home/runner/work/trustee/trustee/integration-tests/src/common.rs:

RVPS is started with tokio::spawn(rvps_future) (returns immediately).
Then KBS built-in AS is initialized and, for remote RVPS, it does a real gRPC connect during init (Agent::new(...)->pool.get()->Channel::connect()), which is effectively one-shot at startup.
If RVPS hasn’t finished binding yet, that connect path races and fails intermittently.
So the timeout increase in PR #1319 is masking this by waiting longer, but the underlying condition is: no deterministic readiness signal/handshake between spawned servers and dependent clients (RVPS and also KBS server start before first test requests).
</blockquote>

fitzthum:
Yes, first time around I found that it was a problem with the RVPS startup. We didn't have problems in built-in mode. Would be better to have a more sophisticated mechanism for checking availability. I will think about that.

Xynnn007:
I think this can be closed by https://github.com/confidential-containers/trustee/pull/1476?

PR #1322, opened 2026-05-07
KBS: rename repository config from 'type' to 'storage_backend_type'

### LINKED ISSUE #1318: Plugin config entry ambiguous
After our kvstorage reshuffle, the config for a plugin looks like this.

```
[[plugins]]
name = "resource"
type = "kvstorage"
```

I find this to be pretty confusing. Particularly, the type field should probably be renamed. This is specifying the storage type (right?), but it seems like it's specifying the plugin type.

Also, it's slightly confusing that the heading is `plugins` when this is the entry for one plugin, but this may just be a limitation of how toml handles arrays.

cc @Xynnn007

Xynnn007:
After our kvstorage reshuffle, the config for a plugin looks like this.
```
[[plugins]]
name = "resource"
type = "kvstorage"
```

It's confusing that type field is specifying the storage type but seems a plugin name.

This commit changes the name `type` to `backend_storage_type` to explicitly show what it means.

Close #1318

### DIFF
--- a/kbs/config/docker-compose/kbs-config.toml
+++ b/kbs/config/docker-compose/kbs-config.toml
@@ -18,7 +18,7 @@ public_key_path = "/opt/confidential-containers/kbs/user-keys/public.pub"
 
 [[plugins]]
 name = "resource"
-type = "kvstorage"
+storage_backend_type = "kvstorage"
 
 [storage_backend]
 storage_type = "LocalFs"
--- a/kbs/config/kbs-config-grpc.toml
+++ b/kbs/config/kbs-config-grpc.toml
@@ -21,4 +21,4 @@ dir_path = "/opt/confidential-containers/storage"
 
 [[plugins]]
 name = "resource"
-type = "kvstorage"
+storage_backend_type = "kvstorage"
--- a/kbs/config/kbs-config-intel-trust-authority.toml
+++ b/kbs/config/kbs-config-intel-trust-authority.toml
@@ -22,4 +22,4 @@ dir_path = "/opt/confidential-containers/storage"
 
 [[plugins]]
 name = "resource"
-type = "kvstorage"
+storage_backend_type = "kvstorage"
--- a/kbs/config/kbs-config.toml
+++ b/kbs/config/kbs-config.toml
@@ -33,13 +33,13 @@ dir_path = "/opt/confidential-containers/storage"
 
 [[plugins]]
 name = "resource"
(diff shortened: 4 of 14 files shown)


mythi:
> After our kvstorage reshuffle, the config for a plugin looks like this.

I believe kvstorage resuffle did not change this so it's been there longer that.

Can you clarify how this works after the PR because I can see there's also something like:

```toml
[backend_storage]
storage_type = "LocalFs"
```

also I see the external plugin has some `backends` config.

Xynnn007:
@mythi As the example, 
```toml
[[plugins]]
name = "resource"
type = "kvstorage"
```

the field `type` is now in `resource` plugin to specify which backend resource storage is used, e.g. `Vault`, `kvstorage` (following the unified storage backend), `alibaba` (Alibaba Cloud KMS).

After the PR, it should be like
```toml
[[plugins]]
name = "resource"
backend_storage_type = "kvstorage"
```

another example is 
```toml
[[plugins]]
name = "resource"
backend_storage_type = "Vault"
```

kbs/config/docker-compose/kbs-config.toml
```diff
@@ -18,7 +18,7 @@ public_key_path = "/opt/confidential-containers/kbs/user-keys/public.pub"
 
 [[plugins]]
 name = "resource"
-type = "kvstorage"
+backend_storage_type = "kvstorage"
```
mkulke: can this be "storage_backend_type" or just "storage_type"? "backend" indicates an implementation, "type" indicates an abstract class, I think kvstorage is the latter (`trait KeyValueStore`)

Xynnn007: yes `storage_backend_type` could be good.

mythi: and users would then have something like?

```toml
[[plugins]]
name = "resource"
storage_backend_type = "kvstorage"

[backend_storage]
storage_type = "LocalFs"
```

Xynnn007: currently yes. maybe 
```toml
[[plugins]]
name = "resource"
resource_backend_type = "kvstorage"

[backend_storage]
storage_type = "LocalFs"
```
?

fitzthum (approved):
LGTM 


I think the overall `storage_backend` entry could possibly be clarified, but that's a separate question.

mkulke (approved):

PR #1327, opened 2026-05-09
KBS/AS: add unified storage overwriting logic

Xynnn007:
This is a patch based on #1312. We could use a global storage config, and set concrete overwrite with types.

This commit also updates related documents and examples.

cc @lmilleri


fitzthum (approved):
LGTM

Xynnn007:
let me get this merged as I see @lmilleri showed thumb up

PR #1328, opened 2026-05-10
TPM: Update attestation policy to check trusted AK against reference values

iroykaufman:
This pull request simplifies the TPM verifier configuration by removing support for pre-configured trusted AK (Attestation Key) directories and instead requiring trusted AKs to be set as reference values and checked via attestation policy.

**Core changes:**

* Removed support for loading trusted AK public keys from a directory.

* Updated the default TPM policy to check that `input.tpm.ak_public` is in the trusted AK reference value.

* Revised the TPM verifier documentation to remove the section on directory-based configuration.

* Modified evidence parsing to include the `ak_public` field in the claim.


deps/verifier/src/tpm/mod.rs
```diff
@@ -288,95 +209,50 @@ impl Verifier for TpmVerifier {
             .to_quote()
             .context("Failed to convert quote from string format")?;
 
-        let ak_public = guest_ev.ak_public;
+        let ak_public = guest_ev.ak_public.clone();
```
Xynnn007: We might not need this `.clone()`

Xynnn007 (commented):
Thanks for this! Another place should be changed is [the doc](https://github.com/confidential-containers/trustee/blob/main/attestation-service/docs/tcb_claims.md#tpm) 

Also, cc @bpradipt to take a look

attestation-service/src/ear_token/ear_default_policy_cpu.rego
```diff
@@ -312,6 +312,8 @@ executables := 3 if {
 	input.tpm
 
 	input.tpm.pcr11 in query_reference_value("tpm_pcr11")
+
+	input.tpm.ak_public in query_reference_value("trusted_ak")
```
fitzthum: This reference value should probably be called `trusted_aks`. It is referring to a list of AKs (hence the `in` operator).

Also,  I'm not sure if this check should be in just one of the vectors or in all of the TPM-related ones.

deps/verifier/src/lib.rs
```diff
@@ -54,9 +54,6 @@ pub struct VerifierConfig {
     #[cfg(feature = "nvidia-verifier")]
     nvidia_verifier: Option<nvidia::NvidiaVerifierConfig>,
 
-    #[cfg(feature = "tpm-verifier")]
-    tpm_verifier: Option<tpm::config::TpmVerifierConfig>,
```
fitzthum, line 58: I suspect we're going to need this config later for some other stuff, so you might want to keep it with an empty config struct.

fitzthum (commented):
Looks pretty good. Some minor comments.

mkulke (approved):
looks good. it makes a lot of sense to use the reference values

bpradipt:
@iroykaufman the tpm related test files under kbs/test will need changes as well.

deps/verifier/src/tpm/mod.rs
```diff
@@ -288,95 +209,50 @@ impl Verifier for TpmVerifier {
...
         Ok(vec![(claims, "cpu".to_string())])
     }
 }
 
-pub fn parse_tee_evidence(quote: &VtpmQuote) -> TeeEvidenceParsedClaim {
+pub fn parse_tee_evidence(guest_ev: &Evidence) -> TeeEvidenceParsedClaim {
+    let quote = guest_ev
+        .tpm_quote
+        .to_quote()
+        .expect("Failed to convert quote from string format");
```
bpradipt: Looks like `to_quote` is called twice now, once in evaluate and then again in this function.

iroykaufman:
Thanks for the comments. I just updated the PR addressing all comments.
Let me know if something is missing

Xynnn007 (approved):

PR #1332, opened 2026-05-11
key-value-storage: add redis backend

### LINKED ISSUE #1106: [KV Storage]  Add Redis read/write plugin
This requirement stems from the decoupling of computation and storage in #1092. For most storage needs, a database can handle it. However, for session state—which resides in the KBS and requires maintaining remote attestation state—rapid processing in high-concurrency scenarios may be necessary. This issue aims to bring in redis client as one key value storage plugin to fullfill such case.

Xynnn007:
Redis is useful to store the in-memory data, like the kbs-sessions. This patch includes the code, README.md and the unit cases, also the updates to the places where refers to storage config.

Close #1106


fitzthum (approved):
LGTM

fitzthum:
@confidential-containers/trustee-maintainers PTAL

Xynnn007:
Let me get this merged until the next Mon if no blockers. I will start to test redis (kbs-session)+postgres(other persistent storage) helm deployment. Any errors we face can be resolved by PR ad-hoc.

mkulke (commented):
Is this compliant with CNCF requirements if we introduce a dependency on Redis?

Redis had some licensing journey in recent years (currently AGPL afaik), which is prohibitive for some deployments.

I think we should be able to switch to Valkey, which is a popular fork of Redis, before it started changing its license.

fitzthum:
Looks like the redis crate itself is BSD-3 which is compatible with Apache-2 so I think we can move forward here, although I wouldn't mind supporting valkey as well.

mkulke:
> Looks like the redis crate itself is BSD-3 which is compatible with Apache-2 so I think we can move forward here, although I wouldn't mind supporting valkey as well.

indeed, the crate should work fine with valkey. If we plan to add redis to the reference helm charts, we should probably use valkey instead.

deps/key-value-storage/src/redis/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+            .keys(pattern)
+            .await
+            .map_err(|e| KeyValueStorageError::ListKeysFailed { source: e.into() })?;
+
+        Ok(keys
+            .into_iter()
+            .filter_map(|key| self.strip_namespace(&key).map(ToString::to_string))
+            .collect())
+    }
+
```
mkulke, line 123: the prefix to strip is constructed on the heap (format => string) for each key, this can be simplified:

```suggestion
        let prefix = format!("{}:", self.namespace)

        Ok(keys
            .into_iter()
            .map(|key| key.strip_prefix(&prefix).unwrap_or(key).into())
            .collect())
    }

```

Xynnn007:
Link error is not related to this Pr. Fixed in #1369

Xynnn007:
@mkulke PTAL.

mythi:
> > Looks like the redis crate itself is BSD-3 which is compatible with Apache-2 so I think we can move forward here, although I wouldn't mind supporting valkey as well.
> 
> indeed, the crate should work fine with valkey. If we plan to add redis to the reference helm charts, we should probably use valkey instead.

On that note, some places in this PR could refer to "Redis protocol" rather than just "Redis" to suggests this is not tied to any particular implementation.

Xynnn007:
> > > Looks like the redis crate itself is BSD-3 which is compatible with Apache-2 so I think we can move forward here, although I wouldn't mind supporting valkey as well.
> > 
> > 
> > indeed, the crate should work fine with valkey. If we plan to add redis to the reference helm charts, we should probably use valkey instead.
> 
> On that note, some places in this PR could refer to "Redis protocol" rather than just "Redis" to suggests this is not tied to any particular implementation.

Updated the README.md

deps/key-value-storage/README.md
```diff
@@ -50,6 +50,15 @@ Stores key-value pairs in a PostgreSQL database table.
 
 See [Postgres README](src/postgres/README.md) for more details.
 
+### Redis (`Redis`)
```
mkulke: I agree with Mikko's suggestion that we should maybe point out that it works with Redis-compatible stores like Valkey, elasticache, etc.

mkulke (approved):

PR #1333, opened 2026-05-11
verifier: restrict intel_dcap module items to crate visibility

mythi:
intel_dcap is not meant to be public. It's only serving SGX/TDX verifiers with helper functions and types. Change to pub(crate) to reflect.

Similarly, restrict TcbUpdateType, QcnlConfig, set_qcnl_config, ecdsa_quote_verification and extend_using_custom_claims to pub(crate) for the same reason.


fitzthum (approved):
Ok

Xynnn007 (approved):

PR #1289, opened 2026-04-22
verifier: tdx/az_tdx_vtpm: add platform_instance_id claim 

mythi:

PR #1335, opened 2026-05-12
verifier: intel_dcap: add PlatformInfo parsing and move to use it

mythi:
#1289 focused on platfom instance ID but new needs came in shortly after so just expose full Platform Info to callers. 

PlatformInfo is a struct carrying all fields from the SGX extension (OID 1.2.840.113741.1.13.1) of a PCK certificate: fmspc, pceid, 16 TCB SVN components, pcesvn, cpusvn, sgx_type, and the Platform CA-only fields (platform_instance_id, dynamic_platform, cached_keys, smt_enabled).

parse_platform_info() takes raw PEM bytes from the QE certification data field and parses all three certs in the chain. The CA type (is_platform_ca) is derived from the intermediate cert subject CN and cross-checked against the presence of platform_instance and configuration in the leaf cert's SGX extensions.

generate_parsed_claim() is updated to take Option<&PlatformInfo> instead of raw cert bytes; tdx/mod.rs and az_tdx_vtpm/mod.rs call parse_platform_info() before invoking it. The old
platform_instance_id_from_pck_leaf_cert() is removed.

Assisted-by: Copilot:claude-sonnet-4.6

### DIFF
--- a/deps/verifier/src/az_tdx_vtpm/mod.rs
+++ b/deps/verifier/src/az_tdx_vtpm/mod.rs
@@ -13,7 +13,9 @@ use super::az_snp_vtpm::{
 use super::tdx::claims::generate_parsed_claim;
 use super::tdx::quote::{parse_tdx_quote, parse_tdx_quote_certification, Quote as TdQuote};
 use super::{TeeClass, TeeEvidence, TeeEvidenceParsedClaim, Verifier};
-use crate::intel_dcap::{ecdsa_quote_verification, extend_using_custom_claims};
+use crate::intel_dcap::{
+    ecdsa_quote_verification, extend_using_custom_claims, pck::parse_platform_info,
+};
 use crate::{InitDataHash, ReportData};
 use anyhow::{bail, Context, Result};
 use async_trait::async_trait;
@@ -66,8 +68,9 @@ impl Verifier for AzTdxVtpm {
         let pck_certs = parse_tdx_quote_certification(evidence.td_quote(), &td_quote)?
             .qe_certification_data
             .certificates;
+        let platform_info = parse_platform_info(&pck_certs)?;
 
-        let mut claim = generate_parsed_claim(td_quote, None, Some(&pck_certs))?;
+        let mut claim = generate_parsed_claim(td_quote, None, &platform_info)?;
         extend_claim(&mut claim, &tpm_quote)?;
         extend_using_custom_claims(&mut claim, custom_claims)?;
 
--- a/deps/verifier/src/intel_dcap/pck.rs
+++ b/deps/verifier/src/intel_dcap/pck.rs
@@ -8,11 +8,13 @@
 //! and are present in both TDX and SGX PCK certificate chains.
 //! See "Intel® SGX PCK Certificate and Certificate Revocation List Profile Specification".
 
-use anyhow::{Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use asn1_rs::{oid, DerSequence, Enumerated, FromDer, Oid};
 use x509_parser::prelude::*;
 
 const DCAP_SGX_EXTENSIONS: Oid<'static> = oid!(1.2.840 .113741 .1 .13 .1);
+const PCK_PLATFORM_CA_CN: &str = "Intel SGX PCK Platform CA";
+const PCK_PROCESSOR_CA_CN: &str = "Intel SGX PCK Processor CA";
 
 #[derive(Debug, PartialEq, DerSequence)]
 struct OidAndString<'a> {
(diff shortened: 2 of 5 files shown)


deps/verifier/src/intel_dcap/pck.rs
```diff
@@ -92,68 +92,205 @@ struct SgxExtension<'a> {
...
+    pub dynamic_platform: Option<bool>,
+    pub cached_keys: Option<bool>,
+    pub smt_enabled: Option<bool>,
 }
 
 /// Parse all PEM-encoded certificates from a PCK certificate chain.
 /// The Intel cert chain ordering is leaf (index 0), intermediate CA, root CA.
-pub(crate) fn parse_pck_pem_certs(pem_certs: &[u8]) -> Result<Vec<Pem>> {
-    Pem::iter_from_buffer(pem_certs)
+fn parse_pck_pem_certs(pem_certs: &[u8]) -> Result<Vec<Pem>> {
```
Xynnn007: Let's define a new struct named `PckCertificateChain` like
```rust
struct PckCertificateChain {
    intermediate: Pem,
    leaf: Pem,
    root: Pem,
}
```
to explicitly show the structure. In this way we can avoid using `[0]` like indexs which can easily cause panic.

deps/verifier/src/intel_dcap/pck.rs
```diff
@@ -92,68 +92,205 @@ struct SgxExtension<'a> {
...
+        .subject()
+        .iter_common_name()
+        .next()
+        .context("PCK intermediate CA cert has no Common Name")?
+        .as_str()
+        .context("PCK intermediate CA CN is not valid UTF-8")?;
+
+    let is_platform_ca = match cn {
+        "Intel SGX PCK Platform CA" => true,
+        "Intel SGX PCK Processor CA" => false,
```
Xynnn007: can we make the two as consts?

deps/verifier/src/intel_dcap/pck.rs
```diff
@@ -92,68 +92,205 @@ struct SgxExtension<'a> {
...
-    let ext = cert
+/// Parse the SGX extensions from a PCK certificate chain and return owned platform information.
+///
+/// The CA type is derived from the intermediate cert subject CN, which also determines
+/// which SGX extension fields are present. Absent SGX extensions is an error — PCK
+/// leaf certificates are always expected to carry them.
+pub(crate) fn parse_platform_info(pem_certs: &[u8]) -> Result<PlatformInfo> {
+    let pems = parse_pck_pem_certs(pem_certs)?;
+
+    let intermediate = pems[1]
```
Xynnn007: See previous suggestion to make a PckCertificate structure

deps/verifier/src/intel_dcap/pck.rs
```diff
@@ -92,68 +92,205 @@ struct SgxExtension<'a> {
...
+        .as_str()
+        .context("PCK intermediate CA CN is not valid UTF-8")?;
+
+    let is_platform_ca = match cn {
+        "Intel SGX PCK Platform CA" => true,
+        "Intel SGX PCK Processor CA" => false,
+        other => bail!("unexpected PCK intermediate CA CN: {other}"),
+    };
+
+    let leaf = pems[0]
```
Xynnn007: See previous suggestion to make a PckCertificate structure

deps/verifier/src/intel_dcap/pck.rs
```diff
@@ -92,68 +92,205 @@ struct SgxExtension<'a> {
...
+        tcb.comp15.val,
+        tcb.comp16.val,
+    ];
+
+    let cpusvn: [u8; 16] = tcb.cpusvn.s.try_into().context("cpusvn is not 16 bytes")?;
+    let sgx_type = sgx.sgxtype.e.0 as u8;
+
+    // Cross-check: the CA type from the intermediate cert must agree with the presence
+    // of platform_instance and configuration in the leaf cert's SGX extensions.
+    let (platform_instance_id, dynamic_platform, cached_keys, smt_enabled) = if is_platform_ca {
```
Xynnn007: Let's use 
```rust
let X = match is_platform_ca {
    true => {...}
    false => {...}
}
```

pattern instead of `let  X = if {} else {}` for readibilities.

deps/verifier/src/az_tdx_vtpm/mod.rs
```diff
@@ -66,8 +68,9 @@ impl Verifier for AzTdxVtpm {
         let pck_certs = parse_tdx_quote_certification(evidence.td_quote(), &td_quote)?
             .qe_certification_data
             .certificates;
+        let platform_info = parse_platform_info(&pck_certs)?;
 
-        let mut claim = generate_parsed_claim(td_quote, None, Some(&pck_certs))?;
+        let mut claim = generate_parsed_claim(td_quote, None, Some(&platform_info))?;
```
Xynnn007: Looks like the third parameter is always given as `Some(..)`. Do you consider to make it `platform_info: &PlatformInfo,` than `platform_info: Option<&PlatformInfo>,`?

Xynnn007 (commented):
Thanks @mythi The code looks good. Let's make it better.

deps/verifier/src/az_tdx_vtpm/mod.rs
mythi: `test_generate_parsed_claim()` uses `None` and  `Option<>` is consistent with the eventlog.

mythi:
> Thanks @mythi The code looks good. Let's make it better.

Thanks! I've updated the code.

deps/verifier/src/az_tdx_vtpm/mod.rs
Xynnn007: `cc_eventlog` is `Option` in the evidence body thus we need `Option` here, while `platform_info` always exists due to code logic

deps/verifier/src/intel_dcap/pck.rs
```diff
@@ -92,68 +94,223 @@ struct SgxExtension<'a> {
...
+        .and_then(|pems| {
+            if pems.len() != 3 {
+                bail!(
+                    "PCK cert chain must contain exactly 3 certificates (leaf, intermediate CA, root CA), got {}",
+                    pems.len()
+                );
+            }
+            // Safe: length verified above.
+            Ok(pems.try_into().unwrap())
+        })?;
```
Xynnn007, line 151: we can directly use `try_into` w/o length judgement.

```suggestion
    let [leaf, intermediate, root] = Pem::iter_from_buffer(pem_certs)
        .collect::<Result<Vec<Pem>, _>>()
        .context("failed to parse PCK PEM certificate chain")
        .try_into()
        .map_err(|v: Vec<Pem>| {
        anyhow::anyhow!(
            "PCK cert chain must contain exactly 3 certificates (leaf, intermediate CA, root CA), got {}",
            v.len()
        )
    })?;
```

mythi: I know. It was done because I wanted avoid the suggested pattern.

mythi: updated per feedback

deps/verifier/src/az_tdx_vtpm/mod.rs
mythi: changed to `platform_info: &PlatformInfo`

Xynnn007 (approved):
LGTM. Thanks @mythi

fitzthum (approved):
LGTM

mythi:
two approvals so merging.

Issue #1334, opened 2026-05-12
DCAP verification collateral improvements


mythi:
I'm drafting a plan for visibility that I'm going to follow to improve DCAP quote verification collateral handling. The goal is to stop using "QCNL quote provider library". It's currently used via https://github.com/confidential-containers/trustee/blob/9cd45e15cacceb142ae8ac9982bbbcd9b080b786/deps/verifier/src/intel_dcap/mod.rs#L127-L136 to populate `QuoteCollateral`. We can do this in pure Rust very easily and most importantly, add support for "offline file based collateral" that QCNL does not support.

- [x] #1333 
- [x] #1335 (a follow-up to #1289 to  parse full platform info (e.g., "fmspc") to be used with collateral handling)
- [x] #1336
- [x] #1368 
   - [x] RFC on CollateralService trait + Pcs impl for it 
   - [x] Make the verification collateral available to `ecdsa_quote_verification()`
   - [x] Add file based offline collateral using JSON provided by [Intel's PCS client tool](https://github.com/intel/confidential-computing.tee.dcap/tree/main/tools/PcsClientTool)
   - [x] Drop QCNL from the builds and deployments

mythi:
the tasks are now complete so closing. @Xynnn007 and others, thanks for the help with reviews!

PR #1336, opened 2026-05-12
verifier: unify SGX/TDX quote parsing under intel_dcap

mythi:
This may look big but it's mostly just combining types from SGX and TDX verifiers under intel_dcap.


deps/verifier/src/intel_dcap/mod.rs
```diff
@@ -20,6 +20,8 @@ mod claims;
 mod error;
 #[cfg(any(feature = "tdx-verifier", feature = "sgx-verifier"))]
 pub(crate) mod pck;
+#[cfg(any(feature = "tdx-verifier", feature = "sgx-verifier"))]
```
Xynnn007, line 23: Not related to this PR.

It's acceptable to have a helper feature like `__intel_dcap` in Cargo.toml, which will be enabled by either `tdx-verifier` or `sgx-verifier`. And here we could simply have 
```
#[cfg(feature = "__intel_dcap")]
```

This can be done in a separate PR then.

deps/verifier/src/intel_dcap/pck.rs
```diff
@@ -3,6 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+#![allow(dead_code)]
```
Xynnn007: Let's not add this clippy flag. Instead, use `_` prefix if one variable is not used explicitly and remove the functions that is never used.

deps/verifier/src/intel_dcap/quote.rs
```diff
@@ -0,0 +1,733 @@
...
+    pub(crate) cpu_svn: [u8; 16],
+    /// ( 16) Which fields defined in SSA.MISC
+    pub(crate) misc_select: [u8; 4],
+    /// ( 20)
+    pub(crate) reserved1: [u8; 12],
+    /// ( 32) ISV assigned Extended Product ID
+    pub(crate) isv_ext_prod_id: [u8; 16],
+    /// ( 48) Any special capabilities the enclave possesses
+    pub(crate) attributes_flags: [u8; 8],
+    pub(crate) attributes_xfrm: [u8; 8],
```
Xynnn007, line 108: just out of curiosity: do we have a comment upon `xfrm`? As I see other fields all have. but this is not a big deal.

deps/verifier/src/intel_dcap/quote.rs
```diff
@@ -0,0 +1,733 @@
...
+    pub(crate) report_data: [u8; 64],
+}
+
+impl fmt::Display for SgxReportBody {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "
+SGX Report Body:
+\tcpu_svn:\t{}
```
Xynnn007, line 138: Note that this Display has different format with `QuoteHeader`'s. The one of `QuoteHeader` uses upper case while this uses lower snake case. Please try to use a common format.

deps/verifier/src/intel_dcap/quote.rs
```diff
@@ -0,0 +1,733 @@
...
+impl fmt::Display for QuoteV5Type {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            QuoteV5Type::TDX10 => writeln!(f, "Quote v5 Type: TDX 1.0"),
+            QuoteV5Type::TDX15 => writeln!(f, "Quote v5 Type: TDX 1.5"),
+        }
+    }
+}
+
+#[allow(dead_code)]
```
Xynnn007: Still. Let's try to get rid of the macro

deps/verifier/src/intel_dcap/quote.rs
```diff
@@ -0,0 +1,733 @@
...
+                Quote::V5 { body, .. } => match body {
+                    QuoteV5Body::Tdx10(body) => &body.$r,
+                    QuoteV5Body::Tdx15(body) => &body.$r,
+                },
+            }
+        }
+    };
+}
+
+#[allow(dead_code)]
```
Xynnn007: ditto

deps/verifier/src/intel_dcap/quote.rs
```diff
@@ -0,0 +1,733 @@
...
+impl Quote {
+    tdx_body_field!(report_data);
+    tdx_body_field!(mr_config_id);
+    tdx_body_field!(rtmr_0);
+    tdx_body_field!(rtmr_1);
+    tdx_body_field!(rtmr_2);
+    tdx_body_field!(rtmr_3);
+    tdx_body_field!(td_attributes);
```
Xynnn007: btw I think it's fine to make the members of the quote structure public. As `verifier` could be used by other projects, and this will help external developers to get benefit from the TDX ser/de-ser implementations.

deps/verifier/src/intel_dcap/quote.rs
```diff
@@ -0,0 +1,733 @@
...
+    if !is_sgx_v3 {
+        let qe_report_cert_data_type: u16 = data.gread::<u16>(&mut offset)?;
+
+        // Expect QE Report Certification Data
+        if qe_report_cert_data_type != QE_REPORT_CERT_DATA_TYPE {
+            bail!("expected cert data type {QE_REPORT_CERT_DATA_TYPE}, got {qe_report_cert_data_type}");
+        }
+
+        // Skip length for now
+        _ = data.gread::<u32>(&mut offset)?;
```
Xynnn007, line 464: Do we need any comparation for assertion by using this left value?

deps/verifier/src/intel_dcap/quote.rs
```diff
@@ -0,0 +1,733 @@
...
+pub(crate) fn parse_quote(quote_bin: &[u8]) -> Result<Quote> {
+    let header = quote_bin
+        .pread::<QuoteHeader>(0)
+        .context("parse quote header failed")?;
+
+    let version = u16::from_le_bytes(header.version);
+    let tee_type = u32::from_le_bytes(header.tee_type);
+    let att_key_type = u16::from_le_bytes(header.att_key_type);
+
+    if att_key_type != 2 {
```
Xynnn007: Can we make this `2` into a const?

deps/verifier/src/intel_dcap/quote.rs
```diff
@@ -0,0 +1,733 @@
...
+        bail!("unsupported att_key_type {att_key_type}, expected 2 (ECDSA-256-with-P-256)");
+    }
+
+    match (version, tee_type) {
+        (3, 0x00000000) => {
```
Xynnn007: Do you think we need to make things like `0x00000000`, `0x00000081` into consts? As they are referenced in many places, like codes/ comments.

deps/verifier/src/sgx/mod.rs
```diff
@@ -52,11 +49,12 @@ impl Verifier for SgxVerifier {
...
-    quote_body
-        .pread::<sgx_quote3_t>(0)
-        .map_err(|e| anyhow!("Parse SGX quote failed: {:?}", e))
+pub(crate) fn parse_sgx_quote(quote: &[u8]) -> Result<Quote> {
+    let q = parse_quote(quote)?;
+    match &q {
+        Quote::V3 { .. } => Ok(q),
+        _ => bail!("expected SGX quote (v3), got non-SGX quote"),
+    }
 }
```
Xynnn007: Looks like this function is not needed. We could directly use the following like logic to replace line 76 - 79
```rust
let quote = parse_quote(quote)?;
let Quote::V3 { body, ... } = quote else {
    bail!("expected SGX quote (v3), got non-SGX quote");
}

// do things with body
```

deps/verifier/src/tdx/quote.rs
```diff
@@ -1,588 +1,12 @@
...
-        assert_eq!(
-            claims, expected_output,
-            "Unexpected verification output for {quote}"
-        );
+/// Parse a TDX ECDSA quote (v4 or v5). Returns an error for SGX v3 quotes.
+pub(crate) fn parse_tdx_quote(quote_bin: &[u8]) -> anyhow::Result<Quote> {
+    let quote = parse_quote(quote_bin)?;
+    match &quote {
+        Quote::V4 { .. } | Quote::V5 { .. } => Ok(quote),
+        Quote::V3 { .. } => anyhow::bail!("expected TDX quote (v4/v5), got SGX quote (v3)"),
```
Xynnn007: Also for this. we could use `matches!` macro upon the result of `parse_quote` to judge its type. Because `parse_tdx_quote` couples two capibilities: 1) parse a quote 2) judge its type

deps/verifier/src/intel_dcap/quote.rs
mythi: I added these last minute when I noticed warnings when only `sgx-verifier` is enabled. I'll try to think something else. Maybe the helper feature you suggested can help here.

deps/verifier/src/sgx/mod.rs
mythi: True, the goal in this PR was to keep the callers unchanged (as much as possible).

deps/verifier/src/intel_dcap/pck.rs
mythi: fixed by [facc68e](https://github.com/confidential-containers/trustee/pull/1336/commits/facc68ef6cf6c5a69e662bb4db2fd7206606df2b)

deps/verifier/src/intel_dcap/quote.rs
mythi: should be aligned now

deps/verifier/src/intel_dcap/quote.rs
mythi: updated

deps/verifier/src/intel_dcap/quote.rs
mythi: fixed

deps/verifier/src/intel_dcap/quote.rs
mythi: fixed

deps/verifier/src/intel_dcap/quote.rs
mythi: I could not find any value to check (this is also moving from old code wtith no changes)

deps/verifier/src/intel_dcap/quote.rs
mythi: would you be OK if this is just documented that the warnings trigger when `tdx-verifier` is not enabled. I tried to include a feature gate for these but it wasn't pretty. `dead_code` seems to be the easiest approach

deps/verifier/src/intel_dcap/quote.rs
mythi: I'm thinking of doing something like this once the re-work in #1334 settles

deps/verifier/src/intel_dcap/quote.rs
mythi: `dead_code` macros are now gone.

mythi:
@Xynnn007 I have your feedback addressed.

deps/verifier/src/sgx/mod.rs
Xynnn007: The function signature has been changed. we do not need this nested call.

deps/verifier/src/sgx/mod.rs
```diff
@@ -75,10 +73,14 @@ async fn verify_evidence(
         .context("Evidence's identity verification error.")?;
 
     let quote = parse_sgx_quote(&quote_bin)?;
+    let (report_data, config_id) = match &quote {
+        Quote::V3 { body, .. } => (body.report_data, body.config_id),
+        _ => unreachable!("parse_sgx_quote guarantees V3"),
+    };
```
Xynnn007, line 72: Let's avoid using `unreachable!` macros and also do not need `parse_sgx_quote` layer.

```rust
let quote = parse_quote(quote)?;
let Quote::V3 { body, ... } = quote else {
    bail!("expected SGX quote (v3), got non-SGX quote");
};

// do things with body.report_data and  body.config_id
```

deps/verifier/src/sgx/claims.rs
```diff
@@ -57,54 +56,48 @@ macro_rules! parse_claim {
     };
 }
 
-pub fn generate_parsed_claims(quote: sgx_quote3_t) -> Result<TeeEvidenceParsedClaim> {
+pub fn generate_parsed_claims(quote: &Quote) -> Result<TeeEvidenceParsedClaim> {
```
Xynnn007: We could have one overall `generate_parsed_claims` function under the definition of `enum Quote` in `intel_dcap`. It will generate parsed claims based on the concret enum type, s.t.
1. SGX
2. TDX, with also `platform_info` parsing logic inside.

mythi: some later PR perhaps?

Xynnn007: Ok. If you do not mind, I can help with some refactor work with the current shape

deps/verifier/src/sgx/mod.rs
mythi: Dropped `parse_sgx_quote()` and `parse_tdx_quote()`.

deps/verifier/src/sgx/mod.rs
mythi: updated. PTAL

Xynnn007 (approved):

Xynnn007:
looks like the linter cries - needs some manual fix

deps/verifier/src/sgx/claims.rs
mythi: That's fine. `src/tdx/claims.rs` already gives a pretty good baseline to add `V3` arm.,

mythi:
> looks like the linter cries - needs some manual fix

this should be fixed now too.

PR #1338, opened 2026-05-13
Verifier/KBS: Use SHA-512 for report data / runtime_data calculation in IBM SEL

BbolroC:
Use the more secure SHA-512 algorithm for the `selected-hash-algorithm` field in `extra_params` for IBM SEL.
Update the verifier validation accordingly. The logic for configuring `extra_params` follows the pattern used by Intel Trust Authority.

In summary:
1. Use SHA-512 for report data in IBM SEL via extra_params
2. Select SHA-512 as runtime data hash algorithm for IBM SE in KBS

Depends-on: https://github.com/confidential-containers/guest-components/pull/1469

### DIFF
--- a/deps/verifier/src/se/ibmse.rs
+++ b/deps/verifier/src/se/ibmse.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
-use crate::{ReportData, TeeEvidence, TeeEvidenceParsedClaim, ToHex};
+use crate::{regularize_data, ReportData, TeeEvidence, TeeEvidenceParsedClaim, ToHex};
 use anyhow::{anyhow, Context, Result};
 use core::result::Result::Ok;
 use openssl::encrypt::{Decrypter, Encrypter};
@@ -23,6 +23,9 @@ use thiserror::Error;
 use tracing::{debug, info, warn};
 
 const DEFAULT_CERTS_OFFLINE_VERIFICATION: &str = "false";
+
+/// Size of report data in IBM SE attestation (64 bytes)
+const SE_REPORT_DATA_SIZE: usize = 64;
 
 const DEFAULT_SE_HOST_KEY_DOCUMENTS_ROOT: &str = "/run/confidential-containers/ibmse/hkds";
 
@@ -230,13 +233,19 @@ impl SeVerifierImpl {
 
         // Validate runtime_data_digest if provided
         if let ReportData::Value(expected_report_data) = expected_report_data {
+            let expected_report_data = regularize_data(
+                expected_report_data,
+                SE_REPORT_DATA_SIZE,
+                "USER_DATA",
+                "IBM SE",
+            );
             let report_data = se_response
                 .user_data
-                .get(..48)
+                .get(..expected_report_data.len())
                 .context("Failed to get report_data section from USER_DATA")?;
-            if report_data != *expected_report_data {
+            if report_data != expected_report_data {
                 return Err(SeError::UserDataMismatch {
-                    expected: expected_report_data.to_vec(),
+                    expected: expected_report_data,
                     actual: report_data.to_vec(),
                 }
                 .into());
@@ -438,8 +447,8 @@ mod tests {
     fn test_user_data_validation_success() {
         let verifier = create_test_verifier();
 
-        // Create test data - SHA-384 is 48 bytes
-        let report_data = vec![0x05; 48];
+        // Create test data - SHA-512 is 64 bytes
+        let report_data = vec![0x05; SE_REPORT_DATA_SIZE];
 
         // Build user_data: first 48 bytes are report_data, rest can be anything
         let mut user_data = report_data.clone();
@@ -486,11 +495,11 @@ mod tests {
     fn test_user_data_validation_mismatch() {
         let verifier = create_test_verifier();
 
-        // Create test data - SHA-384 is 48 bytes
-        let report_data = vec![0x05; 48];
+        // Create test data - SE report data is 64 bytes
+        let report_data = vec![0x05; SE_REPORT_DATA_SIZE];
 
-        // Build user_data with WRONG report_data in first 48 bytes
-        let mut user_data = vec![0xFF; 48]; // Wrong report data
+        // Build user_data with WRONG report_data
+        let mut user_data = vec![0xFF; SE_REPORT_DATA_SIZE]; // Wrong report data
         user_data.extend_from_slice(&[0xAA; 16]); // Add some extra data
 
         let nonce = vec![0x09; 16];
@@ -534,7 +543,7 @@ mod tests {
         let verifier = create_test_verifier();
 
         // Build user_data with some report_data
-        let report_data = vec![0x05; 48];
+        let report_data = vec![0x05; SE_REPORT_DATA_SIZE];
         let mut user_data = report_data.clone();
         user_data.extend_from_slice(&[0xBB; 16]);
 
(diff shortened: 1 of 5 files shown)


kbs/src/attestation/coco/builtin.rs
```diff
@@ -112,13 +112,43 @@ impl Attest for BuiltInCoCoAs {
             }
             _ => make_nonce().await?,
         };
-
-        let challenge = Challenge {
-            nonce,
-            extra_params: serde_json::Value::String(String::new()),
+        let extra_params = match tee {
+            Tee::Se if tee_parameters.is_null() => serde_json::Value::String(String::new()),
```
Xynnn007: Looks like this line shares a same result with `_ =>`, so maybe we could mark
```
Tee::Se if !tee_parameters.is_null() => {...}
_ => serde_json::Value::String(String::new()),
```

Xynnn007 (commented):
We might also need to change `grpc-as` to use sha512 to align with https://github.com/confidential-containers/guest-components/pull/1469/changes

kbs/src/attestation/coco/builtin.rs
BbolroC: Genius, bro. I will update the PR. 👍

attestation-service/src/lib.rs
```diff
@@ -3,6 +3,46 @@
 //! # Features
 //! - `rvps-grpc`: The AS will connect a remote RVPS.
 
+pub const SUPPORTED_HASH_ALGORITHMS_JSON_KEY: &str = "supported-hash-algorithms";
+pub const SELECTED_HASH_ALGORITHM_JSON_KEY: &str = "selected-hash-algorithm";
```
Xynnn007: Note that [there](https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/intel_trust_authority/mod.rs#L24-L25) are also declarations . Maybe we could move that part under `attestation/mod.rs`

attestation-service/src/lib.rs
```diff
@@ -3,6 +3,46 @@
...
+                    .as_str()
+                    .is_some_and(|algorithm| algorithm.eq_ignore_ascii_case(&needed_algorithm))
+            })
+        })
+        .map(|_| {
+            serde_json::json!({
+                SELECTED_HASH_ALGORITHM_JSON_KEY: needed_algorithm
+            })
+        })
+        .unwrap_or_else(|| serde_json::Value::String(String::new()))
```
Xynnn007: This will potentially hide the errors when doing the chain calling. Looks like a default value is given whenever there is an error? Note that if a malwared `tee_params` is given, the function will also return a default value, without telling anything wrong. Please ensure if this is expected.

attestation-service/src/lib.rs
```diff
@@ -3,6 +3,46 @@
...
+pub const SUPPORTED_HASH_ALGORITHMS_JSON_KEY: &str = "supported-hash-algorithms";
+pub const SELECTED_HASH_ALGORITHM_JSON_KEY: &str = "selected-hash-algorithm";
+
+/// Generate extra parameters for TEE hash algorithm negotiation.
+///
+/// This function checks if the provided TEE parameters contain supported hash algorithms
+/// and selects a hash algorithm based on the TEE type if available.
+/// Returns a JSON value with the selected algorithm,
+/// or an empty string if negotiation fails or is not applicable.
+pub fn generate_extra_params(tee: Tee, tee_params: &serde_json::Value) -> serde_json::Value {
```
Xynnn007: Also, note that `extra_params` is a conception for KBS's `Challenge` struct. See [the function](https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/backend.rs#L73-L83) and it's proper to move the SE-specific logic to that place.

deps/verifier/src/se/ibmse.rs
```diff
@@ -230,13 +230,15 @@ impl SeVerifierImpl {
 
         // Validate runtime_data_digest if provided
         if let ReportData::Value(expected_report_data) = expected_report_data {
+            let expected_report_data =
+                regularize_data(expected_report_data, 64, "USER_DATA", "IBM SE");
```
Xynnn007: Probably the `64` here could also be made into a const.

kbs/src/attestation/coco/builtin.rs
```diff
@@ -72,11 +63,16 @@ impl Attest for BuiltInCoCoAs {
         let mut verification_requests = vec![];
 
         for evidence in evidence_to_verify {
+            let runtime_data_hash_algorithm = match evidence.tee {
+                Tee::Se => HashAlgorithm::Sha512,
+                _ => HashAlgorithm::Sha384,
+            };
```
Xynnn007, line 69: Looks like we need to think about bringing a mechanism to include this parameter via the header parsing from RCAR protocol than harding coding here. This should be a TODO for me and not for this PR. This place in the PR looks good now.

protos/attestation.proto
```diff
@@ -89,6 +89,7 @@ message ChallengeRequest {
 }
 message ChallengeResponse {
     string attestation_challenge = 1;
+    string extra_params = 2;
```
Xynnn007: We might not need to change the proto here?

attestation-service/src/bin/grpc/mod.rs
```diff
@@ -244,9 +243,16 @@ impl AttestationService for Arc<RwLock<AttestationServer>> {
...
+        // Generate extra_params for hash algorithm negotiation
+        let extra_params = serde_json::from_str::<serde_json::Value>(tee_params)
+            .ok()
+            .map(|params| generate_extra_params(tee, &params).to_string())
+            .unwrap_or_default();
+
         info!("GetAttestationChallenge succeeded.");
         let res = ChallengeResponse {
             attestation_challenge,
+            extra_params,
```
Xynnn007: See comment about `generate_extra_params`, and we do not need the part to be changed.

Xynnn007 (commented):
Thanks @BbolroC  As IBM SE is a little different from other architectures, I tried to get in detail with the code change. There are some comments

attestation-service/src/lib.rs
BbolroC: Let me get back to the old logic then. Thanks!

attestation-service/src/lib.rs
BbolroC: Yep, you are right. I will move them to `attestation/mod.rs`. thanks!

attestation-service/src/lib.rs
BbolroC: True. I didn't realize that the extra_params is determined only with tee and tee_params passed to generate_challenge(). I will update the PR. Thanks!

deps/verifier/src/se/ibmse.rs
BbolroC: Yep, let's use a constant. Thanks!

protos/attestation.proto
BbolroC: Yep, we don't need this. Thanks!

Xynnn007 (approved):
LGTM. Although the PR does two things - 
1.  use sha512 for report data in IBM SEL
2. select sha512 as runtime data hash algorithm for IBM SE in KBS

So a updation on commit message, or reorg into two commits would either be ok

BbolroC:
> LGTM. Although the PR does two things -
> 
> 1. use sha512 for report data in IBM SEL
> 2. select sha512 as runtime data hash algorithm for IBM SE in KBS
> 
> So a updation on commit message, or reorg into two commits would either be ok

I would go with the 1st. Thanks!

PR #1345, opened 2026-05-15
kbs: zeroize JWE key material on drop

Cropi:
The JWE encryptor in `kbs/src/jwe.rs` left key material in heap memory
after operations completed — dropped `Vec<u8>` allocations are not
zeroed by the allocator, leaving bytes readable until reused.

- Wrap the ECDH shared secret `z` and its derived wrapping key in
  `Zeroizing<Vec<u8>>`
- Remove `cek.to_vec()` in the ECDH paths — `GenericArray` derefs to
  `&[u8]` directly so the heap copy was unnecessary
- Wrap the RSA symmetric key in `Zeroizing<Vec<u8>>` after generation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,6 +76,7 @@ time = { version = "0.3.47", features = ["std"] }
 tracing = "0.1.44"
 tracing-subscriber = { version = "0.3.23", features = ["env-filter"] }
 uuid = { version = "1.23.1", features = ["v4"] }
+zeroize = "1"
 
 [patch.crates-io]
 sev = { git = "https://github.com/virtee/sev", rev = "d487809188c3c92e397bc92e2cb2a85eb982c980" }
--- a/kbs/Cargo.toml
+++ b/kbs/Cargo.toml
@@ -47,7 +47,11 @@ vault = ["vaultrs"]
 actix = "0.13.5"
 actix-web = { workspace = true, features = ["openssl"] }
 actix-web-httpauth.workspace = true
-aes-gcm = "0.10.1"
+aes-gcm = { version = "0.10.1", features = ["zeroize"] }
+# Direct deps to activate their `zeroize` feature: `aes-gcm`'s own `zeroize`
+# feature does not propagate down to `aes` or `polyval`. Do not remove.
+aes = { version = "0.8", features = ["zeroize"] }
+polyval = { version = "0.6", features = ["zeroize"] }
 aes-kw = "0.3.0"
 anyhow.workspace = true
 async-trait.workspace = true
@@ -66,6 +70,7 @@ kms = { workspace = true, default-features = false }
 mobc = { workspace = true, optional = true }
 p256 = { workspace = true, features = ["ecdh"] }
 p521 = { workspace = true, features = ["ecdh"] }
+zeroize.workspace = true
 policy-engine.path = "../deps/policy-engine"
 prometheus = "0.14.0"
 prost = { workspace = true, optional = true }
--- a/kbs/src/jwe.rs
+++ b/kbs/src/jwe.rs
@@ -16,15 +16,16 @@ use p256::elliptic_curve::sec1::FromEncodedPoint;
 use rsa::{rand_core::RngCore, sha2::Sha256, BigUint, Oaep, Pkcs1v15Encrypt, RsaPublicKey};
 use serde_json::{json, Map};
 use tracing::warn;
+use zeroize::Zeroizing;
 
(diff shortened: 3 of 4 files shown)


kbs/src/jwe.rs
Xynnn007: do we also need to Zeroize this `cek`?

Xynnn007 (commented):
Thanks for this! Only one comment

kbs/src/jwe.rs
Cropi: that's a nice catch! I do believe it needs to be wrapped in both cases. Fixed it.

kbs/src/jwe.rs
```diff
@@ -223,7 +226,7 @@ fn ecdh_es_a256kw_p256(x: String, y: String, mut payload_data: Vec<u8>) -> Resul
     };
 
     // 3. Encrypt content with CEK
-    let mut cek_cipher = Aes256Gcm::new(GenericArray::from_slice(&cek));
+    let mut cek_cipher = Aes256Gcm::new(&*cek);
```
Xynnn007, line 225: do we need to mark `cek_cipher` as `Zeroize`? Because it's also confidential

Cropi: I tried it but `AesGcm` doesn't implement `Zeroize`, so it won't compile.
```
error[E0277]: the trait bound `AesGcm<Aes256, UInt<..., ...>>: DefaultIsZeroes` is not satisfied
   --> kbs/src/jwe.rs:229:26
    |
229 |     let mut cek_cipher = Zeroizing::new(Aes256Gcm::new(&*cek));
    |                          ^^^^^^^^^ unsatisfied trait bound
    |
    = help: the trait `DefaultIsZeroes` is not implemented for `AesGcm<aes_gcm::aes::Aes256, UInt<UInt<UInt<UInt<UTerm, B1>, B1>, B0>, B0>>`
    = help: the following other types implement trait `DefaultIsZeroes`:
```
And maybe for good reasons. I did some digging in the aes-gcm create.

`cek` is wrapped in `Zeroizing<>` so the raw key bytes are zeroed on drop.

For `cek_cipher`, `Aes256Gcm` has two fields that hold persistent sensitive material derived from `cek`: the AES-256 round key schedule (`cipher: Aes256`) and the GHASH key (`ghash: GHash(Polyval)`). Enabling `aes/zeroize` adds a `Drop` impl to `Aes256` that wipes the key schedule, and `polyval/zeroize` does the same for the GHASH state in `Polyval`. Both fire automatically when `cek_cipher` drops.

The awkward part is that `aes` and `polyval` have to be added as direct deps just to flip those feature flags, because `aes-gcm`'s own `zeroize` feature doesn't propagate down to them. IMHO their upstream should propagate zerorize down to the bottom (aes-gcm -> ghash ->polyval), but that is just a wild guess I may not know the full story.
I will add a new commit to try address this, let me know if that makes sense or it should be discarded.

Xynnn007: Thanks for the investigation! I agree with (I saw you have updated the PR already)

1. let's keep `cek` in `Zeroizing<>`.
2. Temporarily get `aes` and `polyval`.
3. Raise an issue to track this thread. When upstream `Aes256Gcm` implements `Zeroize`, we could remove the two crates from `Cargo.toml`.

kbs/Cargo.toml
```diff
@@ -47,7 +47,9 @@ vault = ["vaultrs"]
 actix = "0.13.5"
 actix-web = { workspace = true, features = ["openssl"] }
 actix-web-httpauth.workspace = true
-aes-gcm = "0.10.1"
+aes-gcm = { version = "0.10.1", features = ["zeroize"] }
+aes = { version = "0.8", features = ["zeroize"] }
```
Xynnn007, line 53: Let's add a comment upon the two dependencies to prevent from removing them unexpectedly when find they are not direcctly referenced in the code

Cargo.toml
```diff
@@ -45,6 +45,7 @@ mobc = "0.9.0"
 openssl = "0.10.79"
 p256 = "0.13.2"
 p521 = "0.13.2"
+zeroize = "1"
```
Xynnn007: Can we put the item in alphabet order?

Xynnn007 (approved):
Thanks for this!

Cropi:
> Thanks for this!

Thank you for the feedback. I modified the PR per your suggestions. I can create a tracker if the PR gets merged, so we can get back later to this topic once upstream of those dependencies resolves the issue.

Xynnn007:
@Cropi Thanks for this! I get it merged.

PR #1346, opened 2026-05-15
kv-storage: fix inverted overwrite logic in LocalJson backend

sergio-correia:
The overwrite guard in LocalJson::set() had the condition inverted: `if parameters.overwrite && items.contains_key(key)` returned AlreadyExists without writing, which is the opposite of the intended behavior and inconsistent with all other backends (Memory, LocalFs, PostgreSQL).

The consequence is that set() with overwrite=true silently discards updates to existing keys (returning AlreadyExists), while set() with overwrite=false silently overwrites existing keys. This breaks session state persistence, admin policy updates, and resource writes for any deployment using the LocalJson storage backend.


fitzthum (approved):
LGTM

if our main deployment types were using JSON backend I might try to sneak this into the upcoming release, but they use LocalFs, which is probably how this snuck in in the first place

Xynnn007 (approved):
Thanks @sergio-correia LGTM!

PR #1353, opened 2026-05-18
verifier: intel_dcap: treat supplemental data size mismatch as an error

mythi:
A size mismatch between the QVL's reported supplemental data size and sizeof(sgx_ql_qv_supplemental_t) previously fell back to passing None to tee_verify_quote, leaving supp_data zero-initialised. prepare_custom_claims_map() was then called unconditionally on that zeroed struct, silently emitting false claims: timestamps, zero CRL numbers, empty advisory_ids, etc. Policies checking these values would operate on garbage data.

Treat the mismatch as a hard error instead. Also simplify and reorganize the code to make it more readable.

### DIFF
--- a/deps/verifier/src/intel_dcap/mod.rs
+++ b/deps/verifier/src/intel_dcap/mod.rs
@@ -78,13 +78,6 @@ pub(crate) fn set_qcnl_config(c: Option<QcnlConfig>) -> Result<(), std::io::Erro
 }
 
 pub(crate) async fn ecdsa_quote_verification(quote: &[u8]) -> anyhow::Result<Map<String, Value>> {
-    let mut supp_data: sgx_ql_qv_supplemental_t = Default::default();
-    let mut supp_data_desc = tee_supp_data_descriptor_t {
-        major_version: 0,
-        data_size: 0,
-        p_data: &mut supp_data as *mut sgx_ql_qv_supplemental_t as *mut u8,
-    };
-
     // Call DCAP quote verify library to set QvE loading policy to multi-thread
     // We only need to set the policy once; otherwise, it will return the error code 0xe00c (SGX_QL_UNSUPPORTED_LOADING_POLICY)
     static INIT: std::sync::Once = std::sync::Once::new();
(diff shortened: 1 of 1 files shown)


mythi:
ready for review (I got testing completed)

deps/verifier/src/intel_dcap/mod.rs
```diff
@@ -102,27 +95,27 @@ pub(crate) async fn ecdsa_quote_verification(quote: &[u8]) -> anyhow::Result<Map
...
-            "tee_get_quote_supplemental_data_size failed: {}",
+    let (_, supp_size) = tee_get_supplemental_data_version_and_size(quote).map_err(|e| {
+        anyhow!(
+            "tee_get_supplemental_data_version_and_size failed: {}",
             describe_error(e)
-        ),
+        )
+    })?;
+
+    let expected_size = mem::size_of::<sgx_ql_qv_supplemental_t>() as u32;
```
Xynnn007, line 105: Can this be defined as a const? looks like the size never changes as `sgx_ql_qv_supplemental_t` is a FFI C type.

Xynnn007 (approved):
LGTM. Only one nit

deps/verifier/src/intel_dcap/mod.rs
mythi: AFAUI, it can change based on the library version we choose in Trustee and generate the bindings. The mismatch can happen if the runtime library reports different size. 

We have improved this in Trustee by pinning the crate + build/runtime library versions.

fitzthum (approved):
LGTM

deps/verifier/src/intel_dcap/mod.rs
Xynnn007: do you mean that different dcap dep version will probably have different sizes for `sgx_ql_qv_supplemental_t `?

mythi: That's possible yes but it hasn't changed that often. The [spec](https://download.01.org/intel-sgx/sgx-dcap/1.26/linux/docs/Intel_TDX_DCAP_Quoting_Library_API.pdf) appendix A.2 has it documented.

mythi:
Good to merge?

PR #1348, opened 2026-05-15
Add configurable TLS security profiles

lmilleri:
Add support for Mozilla TLS security profiles (old, intermediate, modern, custom) and custom TLS configuration options.

Fix issue #1347

Issue #1347, opened 2026-05-15
[RFC] Add configurable TLS profile support


lmilleri:
## Problem

Trustee currently has hardcoded TLS configuration using `mozilla_intermediate_v5`.
This prevents:
   1. Dynamic inheritance of TLS settings from platform-level configuration (e.g., OpenShift APIServer TLS profiles)
   2. Alignment with centrally managed TLS security policies for Post-Quantum Cryptography (PQC) readiness
   3. Support for custom TLS profiles where customers need to disable specific algorithms

## Background

Some platforms (e.g. OpenShift 4.22) require all components to dynamically consume TLS settings from central configuration sources. Customers must be able to opt into PQC-resilient ciphers across the entire platform by adjusting documented knobs, and components must respect custom TLS profiles.

While Trustee already has PQC auto-detection logic for TLS groups, the TLS version and cipher configuration remain hardcoded, preventing proper integration with platform TLS policies.

### 1. Extend http_server section with new TLS profile fields

tls_profile = "intermediate"                                                                               # or "old", "modern", "custom"
tls_min_version = "1.2"                                                                                      # optional, for custom profile
tls_max_version = "1.3"                                                                                     # optional, for custom profile
tls_ciphers = "TLS_AES_256_GCM_SHA384:TLS_AES_128_GCM_SHA256"  # optional
tls_groups = "X25519MLKEM768:X25519:secp256r1"                                  # optional

### 2. Modify `tls_config()` function

Replace hardcoded `mozilla_intermediate_v5` with profile-based configuration

### Case 1: Default Configuration (No TLS fields specified)

Behavior:
- Uses tls_profile = "intermediate" (default)
- TLS versions: 1.2 and 1.3
- Ciphers: Mozilla Intermediate v5 preset
- Groups: Auto-detected PQC (if available) + classical fallback
- No additional fields needed

### Use Case 2: Legacy Client Support

Behavior:
- TLS versions: 1.0, 1.1, 1.2, 1.3 (maximum compatibility)
- Ciphers: Mozilla Old v5 preset (includes weaker ciphers)
- Groups: Auto-detected PQC + classical

###  Use Case 3: Modern/PQC-Ready Configuration

Behavior:
- TLS versions: 1.3 only
- Ciphers: Mozilla Modern v5 (only TLS 1.3 ciphers like TLS_AES_256_GCM_SHA384)
- Groups: Auto-detected PQC + classical

### Use Case 4: Custom Profile - Full Control

[...]

PR #1354, opened 2026-05-18
Add PQC-aware TLS for KBS HTTP and gRPC TLS across KBS, attestation-service, and RVPS

Cropi:
## Summary

This PR addresses two related gaps: KBS's HTTP TLS has no way to enforce
post-quantum cryptography, and none of the service-to-service gRPC channels
support TLS at all.

Related to #1347. See the overlap note at the end.

### Shared `deps/tls-config` crate

A small new crate centralising TLS logic shared by all three services:

- **PQC group detection** — probes OpenSSL at startup for supported hybrid
  groups (`X25519MLKEM768`, `SecP256r1MLKEM768`, `SecP384r1MLKEM1024`).
- **`configure_pqc_groups(builder, require_pqc)`** — applies PQC groups with
  classical fallback, or enforces PQC-only + TLS 1.3 minimum when
  `require_pqc = true`.
- **gRPC TLS helpers** — tonic `ClientTlsConfig` / `ServerTlsConfig`
  wrappers behind the `grpc` feature flag, eliminating duplicated
  certificate-loading boilerplate across services.
- 12 unit tests covering all public entry points.

`X448MLKEM1024` is omitted as it appears to still be experimental.

### KBS HTTP TLS

- New `require_pqc` config field (default `false`). When `true`, startup fails
  if PQC is unavailable, TLS 1.3 minimum is enforced, and classical groups are
  dropped.
- Switches the rustls provider from `ring` to `aws-lc-rs`.

### gRPC TLS

- **KBS → AS**: new `tls_mode` / `ca_cert_path` fields on `GrpcConfig`.
- **grpc-as**: `--tls-cert` / `--tls-key` for server-side TLS.
- **restful-as**: adds `--require-pqc`; renames `--https-pubkey-cert` /
  `--https-prikey` to `--tls-cert` / `--tls-key` (old names kept as aliases).
- **RVPS client and server**: same `tls_mode` / `ca_cert_path` pattern.

All gRPC clients validate the URI scheme against `tls_mode` to prevent silent
plaintext fallback.

## Backward compatibility

All new fields default to their pre-existing behaviour. Existing config files
and CLI invocations work without modification.

## Relation to #1347 and #1348

PR #1348 covers overlapping ground from the operator-profile angle —
complementary rather than conflicting. Happy to coordinate with @lmilleri and
maintainers on sequencing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -604,6 +604,7 @@ dependencies = [
  "regorus",
  "rsa",
  "rstest",
+ "rustls",
  "serde",
  "serde_json",
  "serde_json_canonicalizer",
@@ -616,6 +617,7 @@ dependencies = [
  "testing_logger",
  "thiserror 2.0.18",
  "time",
+ "tls-config",
  "tokio",
  "toml",
  "tonic",
@@ -3649,6 +3651,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "time",
+ "tls-config",
  "tokio",
  "toml",
  "tonic",
@@ -5302,6 +5305,7 @@ dependencies = [
  "prost",
  "roxmltree",
  "rstest",
+ "rustls",
  "serde",
  "serde_json",
  "serial_test",
@@ -5310,6 +5314,7 @@ dependencies = [
  "sled",
  "strum 0.28.0",
  "tempfile",
+ "tls-config",
  "tokio",
  "tonic",
  "tonic-prost",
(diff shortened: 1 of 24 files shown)


lmilleri:
Hi @Cropi, how do you suggest to proceed? Would it be easy for you to rebase after my PR is merged?

Cropi:
> Hi @Cropi, how do you suggest to proceed? Would it be easy for you to rebase after my PR is merged?

Yes, I can do that. I just wanted to inform you as we are both trying to resolve a similar issue but from different POVs.

PR #1356, opened 2026-05-19
verifier/nvidia: fix json claim test to use Value than String

Xynnn007:
Sometimes the assertion of equal string will fail because the serialization result for a json object could in an unexpected order. Thus we fix this by comparing the two json Value object, which will not have order issue.

### DIFF
--- a/deps/verifier/src/nvidia/mod.rs
+++ b/deps/verifier/src/nvidia/mod.rs
@@ -390,6 +390,7 @@ impl Verifier for Nvidia {
 
 #[cfg(test)]
 mod tests {
+    use assert_json_diff::assert_json_eq;
     use rstest::rstest;
     use tracing_subscriber::{fmt, EnvFilter};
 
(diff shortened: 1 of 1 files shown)


deps/verifier/src/nvidia/mod.rs
```diff
@@ -436,12 +436,13 @@ mod tests {
...
 
-        let expected_claim =
-            include_str!("../../test_data/nvidia/hopperAttestationReport-claims.txt");
+        let expected_claim: serde_json::Value = serde_json::from_str(
+            include_str!("../../test_data/nvidia/hopperAttestationReport-claims.txt").trim(),
+        )
+        .expect("hopperAttestationReport-claims.txt must be valid JSON");
 
-        assert_eq!(expected_claim.to_string(), json);
+        assert_eq!(expected_claim, value);
```
mythi: we could start using `assert_eq_json!()` more where applicable.

mythi (approved):
LGTM

fitzthum (approved):
LGTM

PR #1358, opened 2026-05-19
Promote SNP parsed claims

### LINKED ISSUE #1364: Add hardware-type field to verifier output for precise hardware identification
## Background

The current attestation flow uses the `Tee` enum (from `kbs_types`) to identify the verifier plugin and the `tee_class` field (`cpu` / `gpu` / `switch` / `ppcie`) to categorize the hardware type for policy selection.

However, both identifiers are too coarse-grained to distinguish between specific hardware generations or product lines that may have different security properties, reference values, or attestation semantics.

This limitation was discussed in [#1350](https://github.com/confidential-containers/trustee/issues/1350) as a prerequisite for adding an accurate `hardware-type` claim to EAR token appraisals.

---

## Problem Statement

`Tee` enum does not distinguish hardware generations

| Tee + TeeClass variant | Actual hardware covered | Missing granularity |
|-------------|------------------------|---------------------|
| `Tee::Nvidia` + `gpu` | NVIDIA Hopper, NVIDIA Blackwell | Cannot tell Hopper from Blackwell |

For example, a node with one Hopper GPU and one Blackwell GPU will both be processed by the NVIDIA verifier and tagged with the same `Tee::Nvidia` variant. The broker and the downstream token consumer have no reliable way to know which appraisal corresponds to which architecture without inspecting raw evidence claims.

---

## Proposed Solution

```json
{
  "submods": {
    "gpu0": {
      "ear.trustee.hardware-type": "nvidia-hopper"
    },
    "gpu1": {
      "ear.trustee.hardware-type": "nvidia-blackwell"
    }
  }
}
```
- [ ] The broker can access `hardware_type` when constructing the EAR token
- [ ] A list of canonical device-type strings is documented

---

## Related Work

- [#1350](https://github.com/confidential-containers/trustee/issues/1350) — Discussion on EAR submod labels and tee-type claims
- [draft-ietf-rats-ear-03](https://www.ietf.org/archive/id/draft-ietf-rats-ear-03.txt) — EAR specification

Xynnn007:
~~This is a simple PR to add a hardware-type extension to EAR submod.~~

~~Also, draft a Trustee EAR profile that we can work on the following.~~

~~Close https://github.com/confidential-containers/trustee/issues/1364~~

This PR adds all fields of SNP report to the claims with nested structure, to provide a more flexible way to customize their policies. Also updates the documents and polcies.

### DIFF
--- a/attestation-service/docs/tcb_claims.md
+++ b/attestation-service/docs/tcb_claims.md
@@ -225,27 +225,97 @@ along with the full set of detached claims.
 
 ## AMD SEV-SNP
 
-- `snp.measurement` Launch Digest covering initial guest memory
-- `snp.platform_smt_enabled`:  Whether Simultaneous Multithreading is enabled on the system
-- `snp.platform_tsme_enabled`: Whether Transparent SME is enabled on the system
-- `snp.policy_abi_major`: Minimum ABI major version allowed for guest
-- `snp.policy_abi_minor`: Minimum ABI minor version allowed for guest
-- `snp.policy_debug_allowed`: Whether SNP debug features are allowed for guest
-- `snp.policy_migrate_ma`: Whether migration agent can be connected to guest
-- `snp.policy_single_socket`: Whether guest can be activated only on one socket
-- `snp.policy_smt_allowed`: Whether guest can run on a system with SMT enabled
-- `snp.reported_tcb_bootloader`: Reported SVN of ASP bootloader
-- `snp.reported_tcb_microcode`: Reported microcode version
-- `snp.reported_tcb_snp`: Reported SVN of SNP Firmware
-- `snp.reported_tcb_tee`: Reported SVN of ASP OS
+Existing flat SNP claims are preserved for policy compatibility. New detailed claims are additive under `snp.evidence`. When a flat claim and a detailed claim describe the same value, both paths are listed on the same line below.
 
-The claims map only includes the reported TCB version.
-An SEV-SNP Attestation Report contains four sets of TCB version information.
-Often all four values are the same, but sometimes the reported TCB might lag
-behind the true firmware version. This is done to minimize churn of policies
-and certificates while the provider updates to provisional firmware.
-The actual firmware must always be newer than or equal to the reported TCB.
-Generally, policies should be evaluated against the reported TCB.
+- `snp.generation`: string. Processor generation derived from the attestation report. One of `milan` (3rd Gen EPYC), `genoa` (4th Gen EPYC), or `turin` (5th Gen EPYC).
+
+### Guest policy
+
+The guest policy is set by the guest owner at launch and cannot be changed for the lifetime of the guest.
+
+- `snp.policy_abi_major` / `snp.evidence.policy.abi_major`: u64. Minimum SNP firmware ABI major version required for this guest.
+- `snp.policy_abi_minor` / `snp.evidence.policy.abi_minor`: u64. Minimum SNP firmware ABI minor version required for this guest.
+- `snp.policy_smt_allowed` / `snp.evidence.policy.smt_allowed`: Boolean. Whether the guest may run on a host with SMT enabled.
+- `snp.policy_migrate_ma` / `snp.evidence.policy.migrate_ma_allowed`: Boolean. Whether the guest may be associated with a migration agent.
+- `snp.policy_debug_allowed` / `snp.evidence.policy.debug_allowed`: Boolean. Whether SNP debug features are allowed for the guest.
+- `snp.policy_single_socket` / `snp.evidence.policy.single_socket_required`: Boolean. Whether the guest may be activated on multiple sockets.
+- `snp.evidence.policy.cxl_allowed`: Boolean. Whether CXL may be populated with devices or memory.
+- `snp.evidence.policy.mem_aes_256_xts`: Boolean. Memory encryption mode requirement. `false`: AES-128-XEX or AES-256-XTS is allowed; `true`: AES-256-XTS is required.
+- `snp.evidence.policy.rapl_dis`: Boolean. Running Average Power Limit (RAPL) requirement.
+- `snp.evidence.policy.ciphertext_hiding`: Boolean. Ciphertext hiding requirement.
+- `snp.evidence.policy.page_swap_disabled`: Boolean. Guest access to page-move/swap commands. `false`: SNP page-move/swap commands are allowed; `true`: guest access to `SNP_PAGE_MOVE`, `SNP_SWAP_OUT`, and `SNP_SWAP_IN` is disabled.
+
+### Platform info
+
+- `snp.platform_smt_enabled` / `snp.evidence.plat_info.smt_enabled`: Boolean. Whether Simultaneous Multithreading is enabled on the host.
+- `snp.platform_tsme_enabled` / `snp.evidence.plat_info.tsme_enabled`: Boolean. Whether Transparent SME is enabled on the host.
+- `snp.evidence.plat_info.ecc_enabled`: Boolean. Whether the platform is currently using ECC memory.
+- `snp.evidence.plat_info.rapl_disabled`: Boolean. Whether the RAPL feature is disabled on the platform.
+- `snp.evidence.plat_info.ciphertext_hiding_enabled`: Boolean. Whether ciphertext hiding is enabled on the platform.
+- `snp.evidence.plat_info.alias_check_complete`: Boolean. Whether alias detection has completed since the last system reset with no aliasing addresses found.
+- `snp.evidence.plat_info.tio_enabled`: Boolean. Whether SEV-TIO is enabled on the platform.
+
+### Key info (`snp.evidence.key_info`)
+
+- `snp.evidence.key_info.author_key_en`: Boolean. Whether the author key digest is present in `author_key_digest`. `true`: digest is present; `false`: digest is zero.
+- `snp.evidence.key_info.mask_chip_key`: Boolean. Whether the report signature is masked. `false`: firmware signs with VCEK or VLEK; `true`: the signature field is zeroed instead of signing.
+- `snp.evidence.key_info.signing_key`: u32. Key used to sign this report. `0`: VCEK; `1`: VLEK; `7`: none (unsigned report).
+
+### TCB versions
+
+Four TCB sets are exposed. Each is an object with the subfields below. The same subfields and types exist on `current_tcb`, `reported_tcb`, `committed_tcb`, and `launch_tcb`.
+
+- `snp.evidence.current_tcb`: object. Current platform TCB at report generation time.
+- `snp.evidence.reported_tcb`: object. TCB version used to derive the VCEK/VLEK that signed this report. **Policies should generally evaluate against this set.**
+- `snp.reported_tcb_bootloader` / `snp.evidence.reported_tcb.bootloader`: u8. Reported PSP bootloader SVN.
+- `snp.reported_tcb_tee` / `snp.evidence.reported_tcb.tee`: u8. Reported PSP operating-system SVN.
+- `snp.reported_tcb_snp` / `snp.evidence.reported_tcb.snp`: u8. Reported SNP firmware SVN.
+- `snp.reported_tcb_microcode` / `snp.evidence.reported_tcb.microcode`: u8. Reported microcode patch level.
+- `snp.evidence.committed_tcb`: object. Committed TCB version.
+- `snp.evidence.launch_tcb`: object. Platform TCB at guest launch or import time.
+
+Each TCB object contains the following fields (SVN / patch level):
+
+- `fmc`: u8 | null. FMC firmware SVN. Present on Turin (`generation` = `turin`); `null` on Milan and Genoa.
+- `bootloader`: u8. PSP bootloader SVN.
+- `tee`: u8. PSP operating-system SVN.
+- `snp`: u8. SNP firmware SVN.
+- `microcode`: u8. Lowest current microcode patch level across all cores.
+
+Often all four TCB values are the same, but the reported TCB may lag behind the true firmware version to minimize churn of policies and certificates while the provider updates to provisional firmware. The actual firmware must always be newer than or equal to the reported TCB.
+
+Firmware version objects:
+
+- `snp.evidence.current`: object. Current SNP firmware version.
+- `snp.evidence.current.major`: u8. Major version component.
+- `snp.evidence.current.minor`: u8. Minor version component.
+- `snp.evidence.current.build`: u8. Build version component.
+- `snp.evidence.committed`: object. Committed SNP firmware version.
+- `snp.evidence.committed.major`: u8. Major version component.
+- `snp.evidence.committed.minor`: u8. Minor version component.
+- `snp.evidence.committed.build`: u8. Build version component.
+
+### Measurements and identifiers
+
+- `snp.evidence.version`: u32. Attestation report format version. Supported values are 3 through 5.
+- `snp.evidence.guest_svn`: u32. Guest security version number provided at launch.
+- `snp.evidence.family_id`: string (hex). Family ID provided at launch.
+- `snp.evidence.image_id`: string (hex). Image ID provided at launch.
+- `snp.evidence.vmpl`: u32. Requested VMPL for the attestation report. Guest attestation expects `0`.
+- `snp.evidence.sig_algo`: u32. Signature algorithm identifier. `1`: ECDSA P-384 with SHA-384.
+- `snp.measurement` / `snp.evidence.measurement`: string (hex). Launch digest covering initial guest memory (48 bytes).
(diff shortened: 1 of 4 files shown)


attestation-service/docs/ietf-rats-ear-profile.md
```diff
@@ -0,0 +1,217 @@
...
+| `az-snp-vtpm` | Azure SNP guest with vTPM | Guest launch integrity; vTPM identity; and vTPM-reported configuration or measured boot-chain state |
+| `az-tdx-vtpm` | Azure TDX guest with vTPM | Guest initial memory integrity; vTPM identity; and vTPM-reported configuration or measured boot-chain state |
+| `arm-cca` | Arm Confidential Compute Architecture guest / realm material as attested | Realm initial measurement and content integrity |
+| `ibm-se` | IBM Secure Execution | Measured workload image and guest configuration; firmware and platform trust state for the secure partition |
+| `hygon-csv` | Hygon CSV CVM | Confidential VM launch measurement and guest memory integrity; platform firmware and TCB state |
+| `hygon-dcu` | Hygon Deep Computing Unit | Device firmware and security-version posture; device identity |
+| `nvidia-hopper` | NVIDIA Hopper-class GPU attestation | Hopper GPU device firmware and security state |
+| `nvidia-blackwell` | NVIDIA Blackwell-class GPU attestation | Blackwell GPU device firmware and security state |
```
fitzthum: Does it make sense to have separate entries for different generations? We don't do that for CPUs.

attestation-service/docs/ietf-rats-ear-profile.md
```diff
@@ -0,0 +1,217 @@
...
+| `nvidia-switch-ls10` | NVIDIA LS10 switch attestation | Switch firmware and security posture; device or fabric identity |
```
fitzthum: probably just call this `nvidia-nvswitch`

deps/verifier/src/lib.rs
```diff
@@ -49,6 +50,75 @@ pub(crate) mod intel_dcap;
 #[cfg(feature = "tpm-verifier")]
 pub mod tpm;
 
+#[derive(
+    Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Hash, EnumString, Display, AsRefStr, Copy,
+)]
+pub enum HardwareType {
```
fitzthum: I feel like maybe we should just have the hardware type be a string that is defined by the verifier rather than an enum that we need to centrally control

fitzthum (commented):
Haven't really had time to review yet, but I have a few questions about the scope/granularity of the hw type. 

Also, I shoudl probably return to https://github.com/confidential-containers/trustee/pull/664 at some point

attestation-service/docs/ietf-rats-ear-profile.md
Xynnn007: When complex topologies exist, it's possible to distinguish different mounted devices. In fact, we also differentiated CPUs, such as AMD SEV and SNP, although SEV has been removed.

attestation-service/docs/ietf-rats-ear-profile.md
Xynnn007: I saw there is `Architecture::LS10` in codes. Is `ls10` the only switch model serves hopper nvlink?

deps/verifier/src/lib.rs
Xynnn007: This is only for better maintaince.  

In fact, we initially attempted to enumerate all supported hardware at the `Tee` layer within `kbs-types`. However, with time passes, the `Tee` became a hybrid of vendor and implementation (for example, `Nvidia` handles both GPUs and switches; while Intel `Tdx` and Intel `Sgx` are two different enums).

Therefore, we needed a separate place to describe individual hardware models.

This allows the trustee to uniformly describe and maintain supported hardware in the documentation. Whenever a vendor needs to add new hardware, they can simply add it here. Verifier ad-hoc support would reduce maintainability.

Xynnn007:
Based on the discussion during the bi-weekly meeting. I think the most un-certainty for the PR is at which granularity we should define tee types. I get some context about the TDX affairs. 
 
They have tdx1.0, tdx1.5 (live migration/updation), tdx2.0 (tee i/o) and more versions. so we can see that different TDX has different capabilities. but they are not sure if version numbers may be iterated, e.g. 2.0 -> 2.1. 

A practical way might be not to distinguish the generations of different TDX for now. Instead, the claim `tdx.quote.type` already tells context about this. For SNP, an optinal is to add another field in the claims to tell the generation.

fitzthum:
Sorry, lost track of this one. I'm not really sure the best approach to the granularity. We could introduce two fields. One for the type and one for the generation. 

I think that until we figure out a scheme, we should just have the type be a string that is set by the verifier.

deps/verifier/src/lib.rs
fitzthum: I think it's easier to maintain if we don't have a central enum tbh. We basically don't have to maintain anything. Verifiers can just report what they want for the type. I know the idea was to have a more clear scheme for how devices are identified, but idk if we actually know what the scheme should be yet.

Xynnn007:
The primary goal is to enable users to distinguish between different specific hardware from the claims, as different hardware protects different software and hardware. Another approach is to add a new field to the claims, without needing to ascend the hierarchy. This is because the rego policy supports syntax like 

```
input[_]["generation"] == "turin"
```

- For SNP, another claim named `generation` is added in the commit, to tell `turin`, `milan` or so.
- For TDX, we already have the field `type` for TDX Quote v5.
- Similarly, for Nvidia, we could have `generation` field, to tell `hopper`, `blackwell` or later generations. But looks like the Nvidia claims need another round of optimization. Some of the fields are not necessary to be exposed to policies, like `x-nvidia-gpu-vbios-rim-measurements-available` which should always be asserted. 


anyway, for TDX/SGX/SNP, which already have different `Tee`, simply adding the Intel logo doesn't mean much to users; however, for NVIDIA, the distinction is more significant because it currently has a single Tee to represent various different hardware components.

Let me make this PR simpler and start to diving to nvidia things.

fitzthum:
Ok. Sounds good.

@hgowda-amd @amd-aliem you might want to take a look at this.

Updating the tcb claims does have some downstream impact so let's make sure we have the best possible set of claims.

attestation-service/docs/tcb_claims.md
```diff
@@ -225,27 +225,92 @@ along with the full set of detached claims.
...
+- `snp.evidence.key_info.signing_key`: u32. Key used to sign this report. `0`: VCEK; `1`: VLEK; `7`: none (unsigned report).
+
+### TCB versions
+
+Four TCB sets are exposed. Each is an object with the subfields below. The same subfields and types exist on `current_tcb`, `reported_tcb`, `committed_tcb`, and `launch_tcb`.
+
+- `snp.evidence.current_tcb`: object. Current platform TCB at report generation time.
+- `snp.evidence.reported_tcb`: object. TCB version used to derive the VCEK/VLEK that signed this report. **Policies should generally evaluate against this set.**
+- `snp.evidence.committed_tcb`: object. Committed TCB version.
+- `snp.evidence.launch_tcb`: object. Platform TCB at guest launch or import time.
```
fitzthum, line 275: In the past, I deliberately left out some of these for simplicity's sake. I think we probably should just report one, but it's quite confusing to determine which one is the most important. I actually think the reported tcb is not the best one, despite that being what we used in the past.

hgowda-amd:
@Xynnn007  we would like to discuss the changes made in this PR internally and evaluate if we have any impact or if these changes suggested makes sense. Please give me a couple of days to discuss it with my teammates and get back to you. 
I would appreciate if this PR can be put on hold and not merged until then.

hgowda-amd:
@Xynnn007 Change to format of the generated token would break any relying party using SNP attestation with coco. Do we need to plan to convey that message to any coco customers? or would the release notes suffice this purpose?

deps/verifier/src/snp/mod.rs
```diff
@@ -597,30 +598,100 @@ pub(crate) fn verify_report_tcb(
...
-
-        // platform info
-        "platform_tsme_enabled": report.plat_info.tsme_enabled(),
-        "platform_smt_enabled": report.plat_info.smt_enabled(),
-
-        // measurements
-        "measurement": hex::encode(report.measurement),
-        "report_data": hex::encode(report.report_data),
         "init_data": hex::encode(report.host_data),
+        "report_data": hex::encode(report.report_data),
```
hgowda-amd, line 682: do we need to include every single field in the attestation displayed in the token as well even if there are no rules for it in the policy?

deps/verifier/src/snp/mod.rs
```diff
@@ -597,30 +598,100 @@ pub(crate) fn verify_report_tcb(
...
+                "rapl_dis": report.policy.rapl_dis(),
+                "ciphertext_hiding": report.policy.ciphertext_hiding(),
+                "page_swap_disabled": report.policy.page_swap_disabled(),
+            },
+            "family_id": hex::encode(report.family_id),
+            "image_id": hex::encode(report.image_id),
+            "vmpl": report.vmpl,
+            "sig_algo": report.sig_algo,
+            "current_tcb": {
+                "fmc": report.current_tcb.fmc,
```
hgowda-amd: "fmc" TCB measurement is available only in Turin. would this line throw NPE on milan or genoa?

deps/verifier/src/snp/mod.rs
Xynnn007: Yes, I hope so. At least fields that aren't evaluated (e.g., determining a value or size) in vendor-specific code logic (not just verifiers, but also services like TDX/CCA that connect to a remote hardware vendor's service to check firmware versions, SVN, etc.) need to be extracted into the claims.

The given default policy doesn't necessarily need to cover these fields, as it often requires RVPS to provide reference values, which we can't provide all at once. My next step plan is to provide user documentation to assist users in writing their own remote attestation policies. This way, (advanced) users can set conditions for all the fields they care about with more granularity, with all fields available.

deps/verifier/src/snp/mod.rs
Xynnn007: I added this in the README.md here https://github.com/confidential-containers/trustee/pull/1358/changes#diff-f396654b0e3e0b311c4049fdc3e98825b3f05dc44ecd46e777cfad2a4905e5beR274

it will be `null` for milan or genoa

attestation-service/docs/tcb_claims.md
hgowda-amd: Typically we would use Reported TCB since that is what would be used to fetch the VCEK certificate. I can see customers wanting to check for current TCB. Committed and launch TCB I am not really sure if it has any significance to be added in the report. I can reach out to some folks about this internally and get back.

deps/verifier/src/snp/mod.rs
```diff
@@ -991,12 +1062,21 @@ mod tests {
         let attestation_report = AttestationReport::from_bytes(VCEK_REPORT).unwrap();
 
         // Generate parsed claims
-        let claims = parse_tee_evidence(&attestation_report);
+        let claims = parse_tee_evidence(&attestation_report, ProcessorGeneration::Milan);
```
hgowda-amd, line 1087: Is there a reason we are hardcoding this?

deps/verifier/src/snp/mod.rs
hgowda-amd: sounds good. thanks for the clarification.

deps/verifier/src/snp/mod.rs
Xynnn007: The evidence `VCEK_REPORT` used by the unit test does not have `cpuid_fam_id` field, but the api of `parse_tee_evidence` needs the parameter. it's just used to call the `parse_tee_evidence` function and not used in the test logic.

A better way is to tidy the functions of SNP, to make the api clean for the black boxes.

I can do this in the PR.

amd-aliem:
@Xynnn007 
What will currently happen if a user misses the release notes and doesn't update their policy? Is there any kind of policy validation for missing fields if a user tries to load their old policy file with the old fields? I'm worried about the old policy fields silently failing, and no longer checking the fields they were configured for.

If this is a valid concern, maybe we could leave the old fields and just do an additive change? Still update the default policy example, but token could work with both old and new policies. Or add some validation during policy load time.

Xynnn007:
@amd-aliem Yes. This is a reasonable concern. Note that the old fields are flattened, and the new format is nested. The  reason is that we also want to make the claim's structure consistent with the original evidence, and also bring those 'initdata', 'report_data' fields into a separate place isolated with original evidence fields.

In future, I want to move all `evidence` fields to EAR's [attester claim](https://www.ietf.org/archive/id/draft-ietf-rats-ear-04.html#name-attester-claims) to follow the standard, which will depend on the current shape.

So what about this:

1. Adds the whole `evidence` sub map along side into the original parsed claims, so this will not influence any current behavior. This shape would exist some time in the release cycle.
2. Document about the new structure into the README.md, and mention that the original fields **MAY BE REMOVED** in future in both docs and the code (by `warn` log).
3. In the step when applying to "attester claim", we can move the whole `evidence` map to that. Then state to remove original claims shape in the EAR in the next release.

This is a step-by-step changing, towards a standard way but leave time for users to know.

Xynnn007:
@amd-aliem Updated with add-only way. PTAL

hgowda-amd:
> @amd-aliem Yes. This is a reasonable concern. Note that the old fields are flattened, and the new format is nested. The reason is that we also want to make the claim's structure consistent with the original evidence, and also bring those 'initdata', 'report_data' fields into a separate place isolated with original evidence fields.
> 
> In future, I want to move all `evidence` fields to EAR's [attester claim](https://www.ietf.org/archive/id/draft-ietf-rats-ear-04.html#name-attester-claims) to follow the standard, which will depend on the current shape.
> 
> So what about this:
> 
> 1. Adds the whole `evidence` sub map along side into the original parsed claims, so this will not influence any current behavior. This shape would exist some time in the release cycle.
> 2. Document about the new structure into the README.md, and mention that the original fields **MAY BE REMOVED** in future in both docs and the code (by `warn` log).
> 3. In the step when applying to "attester claim", we can move the whole `evidence` map to that. Then state to remove original claims shape in the EAR in the next release.
> 
> This is a step-by-step changing, towards a standard way but leave time for users to know.

Hi @Xynnn007 ,
If this is eventually going to be changed to strictly adhere to the EAR token format, do you think we should proceed with this PR at this point?

From the AMD side, we’re still working through which claims need to be matched and how they should be evaluated. Because of that, I don’t think adding all the fields from the report to the default policy is the right approach yet.

If you’re okay with it, could we hold off on merging this PR until we come back with a proposal? That would help ensure that both the token format and the set of claims being evaluated remain stable, since changes like these can have a direct impact on relying parties.

Xynnn007:
> From the AMD side, we’re still working through which claims need to be matched and how they should be evaluated. Because of that, I don’t think adding all the fields from the report to the default policy is the right approach yet.
> 
> If you’re okay with it, could we hold off on merging this PR until we come back with a proposal? That would help ensure that both the token format and the set of claims being evaluated remain stable, since changes like these can have a direct impact on relying parties.

That's fair.

fitzthum:
@hgowda-amd anything happening on this one?

hgowda-amd:
> @hgowda-amd anything happening on this one?

Yes, I think it would be better to put this PR on hold for now, since we’re expecting changes in the near future that will require us to update the policy format anyway. Given that this is a breaking change for relying parties, it would be better to minimize the frequency of these kinds of changes.

fitzthum:
btw @Xynnn007 why introduce the evidence object here? Most of the verifiers don't have any equivalent.

Xynnn007:
> btw @Xynnn007 why introduce the evidence object here? Most of the verifiers don't have any equivalent.

New EAR token [claims](https://www.ietf.org/archive/id/draft-ietf-rats-ear-04.html#section-3.1) have been updated, with `attester-claims` and `verifier-claims`.

I want to put all claims parsed from original evidence into `attester-claims` (defined as "A map containing claims with Attester authority, i.e., claims extracted from the appraised Evidence.")

And for `init_data`, `report_data` and other fields, we put into `verifier-claims` (defined as "A map containing claims with Verifier authority, i.e., claims added by the Verifier during appraisal and derived from evaluating the Appraisal Policy for Evidence.")

This is a pre-work for SNP for `attester-claims`.

fitzthum:
If we want to adopt `attester-claims` and `verifier-claims` (which I am somewhat skeptical about), maybe we should do that in a separate PR across all the architectures.

Xynnn007:
> If we want to adopt `attester-claims` and `verifier-claims` (which I am somewhat skeptical about), maybe we should do that in a separate PR across all the architectures.

That makes sense. Let me
1. wait for AMD new changing PR and keep this one pending
2. rebase with this one
3. then update ear v4

PR #1359, opened 2026-05-20
KBS | Promote the configuration error printing

Xynnn007:
This is based on #1239 

The core idea is from https://github.com/confidential-containers/trustee/pull/1239#discussion_r3264201873
'''
The goal is to inform the user that the current configuration will not take effect. Could I do this by mandating that the input configuration format must conform to certain fields, otherwise throwing an error and referencing a configuration file format documentation?

Because if we set a precedent here—meaning that every time the configuration file is modified, we need to record historical configuration items in the code—we'll need to do the same elsewhere in the future. This will result in a lot of redundant code. My suggestion is to avoid saying "which are no longer supported" and instead directly say "the current format is as follows, there are configuration errors, startup failed, and modifications are needed."
'''

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3654,6 +3654,7 @@ dependencies = [
  "chrono",
  "clap",
  "config",
+ "const_format",
  "cryptoki",
  "educe",
  "hex",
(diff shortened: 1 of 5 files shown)


fitzthum (approved):
LGTM

Eventually we should move that config doc to the website

PR #1360, opened 2026-05-20
 kbs: add insecure options for admin/attestation token verification

Xynnn007:
This patch adds three switches of insecure options for admin/attestation token verification
- Allow fetch verification public keys from HTTP endpoints for Attestation Token
- Allow fetch verification public keys from HTTP endpoints for Admin Token
- Skip `jwk` field's endorsement check for Admin Token

These are useful in dev cases.

Relies on #1239

### DIFF
--- a/kbs/docs/admin.md
+++ b/kbs/docs/admin.md
@@ -79,7 +79,8 @@ This lets you keep identity issuance external while keeping KBS-side authorizati
 `[admin.authentication.bearer_jwt]` accepts:
 
 - `identity_providers` (array): list of trusted identity providers
-- `insecure_public_key_uri` (optional bool, default `false`): allow fetching verification keys over plaintext `http://`
+- `insecure_public_key_uri` (optional bool, default `false`): allow fetching verification keys over plaintext `http://`. Applies to configured `public_key_uri` and `jwk_set_uri` values, and to any `jwks_uri` returned by OpenID discovery for remote `jwk_set_uri` entries
+- `insecure_header_jwk` (optional bool, default `false`): skip endorsement checks for JWT header-embedded JWK keys (signature is still verified, testing only)
 
 Each `identity_providers` entry:
 
--- a/kbs/docs/admin.md
+++ b/kbs/docs/admin.md
@@ -79,6 +79,7 @@ This lets you keep identity issuance external while keeping KBS-side authorizati
 `[admin.authentication.bearer_jwt]` accepts:
 
 - `identity_providers` (array): list of trusted identity providers
+- `insecure_public_key_uri` (optional bool, default `false`): allow fetching verification keys over plaintext `http://`
 
 Each `identity_providers` entry:
 
@@ -93,8 +94,11 @@ Each entry must provide at least one of `public_key_uri` or `jwk_set_uri`.
 
 - `file://...` or local path (for example `./keys/admin.jwks`): JWKS JSON file, read directly
 - `https://...`: remote JWKS URL or OpenID issuer base URL (see note below)
+- `http://...`: same as `https://`, only when `insecure_public_key_uri=true`
 
 > [!NOTE]
 > For remote `jwk_set_uri` values, KBS first tries to load JWKS from the configured URL
 > directly (for example a `/jwks/` endpoint). If that fails, it falls back to OpenID discovery
 > at `{uri}/.well-known/openid-configuration` and loads keys from the returned `jwks_uri`.
+> Plaintext `http://` is rejected unless `insecure_public_key_uri=true`, including on
+> URLs returned by discovery.
--- a/kbs/src/admin/authentication/bearer_jwt.rs
+++ b/kbs/src/admin/authentication/bearer_jwt.rs
@@ -21,9 +21,18 @@ pub struct BearerJwtConfig {
 
     /// Allow loading admin verification keys over plaintext HTTP.
     ///
-    /// When false, any HTTP fetch for key material is rejected. Keep disabled by default and only
-    /// enable in controlled environments.
+    /// When false, any HTTP fetch for key material is rejected: configured `public_key_uri`
+    /// and `jwk_set_uri` values, and any `jwks_uri` returned by OpenID discovery for remote
+    /// `jwk_set_uri` entries. Keep disabled by default and only enable in controlled environments.
     pub insecure_public_key_uri: bool,
+
+    /// Whether to skip endorsement checks of header-embedded JWK keys.
+    /// Signature is still verified.
+    /// This should only be set to true for testing.
+    ///
+    /// When false, header-embedded `jwk` keys must include an `x5c` certificate chain
+    /// that can be validated against configured trusted certificates (if any).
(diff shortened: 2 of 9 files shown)


kbs/docs/admin.md
```diff
@@ -92,5 +93,6 @@ Each entry must provide at least one of `public_key_uri` or `jwk_set_uri`.
 Supported source formats:
 
 - `https://...` (remote fetch)
+- `http://...` (remote fetch, only when `insecure_public_key_from_uri=true`)
```
mythi: I think this is an "expensive" developer feature to maintain. Can you comment on why manual download and `file://` or developers setting custom CA bundles to `https://` do not work as options?

Xynnn007: This is usable during development and deployment. As mentioned earlier, some locally deployed services expose relevant APIs, allowing KBS to download them automatically upon startup. KBS employs this insecure switch design in many places. Regarding maintainability, the maintenance of the relevant code is currently my responsibility and will continue to be my responsibility.

Adding this description to the README.md is not good, as it's enough to tell users the existance of this feature without too much details.

mythi: OK, but I did not see the answer to my question.

Xynnn007: It also works. But some auth services usually roll-out their keys, so an old key downloaded may expired when new KBS deploys. We need to manually download them and then replace. To sum-up, the process needs extra manual steps.

Xynnn007:
Added another commit for `attestation token` part to allow `http://` jwk set, which will help to connect to an http `.well-known` to get jwk sets.

mythi:
My opinion on this PR has not changed. It's an "expensive" developer feature to be added to the config especially when there's no real blockers that justify these changes.

Xynnn007:
> My opinion on this PR has not changed. It's an "expensive" developer feature to be added to the config especially when there's no real blockers that justify these changes.

Cannot agree with the point. It's disabled by default and requires explicit user enablement. Just like KBS currently supports both HTTP and HTTPS startup.

In many enterprise environments and development/testing scenarios, where users cannot provide a CA, this can be very helpful. You mentioned using configmaps for configuration, but this requires users to manually update the jwk set instead of automatically fetching updates.

It's important to understand that introducing manual updates is very cumbersome.

mythi:
> > My opinion on this PR has not changed. It's an "expensive" developer feature to be added to the config especially when there's no real blockers that justify these changes.
> 
> Cannot agree with the point. It's disabled by default and requires explicit user enablement. Just like KBS currently supports both HTTP and HTTPS startup.

What was done in the past does not necessarily mean it needs to be followed.
 
> In many enterprise environments and development/testing scenarios, where users cannot provide a CA, this can be very helpful. You mentioned using configmaps for configuration, but this requires users to manually update the jwk set instead of automatically fetching updates.
> 
> It's important to understand that introducing manual updates is very cumbersome.

This PR has changed from admin only topic to something else. My comments have been related to that admin http jwkset part only. How likely it is that these jwksets need to be updated regularly during testing?

Can we keep the two separate PRs so that one topic does not block the other.

kbs/src/crypto/jwk.rs
```diff
@@ -29,37 +32,24 @@ pub(crate) struct OpenIDConfig {
...
-                .await
-                .map_err(|e| JwksGetError::AccessFailed(e.to_string()))?
-                .json::<JwkSet>()
-                .await
-                .map_err(|e| JwksGetError::FailedToGetKeyMaterial {
-                    source: Into::<anyhow::Error>::into(e).context("failed to get JWK set"),
-                })
+        "https" | "http" => {
+            ensure_http_allowed(url.scheme(), insecure_public_key_from_uri)?;
+            read_jwk_from_remote(&url, uri, insecure_public_key_from_uri).await
```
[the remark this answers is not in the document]
Xynnn007: Let's consider this later.

Xynnn007:
> > In many enterprise environments and development/testing scenarios, where users cannot provide a CA, this can be very helpful. You mentioned using configmaps for configuration, but this requires users to manually update the jwk set instead of automatically fetching updates.
> > It's important to understand that introducing manual updates is very cumbersome.
> 
> This PR has changed from admin only topic to something else. My comments have been related to that admin http jwkset part only. How likely it is that these jwksets need to be updated regularly during testing?
> 
> Can we keep the two separate PRs so that one topic does not block the other.

Not just testing, but inner deployments

I get attestation-token things into https://github.com/confidential-containers/trustee/pull/1434.

About admin part, I believe this follows the industry-standard secure-by-default pattern: Redis `protected-mode` defaults to yes and only disables when explicitly set to no in config. Similarly, this PR keeps things disabled by default, requiring an explicit configuration change to enable. [OWASP Proactive Controls C5](https://devguide.owasp.org/en/04-design/02-web-app-checklist/01-secure-by-default/) states "the default settings are always the most secure option," and CISA's Secure by Design guidance confirms "out-of-the-box, products should be secure." Defaulting to off and requiring explicit opt-in does not introduce risk—it acknowledges legitimate use cases while ensuring they are intentional and auditable. Removing this option would only push users toward less visible workarounds.


Completely blocking the path is not what we want; a better approach is to provide a door that users can choose to open or close.

kbs/src/admin/authentication/bearer_jwt.rs
```diff
@@ -18,12 +18,30 @@ use crate::crypto::jwt::JwtVerifier;
 #[serde(default, deny_unknown_fields)]
 pub struct BearerJwtConfig {
     pub identity_providers: Vec<IssuerConfig>,
+
+    /// Allow loading admin verification keys over plaintext HTTP.
+    ///
+    /// When false, any HTTP fetch for key material is rejected: configured `public_key_uri`
+    /// and `jwk_set_uri` values, and any `jwks_uri` returned by OpenID discovery for remote
+    /// `jwk_set_uri` entries. Keep disabled by default and only enable in controlled environments.
+    pub insecure_public_key_from_uri: bool,
```
mkulke: nit: for consistency: "insecure_public_key_uri"

kbs/src/token/mod.rs
```diff
@@ -60,6 +62,14 @@ pub struct AttestationTokenVerifierConfig {
     /// Default: false
     #[serde(default = "bool::default")]
     pub insecure_header_jwk: bool,
+
+    /// Allow loading attestation-token verification keys over plaintext HTTP.
```
mkulke: nit: plaintext is proably the wrong reference here. a public key is by definition not a secret, however HTTPS is required for authenticity/integrity of the public key.

Xynnn007: deleted `plaintext`

fitzthum:
Yeah fwiw I am fine with adding this. We should be careful with adding insecure options, but I think this one is reasonable. This is quite a niche use case, but if you're trying to setup a demo or test that uses an IDP, you probably are not going to have HTTP setup. You could always specify the keys directly in this case, but that changes the flow a bit.

PR #1368, opened 2026-05-22
verifier: intel_dcap: implement CollateralService and drop libsgx-dcap-default-qpl

mythi:
`libsgx-dcap-default-qpl` and the Rust wrappers for pulling DCAP quote verification collateral has problems such as:

1. "offline" support is not possible without getting a PCCS (caching) service to run
2.  async of `tee_qv_get_collateral()` not handled well
3.  configuration requires a file write to `QCNL_CONF_PATH`
4. https://github.com/intel/confidential-computing.tee.dcap/pull/372

Following the DCAP design where "quote provider library" can be a custom implementation and `libsgx-dcap-default-qpl` is the reference for PCCS/PCS based collateral, define `CollateralService` and add `impl Pcs` for it. Currently, `CollateralService` adds support for verification collateral only.


deps/verifier/src/intel_dcap/collateral_service.rs
```diff
@@ -0,0 +1,342 @@
...
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub advisory_ids: Option<Vec<String>>,
+}
+
+/// TCB SVN components for a single TCB level.
+#[derive(Serialize, Deserialize)]
+pub struct Tcb {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub sgxtcbcomponents: Option<TcbComponentList>,
```
Xynnn007, line 129: just nit. Can we use `sgx_tcb_components` as variable name and add a `serde(rename)` macro? ditto for the other fields inthis struct `Tcb`

deps/verifier/Cargo.toml
```diff
@@ -63,7 +63,7 @@ csv-rs = { git = "https://github.com/openanolis/csv-rs", rev = "b67a07e", option
     "network",
 ] }
 eventlog = { path = "../eventlog", optional = true }
-hex.workspace = true
+hex = { version = "0.4.3", features = ["serde"] }
```
Xynnn007: we could reuse the version `hex = { workspace = true, features = ["serde"] }`

deps/verifier/Cargo.toml
```diff
@@ -98,6 +98,7 @@ http-cache-reqwest = { version = "1.0.0-alpha.6", default-features = false, feat
     "manager-moka",
     "url-standard",
 ] }
+chrono = { version = "0.4", features = ["serde"] }
```
Xynnn007: we could use the workspace one
`chrono = { workspace = true, features = ["serde"] }`

deps/verifier/src/intel_dcap/collateral_service.rs
```diff
@@ -0,0 +1,342 @@
...
+
+    #[serde(default)]
+    pub tdx_tcbinfo: Option<TcbInfoJson>,
+    #[serde(default)]
+    pub tdx_tcbinfo_early: Option<TcbInfoJson>,
+}
+
+#[derive(Deserialize, Serialize)]
+pub struct PckCaCrl {
+    #[serde(rename = "processorCrl")]
```
Xynnn007: We could use `#[serde(rename_all = "camelCase")]` upon the whole struct instead field-specific ones

deps/verifier/src/intel_dcap/collateral_service.rs
```diff
@@ -0,0 +1,342 @@
...
+    pub processor_crl: String,
+    #[serde(rename = "platformCrl", default)]
```
Xynnn007: ditto

deps/verifier/src/intel_dcap/collateral_service.rs
```diff
@@ -0,0 +1,342 @@
...
+    pub platform_crl: Option<String>,
+}
+
+#[derive(Deserialize, Serialize)]
+pub struct PckCrlIssuerChains {
+    #[serde(rename = "PROCESSOR", default)]
+    pub processor: Option<String>,
+    #[serde(rename = "PLATFORM", default)]
```
Xynnn007: `#[serde(rename_all = "UPPERCASE")]` can be used upon the whole struct

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,435 @@
...
+    #[error("Response error: {0}")]
+    Response(#[from] reqwest::Error),
+    #[error("Send error: {0}")]
+    Send(#[from] reqwest_middleware::Error),
+    #[error("URL parse error: {0}")]
+    UrlParse(#[from] url::ParseError),
+    #[error("Chain decoding error: {0}")]
+    ChainDecoding(#[from] std::string::FromUtf8Error),
+    #[error("Root CA CRL distribution point URL not found in certificate chain")]
+    MissingCrlDistributionPoint,
```
Xynnn007: This enum seems overlapped with `Missing(String)`?

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,435 @@
...
+    Missing(String),
+}
+
+pub struct Pcs {
+    inner: PcsInner,
+    tcb_update_type: TcbUpdateType,
+}
+
+impl Pcs {
+    pub fn new(config: &QcnlConfig) -> anyhow::Result<Self> {
```
Xynnn007: Let's also mark this return type to `Result<Self, PcsError>`. As all `pub`-ed functions in this module will share a same error type.

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,435 @@
...
+        let mut params: HashMap<&str, String> = HashMap::new();
+        let chain_header = Self::cert_chain_header(&ct);
+
+        let mut endpoint = match ct {
+            CollateralType::TcbInfo(tee, fmspc) => {
+                params.insert("fmspc", hex::encode_upper(fmspc));
+                params.insert(
+                    "update",
+                    match self.tcb_update_type {
+                        TcbUpdateType::Early => "early",
```
Xynnn007: just nit. We can use strum to impl a `to_string` for `TcbUpdateType` to avoid ad-hoc to string here

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,435 @@
...
+    }
+}
+
+impl CollateralService for Pcs {
+    type Error = PcsError;
+
+    async fn get(&self, ct: CollateralType<'_>) -> Result<CollateralData, PcsError> {
+        match &self.inner {
+            PcsInner::File { path } => self.get_from_file(path, ct).await,
+            PcsInner::Http { url, is_pcs } => self.get_from_http(url, *is_pcs, ct).await,
```
Xynnn007: Note that the `url`, `is_pcs` is already the member of `self`, but we use them as parameters of the `self` function - meaning that inside `get_from_http` we can still access the variables via `self`. This design pattern looks not clean. but not a great deal, we can make it better then

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,435 @@
...
+fn url_decode_bytes(s: &str) -> Result<Vec<u8>, PcsError> {
+    Ok(urldecode(s)
+        .map_err(PcsError::ChainDecoding)?
+        .into_owned()
+        .into_bytes())
+}
+
+#[cfg(test)]
+mod tests {
+    #[cfg(feature = "openssl")]
```
Xynnn007: Let's mark `openssl` as a dependency for both feature = "tdx-verifier" and feature = "sgx-verifier". Then we can test it.

Btw, can we have a test with `Pcs::` from local file, and do `trait CollateralService` for it?

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,435 @@
...
+pub struct Pcs {
+    inner: PcsInner,
+    tcb_update_type: TcbUpdateType,
+}
+
+impl Pcs {
+    pub fn new(config: &QcnlConfig) -> anyhow::Result<Self> {
+        let tcb_update_type = config.tcb_update_type.clone();
+
+        if config.collateral_service.starts_with("file://") {
```
Xynnn007: According to the document, we only allow `file://` and `http(s)://`, right? Do you think it a good idea to add a `README.md` under `intel_dcap` to show some information about the module? As this is somehow user-facing.

deps/verifier/src/intel_dcap/mod.rs
```diff
@@ -56,36 +55,25 @@ impl Default for QcnlConfig {
...
-            }
-            _ => warn!("DCAP QCNL configuration failed due to an unknown error."),
-        })
-        .inspect(|_| debug!("DCAP QCNL configuration was written to $QCNL_CONF_PATH."))
+impl QcnlConfig {
+    /// Build a [`pcs::Pcs`] collateral client from the configured `collateral_service` URL.
+    ///
+    /// A `file://` URL selects the offline file backend; any other URL selects
+    /// the PCS/PCCS HTTPS backend.
+    #[cfg(any(feature = "tdx-verifier", feature = "sgx-verifier"))]
```
Xynnn007: Why do we need this feature flag?

I noticed that there are many feature flags in the module. Usually, we prefer to put all codes inside a mod and give feature flag upon the module. But this can be optimized later.

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,435 @@
...
+    MissingCrlDistributionPoint,
+    #[error("I/O error: {0}")]
+    Io(#[from] std::io::Error),
+    #[error("JSON error: {0}")]
+    Json(#[from] serde_json::Error),
+    #[error("Hex decode error: {0}")]
+    Hex(#[from] hex::FromHexError),
+    #[error("{0}")]
+    Missing(String),
+}
```
Xynnn007, line 95: I've noticed that `anyhow::Error` and `thiserror::Error` are mixed by `pub` and `pub(crate)` functions. Usually we'd prefer to have a common error interface. Using purely `anyhow::{Result, Error}` would be a good choice and we do not need this error enum.

Xynnn007 (commented):
Some nits. 

I've met a concurrency limit when using DCAP to verify quotes. This might be related to rust-c binding and different co-routine models for rust(tokio) and C. qpl-replacement would help I think. So this will be a very great work for performance optimization.

There is a question: do you have plans to rust-ify quote-verification logic?

mythi:
> I've met a concurrency limit when using DCAP to verify quotes. This might be related to rust-c binding and different co-routine models for rust(tokio) and C. qpl-replacement would help I think. So this will be a very great work for performance optimization.

I'm curious about this. I did benchmark the new implementation and it was marginally faster than the old/existing C based. What do you mean by concurrency limit? Or you mean quote verification process overall?

deps/verifier/src/intel_dcap/pcs.rs
mythi: > Btw, can we have a test with `Pcs::` from local file, and do `trait CollateralService` for it?

The file is rather big so I dropped that test. I can see if I could strip it manually (the intel tool gives me Tcb Info for all platforms) to get it committed in the repo. I wanted to start something simple and have only one impl.

deps/verifier/src/intel_dcap/collateral_service.rs
mythi: currently, I've tried to keep these so that I can verify the signature after a round-trip deserialize+serialize. many things could be changed/simplified if only deser if seen sufficient

Xynnn007:
> I'm curious about this. I did benchmark the new implementation and it was marginally faster than the old/existing C based. What do you mean by concurrency limit? Or you mean quote verification process overall?

I get an im-perfect test here https://github.com/Xynnn007/AS-Performance/issues/1

If throughput growth is not linearly correlated with computing resource growth, then some internal programming model must be the bottleneck. Historically, we've tried timestamping various points within the CoCoAS, and found that the time difference before and after calling the C-Rust binding function `intel_tee_quote_verification_rs` was almost identical.

deps/verifier/src/intel_dcap/pcs.rs
mythi: I fixed some of these and will take another review to see if they can be simplified. I prefer keeping the std error trait bound for CollateralService. `new()` moved to `PcsError` as well.

deps/verifier/src/intel_dcap/pcs.rs
mythi: Will add docs once the code is fine. I originally did not plan to add `file://` in this PR yet but changed my mind.

The question for review is do you think it's fine to re-purpose `collateral_service` for `file://` as well. QcnlConfig (we can rename that later) structure uses the same fields as SGX QCNL library uses so I wanted to keep that "compatibility" here.

mythi:
> > I'm curious about this. I did benchmark the new implementation and it was marginally faster than the old/existing C based. What do you mean by concurrency limit? Or you mean quote verification process overall?
> 
> I get an im-perfect test here [Xynnn007/AS-Performance#1](https://github.com/Xynnn007/AS-Performance/issues/1)
> 
> If throughput growth is not linearly correlated with computing resource growth, then some internal programming model must be the bottleneck. Historically, we've tried timestamping various points within the CoCoAS, and found that the time difference before and after calling the C-Rust binding function `intel_tee_quote_verification_rs` was almost identical.

OK. I implemented a micro-benchmark using Criterion for `ecdsa_quote_verification()` that compared the existing `QuoteCollateral` method and what's in this PR. Both had their "caching" enabled in and included in the run.

I've worked on a POC for full Rust based quote verification and I can try to benchmark that too but I first need to get all these changes merged (they have been built such that they support replacing the existing `ecdsa_quote_verification()` easily).

deps/verifier/src/intel_dcap/mod.rs
mythi: These are now all gone. Resolving.

deps/verifier/src/intel_dcap/pcs.rs
mythi: I've now simplified these errors quite a bit and added a unified `Collateral` error for missing collateral error that is the same independent of the source.

deps/verifier/src/intel_dcap/pcs.rs
mythi: Updated docs.

mythi:
The latest version is tested using PCS (quote 4) and Ali PCS (quote 5) and offline collateral and should address all important feedback comments. 

I'm keeping that `QcnlConfig` struct name but we can change it later.

deps/verifier/src/intel_dcap/pcs.rs
Xynnn007: apply for the openssl thing, please

attestation-service/docs/config.md
```diff
@@ -168,17 +168,20 @@ Available when the `tdx-verifier`, `sgx-verifier`, or `az-tdx-vtpm-verifier` fea
...
 [dcap-1]: #qcnlconfig
 
 ###### QcnlConfig
 
-| Property              | Type    | Description                                           | Required | Default                                                    |
-|-----------------------|---------|-------------------------------------------------------|----------|------------------------------------------------------------|
-| `collateral_service`  | String  | URL of the Intel PCS collateral service               | No       | `https://api.trustedservices.intel.com/sgx/certification/v4/` |
-| `use_secure_cert`     | Boolean | Whether to use secure certificates                    | No       | -                                                          |
-| `tcb_update_type`     | String  | TCB update type: `"early"` or `"standard"`            | No       | `"early"`                                                  |
+| Property             | Type    | Description                                          | Required | Default                                                      |
```
Xynnn007, line 177: Add the documents also to kbs/docs/config.md

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,430 @@
...
+            mode: CacheMode::ForceCache,
+            manager: PCS_CACHE_MANAGER.clone(),
+            options: HttpCacheOptions {
+                cache_status_headers: true,
+                max_ttl: Some(PCS_CACHE_MAX_AGE),
+                ..Default::default()
+            },
+        }))
+        .build()
+});
```
Xynnn007, line 50: Note that a global `PCS_CLIENT` can cause lock contention, leading to lower concurrency performance. This can be optimized; it's recommended to include a `TODO` stub in the code.

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,430 @@
...
+
+struct Cached {
+    data: Arc<PcsCollaterals>,
+    mtime: SystemTime,
+}
+
+static FILE_CACHE: LazyLock<RwLock<HashMap<PathBuf, Cached>>> =
```
Xynnn007, line 57: let's use  `tokio::sync::Rwlock` instead of `sync::Rwlock`. The mixing of async and sync primitives will cause UB.

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,430 @@
...
+    }
+}
+
+fn url_decode_bytes(s: &str) -> Result<Vec<u8>, PcsError> {
+    Ok(urldecode(s)?.into_owned().into_bytes())
+}
+
+#[cfg(test)]
+mod tests {
+    #[cfg(feature = "openssl")]
```
Xynnn007: This feature flag too. Let's add openssl under dcap's feature deps.

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,430 @@
...
+    UrlParse(#[from] url::ParseError),
+    #[error("I/O error: {0}")]
+    Io(#[from] std::io::Error),
+    #[error("Decoding error: {0}")]
+    Decoding(String),
+    #[error("Collateral not available from {0}")]
+    Collateral(String),
+    #[error("Internal error: {0}")]
+    Internal(&'static str),
+}
```
Xynnn007, line 83: This error is an over-design. Because the upper-level caller will wrap this error with anyhow, the error presented to the upper level will be in the form of

"Decoding error: ", which still doesn't really tell the developer what happened. My suggestions are:

1. Either don't use thiserror and replace it with anyhow for everything, providing detailed error context using context.

2. Or keep thiserror, but provide more detailed error information.

Xynnn007 (commented):
we are getting near to goal.

deps/verifier/src/intel_dcap/pcs.rs
mythi: Going with 2.

attestation-service/docs/config.md
mythi: Will add a pointer to this doc to avoid double maintenance.

deps/verifier/src/intel_dcap/pcs.rs
mythi: Curious what needs to change here? LazyLock is taken once on init. Reqwest Client handles multithreading.

Xynnn007: There is a lock behind lazy lock. Everytime the PCS_CLIENT is called, the lock will help keep different callers mutually exclusive.

deps/verifier/src/intel_dcap/collateral.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    bail!("no URI found in CRL Distribution Points of root CA cert")
+}
+
+/// Fetch all verification collateral from PCS and return a [`QuoteCollateral`]
+/// ready for use with `tee_verify_quote`.
+///
+/// The Root CA CRL distribution point URL is read from the intermediate CA certificate
+/// in the QE Identity issuer chain, which is the authoritative source per the Intel
+/// DCAP certificate profile specification.
```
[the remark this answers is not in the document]
mythi: fixed

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,430 @@
...
+    pub fn new(config: &QcnlConfig) -> Result<Self, PcsError> {
+        let tcb_update_type = config.tcb_update_type.clone();
+
+        if config.collateral_service.starts_with("file://") {
+            let url = url::Url::parse(&config.collateral_service)?;
+            return Ok(Self {
+                inner: PcsInner::File {
+                    path: PathBuf::from(url.path()),
+                },
+                tcb_update_type,
```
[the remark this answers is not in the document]
mythi: fixed

deps/verifier/src/intel_dcap/pcs.rs
mythi: updated

deps/verifier/src/intel_dcap/pcs.rs
mythi: dropped. openssl under dcap is not correct either.

deps/verifier/src/intel_dcap/pcs.rs
mythi: good catch, fixed

attestation-service/docs/config.md
mythi: added a pointer to AS docs

mythi:
rebased to fix merge conflicts

deps/verifier/src/intel_dcap/pcs.rs
mythi: I don't think that is the case. From what I've read, it locks on first init but not after that.

Xynnn007: Ok. I think I mixed `OnceLock` and `OnceLock<Mutex<..>>` which we usually use. You are rignt.

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,430 @@
...
+use super::collateral_service::{
+    CollateralData, CollateralService, CollateralType, IntelTee, PcsCollaterals,
+    PlatformCollaterals,
+};
+use super::{QcnlConfig, TcbUpdateType};
+
+const PCS_CACHE_MAX_AGE: Duration = Duration::from_secs(168 * 3600);
+
+static PCS_CACHE_MANAGER: LazyLock<MokaManager> =
+    LazyLock::new(|| MokaManager::new(MokaCacheBuilder::new(1024).build()));
```
Xynnn007: Note that if this is only used in `PCS_CLIENT`, we can move into `PCS_CLIENT`'s `|| {..}` function. As a `clone` will keep the original object occupying memory w/o using it.

deps/verifier/src/intel_dcap/pcs.rs
Xynnn007: Btw why we need a HashMap here? Looks like the RwLock here is only used to track the mtime of the collateral file?

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,430 @@
...
+                    }
+                    (IntelTee::Tdx, _) => &col.tdqeidentity,
+                    (IntelTee::Sgx, TcbUpdateType::Early) => {
+                        col.qeidentity_early.as_ref().unwrap_or(&col.qeidentity)
+                    }
+                    (IntelTee::Sgx, _) => &col.qeidentity,
+                };
+
+                Ok(CollateralData {
+                    body: serde_json::to_vec(body)?,
```
Xynnn007, line 236: The error conversion will make here an error message `failed to decode collateral data`, but looks like it's serialization

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,430 @@
...
+}
+
+fn url_decode_bytes(s: &str) -> Result<Vec<u8>, PcsError> {
+    Ok(urldecode(s)?.into_owned().into_bytes())
+}
+
+#[cfg(test)]
+mod tests {
+    #[tokio::test]
+    #[ignore = "run with: cargo test -p verifier --no-default-features --features tdx-verifier,openssl -- tcb_info_signature --include-ignored"]
```
Xynnn007: `tdx-verifier` feature's test is not explicitly showed in the `Cargo.toml`. another way is directly put openssl in `[dev-dependencies]` of `Cargo.toml`.

mythi: are you asking this to be changed?

Xynnn007: right

deps/verifier/src/intel_dcap/pcs.rs
mythi: It's not used currently but helps to support different files for different scenarios (like separate files for each sgx/tdx/az-vtpm)

deps/verifier/src/intel_dcap/pcs.rs
mythi: done. leftover from incremental changes. `clone()` uses `Arc` it seems so it was cheap but good to be fixed

deps/verifier/src/intel_dcap/pcs.rs
mythi: got rid of that

deps/verifier/src/intel_dcap/pcs.rs
mythi: added

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,429 @@
...
+                    (IntelTee::Sgx, TcbUpdateType::Standard) => entry.sgx_tcbinfo.as_ref(),
+                }
+                .ok_or_else(|| {
+                    debug!("no {:?} tcbinfo for fmspc {:02x?} in {source}", tee, fmspc);
+                    PcsError::Collateral(source.to_string())
+                })?;
+
+                Ok(CollateralData {
+                    // Safe: we just deserialized this from JSON, so it round-trips without error.
+                    body: serde_json::to_vec(tcb_info).expect("serialize TcbInfoJson"),
```
Xynnn007: Let's avoid `expect`. it will cause panic and influence upper layer. One way is to add one enum of PcsError about serialization, different from deserialization. Then use `map_err` or something here.

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,429 @@
...
+                    (IntelTee::Tdx, _) => &col.tdqeidentity,
+                    (IntelTee::Sgx, TcbUpdateType::Early) => {
+                        col.qeidentity_early.as_ref().unwrap_or(&col.qeidentity)
+                    }
+                    (IntelTee::Sgx, _) => &col.qeidentity,
+                };
+
+                Ok(CollateralData {
+                    // Safe: we just deserialized this from JSON, so it round-trips without error.
+                    body: serde_json::to_vec(body).expect("serialize QeIdentity"),
```
Xynnn007: expect here too

deps/verifier/src/intel_dcap/pcs.rs
Xynnn007: Then you need to remove the ignore message here

deps/verifier/src/intel_dcap/pcs.rs
mythi: An error for something where an error cannot happen conflicts with your earlier feedback that `PcsError` is over-engineered. What approach would be acceptable to you?

deps/verifier/src/intel_dcap/pcs.rs
mythi: will drop `,openssl` on next push

deps/verifier/src/intel_dcap/pcs.rs
Xynnn007: Oh. The comment makes no sense here. The `to_vec` behavior is not related to deserilization

mkulke: I would also ecommend avoiding panics. an invariant can change after refactoring and serialization can still fail even when deserialization worked, depending on serde directives.

the deserialization is not local in this fn (maybe you can do this here and then clone or use a reference to the original data), so I would return an error type even if it is a programmer error.

mythi: it refers to `let pc: PlatformCollaterals = serde_json::from_str(&text)?;` which returns a decode error if the input data is not valid (including `TcbInfoJson` and  `QeIdentity`). My thinking is that doing the inverse cannot error once we've reached the point here: `body: serde_json::to_vec(tcb_info).expect("serialize TcbInfoJson")`

mkulke: I understand, but in the future someone could change that code or put an arbitrary serialization rule on a struct field and then it would be fallible. IMO it's good code hygiene to not use panics if we can encode the invariants in the type system.

> Serialization can fail if T’s implementation of Serialize decides to fail, or if T contains a map with non-string keys.

mythi: (sorry, I did not see your earlier comment when I replied).

I don't insist keeping `expect()`  if there was another acceptable solution.

From what I can see, at least I cannot do 
> PcsError about serialization, different from deserialization

very easily either.

mkulke: wouldn't that work?

```rust
let body = serde_json::to_vec(tcb_info).or(PcsError::Internal("tcb info serialization failed"))?;
```

mythi: Works for me. I pushed an alternative that adds `Json` error back as I missed this comment again. Either way is fine

mkulke (approved):

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -0,0 +1,423 @@
...
+        .build()
+});
+
+struct Cached {
+    data: Arc<PcsCollaterals>,
+    mtime: SystemTime,
+}
+
+static FILE_CACHE: LazyLock<RwLock<HashMap<PathBuf, Cached>>> =
+    LazyLock::new(|| RwLock::new(HashMap::new()));
```
fitzthum, line 58: We may want to use a centralized thing for this at some point, but fine for now.

fitzthum (approved):
LGTM. one note 

It would be nice if we had TEE runners for this repo for changes like this.

mythi:
> It would be nice if we had TEE runners for this repo for changes like this.

Agree. For TDX we have it on our backlog but haven't gotten to it yet

PR #1369, opened 2026-05-25
ci: fix link check for ascii-code.com

Xynnn007:
Error like https://github.com/confidential-containers/trustee/actions/runs/26378352326/job/77642724758

The link is unstable to access so we skip.


mythi (approved):
I may have asked this before but do we want to keep the link checker on PR path?

Xynnn007:
> I may have asked this before but do we want to keep the link checker on PR path?

Good question. From my side it helps when we change/add the documents, and get some error links. But also it often fails to check the websites with things like rate limit. So currently we are trying to add the websites to skip. Do you have some better ideas?

PR #1373, opened 2026-05-25
verifieri/hygon-dcu: emit one claim per attestation report

### LINKED ISSUE #1372: Hygon DCU | Bring Different device claims to separate EAR submodule
Now a guest with multiple DCU devices, will generate only one submod. Let's bring them into different submodules. This will help to identify different DCU devices on one platform. Furthermore, if new version of DCU comes, we can identify different version DCUs on the same guest.

Xynnn007:
Return a separate (claims, dcu) pair for each DCU report instead of bundling all reports under numeric keys in a single hygondcu object. This aligns with other multi-device verifiers and yields per-device policy evaluation and EAR submods (dcu0, dcu1, ...). Breaking change: policy and token consumers must use flat paths such as input.hygondcu.body.chip_id instead of input.hygondcu["0"].body.chip_id. Update tcb_claims and attestation_token docs accordingly.

Close #1372

### DIFF
--- a/attestation-service/docs/attestation_token.md
+++ b/attestation-service/docs/attestation_token.md
@@ -11,6 +11,7 @@ Each submod is given a generic key constructed from the CPU class and the device
 For instance, the CPU submod will be called `cpu0`.
 If a GPU has been attested as part of the guest, there will be a `gpu0` submod.
 If there is more than one GPU, there will be additional `gpuN` submods.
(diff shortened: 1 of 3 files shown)


mythi (approved):
LGTM and aligns with others

fitzthum (approved):
LGTM

PR #1375, opened 2026-05-25
kbs: fix insecure_key for AttestationTokenVerifierConfig

Xynnn007:
Old name `insecure_key` is confusing. Actually, the option controls
whether the endorsement of the jwk field of the token will be checked.

The old comment and document for insecure_key item in
AttestationTokenVerifierConfig is not clear enough. This commit fixes
the documents and changes the config name.

### DIFF
--- a/kbs/src/token/mod.rs
+++ b/kbs/src/token/mod.rs
@@ -44,18 +44,22 @@ pub struct AttestationTokenVerifierConfig {
     #[serde(default)]
     pub trusted_jwk_sets: Vec<String>,
 
-    /// Whether the token signing key is (not) validated.
-    /// If true, the attestation token can be modified in flight.
-    /// This should only be set to true for testing.
-    /// While the token signature is still validated, the provenance of the
-    /// signing key is not checked and the key could be replaced.
+    /// Whether to skip endorsement checks for a JWK embedded in the JWT header.
     ///
-    /// When false, the key must be endorsed by the certificates or JWK sets
-    /// specified above.
+    /// When false (default) and the token header carries a `jwk`, the key must be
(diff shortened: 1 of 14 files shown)


kbs/src/token/mod.rs
```diff
@@ -44,18 +44,22 @@ pub struct AttestationTokenVerifierConfig {
...
+    /// When false (default) and the token header carries a `jwk`, the key must be
+    /// backed by an `x5c` certificate chain that chains to [`trusted_certs_paths`].
+    /// When true, that `jwk` is used directly to verify the signature without
+    /// checking its provenance; the signature itself is still validated.
+    ///
+    /// This only affects header-embedded `jwk` keys. Tokens that identify the
+    /// signing key via `kid` are still resolved from [`trusted_jwk_sets`].
+    ///
+    /// Should only be set to true for testing: an attacker who can replace the
+    /// header `jwk` can make KBS accept tokens signed with an arbitrary key.
```
mythi, line 58: Just an observation: I believe people mostly use configs from `kbs/config/` as is and all of those files set this to true without any warning. The comments and config guide on default being false is only partly true in that case.

Do we have docs somewhere that help users to generate the signing key and set the certs here?

mythi (approved):
ok

kbs/src/token/mod.rs
Xynnn007: nice catch. We do not have the docs now. Let me add and tidy the docs a little in latter PRs.

Xynnn007: https://github.com/confidential-containers/trustee/issues/1377

fitzthum (approved):
LGTM

PR #1380, opened 2026-05-26
ci: optimize image building and docker e2e test

Xynnn007:
Now we have same image building process in two places:
1. Build staged images matrix
2. Docker compose e2e test

This will waste time and computation resource. This PR will try two combine the two process into one:
1. Parallely build the images
2. Once all images required by docker-compose, the docker-compose e2e test starts.

Assisted-by: Cursor

### DIFF
--- a/.github/actions/build-single-image/action.yml
+++ b/.github/actions/build-single-image/action.yml
@@ -27,6 +27,10 @@ inputs:
     description: Whether to push the built image to GHCR
     required: false
     default: "false"
+  load:
+    description: Load the built image into the local Docker engine (for docker save / compose e2e)
+    required: false
+    default: "false"
 
 runs:
   using: composite
@@ -60,6 +64,8 @@ runs:
         build_option_args=()
         if [ "${{ inputs.push }}" = "true" ]; then
           build_option_args+=(--push)
+        elif [ "${{ inputs.load }}" = "true" ]; then
+          build_option_args+=(--load)
         fi
         docker buildx build --provenance false \
           "${platform_args[@]}" \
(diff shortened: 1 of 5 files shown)


.github/actions/build-single-image/action.yml
```diff
@@ -60,6 +64,8 @@ runs:
         build_option_args=()
         if [ "${{ inputs.push }}" = "true" ]; then
           build_option_args+=(--push)
+        elif [ "${{ inputs.load }}" = "true" ]; then
+          build_option_args+=(--load)
```
fitzthum, line 68: Is this necessary? I thought that images built with `-t` were automatically loaded.

.github/workflows/workflow-call-build-docker-e2e-images.yml
```diff
@@ -0,0 +1,153 @@
...
+      IMAGE_PREFIX: ghcr.io/confidential-containers/staged-images
+    steps:
+      - name: Download built images
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          pattern: docker-e2e-image-${{ matrix.platform_slug }}-*
+          path: ${{ runner.temp }}/docker-e2e-images
+          merge-multiple: true
+
+      - name: Load and repackage images
```
fitzthum: Is this just for reconciling the names of the images? I think there are some improvements that can be made there, but not part of this PR.

fitzthum (approved):
LGTM

.github/actions/build-single-image/action.yml
Xynnn007: Yes. For `docker build`, `-t` is enough; while for `docker buildx build`, explicit `--load` is needed.

.github/workflows/workflow-call-build-docker-e2e-images.yml
Xynnn007: It's to combine KBS-AS-GRPC, AS-GRPC and RVPS images into one tar package. Note that they are built separately before this job, and they need to serve the docker-compose test altogether in later steps. Thus we need to package one package for each arch with three images.

PR #1381, opened 2026-05-26
verifier: centralize DCAP quote claim generation on Quote

Xynnn007:
Move SGX/TDX quote field serialization into Quote::generate_parsed_claim so platform_instance_id and td_attributes parsing live in one place. Preserve policy-stable claim layouts (flat SGX header/body, nested TDX quote). Drop duplicate parse_claim macros and per-verifier PCK parsing.

cc @mythi

### DIFF
--- a/deps/verifier/src/az_tdx_vtpm/mod.rs
+++ b/deps/verifier/src/az_tdx_vtpm/mod.rs
@@ -11,9 +11,7 @@ use super::az_snp_vtpm::{
     TpmQuote,
 };
 use super::intel_dcap::quote::{parse_quote, Quote as TdQuote};
-use super::intel_dcap::{
-    ecdsa_quote_verification, extend_using_custom_claims, pck::parse_platform_info,
-};
+use super::intel_dcap::{ecdsa_quote_verification, extend_using_custom_claims};
 use super::tdx::claims::generate_parsed_claim;
 use super::{TeeClass, TeeEvidence, TeeEvidenceParsedClaim, Verifier};
 use crate::{InitDataHash, ReportData};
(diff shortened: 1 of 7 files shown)


deps/verifier/src/az_tdx_vtpm/mod.rs
```diff
@@ -68,10 +66,7 @@ impl Verifier for AzTdxVtpm {
         let pcr_refs: Vec<&[u8; 32]> = pcrs.iter().collect();
         verify_init_data(expected_init_data_hash, &pcr_refs)?;
 
-        let platform_info =
-            parse_platform_info(&td_quote.cert_data().qe_certification_data.certificates)?;
-
-        let mut claim = generate_parsed_claim(&td_quote, None, &platform_info)?;
+        let mut claim = generate_parsed_claim(&td_quote, None)?;
```
mythi, line 69: this is slightly problematic since I'm going to need platform_info for pulling verification collateral in #1368

mythi (commented):
thanks for getting this started.

deps/verifier/src/az_tdx_vtpm/mod.rs
Xynnn007: it's still available via `pck::parse_platform_info`.

mythi: true but no need to parse them twice, right?

Xynnn007: I just reviewed your code. There are some overlapping. So let me convert this to draft and I will refactor after that one gets merged/

mythi: my PR is not touching claims.rs files to there's rather minimal overlap if we keep platform_info as is but I agree it's easier to wait with this one

PR #1383, opened 2026-05-26
KBS Protocol | Add PQC algorithm support - Prototype/iss 1271

grant-arqit:
This is the initial draft implementation to enable PQC algorithm support across CoCo. These changes reflect those required on the Trustee side, and is enabled via a compile time feature flag.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ educe = "0.7"
 ear = "0.5.0"
 hex = "0.4.3"
 kbs_protocol = { git = "https://github.com/confidential-containers/guest-components.git", rev = "da8d93f2797088a5f0636c8c1eeb31da73784fe8", default-features = false }
-kbs-types = "0.15.0"
+kbs-types = "0.18.0"
 kms = { git = "https://github.com/confidential-containers/guest-components.git", rev = "da8d93f2797088a5f0636c8c1eeb31da73784fe8", default-features = false }
 jsonwebtoken = { version = "10", default-features = false, features = [
     "aws_lc_rs",
--- a/kbs/Cargo.toml
+++ b/kbs/Cargo.toml
@@ -43,8 +43,14 @@ nebula-ca-plugin = []
 # Use HashiCorp Vault KV v1 as KBS backend
 vault = ["vaultrs"]
 
+
 # Use AWS Secrets Manager as KBS resource backend
 aws = ["aws-config", "aws-sdk-secretsmanager"]
+# Experimental Post-Quantum Cryptography support for the resource-response
+# JWE path. Implements draft-ietf-jose-pqc-kem-05 (ML-KEM-768+A192KW). The
+# IETF draft is not finalized and the wire format may change; do not enable
+# in production.
+pqc-experimental = ["dep:ml-kem", "dep:sha3-kmac"]
 
 # Use GCP Secret Manager as KBS resource backend
 gcp = ["google-cloud-secretmanager-v1"]
@@ -111,6 +117,8 @@ uuid = { workspace = true, features = ["serde"] }
 openssl.workspace = true
 az-cvm-vtpm = { version = "0.8.0", default-features = false, optional = true }
 vaultrs = { version = "0.8.0", optional = true }
+ml-kem = { version = "0.3.2", features = ["getrandom"], optional = true }
+sha3-kmac = { version = "0.3.0", optional = true }
 aws-config = { version = "1", features = [
     "behavior-version-latest",
 ], optional = true }
--- a/kbs/Makefile
+++ b/kbs/Makefile
@@ -1,5 +1,6 @@
 AS_TYPE ?= coco-as
 ALIYUN ?= false
+PQC_EXPERIMENTAL ?= false
 NEBULA_CA_PLUGIN ?= false
 VAULT ?= true
 GCP ?= false
@@ -73,6 +74,10 @@ endif
 
 ifeq ($(EXTERNAL_PLUGIN), true)
   override FEATURES := $(FEATURES),external-plugin
+endif
+
+ifeq ($(PQC_EXPERIMENTAL), true)
+  override FEATURES := $(FEATURES),pqc-experimental
 endif
 
 ifeq ($(CLI_FEATURES),sample_only)
--- a/kbs/README.md
+++ b/kbs/README.md
@@ -141,6 +141,16 @@ The KBS can use different backend storage. `LocalFs` will always be builtin.
 are `true` or `false` (by defult). Please refer to [the document](docs/config.md#repository-configuration)
 for more details.
 
+## Experimental Features
+
+PQC algorithm support for the KBS protocol as a replacement for classic cryptographic methods is provided here as an experimental compile-time feature. These can be enabled by building with the additional [PQC_EXPERIMENTAL=?] parameter set as required, e.g.
+
+```shell
+PQC_EXPERIMENTAL=true make
+```
+
+**Please note, enabling this feature extends support for the designated PQC algorithm, but will still support classic algorithms. However enabling this feature in guest components will require the feature to be enabled here to support resource requests with the new TeePubKey::AKP type.**
+
 ## References
 
 ### Attestation Protocol
--- a/kbs/src/api_server.rs
+++ b/kbs/src/api_server.rs
@@ -457,6 +457,7 @@ pub(crate) async fn api(
                     .map_err(|e| Error::PluginInternalError { source: e })?
                 {
                     let public_key = core.token_verifier.extract_tee_public_key(claims)?;
+
                     let jwe =
                         jwe(public_key, response).map_err(|e| Error::JweError { source: e })?;
                     let res = serde_json::to_string(&jwe)?;
--- a/kbs/src/jwe/akp.rs
+++ b/kbs/src/jwe/akp.rs
@@ -1,0 +1,219 @@
+// Copyright (c) 2026 Trustee contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Experimental Post-Quantum Cryptography extension for the KBS
+//! resource-response JWE path.
+//!
+//! Implements `ML-KEM-768+A192KW` per draft-ietf-jose-pqc-kem-05.
+//!
+//! Wire format is not finalized — the IETF draft is in WG-adopted
+//! Standards Track but not yet RFC. The KDF FixedInfo encoding (X input
+//! to KMAC256) is underspecified by the draft; we follow the RFC 7518
+//! §4.6.2 ConcatKDF precedent, omitting PartyUInfo/PartyVInfo as the PQ
+//! draft directs. Verify against reference implementations and test
+//! vectors when those emerge.
+
+use aes_gcm::{
+    aead::{AeadInOut, Generate},
+    Aes256Gcm, KeyInit, Nonce,
+};
+use aes_kw::KwAes192;
+use anyhow::{anyhow, bail, Context, Result};
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
+use kbs_types::{ProtectedHeader, Response};
+use ml_kem::{Encapsulate, EncapsulationKey, Key, MlKem768};
+use serde_json::{Map, Value};
+use sha3_kmac::Kmac256;
+use zeroize::Zeroizing;
+
+/// `kty` value for the Algorithm Key Pair (AKP) key type per
+/// draft-ietf-jose-pqc-kem-05 §10.
+pub const AKP_KTY: &str = "AKP";
+
+/// Algorithm identifier for ML-KEM-768 with AES-192 key wrap.
+pub const ML_KEM_768_A192KW_ALGORITHM: &str = "ML-KEM-768+A192KW";
+
+/// AES-256-GCM content encryption, matches existing classical paths.
+const AES_GCM_256_ALGORITHM: &str = "A256GCM";
+
+/// ML-KEM-768 encapsulation-key length in bytes (FIPS 203).
+const ML_KEM_768_ENCAP_KEY_LEN: usize = 1184;
+
+/// KMAC256-based KDF per draft-ietf-jose-pqc-kem-05 §5.1.
+///
+/// `KMAC256(K = shared_secret, X = AlgorithmID || SuppPubInfo,
+///          L = out_len_bytes·8 bits, S = "")`
+///
+/// AlgorithmID = 4-byte BE length(alg) || alg.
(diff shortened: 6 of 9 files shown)


kbs/src/akp.rs
```diff
@@ -0,0 +1,237 @@
...
+    // 1. Decode and parse the recipient's encapsulation key.
+    let pub_key_bytes = URL_SAFE_NO_PAD
+        .decode(pub_key_b64)
+        .context("base64 decode AKP pub_key failed")?;
+    if pub_key_bytes.len() != ML_KEM_768_ENCAP_KEY_LEN {
+        return Err(anyhow!(
+            "ML-KEM-768 encapsulation key has wrong length: got {}, expected {}",
+            pub_key_bytes.len(),
+            ML_KEM_768_ENCAP_KEY_LEN,
+        ));
```
Xynnn007: we can use `bail!("..")` to simplify `return Err(anyhow!("..."))`

kbs/src/lib.rs
```diff
@@ -25,4 +25,7 @@ pub mod admin;
 pub mod http;
 pub mod jwe;
 
+#[cfg(feature = "pqc-experimental")]
+pub mod akp;
```
Xynnn007: Let's try to modularize the codes.

My idea is to turn `jwe` into a directory mod. Put all original `jwe.rs` things to `jwe/mod.rs`. Then create `jwe/akp.rs` under it. Then claim `mod akp;` as `#[cfg(feature = "pqc-experimental")]`. Do you have better ideas?

Xynnn007 (commented):
Looks cool. Some nits at first round review. Btw, it would be good to add the experimental information in [README.md](https://github.com/confidential-containers/trustee/blob/main/kbs/README.md)

kbs/src/lib.rs
grant-arqit: That sounds like a good idea @Xynnn007. Do you think its worth keeping the jwe.rs code out of mod.rs in a separate file still, keeping the mod.rs content minimal e.g.

<pre>
└─ jwe/
    ├─ mod.rs
    ├─ jwe.rs
    └─ akp.rs
</pre>

grant-arqit: Actually on reflection, I think you're right about putting existing jwe.rs content into jwe/mod.rs. Namespacing gets a bit janky otherwise and it keeps the current default implementation in the obvious place.

grant-arqit:
Latest commits integrate updated kbs-types definition for TeePubKey implemented in  [PR 87](https://github.com/confidential-containers/kbs-types/pull/87). The feature gating for processing PQC worklows is driven by changes in guest-components now. This change will encrypt responses based on any of the TeePubKey types defined in the updated kbs-types

kbs/src/jwe/akp.rs
```diff
@@ -0,0 +1,217 @@
...
+use rsa::rand_core::RngCore;
+use serde_json::json;
+use sha3_kmac::Kmac256;
+
+/// `kty` value for the Algorithm Key Pair (AKP) key type per
+/// draft-ietf-jose-pqc-kem-05 §10.
+pub const AKP_KTY: &str = "AKP";
+
+/// Algorithm identifier for ML-KEM-768 with AES-192 key wrap.
+pub const ML_KEM_768_A192KW_ALGORITHM: &str = "ML-KEM-768+A192KW";
```
Xynnn007, line 35: Just out of curiosity: why not use MLKEM1024+A256KW?

kbs/src/jwe/akp.rs
```diff
@@ -0,0 +1,217 @@
...
+    // 5. Protected header carries the KEM ciphertext in the `ek` parameter
+    // per draft-ietf-jose-pqc-kem §6.2.
+    let ek_b64 = URL_SAFE_NO_PAD.encode(kem_ciphertext.as_slice());
+    let protected = ProtectedHeader {
+        alg: ML_KEM_768_A192KW_ALGORITHM.to_string(),
+        enc: AES_GCM_256_ALGORITHM.to_string(),
+        other_fields: json!({ "ek": ek_b64 })
+            .as_object()
+            .unwrap()
+            .clone(),
```
Xynnn007: We usually try to avoid `unwrap`s in the code. If possible, can we make a Map and initialize in place?

kbs/src/api_server.rs
```diff
@@ -453,6 +454,16 @@ pub(crate) async fn api(
...
+
+                    if let TeePubKey::AKP { alg: _, public_key } = public_key {
+                        let jwe = crate::jwe::akp::ml_kem_768_a192kw(&public_key, response)
+                            .map_err(|e| Error::JweError { source: e })?;
+                        let res = serde_json::to_string(&jwe)?;
+                        return Ok(HttpResponse::Ok()
+                            .content_type("application/json")
+                            .body(res));
+                    }
+
```
Xynnn007: We could move this logic to https://github.com/confidential-containers/trustee/pull/1383/changes#diff-d4136bad5e1f864fe9ef260a347b0ea1c01a1c8ce4221bb380913f5e91a356e3R348

at that place, use feature flag under `TeePubKey::AKP` to either call akp handler, or raise error about "feature not enabled"

kbs/Cargo.toml
```diff
@@ -104,6 +104,8 @@ uuid = { workspace = true, features = ["serde"] }
 openssl.workspace = true
 az-cvm-vtpm = { version = "0.8.0", default-features = false, optional = true }
 vaultrs = { version = "0.8.0", optional = true }
+ml-kem = { version = "0.3.2", features = ["getrandom"] }
+sha3-kmac = { version = "0.3.0" }
```
Xynnn007: We can mark the two crates as `optional`, and move them under a feature like `pqc-experimental` or so. Then the Makefile could enable this feature by `PQC_EXPERIMENTAL` mentioned in the README.md

kbs/src/jwe/mod.rs
```diff
@@ -1,7 +1,9 @@
-// Copyright (c) 2024 by Alibaba.
+// Copyright (c) 2026 Trustee contributors.
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+pub mod akp;
```
Xynnn007, line 6: let's add a feature flag upon this module

Xynnn007 (commented):
Some more ideas;

About the algorithm implementation, only one question (not blocker) is about scheme candidate.

More ideas are about code organization

kbs/src/jwe/akp.rs
grant-arqit: There was no technical reason for this. It's the default for TLS Hybrid and while MLKEM1024+A256KW provides higher assurance ML-KEM-768 which is tied to AES-192 for the key wrap should be adequate for general use. Let me know if you think otherwise.

kbs/src/api_server.rs
grant-arqit: That's a good call. I'll update the implementation.

kbs/src/jwe/mod.rs
grant-arqit: Yes I'll re-apply that. I removed the feature gating in the last commit, relying on the feature flag from guest components. Was a bit uncertain about that, but as you've also picked up on this I'll re-apply the overall feature flag.

kbs/src/jwe/akp.rs
Xynnn007: no strong preference. I assumed that there is some hidden rules, but looks like no. Probably we need some comments upon the algorithm to help newbie like me to understand the reason?

kbs/src/jwe/mod.rs
```diff
@@ -343,11 +344,15 @@ pub fn jwe(tee_pub_key: TeePubKey, payload_data: Vec<u8>) -> Result<Response> {
             (P256_CURVE, ECDH_ES_A256KW) => ecdh_es_a256kw_p256(x, y, payload_data),
             (P521_CURVE, ECDH_ES_A256KW) => ecdh_es_a256kw_p521(x, y, payload_data),
             (crv, alg) => bail!("curve {crv} and algorithm {alg} is not supported"),
-        },
-        TeePubKey::AKP { alg, public_key } => match &alg[..] {
+        }
+        
+        #[allow(unused_variables)]
+        TeePubKey::AKP { alg, public_key } => match &alg[..] {
```
Xynnn007: Looks like we are using a macro here to avoid linter crying. Usually we are trying to avoid macros. Two ways from my side are

1. mark `TeePubKey::AKP` enum in `kbs-types` under a `"pqc-experimental"` feature flag. Then we can put feature upon the match arm here.
2. use `TeePubKey::AKP { alg, public_key: _public_key } => match &alg[..]` and then `akp::ML_KEM_768_A192KW_ALGORITHM => akp::ml_kem_768_a192kw(&_public_key, payload_data),`

wdyt?

Xynnn007 (approved):
Last comments from my side. Others LGTM!

kbs/src/jwe/mod.rs
grant-arqit: My preference would be option 2 here. The fundamental change to `kbs-types` is an extension of `TeePubKey` and so in itself is not experimental. Are there any consumers of this type outside of `guest-components` and `Trustee` which could have similar match operations which could be impacted by this extension? Does this matter or is it expected they would need to resolve this as part of moving to the new version of `kbs-types`.

I'll make the change along the lines of option 2 for now to get rid of the macro.

kbs/src/jwe/mod.rs
```diff
@@ -1,7 +1,10 @@
-// Copyright (c) 2024 by Alibaba.
```
fitzthum, line 1: Don't change the existing notice. You can add one below if you want to.

kbs/README.md
```diff
@@ -141,6 +141,14 @@ The KBS can use different backend storage. `LocalFs` will always be builtin.
 are `true` or `false` (by defult). Please refer to [the document](docs/config.md#repository-configuration)
 for more details.
 
+## Experimental Features
+
+PQC algorithm support for the KBS protocol as a replacement for classic cryptographic methods is provided here as an experimental compile-time feature. These can be enabled by building with the additional [PQC_EXPERIMENTAL=?] parameter set as required, e.g.
```
fitzthum, line 146: Can you clarify the compatibility of pqc trustee and guest-components. If this feature is enabled, you must use pqc guest components only?

fitzthum (commented):
Sorry I took so long to review this. Getting into it a bit now. A couple first comments. Generally looks pretty good.

kbs/README.md
grant-arqit: > Can you clarify the compatibility of pqc trustee and guest-components. If this feature is enabled, you must use pqc guest components only?

No, enabling the feature allows pairing with guest-components with the same feature enabled, but if not enabled in guest-components, requests with RSA or EC public keys will still be processed. 

However if the feature is enabled in guest-components, sending resource requests with an AKP public key and not enabled here in the trustee, the request will fail. Will amend this README.md to clarify.

kbs/src/jwe/mod.rs
grant-arqit: Have reverted copyright notice

kbs/README.md
grant-arqit: Have now added additional qualifiers for operating with PQC experimental feature enabled.

grant-arqit:
I've added an additional check in CI for the PQC experimental feature. The update to RustCrypto broke some of the experimental code. Any similar upgrades once this PR is merged should be protected with additional CI checks. Will look to add similar to guest-components which is similarly at risk to dependency updates when this feature is not enabled for a CI check.

kbs/Makefile
```diff
@@ -1,5 +1,7 @@
 AS_TYPE ?= coco-as
 ALIYUN ?= false
+false
```
Xynnn007: What's this false for?

Cargo.toml
```diff
@@ -35,8 +35,8 @@ const_format = "0.2.36"
 educe = "0.7"
 ear = "0.5.0"
 hex = "0.4.3"
+kbs-types = { git = "https://github.com/grant-arqit/kbs-types.git", rev = "2009cbbc2b2c7c069317057deac0a718ca783a6f" }
```
Xynnn007: I guess this should be replaced once @tylerfanelli helps to cut a release. but before that, use a rev of https://github.com/confidential-containers/kbs-types is ok also

kbs/Makefile
grant-arqit: Looks like that got added from an earlier commit. Will remove.

kbs/src/jwe/akp.rs
```diff
@@ -0,0 +1,219 @@
...
+        let kem_ct_bytes = URL_SAFE_NO_PAD.decode(ek_b64).expect("decode ek");
+        let shared_secret = decap_key
+            .decapsulate_slice(&kem_ct_bytes)
+            .expect("decapsulate");
+
+        let kwk =
+            kmac256_kdf(shared_secret.as_slice(), ML_KEM_768_A192KW_ALGORITHM, 24).expect("kdf");
+        let kwk: [u8; 24] = kwk.try_into().unwrap();
+        let unwrapper = KwAes192::new(&kwk.into());
+        let mut cek = vec![0u8; 32];
```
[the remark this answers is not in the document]
grant-arqit: Part of PQC unit test. Shared secret derived from random encap key value for each run of test. No risk.

Cargo.toml
grant-arqit: Thanks, will leave this in for final review. Will squash commits as final review commit.

kbs/src/jwe/akp.rs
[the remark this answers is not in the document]
grant-arqit: @Xynnn007 I don't have permissions to dismiss this alert. Please can review this and dismiss if you think the above reasoning is valid. Many thanks :-)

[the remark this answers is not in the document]
Xynnn007: ok. let me handle this

rust-toolchain.toml
```diff
@@ -1,2 +1,3 @@
 [toolchain]
 channel = "1.93.0"
+components = ["rust-analyzer"]
```
Xynnn007: What's this for?

grant-arqit: I added this as a recommendation from a vs-code. Helps optimise the plugin with the codebase. Seemed fairly benign, but can remove this if you want it out of the committed code base.

Xynnn007: Yes. let's remove it - or add it in a separate commit

grant-arqit: Removed.

Xynnn007 (approved):

fitzthum (approved):
LGTM 

Thanks @grant-arqit and thanks @Xynnn007 for all the review

PR #1385, opened 2026-05-27
rvps: remove in-toto extractor

Xynnn007:
What's more clean now, is that in-toto protects supplychain security. Usually some inner formats and payloads are protected by in-toto by the community.

The integration for in-toto of the project aims to use the link format, while currently in-toto predicate (attestation) is the target. Before we have a predicate format to use to carry artifact level metadata, it's hard for us to use in-toto to directly carry reference values. We can support it later when there is one.

Also, CoCo now has reference value publishment under development that uses sample format to be protected by slsa/in-toto attestation.


fitzthum (approved):
LGTM

maybe can use rust bindings if we bring this back

mythi (approved):

PR #1396, opened 2026-06-01
Fix the intermittent failures in integration tests

### LINKED ISSUE #969: Intermittent Failures in Integration Tests
I think there is a race condition in the integration tests that sometimes causes them to fail.

Xynnn007:
Close #969

### DIFF
--- a/integration-tests/tests/admin.rs
+++ b/integration-tests/tests/admin.rs
@@ -31,7 +31,7 @@ async fn set_policy(#[case] test_config: KbsConfigType, #[case] valid_key: bool)
     init_tracing();
 
     let mut harness = TestHarness::new(test_config.clone().into()).await?;
-    harness.wait().await;
+    harness.wait().await?;
 
     if !valid_key {
         info!("TEST: replacing admin private key");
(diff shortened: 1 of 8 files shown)


integration-tests/tests/admin.rs
```diff
@@ -31,7 +31,7 @@ async fn set_policy(#[case] test_config: KbsConfigType, #[case] valid_key: bool)
     init_tracing();
 
     let mut harness = TestHarness::new(test_config.clone().into()).await?;
-    harness.wait().await;
+    harness.wait().await?;
```
fitzthum, line 34: `harness.wait()` doesn't return a result tho, does it?

fitzthum (approved):
LGTM. one question

I'm not sure if `file_serial` is necessary, but can't hurt.

integration-tests/tests/admin.rs
Xynnn007: Oh this does. https://github.com/confidential-containers/trustee/pull/1396/changes#diff-04dd9ce25c35da5e7afed8c263d742473d7b5bc5f2dc39b1944d29105e31a714R458-R464

The commit order looks wrong, let me reorg

Xynnn007:
> I'm not sure if `file_serial` is necessary, but can't hurt.

Double checked. It's not necessary. Let me get rid of it.

PR #1398, opened 2026-06-01
Add service health endpoints and harden gRPC/HTTP client robustness

fidencio:
This series makes the Trustee services more observable and more resilient
when run as long-lived deployments behind an orchestrator.

It adds standard health endpoints so liveness/readiness probes can be wired
up, and fixes a handful of robustness issues in the inter-service gRPC and
the outbound HTTP clients that can otherwise cause requests to hang
indefinitely.

These changes are independent of any particular deployment method and are
sent ahead of the Helm chart work so they can be reviewed on their own.

1. **serviceability: add health and readiness endpoints to AS, RVPS and KBS**
   The attestation-service and rvps gRPC servers now register the standard
   gRPC Health service (via `tonic-health`) and report `SERVING` once ready,
   and KBS gains a lightweight `GET /healthz` HTTP route. This lets external
   orchestrators detect a wedged or still-initialising service and restart it.

2. **attestation-service: query RVPS from the runtime that owns the pool**
   The reference-value lookup ran on a helper thread that spun up its own
   current-thread runtime, while the mobc pool and its gRPC channels live on
   the main multi-threaded runtime. Driving `pool.get()` from an unrelated
   runtime delivered wakeups to the wrong reactor and could hang the request
   forever. The query now runs on the handle of the runtime that owns the pool.

3. **attestation-service: recycle stale RVPS gRPC connections**
   Long-lived pooled connections to RVPS could be silently torn down, making
   later queries hang or fail. The channel now uses HTTP/2 keep-alive and a
   capped idle lifetime, and each pooled connection is validated on checkout
   with a short, time-bounded health query before use.

4. **kbs: bound the lifetime of requests on the AS gRPC channel**
   The KBS→AS channel had no timeout. Evidence verification can be slow
   (notably Intel TDX DCAP), but with no upper bound a wedged connection would
   block a request indefinitely. A generous 300s timeout covers the slow path
   while ensuring stuck requests fail cleanly.

5. **verifier: bound NVIDIA NRAS HTTP requests with timeouts**
   The NVIDIA verifier's attestation POST and JWKS fetch used a default
   `reqwest` client with no timeout, so an unreachable NRAS could hang the
   verifier. Explicit timeouts are now set (60s attestation, 30s JWKS).

### DIFF
--- a/attestation-service/src/rvps/grpc.rs
+++ b/attestation-service/src/rvps/grpc.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use mobc::{Manager, Pool};
 use serde::Deserialize;
 use serde_json::Value;
@@ -30,6 +32,30 @@ fn default_address() -> String {
 /// The default pool size for the RVPS client.
 const DEFAULT_RVPS_POOL_SIZE: u64 = 10;
 
+/// Drop and re-establish a pooled connection once it has been idle for this
+/// long, so a connection that died silently while idle is recycled instead of
+/// being handed out to the next query.
+const RVPS_CONN_MAX_IDLE_LIFETIME: Duration = Duration::from_secs(30);
+
+/// Send an HTTP/2 keep-alive ping after this much inactivity on a pooled
+/// connection, so a broken connection is noticed promptly rather than on the
+/// next (possibly much later) query.
+const RVPS_KEEP_ALIVE_INTERVAL: Duration = Duration::from_secs(10);
+
+/// How long to wait for a keep-alive ping acknowledgement before treating the
+/// connection as dead.
+const RVPS_KEEP_ALIVE_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Time budget for the on-checkout connection health probe. It runs on every
+/// pool.get() and only needs a local round-trip to RVPS, so it is kept short.
+const RVPS_HEALTH_CHECK_TIMEOUT: Duration = Duration::from_secs(1);
+
+/// Placeholder reference-value key used to probe connection liveness. RVPS
+/// returns Ok(None) for an unknown key, which confirms the connection is live
+/// without triggering an "Is a directory" error that would otherwise force an
+/// unnecessary reconnect on every pool.get() call.
+const RVPS_HEALTH_CHECK_KEY: &str = "__healthcheck__";
+
 pub struct Agent {
     pool: Pool<GrpcManager>,
 }
@@ -50,6 +76,7 @@ impl Agent {
         let manager = GrpcManager { address };
         let pool = Pool::builder()
             .max_open(DEFAULT_RVPS_POOL_SIZE)
+            .max_idle_lifetime(Some(RVPS_CONN_MAX_IDLE_LIFETIME))
             .build(manager);
 
         // the mobc Pool builder does not establish an actual connection,
@@ -100,7 +127,14 @@ impl Manager for GrpcManager {
     type Error = anyhow::Error;
 
     async fn connect(&self) -> std::result::Result<Self::Connection, Self::Error> {
+        // Called for the initial connection and whenever the pool replaces a
+        // connection that failed check() below, so this is the single place
+        // that observes every (re)connection to RVPS.
+        debug!("establishing RVPS gRPC connection to {}", self.address);
         let channel = Channel::from_shared(self.address.clone())?
+            .keep_alive_while_idle(true)
+            .http2_keep_alive_interval(RVPS_KEEP_ALIVE_INTERVAL)
+            .keep_alive_timeout(RVPS_KEEP_ALIVE_TIMEOUT)
             .connect()
             .await?;
         Ok(ReferenceValueProviderServiceClient::new(channel))
@@ -110,6 +144,16 @@ impl Manager for GrpcManager {
         &self,
         conn: Self::Connection,
     ) -> std::result::Result<Self::Connection, Self::Error> {
-        Ok(conn)
+        let mut c = conn;
+        let req = tonic::Request::new(ReferenceValueQueryRequest {
+            reference_value_id: RVPS_HEALTH_CHECK_KEY.to_string(),
+        });
+        // We only report whether the connection is still usable; the pool is
+        // responsible for discarding it and calling connect() to replace it.
+        match tokio::time::timeout(RVPS_HEALTH_CHECK_TIMEOUT, c.query_reference_value(req)).await {
+            Ok(Ok(_)) => Ok(c),
+            Ok(Err(e)) => Err(anyhow::anyhow!("stale connection: {e}")),
+            Err(_) => Err(anyhow::anyhow!("RVPS health check timed out")),
+        }
     }
 }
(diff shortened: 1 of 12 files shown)


fitzthum:
cc @MatiasVara

kbs/src/attestation/coco/grpc.rs
```diff
@@ -239,6 +239,7 @@ impl Manager for GrpcManager {
 
     async fn connect(&self) -> Result<Self::Connection, Self::Error> {
         let connection = Channel::from_shared(self.as_addr.clone())?
+            .timeout(std::time::Duration::from_secs(300))
```
Xynnn007: Can we make this a const?

attestation-service/src/rvps/grpc.rs
```diff
@@ -110,6 +116,23 @@ impl Manager for GrpcManager {
         &self,
         conn: Self::Connection,
     ) -> std::result::Result<Self::Connection, Self::Error> {
-        Ok(conn)
+        let mut c = conn;
+        // Use a non-empty placeholder key: RVPS returns Ok(None) for unknown keys,
+        // confirming the connection is live without triggering an "Is a directory"
+        // error that would cause an unnecessary reconnect on every pool.get() call.
+        let req = tonic::Request::new(ReferenceValueQueryRequest {
+            reference_value_id: "__healthcheck__".to_string(),
```
Xynnn007: Can we make this ID a const?

attestation-service/src/rvps/grpc.rs
```diff
@@ -50,6 +52,7 @@ impl Agent {
         let manager = GrpcManager { address };
         let pool = Pool::builder()
             .max_open(DEFAULT_RVPS_POOL_SIZE)
+            .max_idle_lifetime(Some(Duration::from_secs(30)))
```
Xynnn007: Can we make this a const, with some inline comments?

attestation-service/src/rvps/grpc.rs
```diff
@@ -101,6 +104,9 @@ impl Manager for GrpcManager {
 
     async fn connect(&self) -> std::result::Result<Self::Connection, Self::Error> {
         let channel = Channel::from_shared(self.address.clone())?
+            .keep_alive_while_idle(true)
+            .http2_keep_alive_interval(Duration::from_secs(10))
```
Xynnn007: Can we make this a const, with some inline comments?

attestation-service/src/rvps/grpc.rs
```diff
@@ -101,6 +104,9 @@ impl Manager for GrpcManager {
...
+            .keep_alive_timeout(Duration::from_secs(5))
```
Xynnn007: Can we make this a const, with some inline comments?

deps/verifier/src/nvidia/mod.rs
```diff
@@ -268,8 +268,9 @@ impl Nvidia {
             "claims_version": "3.0"
         });
 
-        // We can reuse this client for multiple requests, but for now create a new one.
-        let client = reqwest::Client::new();
+        let client = reqwest::Client::builder()
+            .timeout(std::time::Duration::from_secs(60))
```
Xynnn007: Can we make this a const, with some inline comments?

deps/verifier/src/nvidia/nras_jwks.rs
```diff
@@ -18,7 +18,10 @@ pub struct NrasJwks {
 
 impl NrasJwks {
     pub async fn new() -> Result<Self> {
-        let res = reqwest::get(NRAS_JWKS_URL).await?;
+        let client = reqwest::Client::builder()
+            .timeout(std::time::Duration::from_secs(30))
```
Xynnn007: Do you think we would have a same timeout for both NRAS and NRAS jwk?

Xynnn007 (commented):
Thanks @fidencio ! Nice code and nice commit messages. I only have some nits for the timeout consts. Overall LGTM

deps/verifier/src/nvidia/nras_jwks.rs
fidencio: I think so.

kbs/src/attestation/coco/grpc.rs
fidencio: Addressed!

attestation-service/src/rvps/grpc.rs
fidencio: Addressed!

attestation-service/src/rvps/grpc.rs
fidencio: Addressed!

attestation-service/src/rvps/grpc.rs
fidencio: Addressed!

attestation-service/src/rvps/grpc.rs
fidencio: Addressed!

deps/verifier/src/nvidia/mod.rs
fidencio: Addressed!

Xynnn007 (approved):
LGTM. Thanks @fidencio

mythi:
> Evidence verification can be slow (notably Intel TDX DCAP)

Not related to this PR but I noticed some time ago that Trustee's default config does not cache the collateral at all. PCS stopped sending `Cache-Control` headers which enabled caching so to force caching is to add some `"verify_collateral_cache_expire_hours": <value, e.g., 168>,` or get my PR merged which unconditionally sets cache and drops QCNL.

fidencio:
> Not related to this PR but I noticed some time ago that Trustee's default config does not cache the collateral at all. PCS stopped sending `Cache-Control` headers which enabled caching so to force caching is to add some `"verify_collateral_cache_expire_hours": <value, e.g., 168>,` or get my PR merged which unconditionally sets cache and drops QCNL.

Is it okay if we land this one first, @mythi?

This is basically the set of commits that I needed in order to get the helm chart working / passing Kata Containers CI.
Once this lands, and we do the helm chart switch there, things will be less tricky to test end-to-end.

mythi:
I don't mind this goes first, the PRs are not coupled. Just a comment that if you plan to make some changes to `sgx_default_qcnl.conf` files, they become obsolete after my PR gets merged.

attestation-service/src/rvps/grpc.rs
```diff
@@ -110,6 +140,20 @@ impl Manager for GrpcManager {
...
     ) -> std::result::Result<Self::Connection, Self::Error> {
-        Ok(conn)
+        let mut c = conn;
+        let req = tonic::Request::new(ReferenceValueQueryRequest {
+            reference_value_id: RVPS_HEALTH_CHECK_KEY.to_string(),
+        });
+        match tokio::time::timeout(RVPS_HEALTH_CHECK_TIMEOUT, c.query_reference_value(req)).await {
+            Ok(Ok(_)) => Ok(c),
+            Ok(Err(e)) => {
+                warn!("RVPS connection health check failed, reconnecting: {e}");
```
mkulke: we don't attempt a reconnect here, so I would move the log entry to the place where we perform retries. see also the Err arm

fidencio: @mkulke, please, give a second pass.

mkulke (approved):

PR #1409, opened 2026-06-09
key-value-storage: fix return type for localfs and add log for localjson

Xynnn007:
Please see concrete commit messages.


fitzthum (approved):
LGTM

PR #1414, opened 2026-06-10
AS: remove simple token warning

Xynnn007:
Simple token has been removed since 0.16.0, and it's good to remove the warning for now.


fitzthum (approved):
LGTM

PR #1415, opened 2026-06-10
AS: add issuer to ear claims

Xynnn007:
The issuer_name in the attestation token configuration has not been used for EAR up to now. This patch adds an `iss` field in the claims of the ear. This follows RFC7519 about " JSON Web Token (JWT) " where `iss` is a registered claim.


fitzthum (approved):
LGTM

PR #1417, opened 2026-06-11
KV | Add global registry for storage instances

### LINKED ISSUE #1416: Key-Value-Storage | Add a global API for different crates/modules to leverage
We currently have a key-value storage library within the trustee system, which has made significant contributions to [decoupling compute and storage](https://github.com/confidential-containers/trustee/issues/1092). However, as the project progresses, more pull requests have introduced single-instance state for individual AS/KBS instances, disrupting our goal of complete compute-storage decoupling. We need to address this issue fundamentally. Therefore, this issue propose exporting key-value storage as a global variable, which can be directly accessed by various submodules. This convenient programming approach will assist in using the key-value storage module in a simple way and benefit from decoupled storage.

Xynnn007:
Close #1416 

After this PR, I will put PRs to move DCAP / SNP / CSV things to global kv storage.

### DIFF
--- a/attestation-service/src/lib.rs
+++ b/attestation-service/src/lib.rs
@@ -15,9 +15,11 @@ pub use serde_json::Value;
 
 use anyhow::{anyhow, bail, Context, Result};
 use config::Config;
+use key_value_storage::StorageProvider;
 use rvps::RvpsError;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use std::sync::Arc;
 use thiserror::Error;
 use tracing::{debug, info};
 use verifier::{InitDataHash, ReportData, TeeEvidenceParsedClaim};
@@ -121,27 +123,39 @@ pub struct AttestationService {
     config: Config,
     rvps: RvpsClient,
     token_broker: EarAttestationTokenBroker,
+    storage_provider: Arc<dyn StorageProvider>,
 }
 
 impl AttestationService {
     /// Create a new Attestation Service instance.
-    pub async fn new(config: Config) -> Result<Self, ServiceError> {
+    ///
+    /// `storage_provider` mints the namespaces this service (and its verifiers
+    /// and RVPS) need. The caller owns it, so the AS can share a provider with
+    /// a co-located KBS or get a dedicated one in the standalone binaries.
+    pub async fn new(
+        config: Config,
+        storage_provider: Arc<dyn StorageProvider>,
+    ) -> Result<Self, ServiceError> {
+        info!(
+            backend_type = ?config.storage_backend.storage_type,
+            "Attestation Service storage backend"
+        );
+
         let rvps = rvps::initialize_rvps_client(
             &config.rvps_config,
+            storage_provider.clone(),
             config.storage_backend.storage_type,
-            &config.storage_backend.backends,
         )
         .await
         .map_err(ServiceError::Rvps)?;
 
-        let policy_storage = config
-            .storage_backend
-            .backends
-            .to_client_with_namespace(
-                config.storage_backend.storage_type,
+        let policy_storage = storage_provider
+            .get_or_register_with_type(
                 AS_POLICY_STORAGE_NAMESPACE,
+                config.storage_backend.storage_type,
             )
-            .await?;
+            .await
+            .context("initialize policy storage")?;
         let token_broker =
             EarAttestationTokenBroker::new(config.attestation_token_broker.clone(), policy_storage)
                 .await?;
@@ -150,6 +164,7 @@ impl AttestationService {
             config,
             rvps,
             token_broker,
+            storage_provider,
         })
     }
 
@@ -198,6 +213,7 @@ impl AttestationService {
             let verifier = verifier::to_verifier(
                 &verification_request.tee,
                 self.config.clone().verifier_config,
+                self.storage_provider.clone(),
             )
             .await?;
 
@@ -280,7 +296,12 @@ impl AttestationService {
         tee: Tee,
         tee_parameters: String,
     ) -> Result<String> {
-        let verifier = verifier::to_verifier(&tee, self.config.clone().verifier_config).await?;
+        let verifier = verifier::to_verifier(
+            &tee,
+            self.config.clone().verifier_config,
+            self.storage_provider.clone(),
+        )
+        .await?;
         verifier
             .generate_supplemental_challenge(tee_parameters)
             .await
--- a/attestation-service/src/rvps/mod.rs
+++ b/attestation-service/src/rvps/mod.rs
@@ -5,7 +5,7 @@
 
 use std::sync::Arc;
 
-use key_value_storage::{KeyValueStorageStructConfig, KeyValueStorageType};
+use key_value_storage::{KeyValueStorageType, StorageProvider};
 pub use reference_value_provider_service::config::Config as RvpsCrateConfig;
 use reference_value_provider_service::extractors::ExtractorsConfig;
 use serde::Deserialize;
@@ -85,8 +85,8 @@ pub type RvpsClient = Arc<Mutex<dyn RvpsApi + Send + Sync>>;
 #[instrument(skip_all, name = "Initialize RVPS")]
 pub async fn initialize_rvps_client(
     config: &RvpsConfig,
-    unified_storage_type: KeyValueStorageType,
-    storage_config: &KeyValueStorageStructConfig,
+    storage_provider: Arc<dyn StorageProvider>,
+    default_storage_type: KeyValueStorageType,
 ) -> Result<RvpsClient> {
     match config {
         RvpsConfig::BuiltIn {
@@ -95,11 +95,11 @@ pub async fn initialize_rvps_client(
         } => {
             info!("launch a built-in RVPS.");
 
-            // Use RVPS-specific storage if provided, otherwise fall back to defaults
-            let actual_storage_type = storage_type.unwrap_or(unified_storage_type);
+            // Use RVPS-specific storage_type if provided, otherwise fall back to defaults
+            let storage_type = storage_type.unwrap_or(default_storage_type);
 
             Ok(Arc::new(Mutex::new(
-                builtin::BuiltinRvps::new(extractors.clone(), actual_storage_type, storage_config)
+                builtin::BuiltinRvps::new(extractors.clone(), storage_type, storage_provider)
                     .await?,
             )) as RvpsClient)
         }
--- a/deps/key-value-storage/src/provider.rs
+++ b/deps/key-value-storage/src/provider.rs
@@ -1,0 +1,160 @@
+// Copyright (c) 2026 by Alibaba.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! On-demand provider for namespaced key-value storages.
+//!
+//! Instead of threading a concrete storage handle (or the backend config)
+//! through every layer, a component is handed a [`StorageProvider`] and mints
+//! the namespaces it needs in-place, e.g.:
+//!
+//! ```ignore
+//! let policy = provider.get_or_register("attestation_service_policy").await?;
+//! ```
+//!
+//! The provider owns the backend configuration, so callers stay decoupled from
+//! the concrete backend (postgres, redis, local fs, ...). The dependency is
+//! explicit (it shows up in the constructor signature), there is no global
+//! initialization order to get right, and multiple providers with different
+//! configs can coexist in one process (useful for tests and the all-in-one
+//! binary).
+
+use crate::{
+    KeyValueStorageError, KeyValueStorageInstance, KeyValueStorageType, Result,
+    StorageBackendConfig,
+};
+use async_trait::async_trait;
+use std::{collections::HashMap, sync::Arc};
+use tokio::sync::RwLock;
+
+/// Mints namespaced key-value storage instances on demand.
+#[async_trait]
+pub trait StorageProvider: Send + Sync {
+    /// Return the storage for `namespace`, creating it if necessary, using the
+    /// provider's default storage type.
+    ///
+    /// Idempotent: repeated calls for the same namespace return the same
+    /// instance.
+    async fn get_or_register(&self, namespace: &str) -> Result<KeyValueStorageInstance>;
+
+    /// Like [`StorageProvider::get_or_register`] but pins the backend type for
+    /// this namespace, overriding the provider's default.
+    async fn get_or_register_with_type(
+        &self,
+        namespace: &str,
+        storage_type: KeyValueStorageType,
+    ) -> Result<KeyValueStorageInstance>;
+}
+
+/// The default [`StorageProvider`], backed by a single [`StorageBackendConfig`].
+pub struct KvStorageProvider {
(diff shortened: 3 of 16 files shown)


kbs/src/api_server.rs
```diff
@@ -80,16 +82,45 @@ impl ApiServer {
...
+        info!(
+            "Using storage backend type: {:?}",
+            config.storage_backend.storage_type
+        );
+
+        if config
+            .plugins
+            .contains(&PluginsConfig::ResourceStorage(RepositoryConfig::KvStorage))
+        {
+            let resource_plugin_storage = config
```
fitzthum: I don't totally love doing this at the top level of the call tree. It seems kind of redundant to parse the config here, especially as more sub-components start using kvstorage. Can we do this when we initialize the various pieces, like the resource plugin in this case.

fitzthum (commented):
Looks good. One question about initialization

kbs/src/api_server.rs
Xynnn007: Updated and squashed the commits.

deps/key-value-storage/src/global.rs
```diff
@@ -0,0 +1,97 @@
...
+    let config = CONFIG
+        .get()
+        .ok_or(KeyValueStorageError::StorageConfigurationNotSet)?;
+    let storage = config
+        .backends
+        .to_client_with_namespace(config.storage_type, namespace)
+        .await?;
+    storages.insert(namespace.to_string(), storage.clone());
+    Ok(storage)
+}
```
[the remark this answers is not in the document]
Xynnn007: Sounds reasonable. Adds some notes in the code

integration-tests/src/common.rs
```diff
@@ -202,17 +202,17 @@ impl TestHarness {
     }
 
     pub async fn new(test_parameters: TestParameters) -> Result<TestHarness> {
+        // Integration tests share one process; global storage config may only be set once, so we skip the error here as it will always fail
+        // from the second run in this process.
+        let _ = key_value_storage::set_global_config(shared_storage_backend()?);
+        // Clear registry entries and on-disk namespace data between harness instances.
+        key_value_storage::global::clear_all_namespaces().await?;
+
```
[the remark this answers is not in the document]
Xynnn007: Changed the comments.

deps/key-value-storage/src/global/test_utils.rs
```diff
@@ -0,0 +1,26 @@
...
+        .get()
+        .ok_or(KeyValueStorageError::StorageRegistryNotInitialized)?;
+
+    {
+        let mut storages = registry.storages.write().await;
+        for storage in storages.values_mut() {
+            storage.wipe().await?;
+        }
+        storages.clear();
+    }
```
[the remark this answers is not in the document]
Xynnn007: This is ok, because it's in test and always efficient and short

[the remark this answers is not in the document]
Xynnn007: As mentioned, it's test and always efficient. So we can ignore this

deps/key-value-storage/src/global/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/// Global registry of key-value storages.
+static REGISTRY: OnceLock<RegistryInner> = OnceLock::new();
+
+/// Set the global configuration of key-value storages.
+/// This function must be called before any other `register_*` key-value storage operations.
+/// If the global registry is already initialized, this function will return an error.
+pub fn init_global_registry(config: &StorageBackendConfig) -> Result<()> {
+    REGISTRY
```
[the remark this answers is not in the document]
Xynnn007: integration-tests will cover this.

deps/key-value-storage/src/global/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+/// Register a new namespace for a key-value storage with global configuration.
+pub async fn register_namespace(namespace: &str) -> Result<KeyValueStorageInstance> {
+    let registry = REGISTRY
+        .get()
+        .ok_or(KeyValueStorageError::StorageRegistryNotInitialized)?;
+
+    if registry.storages.read().await.contains_key(namespace) {
+        return Err(KeyValueStorageError::StorageNamespaceAlreadyRegistered {
+            namespace: namespace.to_string(),
+        });
```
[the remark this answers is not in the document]
Xynnn007: The duplicate-registration error is a valuable fail-fast check for namespace collisions, and it preserves a clear split between register (create) and get (retrieve).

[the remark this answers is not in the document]
Xynnn007: Mistake clicking thumb up

deps/key-value-storage/src/global/test_utils.rs
```diff
@@ -0,0 +1,26 @@
...
+    let registry = REGISTRY
+        .get()
+        .ok_or(KeyValueStorageError::StorageRegistryNotInitialized)?;
+
+    {
+        let mut storages = registry.storages.write().await;
+        for storage in storages.values_mut() {
+            storage.wipe().await?;
+        }
+        storages.clear();
```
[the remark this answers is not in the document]
Xynnn007: Let me use a different error enum for this.

kbs/src/attestation/backend.rs
```diff
@@ -173,13 +173,14 @@ impl AttestationService {
...
-            .map_err(|e| Error::SessionStorageInitialization { source: e })?;
-
-        let session_map = SessionMap::new(session_storage_backend.clone());
+        let session_storage_backend = key_value_storage::register_namespace_with_specified_type(
+            KBS_SESSION_STORAGE_NAMESPACE,
+            *session_storage_backend_type,
+        )
+        .await
+        .map_err(|e| Error::SessionStorageInitialization { source: e })?;
+
```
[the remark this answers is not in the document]
Xynnn007: Let me use a simpler way - to make `AttestationService::new` `pub(crate)`. Thus the function can only be called via `ApiServer::new()`. `ApiServer::new()` will ensure the global registry is always initialized.

Xynnn007:
Well. The battle with Copilot is over. Good ideas.

attestation-service/src/rvps/mod.rs
```diff
@@ -95,12 +94,11 @@ pub async fn initialize_rvps_client(
         } => {
             info!("launch a built-in RVPS.");
 
-            // Use RVPS-specific storage if provided, otherwise fall back to defaults
-            let actual_storage_type = storage_type.unwrap_or(unified_storage_type);
+            // Use RVPS-specific storage_type if provided, otherwise fall back to defaults
+            let actual_storage_type = storage_type.unwrap_or(storage_backend_config.storage_type);
```
fitzthum: nit: I might prefer re-delaring `storage_type` here rather than `actual_storage_type` but it's fine either way I guess

kbs/src/api_server.rs
```diff
@@ -81,21 +81,31 @@ impl ApiServer {
     }
 
     pub async fn new(config: KbsConfig) -> Result<Self> {
-        let plugin_manager = PluginManager::new(config.plugins.clone(), &config.storage_backend)
+        if key_value_storage::init_global_registry(&config.storage_backend).is_err() {
+            let current_config = key_value_storage::global_registry_config()
+                .map_err(|e| Error::StorageBackendInitialization { source: e })?;
+            if current_config != &config.storage_backend {
```
fitzthum: Isn't the KBS API service one of the first things to get initialized? What kind of config can make this happen?

fitzthum (approved):
LGTM. Two small comments but I think this is generally fine although the storage configs can maybe still be refined a bit.

attestation-service/src/lib.rs
```diff
@@ -126,22 +126,29 @@ pub struct AttestationService {
...
-
-        let policy_storage = config
-            .storage_backend
-            .backends
-            .to_client_with_namespace(
-                config.storage_backend.storage_type,
-                AS_POLICY_STORAGE_NAMESPACE,
-            )
-            .await?;
+        if key_value_storage::init_global_registry(&config.storage_backend).is_err() {
```
mkulke: do we expect the global registry to fail? it looks like it would be a programming error if we would call it twice. we can fail hard here and do not to recover I think (or use some typestate pattern to make sure it cannot be initialized twice, but that might be overkill).

attestation-service/src/rvps/mod.rs
Xynnn007: Good. Applied.

kbs/src/api_server.rs
Xynnn007: It's called in `integration-tests` multiple times. This type of defensive programming can also achieve a certain degree of idempotency, preventing errors from occurring when upper callers are not aware of the situation.

attestation-service/src/lib.rs
Xynnn007: The global registry cannot strictly enforce that the upper layer performs initialization only once. This is because, a built-in AS KBS—which has already been initialized on the KBS side—also calls the same AS entry function, resulting in a second initialization. Therefore, the only feasible approach is to verify whether the initialized configuration matches; an error is raised only if they are different (i.e., the upper layer attempts to initialize the global configuration twice with different settings).

mkulke: if we expect a condition to occur it's not an error. I understand we are raising a real error only after we encounter a faux error and some other condition (equality) is met. We can check "if !config.is_initialized() { config.initialize() }, no?

kbs/src/api_server.rs
mkulke: this sort of defensive/forgiving programming might obscur control flow, which can be problematic for security-sensitive software. IMO in this case clarity and predictability is eventually in the interest of the (API) user. If we encounter something that should be an invariant but it's somehow ambiguous, the best thing is to drop everything and bail out with an error.

if test logic really requires some special treatment a cleanup fn with #[cfg(test)] annotation might help

Xynnn007:
Ok. I met some hard issues about all-in-one model. The `api_server` and `attestation_service` should both face "what if a global registry is already created" and "is it different from mine" question. Let me re-consider the design. Maybe a non-global way is better.

Xynnn007:
Changed the way of implementation, from global variable, to a local variable.

Xynnn007:
@fitzthum @mkulke PTAL

mkulke (approved):
lgtm. nit: maybe it would have been a good idea use an enum instead of &str namespace to reduce the error surface, since the concrete consumers should be known at compile time

```rust
#[derive(EnumString)]
enum StorageConsumer{
    Verifier,
    DcapCache,
    ...
}
```

another alternative would be a trait StorageConsumer that has a `fn to_prefix(&self)`, that Consumer entitities need to implement, but maybe that's excessive typegolfing :)

Xynnn007:
Yes, for your suggestion 1, I once used this way - to define all possible things in a place. But where to define the enum would also be a question. Any place would seem coupled.

But anyway, I prefer the current way a little as it's somehow decoupled, though the caller would face the collision cases but I think we can avoid this by reviewing the code : )

For the two, not sure I understand correctly. Do you mean that every struct needs to use key-value-storage should claim a such thing, especially for `verifier`? I think it's ok, but still if the risk of collision could be tolerant, we do not need this, right?

mkulke:
> Yes, for your suggestion 1, I once used this way - to define all possible things in a place. But where to define the enum would also be a question. Any place would seem coupled.

I would put this enum in a central place, somewhere in the storage code. after all it's not a generic k/v implementation, but specific to this repo and we know exactly which consumers exist, no need to be overly generic.

> For the two, not sure I understand correctly. Do you mean that every struct needs to use key-value-storage should claim a such thing, especially for `verifier`? I think it's ok, but still if the risk of collision could be tolerant, we do not need this, right?

every entity that would consume kv storage would need to impl the StorageConsumer trait, but ofc there could be two consumers who both claim to be "verifier". IMO the central enum is the best option for clarity and avoiding typos.

Xynnn007:
Yes. I once used this "define in central place" way https://github.com/confidential-containers/trustee/pull/1417#discussion_r3404941446, but Tobin does not like it. 

I think this is not a big deal - both ways are ok  to me. If needed, we can bring it back again in future. Noticed that you have given an approval, it's not a blocker, right?

mkulke:
> Yes. I once used this "define in central place" way [#1417 (comment)](https://github.com/confidential-containers/trustee/pull/1417#discussion_r3404941446), but Tobin does not like it.
> 
> I think this is not a big deal - both ways are ok to me. If needed, we can bring it back again in future. Noticed that you have given an approval, it's not a blocker, right?

yes, not a blocker.

deps/key-value-storage/src/provider.rs
```diff
@@ -0,0 +1,234 @@
...
+/// sub-namespaces must be valid for the configured backend.
+struct ScopedStorageProvider {
+    inner: Arc<dyn StorageProvider>,
+    prefix: String,
+}
+
+impl ScopedStorageProvider {
+    fn scoped_namespace(&self, namespace: &str) -> Result<String> {
+        validate_namespace(namespace)?;
+        Ok(format!("{}/{}", self.prefix, namespace))
```
fitzthum: Is this slash problematic for any of the fs backends?

Btw, do we really need centralized logic for the scope? Can't components just use their own namespace? The scope thing seems a little bit orthogonal to the original purpose of this PR.

fitzthum (commented):
Looks pretty good. One comment on the scoping thing.

Some of the comments are superfluous imo but that's ok

deps/key-value-storage/src/provider.rs
Xynnn007: For local fs/json, it will create nested directory structure.

At the underlying level, it essentially adds a prefix to the namespace. I think it's good with `verifier/` prefix to have some logic isolation.

fitzthum: Can't each verifier just use its own namespace? Idk if we need to put so much effort into isolation here. These are our own components.

Xynnn007: well. it's just personal taste. if a blocker, let me remove it

fitzthum: Up to you I guess. Magnus mentioned some stuff about the namespace too. Maybe best to just keep it simple for now and iterate later. Otherwise PR looks good to me.

Xynnn007: Ok. Let me remove the things.

integration-tests/src/common.rs
```diff
@@ -221,6 +215,27 @@ impl TestHarness {
             .map_err(|e| anyhow!("Failed to join reference values path: {:?}", e))?;
         let auth_pubkey_path = work_dir.path().join("auth_pubkey");
 
+        // Each harness gets its own storage directory for isolation. Tests share
+        // one process but run serially, and a fresh storage backend per harness
+        // means no global registry to reset between runs.
```
fitzthum: Seems like this commit doesn't actually change anything? I think we create a new work_dir for each test anyway, so this should already be fine?

fitzthum (commented):
wait one more question. Then I think we are good

integration-tests/src/common.rs
Xynnn007: Yes. It's just a tidy commit. Let me remove the commit.

PR #1420, opened 2026-06-12
verifier: prevent panic in format_rfc3339

hyjiang:
This PR refactors format_rfc3339 in the Intel DCAP claims helper to gracefully support unexpected or invalid/out-of-range unix timestamps by returning fallback defaults and logging a warning instead of inducing a thread panic.

### DIFF
--- a/deps/verifier/src/intel_dcap/claims.rs
+++ b/deps/verifier/src/intel_dcap/claims.rs
@@ -1,3 +1,4 @@
+use anyhow::Context;
 use intel_tee_quote_verification_rs::{sgx_ql_qv_result_t, sgx_ql_qv_supplemental_t};
 use serde_json::{Map, Number, Value};
 use std::ffi::CStr;
@@ -38,24 +39,24 @@ pub(crate) fn prepare_custom_claims_map(
     supp_data: &mut sgx_ql_qv_supplemental_t,
     collateral_expiration_status: u32,
     quote_verification_result: sgx_ql_qv_result_t,
-) -> Map<String, Value> {
+) -> anyhow::Result<Map<String, Value>> {
     let mut claims_map = Map::new();
 
     claims_map.insert(
         "earliest_issue_date".to_string(),
-        Value::String(format_rfc3339(supp_data.earliest_issue_date)),
+        Value::String(format_rfc3339(supp_data.earliest_issue_date)?),
     );
     claims_map.insert(
         "latest_issue_date".to_string(),
-        Value::String(format_rfc3339(supp_data.latest_issue_date)),
+        Value::String(format_rfc3339(supp_data.latest_issue_date)?),
     );
     claims_map.insert(
         "earliest_expiration_date".to_string(),
-        Value::String(format_rfc3339(supp_data.earliest_expiration_date)),
+        Value::String(format_rfc3339(supp_data.earliest_expiration_date)?),
     );
     claims_map.insert(
         "tcb_date".to_string(),
-        Value::String(format_rfc3339(supp_data.tcb_level_date_tag)),
+        Value::String(format_rfc3339(supp_data.tcb_level_date_tag)?),
     );
     claims_map.insert(
         "pck_crl_num".to_string(),
@@ -111,7 +112,7 @@ pub(crate) fn prepare_custom_claims_map(
         Value::String(collateral_expiration_status.to_string()),
     );
     claims_map.insert("advisory_ids".to_string(), get_sa_list(&supp_data.sa_list));
-    claims_map
+    Ok(claims_map)
 }
 
 fn get_sa_list(sa_list: &[c_char; 450]) -> Value {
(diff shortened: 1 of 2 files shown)


deps/verifier/src/intel_dcap/claims.rs
```diff
@@ -130,10 +130,21 @@ fn get_sa_list(sa_list: &[c_char; 450]) -> Value {
...
+                    "failed to format OffsetDateTime to RFC3339: {e:?}, timestamp: {timestamp}"
+                );
+                "1970-01-01T00:00:00Z".to_string()
+            }
+        },
+        Err(e) => {
+            tracing::warn!("invalid/unsupported unix timestamp: {e:?}, timestamp: {timestamp}");
+            "1970-01-01T00:00:00Z".to_string()
+        }
+    }
```
Xynnn007: do you think it's worth an error handling, than using `1970-01-01T00:00:00Z`?

cc @mythi

hyjiang: Good point. Since this is a security-sensitive verification context, failing open with a placeholder timestamp like 1970-01-01T00:00:00Z might hide genuine collateral corruption or validation issues.

I will update the PR.

Xynnn007: btw, you might need to pay attestation to sign-off for commit - which is required from us

deps/verifier/src/intel_dcap/claims.rs
```diff
@@ -129,11 +129,14 @@ fn get_sa_list(sa_list: &[c_char; 450]) -> Value {
...
-        .format(&Rfc3339)
-        .expect("failed to format timestamp.")
+fn format_rfc3339(timestamp: i64) -> anyhow::Result<String> {
+    let odt = OffsetDateTime::from_unix_timestamp(timestamp).map_err(|e| {
+        anyhow::anyhow!("invalid or out-of-range unix timestamp: {e:?}, timestamp: {timestamp}")
+    })?;
+    let formatted = odt.format(&Rfc3339).map_err(|e| {
+        anyhow::anyhow!("failed to format datetime to RFC3339: {e:?}, timestamp: {timestamp}")
+    })?;
+    Ok(formatted)
```
mythi:
```suggestion
    OffsetDateTime::from_unix_timestamp(timestamp)
        .context("invalid or out-of-range unix timestamp")?
        .format(&Rfc3339)
        .context("failed to format datetime to RFC3339)
```

deps/verifier/src/intel_dcap/claims.rs
```diff
@@ -152,7 +155,8 @@ mod tests {
         let mut supp = deserialize_supp_data(&supplemental_data);
 
         let claims =
-            prepare_custom_claims_map(&mut supp, 0, sgx_ql_qv_result_t::SGX_QL_QV_RESULT_OK);
+            prepare_custom_claims_map(&mut supp, 0, sgx_ql_qv_result_t::SGX_QL_QV_RESULT_OK)
+                .unwrap();
```
mythi: I've used `.expect()` in tests.

deps/verifier/src/intel_dcap/claims.rs
```diff
@@ -38,24 +38,24 @@ pub(crate) fn prepare_custom_claims_map(
     supp_data: &mut sgx_ql_qv_supplemental_t,
     collateral_expiration_status: u32,
     quote_verification_result: sgx_ql_qv_result_t,
-) -> Map<String, Value> {
+) -> anyhow::Result<Map<String, Value>> {
```
mythi, line 42: Not sure what the best practice here would be since this now errors the whole verification. One alternative to consider is to just skip adding those four claims and print an error about it.

Xynnn007: I prefer to introduce error handling like the PR here.

hyjiang: The time in quote verification is a security-sensitive behavior. Logging warning messages is often unmonitored. Dynamic omissions broke the token schema contract and can silently bypass validation checks if the downstream policy doesn't explicitly reject missing fields.

mythi: I'm OK with error handling. These errors are unlikely to happen but helps to get rid of `.expect()`.

Xynnn007:
Hi @hyjiang can you help to merge the commits into one?

Xynnn007 (approved):

Xynnn007:
Thanks! @hyjiang Let's see the color of the CI

deps/verifier/src/intel_dcap/claims.rs
hyjiang: Updated.

deps/verifier/src/intel_dcap/claims.rs
hyjiang: Updated.

mythi (approved):
Thanks!

PR #1433, opened 2026-06-17
kbs: add attestation token verification documentation

### LINKED ISSUE #1377: Docs | Add docs about the attestation token verifications
Now the attestation tokens from the AS will be checked in KBS via a RCAR handshake. KBS provides ways to configure the endorsement/trust-anchor to verify the tokens. We need documents for users.

Xynnn007:
Users need guidance on how KBS verifies attestation tokens from the AS, how to configure trust anchors, and how AS signing material maps to KBS [attestation_token] settings. This addresses trustee#1377.

Add kbs/docs/attestation_token_verification.md covering cookie vs bearer verification paths, trust-chain setup, per-AS guidance, and deployment examples. Link to it from config.md, kbs_attestation_protocol.md, cluster.md, quickstart.md, README, and AS config docs.

Close #1377

### DIFF
--- a/kbs/docs/attestation_token_verification.md
+++ b/kbs/docs/attestation_token_verification.md
@@ -1,0 +1,229 @@
+# Attestation Token Verification
+
+After a successful RCAR attestation handshake, the Attestation Service (AS) issues
+an attestation result token (a JWT). KBS uses this token to authorize resource
+requests and to extract the TEE public key for JWE encryption of secret payloads.
+
+This document explains **when** KBS verifies tokens, **how** trust anchors are
+configured, and how AS signing material relates to KBS verification settings.
+
+For the RCAR protocol flow, see [KBS Attestation Protocol](./kbs_attestation_protocol.md).
+For the `[attestation_token]` configuration reference, see
+[config.md](./config.md#attestation-token-configuration).
+For token claim structure, see
+[Attestation Token](../../attestation-service/docs/attestation_token.md).
+
+## When KBS Verifies Tokens
+
+KBS can authenticate resource requests in two ways:
+
+| Method | Request header | Token verification |
+|--------|----------------|-------------------|
+| **Session cookie** | `Cookie: kbs-session-id=...` | KBS looks up the attestation session and reuses the token stored at `/kbs/v0/attest` (the JWT is still verified on each request). |
+| **Bearer token** | `Authorization: Bearer <JWT>` | Full JWT verification via `[attestation_token]` trust anchors. |
+
+The cookie path is used by the `kbs-client` tool, while CoCo typically uses the bearer-token path.
+
+The **Bearer** path is used when:
+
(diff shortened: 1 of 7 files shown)


kbs/docs/attestation_token_verification.md
```diff
@@ -0,0 +1,229 @@
...
+## When KBS Verifies Tokens
+
+KBS can authenticate resource requests in two ways:
+
+| Method | Request header | Token verification |
+|--------|----------------|-------------------|
+| **Session cookie** | `Cookie: kbs-session-id=...` | No JWT re-verification. KBS looks up the attestation session and reuses the token stored at `/kbs/v0/attest`. |
+| **Bearer token** | `Authorization: Bearer <JWT>` | Full JWT verification via `[attestation_token]` trust anchors. |
+
+Now **cookie** path is the default flow.
```
Xynnn007: Here we can see a problem. If HTTPS is not enabled, the cookie can be leaked.

mythi: what `default flow` means here? CDH uses bearer token, right?

Xynnn007: Let me fix the doc

mythi (approved):
LGTM, this is what I had in mind. 

I wonder if it would make sense to add a link to that docker compose setup as the reference if users want to follow the same steps manually.

Xynnn007:
> I wonder if it would make sense to add a link to that docker compose setup as the reference if users want to follow the same steps manually.

it's already here https://github.com/confidential-containers/trustee/pull/1433/changes#diff-757eb9c5e5572ac309734435a81b47d25014f273da3e47c262d4b6630cb43bc8R36

kbs/docs/attestation_token_verification.md
```diff
@@ -0,0 +1,229 @@
...
+the standalone AS config:
+
+```toml
+[attestation_service.attestation_token_broker.signer]
+key_path = "/path/to/token.key"
+cert_path = "/path/to/token-cert-chain.pem"
+```
+
+The `cert_path` PEM chain is embedded in the JWT header as `jwk.x5c`. The root CA
+certificate from that chain must be listed in KBS `trusted_certs_paths`.
```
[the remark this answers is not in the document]
Xynnn007: No need to chage.

mythi:
> > I wonder if it would make sense to add a link to that docker compose setup as the reference if users want to follow the same steps manually.
> 
> it's already here https://github.com/confidential-containers/trustee/pull/1433/changes#diff-757eb9c5e5572ac309734435a81b47d25014f273da3e47c262d4b6630cb43bc8R36

there's no link to `setup.sh` which is what I had in mind but not a big deal at all. It's fine as it is. link checker would notice if we change the file path but it does not notice the if the path given in the doc becomes outdated.

PR #1434, opened 2026-06-17
kbs: resolve trusted JWKS via direct URL or OpenID discovery, with opt-in HTTP

Xynnn007:
Improve how attestation-token verification keys (`trusted_jwk_sets`) are loaded:
- **Direct JWKS URLs**: try loading JWKS straight from the configured URL first,
  and only fall back to OpenID discovery (`.well-known/openid-configuration`)
  if that fails. Previously discovery was the only option.
- **Opt-in HTTP**: new `[attestation_token] insecure_public_key_from_uri` flag
  (default `false`) allows plaintext `http://` sources for controlled/dev
  environments. When off, HTTP is rejected — including any `jwks_uri` returned
  by discovery.
Secure defaults are unchanged (`https://`/`file://`/local paths always work).
Two separate commits; unit-tested without network.

### DIFF
--- a/kbs/docs/config.md
+++ b/kbs/docs/config.md
@@ -114,7 +114,7 @@ The following properties can be set under the `[attestation_token]` section.
 
 | Property              | Type         | Description                                                                                                                       | Default |
 |-----------------------|--------------|-----------------------------------------------------------------------------------------------------------------------------------|---------|
-| `trusted_jwk_sets`    | String Array | Trusted JWKS/OpenID sources (`file://` or `https://`) used to verify attestation tokens                                         | Empty   |
+| `trusted_jwk_sets`    | String Array | Trusted JWKS sources (`file://` or `https://`). Loads JWKS directly when possible, otherwise via OpenID discovery. The keys are used to verify attestation tokens.               | Empty   |
 | `trusted_certs_paths` | String Array | Trusted Certificates file (PEM format) for Attestation Tokens trustworthy verification                                            | Empty   |
 | `extra_teekey_paths`  | String Array | User defined paths to the tee public key in the JWT body                                                                          | Empty   |
 | `insecure_header_jwk`        | Boolean      | Skip `x5c`/`trusted_certs_paths` endorsement for a JWK in the JWT header; signature is still verified | `false` |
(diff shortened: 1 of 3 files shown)


fitzthum:
Does this make sense given that the AS doesn't support HTTPS at all?

Xynnn007:
@fitzthum Yes. This PR is just part of https://github.com/confidential-containers/trustee/pull/1360. and #1360 adds HTTP support.

fitzthum:
You mean #1428 exposes the keys via HTTP?

Either way, my concern is that we don't support HTTPS. It's one thing to allow people to use HTTP for development, but I don't think it's ideal to introduce this when we don't support HTTPS at all. 

Ofc we do trust HTTP communication with the AS today, but this feels a little diffrent.

I do think this is a good idea generally, though. The way we share the signing keys today is clunky. One question is whether we would need to implement a similar endpoint in the KBS to expose the JWKs to a third party. That seems kind of weird, but we can't really allow people outside the cluster to access the AS. 

Also, there's a question about microservices mode vs built-in mode here.

Xynnn007:
@fitzthum I think HTTPS vs openid discovery can be orthogonal, and we can use it in the interaction with KBS later.
Restful AS has already supported HTTPS. Let me add the HTTP for gRPC in a different PR.

For KBS's API. I think we should have a different abstraction than AS EAR Token, which can be in a signed. Or the KBS blackbox is the same as AS.

Different candidates can be
1. replace current JWE with mTLS+certs. Then KBS's API exposes the cert chain/verification keys for the certs.
2. Use TC proposed in https://github.com/confidential-containers/trustee/issues/1132, where we still use JWE thus no PFS is asserted. The KBS's API exposes the verification key of the TC.

fitzthum:
Ok, since we have some HTTPS support already and are adding more then I am fine with this. TBD exactly how we consume it. Let's keep the discussion of the KBS API separate for now.

kbs/docs/config.md
```diff
@@ -106,15 +106,11 @@ Attestation Token configuration controls attestation token verifications. This
...
-For an overview of when verification happens, how trust anchors relate to AS
-token signing, and deployment examples, see
-[Attestation Token Verification](./attestation_token_verification.md).
-
 The following properties can be set under the `[attestation_token]` section.
 
 | Property              | Type         | Description                                                                                                                       | Default |
 |-----------------------|--------------|-----------------------------------------------------------------------------------------------------------------------------------|---------|
-| `trusted_jwk_sets`    | String Array | Trusted JWKS/OpenID sources (`file://` or `https://`) used to verify attestation tokens                                         | Empty   |
+| `trusted_jwk_sets`    | String Array | Trusted JWKS sources (`file://` or `https://`). Loads JWKS directly when possible, otherwise via OpenID discovery.                | Empty   |
```
fitzthum: New comment drops the info about what this is used for. Maybe add back in the part about how this is for verifying attestation tokens.

fitzthum (approved):
LGTM. See nit. Also this raises the same question as the other PR about insecure configs. I guess we could drop the second commit if need be. Let's see what other reviewers think.

Xynnn007:
Dropped the second commit.

PR #1435, opened 2026-06-18
small DCAP updates

mythi:


mkulke (approved):

Xynnn007 (commented):
There is also the code definition. Do we need to delete it, or add a TODO comment upon the code? https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/intel_dcap/mod.rs#L40

Xynnn007 (approved):

Issue #1443, opened 2026-06-24
Helm Chart | Optimize key generation logic


Xynnn007:
Now we are using a job with scripts to generate ephermeral keys for login tokens. We might need to re-consider the logic. Maybe a compilable binary, or some 3rd tools to achieve this.

fitzthum:
For Trustee in CoCo we'll need to generate some of these keys on the host and inject them. It might be handy if we could share some of the logic between this case and the case where keys are generated in the guest. Maybe a rust crate would be helpful here. I think the semi-defunct trustee-cli has some logic for this.

Also, we should think about how the keys are actually shared. Obviously k8s secrets are out for Trustee in CoCo, but we could swap those for sealed secrets.

Xynnn007:
For Trustee in CoCo, a simple way to inject keys would be to bring in provision mechanism. That means when the Trustee is created at first, it belongs to no one. There should be a connection to it to inject something. After that, the blackbox does accept any other new provision connections.

fitzthum:
I'd like to have the bootstrapping involve attestation. The helm chart would set `AA_KBC_PARAMS` or generate init-data pointing to a local KBS that can fulfill sealed secrets.

Xynnn007:
Yes. My idea would also include the attestation. Our difference is either the first requester is trustee in CoCo, or the outer provisioner client.

fitzthum:
I feel like it would be nice to use existing coco features, like sealed-secrets to bootstrap Trustee. You're proposing adding another endpoint to Trustee itself? Can you describe a little more?

Xynnn007:
yup, sealed secret still need an extra trustee service to provision keys/secrets. It will require at least a public IP for the trustee to be deployed to connect to. Thus the chicken-egg problem.

The new API is just for Trustee. No need to be integrated into Trustee, but can be a proxy in front of trustee. When launched, it will serve to receive provision requests. A client will do provision request this API, gathers attestation report and the client will verify it. When verification succeeds, the client will provision the (proxy, trustee) pair with an admin token and something else. Then the trustee starts with provisioned materials.

The trustee can be owned by the provisioner. The risk can be that the trustee in coco could be provisioned by attackers than the real user. A trade-off is that no confidential data is released, but just DOS. Thus it's fine with CoCo's aiming.

fitzthum:
[...]

PR #1447, opened 2026-06-24
Helm: Add e2e test with sample attester

### LINKED ISSUE #1444: Helm Chart | Add e2e test
An e2e test should be included in trustee repo. Imo it should at least cover
1. deploy trustee with helm in a kubernetes cluster
2. use kbs-client to access KBS apis.

Checkbox
- [x] e2e test code with sample attester
- [x] e2e test code with tdx attester
- [ ] e2e test code with snp attester
- [ ] e2e test code with azure attester
- [x] e2e test code with se attester
- [ ] e2e test code with CSV attester
- [ ] e2e test code with Nvidia + any CPU TEE Composite attester
- [ ] e2e test code with CCA attester
- [ ] e2e test code with DCU + any CPU TEE Composite attester
- [x] nightly CI job for TEEs

Xynnn007:
The test uses github runner + kind to set up a k8s cluster. Then, use images built from the PR code to launch a trustee service with helm-chart.

Then use kbs-client to interact with the trustee.

Close #1444

Note that the image building process is reused from the docker e2e test. During implementing this, I get some more ideas about optimizing the current shape - which contains some duplicated tasks. I will take an eye on them in other PRs.

### DIFF
--- a/deployment/helm-chart/e2e/test.sh
+++ b/deployment/helm-chart/e2e/test.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env bash
+# kbs-client checks against an already-deployed Trustee Helm release.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+CHART_DIR="$(cd "${SCRIPT_DIR}/.." && pwd)"
+REPO_ROOT="$(cd "${CHART_DIR}/../.." && pwd)"
+
+KBS_CLIENT="${KBS_CLIENT:-${REPO_ROOT}/target/release/kbs-client}"
+KBS_URL="http://127.0.0.1:8080"
+TEST_RESOURCE_FILE="${SCRIPT_DIR}/fixtures/test-resource.txt"
+RESOURCE_PATH="helm-e2e/test-repo/test-secret"
+
+WORK_DIR="$(mktemp -d)"
+ADMIN_TOKEN_FILE="${WORK_DIR}/admin-token"
+ROUNDTRIP_FILE="${WORK_DIR}/roundtrip.txt"
+PORT_FORWARD_PID=""
+
+cleanup() {
+	local code=$?
+	if [[ -n "${PORT_FORWARD_PID}" ]] && kill -0 "${PORT_FORWARD_PID}" 2>/dev/null; then
+		kill "${PORT_FORWARD_PID}" 2>/dev/null || true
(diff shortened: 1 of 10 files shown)


Xynnn007:
Successful one https://github.com/confidential-containers/trustee/actions/runs/28088717167/job/83162720058?pr=1447

 As this re-uses the docker images and kbs-client built from prvious jobs, the whole time would be ~ 10 min ( ~7 min building + ~3 min e2e test)

deployment/helm-chart/e2e/test.sh
```diff
@@ -0,0 +1,111 @@
...
+
+log() { printf '==> %s\n' "$*"; }
+die() { printf 'error: %s\n' "$*" >&2; exit 1; }
+
+require_cmd() {
+	command -v "$1" >/dev/null 2>&1 || die "missing required command: $1"
+}
+
+wait_for_bootstrap_secret() {
+	local secret_name="trustee-e2e-bootstrap-user-keys"
```
fitzthum, line 39: This bootstrap keys thing may need to change in the future but fine for now I guess.

deployment/helm-chart/e2e/test.sh
```diff
@@ -0,0 +1,111 @@
...
+			return 0
+		fi
+		sleep 2
+	done
+	die "timed out waiting for Secret ${secret_name}"
+}
+
+start_port_forward() {
+	log "port-forward KBS -> 127.0.0.1:18080"
+	kubectl port-forward -n coco-trustee-e2e svc/trustee-e2e-kbs 18080:8080 >/dev/null 2>&1 &
```
fitzthum: Can we not use `127.0.0.1:8080`? That's more inline with what we do elsewhere and the defaults for the kbs-client.

fitzthum (commented):
Looks pretty good. A couple comments.

deployment/helm-chart/e2e/test.sh
Xynnn007: Good. So in which way? Maybe you can share the ideas under https://github.com/confidential-containers/trustee/issues/1443 and we can track the thread

deployment/helm-chart/e2e/test.sh
Xynnn007: Yes - I just used a different port  for local test. Let me change to 8080 and see if ci still works

fitzthum (approved):
LGTM 

i guess next step is hw support.

PR #1448, opened 2026-06-24
ci: add concurrency flag to cancel superseded run

Xynnn007:
With this flag, everytime we push to a same PR will cancel the previous round of job running. This would save a lot of runner resources.


fitzthum (approved):
LGTM

PR #1461, opened 2026-07-03
AS: add HTTPS support to gRPC AS

Xynnn007:
The gRPC Attestation Service could only serve over plaintext HTTP, while the RESTful AS already supported TLS. Add optional `--https-pubkey-cert` and `--https-prikey` arguments so the gRPC server can be started over TLS, mirroring the RESTful AS behavior: TLS is enabled only when both a certificate and a private key are provided, otherwise it stays on plaintext for backward compatibility.

TLS is built on tonic's `ServerTlsConfig` backed by rustls with the ring provider, which is installed explicitly since aws-lc-rs is also linked in via jsonwebtoken.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -639,6 +639,7 @@ dependencies = [
  "regorus",
  "rsa",
  "rstest",
+ "rustls 0.23.41",
  "serde",
  "serde_json",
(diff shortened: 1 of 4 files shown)


fitzthum (approved):
LGTM

PR #1466, opened 2026-07-07
AS: Install libcurl4 on s390x in grpc-as runtime image

BbolroC:
The grpc-as binary dynamically links libcurl.so.4 on s390x (pulled in transitively by the SE verifier), but the runtime stage of the container image did not include the package, causing a startup failure like:

```
grpc-as: error while loading shared libraries: libcurl.so.4:
cannot open shared object file: No such file or directory
```

This commit installs the package unconditionally.

### DIFF
--- a/attestation-service/docker/as-grpc/Dockerfile
+++ b/attestation-service/docker/as-grpc/Dockerfile
@@ -42,8 +42,8 @@ ARG VERIFIER=all-verifier
 
 LABEL org.opencontainers.image.source="https://github.com/confidential-containers/trustee/tree/main/attestation-service"
 
-# Install Openssl Suites
-RUN apt-get update && apt-get install openssl -y && \
+# Install Openssl Suites and libcurl (required by grpc-as on all architectures)
+RUN apt-get update && apt-get install -y openssl libcurl4 && \
     apt install --reinstall ca-certificates && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/{apt,dpkg,cache,log} /tmp/* /var/tmp/*


Xynnn007 (approved):

fidencio:
@BbolroC, was this the only missing library?

attestation-service/docker/as-grpc/Dockerfile
```diff
@@ -42,8 +42,8 @@ ARG VERIFIER=all-verifier
 
 LABEL org.opencontainers.image.source="https://github.com/confidential-containers/trustee/tree/main/attestation-service"
 
-# Install Openssl Suites
-RUN apt-get update && apt-get install openssl -y && \
+# Install Openssl Suites (and libcurl4 on s390x)
+RUN apt-get update && apt-get install -y openssl $([ "${ARCH}" = "s390x" ] && echo libcurl4) && \
```
fidencio: Add it unconditionally, please, as the only reason we don't see it failing on other arches is because $something (that we don't know what) ends up pulling it in as a dep.

BbolroC: Yep, I will update the PR.

BbolroC:
> @BbolroC, was this the only missing library?

Hmm, then I think I need to verify this locally. 😉

fidencio:
>> @BbolroC, was this the only missing library?
>
> Hmm, then I think I need to verify this locally. 😉

Please, do. As the loader will stop on the first library it cannot load, and I'm afraid we would end up with a whack-a-mole situation.

BbolroC:
Here is a deployment result on s390x after only adding `libcurl4`:
```
Every 2.0s: kubectl get po -A                                                                                                                                                                                                          as-grpc-build-01: Tue Jul  7 11:01:07 2026

NAMESPACE      NAME                                       READY   STATUS    RESTARTS      AGE
coco-trustee   trustee-as-5c779cff8-4jg28                 1/1     Running   0             2m43s
coco-trustee   trustee-kbs-5df976959f-vqnn2               1/1     Running   0             2m43s
coco-trustee   trustee-rvps-59496b745f-7z858              1/1     Running   0             2m43s
```

fidencio (approved):
lgtm, thanks @BbolroC!

PR #1467, opened 2026-07-08
helm: support IBM SE material mounting to use PV/PVC

Xynnn007:
The gRPC AS Pod runs the verifiers, so IBM Secure Execution materials (RSA measurement key pair, signing certs, CRLs, host key documents and the SE image header) must be mounted on the AS Pod rather than KBS. The AS deployment previously only exposed `as.extraEnvVars`, leaving no way to attach these volumes.

Add `as.extraVolumes` / `as.extraVolumeMounts` (mirroring the existing KBS hooks), wire them into the AS deployment template, document the s390x SE flow in the chart README, and provide a ready-to-use `scenarios/ibm-se.yaml` override that mounts the materials at their default `/run/confidential-containers/ibmse/` paths.

cc @BbolroC

### DIFF
--- a/deployment/helm-chart/README.md
+++ b/deployment/helm-chart/README.md
@@ -109,6 +109,38 @@ helm upgrade --install trustee ./deployment/helm-chart \
 
 Or use **`scenarios/bring-your-own-keys.yaml`** (adjust Secret name / file paths in the comments there).
 
+### IBM Secure Execution (s390x)
+
+On **s390x**, the **IBM Secure Execution (SE)** verifier needs attestation materials at runtime. Because KBS talks to a **remote `coco_as_grpc` AS**, the verifier runs inside the **AS Pod**, so these materials must be mounted on **AS**, not KBS. (This differs from the builtin-AS kustomize overlay in `kbs/config/kubernetes/overlays/ibm-se`, which mounts them on KBS.)
+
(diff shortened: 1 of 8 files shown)


BbolroC:
Thanks for the PR. The 1st question came to my mind is that:
- Is there any reason to drop the persistent volume-based configuration (https://github.com/confidential-containers/trustee/blob/main/kbs/config/kubernetes/overlays/ibm-se/pv.yaml) for AS? Mounting each sub-directories under `IBM_SE_CREDS_DIR` as a separate k8s secret leads to change the kata containers test code (e.g, https://github.com/kata-containers/kata-containers/blob/7e987cd6c8f0216f879ae9c867204d51725e176e/tests/integration/kubernetes/confidential_kbs.sh#L652-L676).

BbolroC:
As discussed, I will take over this PR from now on. Thanks @Xynnn007 for your permission!

BbolroC:
The current status is tested by https://github.com/kata-containers/kata-containers/pull/12982. Please check out a comment https://github.com/kata-containers/kata-containers/pull/12982#issuecomment-4919016936, too. Thanks!

Xynnn007:
@BbolroC Thanks!

Xynnn007:
tdx helm e2e passed also.

BbolroC (approved):

PR #1470, opened 2026-07-09
ci: only run TDX leg when the tdx-e2e label is added

Xynnn007:
The KBS e2e suite listens for `labeled` pull_request events so the TDX Helm leg can be triggered on demand. Previously any label change re-ran the entire suite, and the workflow-level `cancel-in-progress` meant adding the label also cancelled an in-progress full run for the same PR.

Guard every non-TDX job with `if: github.event.action != 'labeled'` so a `labeled` event only fans out to the TDX leg (and its image-build dependency), and make `cancel-in-progress` skip cancellation on `labeled` events so the label run is purely additive.

This is firstly found in https://github.com/confidential-containers/trustee/pull/1467


fitzthum (approved):
Ok

PR #1474, opened 2026-07-09
chore: update rustls-webpki to resolve GHSA-82j2-j2ch-gfr8

Xynnn007:
aws-sdk-secretsmanager enables a deprecated hyper-0.14/rustls-0.21 connector by default via its `rustls` feature, which pulled in an outdated rustls-webpki 0.101.7 alongside the modern 0.103.13 already used everywhere else. The legacy connector is redundant: the modern `default-https-client` feature provides an aws-lc-rs/rustls-0.23 based HTTPS client on its own.

Disable default features and enable only `default-https-client` and `rt-tokio` (plus `behavior-version-latest`). This removes the duplicate rustls-webpki 0.101.7 (and the transitive rustls 0.21.12, hyper-rustls 0.24.2, tokio-rustls 0.24.1, sct) so the whole workspace resolves to a single rustls-webpki 0.103.13.

Old version rustls-webpki brings GHSA-82j2-j2ch-gfr8

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -640,7 +640,7 @@ dependencies = [
  "regorus",
  "rsa",
  "rstest",
- "rustls 0.23.41",
+ "rustls",
  "serde",
  "serde_json",
  "serde_json_canonicalizer",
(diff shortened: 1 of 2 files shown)


fitzthum (approved):
LGTM. Nice catch

PR #1476, opened 2026-07-10
RVPS: Fix Intermittent Failures in Integration Tests

### LINKED ISSUE #969: Intermittent Failures in Integration Tests

Xynnn007:
Close #969

### DIFF
--- a/attestation-service/src/ear_token/broker.rs
+++ b/attestation-service/src/ear_token/broker.rs
@@ -222,9 +222,9 @@ impl EarAttestationTokenBroker {
                 .policy_engine
                 .evaluate_rego(
                     None,
-                    &tcb_claims_json,
+                    tcb_claims_json,
                     &policy_id,
-                    vec![TRUST_CLAIMS_RULE, EXTENSIONS_RULE],
+                    vec![TRUST_CLAIMS_RULE.to_string(), EXTENSIONS_RULE.to_string()],
                     extensions,
                 )
                 .await?;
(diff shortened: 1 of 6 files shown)


mkulke (approved):
good catch, we probably should investigate the repo systematically for long-running sync blocks that are executed within an async context. I stumbled, over those a couple of times, and they are not trivial to debug/reproduce

Xynnn007:
tbo I am not sure what's the best way to nvestigate the repo systematically. Maybe using AI? To avoid false positives, I currently still prefer a "lazy" approach—relying on user issues and CI—to discover problems.

mkulke:
> tbo I am not sure what's the best way to nvestigate the repo systematically. Maybe using AI? To avoid false positives, I currently still prefer a "lazy" approach—relying on user issues and CI—to discover problems.

Yup, we can assign copilot with such a task. For async/sync interop rusts usually strong compiletime guarantees do not help, its quite tedious to find those.

fitzthum (approved):
I'm not sure we need `#[serial(integration_ports)]` but I guess it doenst hurt

Xynnn007:
> I'm not sure we need `#[serial(integration_ports)]` but I guess it doenst hurt

This is used to only run one test for all test cases with same `integration_ports` label

Xynnn007:
@mkulke Adds the `&str` -> `String` patch

PR #1477, opened 2026-07-10
ci: remove Azure vTPM e2e legs

Xynnn007:
The Azure TDX/SNP vTPM e2e jobs relied on self-hosted `azure-cvm` runners that are no longer available, so drop the two manual-dispatch jobs and the Azure-specific branching in the reusable e2e workflow.

Now the `kbs/test` is still used by other tests, thus not deleted.

cc @mkulke


mkulke (approved):

fitzthum (approved):
LGTM

PR #1481, opened 2026-07-13
ci: Add zizmor GitHub Actions security analysis workflow

supritai:
## Why

GitHub Actions workflows are a common attack surface — misconfigurations such as template injection, over-broad permissions, and unpinned action refs can lead to secret exfiltration or supply-chain compromise. [zizmor](https://github.com/zizmorcore/zizmor) is a static analysis tool purpose-built to audit workflow files for exactly these classes of issues.

## What

Adds `.github/workflows/zizmor.yaml` which runs zizmor in `regular` persona mode on every push to `main`, every pull request, and on a weekly schedule (Mondays 03:00 UTC).

Key properties of the new workflow:
- `permissions: {}` — no token permissions granted to the job itself
- `concurrency` group — cancels superseded runs on the same PR/ref
- All actions pinned to full commit SHAs (same convention already used across this repo)
- `advanced-security: false` — no GHSA/CodeQL upload required; inline PR annotations are used instead
- `persona: regular` — standard ruleset

Also fixes pre-existing security findings surfaced by zizmor:
- `.github/actions/build-single-image/action.yml` — 10× `template-injection` (high): moved all `inputs.*` template expansions into `env:` vars
- `.github/dependabot.yml` — 3× `dependabot-cooldown` (medium): added `cooldown: default-days: 7` to all three ecosystems

## Local validation

zizmor was run locally against the full repo before this PR:

```console
$ zizmor --persona regular .
No findings to report. Good job!

actionlint also reported no issues:

```
actionlint .github/workflows/zizmor.yaml
```
(clean — no output)

---

> Assisted by [IBM Bob](https://github.com/IBM/bob)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: "cargo"
     directory: "/"
@@ -12,6 +14,8 @@ updates:
     open-pull-requests-limit: 3
     allow:
       - dependency-type: direct
+    cooldown:
+      default-days: 7
 
   - package-ecosystem: "github-actions"
     directory: "/"
@@ -21,3 +25,5 @@ updates:
       github-actions:
         patterns:
           - "*"
+    cooldown:
+      default-days: 7
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -1,0 +1,32 @@
+name: GitHub Actions Security Analysis with 🌈zizmor
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+  schedule:
+  - cron: "0 3 * * 1" # Every Monday Weekly
+
+permissions: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  zizmor:
+    name: Run zizmor
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Checkout repository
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by Xynnn007 — APPROVED** [review]
LGTM. Thanks! @supritai

**Review by mkulke — COMMENTED** [review]
thanks. can you address the zizmor findinds in the same PR, they seem rather simple to fix. otherwise we'll have broken CI and have to fix it in a separate PR.

**Comment by supritai** [comment]
> thanks. can you address the zizmor findinds in the same PR, they seem rather simple to fix. otherwise we'll have broken CI and have to fix it in a separate PR.

Sure, that makes sense. Thanks for the suggestion

File: .github/dependabot.yml [hunk 1]
```diff
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
+    cooldown:
+      default-days: 7
```
fitzthum, line 8: This is more of a policy decision than a fix, but I guess it's fine to do it in this PR. We have talked about enabling this in the past.

.github/workflows/zizmor.yaml
```diff
@@ -0,0 +1,32 @@
+name: GitHub Actions Security Analysis with 🌈zizmor
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+  schedule:
+  - cron: "0 3 * * 1" # Every Monday Weekly
```
fitzthum, line 8: Makes sense to have this as a weekly job, but we don't have much practice looking at weekly/nightly stuff.  We'll need to remember to check.

fitzthum (approved):
LGTM. A couple comments but they are non-blocking

.github/dependabot.yml
supritai: sure

mkulke (approved):

PR #1487, opened 2026-07-14
verifier: support az-{snp,tdx}-vtpm evidence v2

mkulke:
This version of the evidence envolope doesn't change the structure of the evidence, but v2 guarantees that the kbs report_data has been provided as "user-data" when retrieving a fresh attestation report.

For v2 evidences we can then assert:

hcl_report.var_data.user-data == kbs_report_data

hcl_report.var_data is bound to the attestation_report by using the var_data digest as "report_data" in the SNP Attestation Report or TD Report. This is already asserted.

With this change we can guarantee the freshness of a report, in previous versions of the evidence freshness was only guaranteed for the TPM quote (which is bound to attestation report, by having the TPM's {AK,EK}pub mixed into the report_data of the Attestation report.

There is a correlating PR on guest-components to produce v2 evidence

There is no automated test environment for the az-*-vtpm TEEs, I did a manual test on TDX and SNP instances

image

image


deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -83,16 +84,54 @@ fn extract_nonce(message: &[u8]) -> Result<Vec<u8>> {
...
+struct VarDataUserData {
+    #[serde(rename = "user-data")]
+    user_data: String,
+}
+
+fn get_user_data(var_data: &[u8]) -> Result<Vec<u8>> {
+    let var_data_str =
+        std::str::from_utf8(var_data).context("Failed to parse var data as UTF-8")?;
+    let var_data_json: VarDataUserData =
+        serde_json::from_str(var_data_str).context("Failed to parse var data as JSON")?;
```
Xynnn007: could the two be combined as `serde_json::from_vec`?

deps/verifier/src/az_tdx_vtpm/mod.rs
```diff
@@ -44,7 +44,8 @@ impl Verifier for AzTdxVtpm {
     /// 3. TPM PCRs' digest matches the digest in the Quote
     /// 4. TD Quote is genuine
     /// 5. TD Report's report_data field matches hashed HCL variable data
-    /// 6. Init data hash matches TPM PCR[INITDATA_PCR]
+    /// 6. User-data field in HCL variable data matches KBS report_data
```
Xynnn007: I saw many places uses `KBS` report_data term (codes and docs). Actually, for verifier (underlying crate), it does not and do not need need to know who's upper layer caller. Not only kbs but also outer codes code refer it. Thus a generic "report_data" term can be ok and we should not mention kbs here.

deps/verifier/src/az_tdx_vtpm/mod.rs
```diff
@@ -88,6 +89,10 @@ impl Verifier for AzTdxVtpm {
 
         verify_hcl_var_data(&hcl_report, &td_quote)?;
 
+        if evidence.version() >= 2 {
```
Xynnn007, line 92: No blockers here. Now the number is up to 2, so this hints more versions will come.

deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -83,16 +84,54 @@ fn extract_nonce(message: &[u8]) -> Result<Vec<u8>> {
...
+    let user_data = hex::decode(var_data_json.user_data)
+        .context("Failed to decode user_data from hex string")?;
+    Ok(user_data)
+}
+
+fn verify_report_data(hcl_report: &HclReport) -> Result<()> {
+    let var_data_hash = hcl_report.var_data_sha256();
+    let snp_report: AttestationReport = hcl_report.try_into()?;
+    if var_data_hash != snp_report.report_data[..32] {
```
Xynnn007: let's make `32` here as a const with notes telling why/where it comes.

Xynnn007 (commented):
Thanks @mkulke only some nits

deps/verifier/src/az_tdx_vtpm/mod.rs
mkulke: yes, if someone wants to implement CCEL then you would need an event_log, etc.

deps/verifier/src/az_snp_vtpm/mod.rs
mkulke: I guess you mean from_slice()?

Xynnn007: oh yes!

deps/verifier/src/az_tdx_vtpm/mod.rs
Xynnn007: There are still places, e.g. 

https://github.com/confidential-containers/trustee/pull/1487/changes#diff-4c820fc8dc8a1b9b5ce7bfa1154c49e22cfbb33a9aea5fe36d66e69bc1b39339R59
https://github.com/confidential-containers/trustee/pull/1487/changes#diff-7991a107111ef4abc07d28b364632a0b909fb0ba8d53f1c81b1d0821e18744e2R112
https://github.com/confidential-containers/trustee/pull/1487/changes#diff-7991a107111ef4abc07d28b364632a0b909fb0ba8d53f1c81b1d0821e18744e2R115
https://github.com/confidential-containers/trustee/pull/1487/changes#diff-7991a107111ef4abc07d28b364632a0b909fb0ba8d53f1c81b1d0821e18744e2R116
https://github.com/confidential-containers/trustee/pull/1487/changes#diff-1d9e829783b1e3200d134b8319162f6fdb4e26a157fe19edcc8deffafa12dbe1R39
https://github.com/confidential-containers/trustee/pull/1487/changes#diff-48e9f1b3fb32acb417bb61347bf14518b12ee6af91d756a0212eba7626e6b900R47

mkulke: should be gone

```
$ git grep -i "kbs report"
$
```

Xynnn007: https://github.com/confidential-containers/trustee/pull/1487/changes#diff-7991a107111ef4abc07d28b364632a0b909fb0ba8d53f1c81b1d0821e18744e2R112-R117

there are some variable/func names left : P

Xynnn007 (commented):
I guess the comment is not delivered correctly:

https://github.com/confidential-containers/trustee/pull/1487/changes#diff-7991a107111ef4abc07d28b364632a0b909fb0ba8d53f1c81b1d0821e18744e2R112-R117

there are some variable/func names left

mkulke:
> I guess the comment is not delivered correctly:
> 
> https://github.com/confidential-containers/trustee/pull/1487/changes#diff-7991a107111ef4abc07d28b364632a0b909fb0ba8d53f1c81b1d0821e18744e2R112-R117
> 
> there are some variable/func names left

the link doesn't work for me indeed, probably best to comment on the code lines directly. btw the rationale for calling this "kbs report data" was because this is the challenge as defined in the "kbs_protocol" that we use in attestation service. in any case it's probably wrong because it's called "kbs runtime-data"? https://github.com/confidential-containers/trustee/blob/86a0aa864ee418c2cf1df4c51e0a40da0bc97f8a/kbs/docs/kbs_attestation_protocol.md?plain=1#L196

deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -83,16 +85,52 @@ fn extract_nonce(message: &[u8]) -> Result<Vec<u8>> {
...
+
+pub(crate) fn verify_user_data(hcl_report: &HclReport, kbs_report_data: &[u8]) -> Result<()> {
+    let var_data = hcl_report.var_data();
+    let user_data = get_user_data(var_data)?;
+    let report_data_len = kbs_report_data.len();
+    if &user_data[0..report_data_len] != kbs_report_data {
+        bail!("User-data in HCL variable data doesn't match report_data");
+    }
+    debug!("HCL user-data verification completed successfully");
+    Ok(())
```
Xynnn007: here they are. In current code routine, the things injected here is kbs's `hash(runtimedata)`. But it's not the key point, the key point is that `verifier` is a underlying crate and it should only know someone give it a reportdata. How the reportdata is generated and what does it mean should not be known by it. It also does not need to know "kbs".

Thus a generic `report_data` name is enough here

Xynnn007: in a way it's semantics decouple and layer design

Xynnn007:
The CI fails. let me figure out what's wrong.

Xynnn007 (approved):

Xynnn007:
ci error is not related to this PR, and it would be fixed in https://github.com/confidential-containers/trustee/pull/1492

deps/verifier/src/az_snp_vtpm/mod.rs
```diff
@@ -115,16 +153,19 @@ impl Verifier for AzSnpVtpm {
 
         verify_tpm_pcrs(tpm_quote)?;
 
-        let var_data_hash = hcl_report.var_data_sha256();
-        let snp_report = hcl_report.try_into()?;
-        verify_report_data(&var_data_hash, &snp_report)?;
+        verify_report_data(&hcl_report)?;
+
+        if evidence.version() >= 2 {
```
fitzthum, line 158: Can a mitm modify the evidence version in flight?

fitzthum (commented):
Nice. One question

deps/verifier/src/az_snp_vtpm/mod.rs
mkulke: the evidence version is merely a hint for the verifier, not some security facility. If you convert evidence from v2 to v1 or v0 the verifier wouldn't assert user-data. If we reach the conclusion that v1 evidence (cached report in VMPL0 + fresh TPM quotes + AKpub linked to report) is unsound then trustee should stop supporting v1 evidence.

At this point I haven't reached this conclusion, though. v1 evidence is following the [documented process](https://learn.microsoft.com/en-us/azure/confidential-computing/how-to-leverage-virtual-tpms-in-azure-confidential-vms).

mythi:
ITA needs v2 support too. ITA doc states for the `user_data` request input: "The ITA will verify that the user-data field of the RuntimeData contains the SHA512 hash of the concatenated nonce and user data".

mkulke:
> ITA needs v2 support too. ITA doc states for the `user_data` request input: "The ITA will verify that the user-data field of the RuntimeData contains the SHA512 hash of the concatenated nonce and user data".

confusingly ITA doesn't need v2. because the v1 evidence for TDX is already including a fresh TD report with report_data populated (the attester has been changed to support ITA), but trustee didn't verify it yet. v2 evidence is basically a signal for the verifier: "the report is fresh (not cached) and you can assert the report_data field"

deps/verifier/src/az_snp_vtpm/mod.rs
fitzthum: Hm ok.

fitzthum (approved):
LGTM

PR #1492, opened 2026-07-15
ci: pin source tarball artifact name in kbs e2e

Xynnn007:
The build-binaries job of the reusable kbs e2e workflow downloaded all artifacts of the run without specifying a name. download-artifact places a single artifact at the workspace root, but once other jobs (e.g. the helm e2e materials build) have already uploaded their artifacts, each artifact ends up in its own subdirectory and the subsequent "tar xzf ./kbs.tar.gz" fails with "No such file or directory". Whether the leg passed depended on job timing.

Give the source tarball artifact an explicit name and download only that artifact, so the layout no longer depends on what else has been uploaded.


mkulke (approved):

fitzthum (approved):
LGTM

PR #1493, opened 2026-07-16
ci: rename tdx-vm group name

Xynnn007:
Now organization level runners would have a common name format, s.t. xxx-vm

for CVM on arch xxx.


fitzthum (approved):
LGTM

PR #1489, opened 2026-07-15
Helm: Add valkey(redis) support for kbs session storage

Xynnn007:

PR #1494, opened 2026-07-16
CI | Avoid cancelling other e2e tests when adding hw tee e2e test label

Xynnn007:
Firstly found in https://github.com/confidential-containers/trustee/pull/1489.

When adding `tdx-e2e` label to trigger tdx e2e tests, all other e2e tests are cancelled. This patch fixes this by moving TEE e2e tests into a separate workflow.

Also, to support future SNP and other platforms, a `tdx-e2e`-like label is too fine-grained thus using `tee-e2e` instead.

### DIFF
--- a/.github/workflows/test-e2e-helm-nightly.yml
+++ b/.github/workflows/test-e2e-helm-nightly.yml
@@ -57,7 +57,7 @@ jobs:
       self-hosted: false
 
   # Real TDX Helm e2e on a self-hosted runner. Unlike the PR suite, nightly runs
-  # this unconditionally (no `tdx-e2e` label gate) so the hardware path is
+  # this unconditionally (no `tee-e2e` label gate) so the hardware path is
   # exercised every day.
   helm-trustee-e2e-tdx:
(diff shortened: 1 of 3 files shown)


Xynnn007:
Good, the only TEE test (TDX) under the label works fine w/o cancelling other jobs https://github.com/confidential-containers/trustee/actions/runs/29488033719/job/87588839690?pr=1494

.github/workflows/test-e2e-helm-nightly.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       self-hosted: false
 
   # Real TDX Helm e2e on a self-hosted runner. Unlike the PR suite, nightly runs
```
fitzthum, line 59: nit: might remove TDX from this line too

fitzthum (approved):
LGTM

.github/workflows/test-e2e-helm-nightly.yml
fitzthum: nvm i'll do it

PR #1497, opened 2026-07-20
Enable CI on SNP 

fitzthum:
Enable CI on SNP VM runner. This will run nightly and on PRs with the `tee-e2e` label.


fitzthum:
Job passing on SNP runner VM

Xynnn007 (approved):
Thanks for this. Let's add it and see tomorrow's dashboard. Btw, there should be a proess inside the e2e test to set attestation policy to ensure it's TEE based. Let me do it then

PR #1498, opened 2026-07-20
snp-verifier: offline certificate store support for multiple VCEKs per hwid

amd-aliem:
Follow-on to https://github.com/confidential-containers/trustee/pull/1142, which enabled specifying a trustee directory which may be preloaded with certs for offline/airgapped trustee deployments.

Current implementation only allows one certificate per hwid path, which makes FW updates more difficult (have to time the trustee cache update together with the fw update). Add support for storing multiple certificates per host/hwid to allow pre-loading certificates.

Notes:
- adds support to append a prefix to certificates under `hwid` path named based on TCB of vcek. Implemented this way to look up the appropriate VCEK rather than search. `vcek/{hwid}/vcek.der` ->`vcek/{hwid}/{tcb_prefix}_vcek.der`. 
  - prefixes are spls in this order: `bl{spl}_tee{spl}_snp{spl}_ucode{spl}[_fmc_{spl}]`
  - SPLs are in decimal, 0-padded to 2 (same as the trustee KDS URL builder code)
- backwards compatible - falls back to legacy non-prefixed name if TCB subdirectory is absent. Docs point out that subdirectory layout is preferred (avoid the extra lookup).

Testing:
- Tested by pinning test host's TCB to several different values with  `snphost config set <tcb values>` and having a set of pods interact with a trustee deployment which had an iptable rule to drop traffic to amd's kds. 

Side note that this `snphost config set <tcb values>` command's main purpose is to help with this FW update case if you're not able to prefetch the next cert before updating - it can pin SPLs to a particular value after updating FW until you're able to locate and fetch the appropriate certs.

### DIFF
--- a/attestation-service/docs/amd-offline-certificate-cache.md
+++ b/attestation-service/docs/amd-offline-certificate-cache.md
@@ -1,4 +1,4 @@
-# Offline AMD VCEK Caching Guide
+# Offline AMD VCEK Store Guide
 
 This document describes the process to pre-load VCEKs (Versioned Chip
 Endorsement Keys) into the trustee environment to allow normal attestation
@@ -40,27 +40,47 @@ Update the attestation configuration file to use a predefined vcek store:
 ```
 
 With the `OfflineStore` configuration specified, Trustee will inspect the
-configured directory for VCEK values following the following format:
+configured directory for VCEK values.
 
+Given an attestation report, trustee will search first:
 ```
+# Preferred (TCB-based filename, when present):
+/opt/confidential-containers/attestation-service/kds-store/vcek/{hwid}/{tcb_prefix}_vcek.der
+```
+
+And if that lookup fails, the legacy flat format will be searched:
+```
+# Fallback (legacy flat layout):
 /opt/confidential-containers/attestation-service/kds-store/vcek/{hwid}/vcek.der
 ```
 
-Where `{hwid}` is the hardware ID of the AMD EPYC server in lowercase hexadecimal.
+Where:
+- `{hwid}` is the hardware ID of the AMD EPYC server in lowercase hexadecimal
+- `{tcb_prefix}` is a TCB-parameter-based filename prefix in the format:
+  - For non-Turin processors: `bl{BL}_tee{TEE}_snp{SNP}_ucode{UCODE}`
+  - For Turin processors: the above with `_fmc{FMC}` appended
+
+  Each parameter is zero-padded to 2 digits (e.g., `bl02_tee00_snp06_ucode21`)
 
 ### 2. Create and Populate VCEK directory
 
-Create a `vcek` directory populated with the following structure:
+Create a `vcek` directory populated with one of the following structures per
+hardware ID. The TCB-prefixed filename is recommended, to allow loading multiple VCEKs per host.
 
 ```
 vcek/
 ├── <Hex hardware ID>/ (ID must be lowercase)
-│   └── vcek.der (cert pre-downloaded from kdsintf.amd.com)
+│   ├── vcek.der                          # flat layout (fallback)
+│   └── <TCB prefix>_vcek.der             # optional, preferred when present
 ├── <Hex hardware ID>/
 │   └── vcek.der
-├── <Hex hardware ID>/
-    └── vcek.der
+└── <Hex hardware ID>/
+    └── <TCB prefix>_vcek.der
 ```
+
+The `<TCB prefix>` is derived from the certificate's TCB parameters and follows
+the format `bl{BL}_tee{TEE}_snp{SNP}_ucode{UCODE}` (with `_fmc{FMC}` appended for Turin
+processors). For example: `bl02_tee00_snp06_ucode21_vcek.der`
 
 Trustee requires one unique certificate per physical AMD EPYC server that the
 KBS will be servicing. The server's hardware ID and certificate's URL can be
@@ -77,10 +97,20 @@ https://kdsintf.amd.com/vcek/{version}/{machine}/{product_name}/{hwid}?{params}
 ```
 
 You may download it from a browser by pasting that URL, or you can run
-the following command on any server with network access to AMD's KDS.
+the following command on the SNP host itself (it derives the URL from the
+local firmware). Requires network access to AMD's KDS.
 ```
-sudo snphost fetch vcek der .
+sudo snphost fetch vek der .
 ```
+
+To fetch from a different machine (e.g. a build box), pass the URL explicitly
+so the cert matches the originating host's firmware rather than the local one:
+```
+sudo snphost fetch vek der . "<vcek-url>"
+```
+
+> [!Note]
+> Older versions of `snphost` use `snphost fetch vcek` instead of `vek`. However it's recommended to update to the latest version of `snphost`.
 
 > [!IMPORTANT]
 > - Note that the VCEK URL is specific to the hardware AMD firmware of the
@@ -94,7 +124,7 @@ sudo snphost fetch vcek der .
 
 Use docker commands to copy your `vcek` folder into the configured directory:
 ```
-sudo docker exec -it trustee-as-1 mkdir -p /opt/confidential-containers/attestation-service/kds-store/
+sudo docker exec trustee-as-1 mkdir -p /opt/confidential-containers/attestation-service/kds-store/
 sudo docker cp ./vcek/ trustee-as-1:/opt/confidential-containers/attestation-service/kds-store/vcek/
 ```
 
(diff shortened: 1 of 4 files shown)


fitzthum:
I have a feeling that @Xynnn007 will suggest using the key-value storage backend for this. You should do that. tbd if that should be part of this PR. Since you are doing some changes specific to the filesystem, I suspect you maybe should switch backends before you get much deeper.

Xynnn007:
Yes I guess key-value storage backend is one step, but we still needs a full workflow for pre-configure/cache. Let's call the full workflow a solution. The key-value storage backend can be a component of the solution.

I believe that s390x meets the same issue for now, which is using local PV for verification materials.

The most important operations for users are
1. how to pre-configure some vceks
2. how to do caching transparently and keep computation-storage decoupled in the cluster

There might be two ways in my mind.
1. Use an out-of-band cache service. 

Naturally computation-storage is decoupled, and verifier only need to connect to it. How to provision vceks depends on how the vcek storage is organized by that service.

2. Use built-in cache service inside AS/verifier.

This is the way now going on. Now we provides a [namespace](https://github.com/confidential-containers/trustee/pull/1417) mechanism for global storage. This is coupled with what the backend storage is. For production, we usually choose postgreSQL. Thus the pre-configuration depdens on some well defined interaction between client and the postgres instance, which could be encapsulated into a tool/script.

amd-aliem:
Hey @Xynnn007 & @fitzthum thanks for the info! I'm not too familiar with the postgreSQL path yet, but from what I gather the key-value storage lets us swap between Postgres, LocalFs, and other backends interchangeably.

So my plan would be:
1. Wire the SNP verifier onto the key-value storage backend, using LocalFs to start (it also fits the docker-compose deployment I'm testing against, which I don't think runs postgres?). 
2. Provide or document a pre-configuration script for loading VCEKs into the LocalFs store as the short-term "full workflow" piece.

Then in a follow-up I'd investigate Postgres as the production backend, and pair it with an AS endpoint, following the same pattern as rvps-tool (or some other solution that is better than a script). 

Does that seem reasonable?

Xynnn007:
@amd-aliem This works fine for short time. For long time, I lean towards to way 1

Some reasons

1. This would help AMD SNP have a general vcek cache service (prototype). It potentially has ability to help different CSPs to set up locally, to accelerate the speed of VCEK getting.
2. This would help trustee deployment totally decoupled with vcek caching mechaism. Note that with different storage backend, we will have different vcek provision steps. If the vcek has a standard way to be cached and provisioned by AMD, what the trustee deployment should do is only configure an API address.
3. One cache service can serve one/multiple cluster(s) inside a same air-gapped env.

fitzthum:
I'm not a fan of an out-of-band caching service.

Ideally pre-provisioning should not be tied to a particular storage backend. During pre-provisioning, the backend should be accessed through the generic KvStorage setters. We could either provide a standalone binary for this or allow pre-provisioning via the config file. 

This might be too much for this PR, though.

mkulke (commented):
is "cache" (e.g. in the PR name) the right term here? I think this implies that something is fetched and then stored to avoid re-fetching the same thing. afaict we don't store anything in this PR?

I think the term "offline store" is used in the code, which seems more fitting. If we don't perform write operations, I think it's fine to pre-seed a deployment with an immutable set of certs in a folder. in a later stage maybe the lifecycle of those certs can be managed via a storage backend.

Xynnn007:
@mkulke This PR does not write any. Currently, SNP verifier configuration can select either to use offline, or to access online KDS. This PR only changes the directory organization for offline. What we discussed might be off the topic.

So, all my comments are based on the assumption that "first local, then online and cache" is the way @amd-aliem wants to go. We can discuss it in a separate thread imo.

deps/verifier/src/snp/mod.rs
```diff
@@ -175,10 +191,19 @@ impl Snp {
...
-            .with_context(|| format!("Failed to read VCEK from offline store at {}", vcek_path))?;
-        Ok(vcek_bytes)
+        let default_vcek_path = format!("{path}/vcek/{hw_id}/vcek.der");
+        let tcb_vcek_path = format!(
+            "{path}/vcek/{hw_id}/{}/vcek.der",
+            format_tcb_dir(&att_report, proc_gen)
+        );
+
+        std::fs::read(&tcb_vcek_path)
+            .or_else(|_| std::fs::read(&default_vcek_path))
```
Xynnn007: here the error of trying to read `tcb_vcek_path` would be hidden. What about
```
.or_else(|e| {
debug!("failed to read vcek from {tcb_vcek_path}, try {default_vcek_path}. Original error: {e}");
std::fs::read(&default_vcek_path)
})
```

Xynnn007 (approved):
LGTM.

attestation-service/docs/amd-offline-certificate-cache.md
```diff
@@ -120,21 +152,24 @@ ssh privileged-user@epyc-host "sudo snphost show vcek-url" >> urls.txt
 
 On a system with network access to AMD KDS:
 ```bash
-
-# Create the vcek folder that will be copied to trustee
 mkdir vcek
+qval() { grep -oP "$1SPL=\K[0-9]+" <<<"$2"; }
```
lmilleri, line 155: nit: what are $1 and $2 parameters? Need to be documented?

deps/verifier/src/snp/mod.rs
```diff
@@ -107,6 +107,22 @@ fn init_cache_manager() -> MokaManager {
...
+// caching (only used for OfflineStore option)
+fn format_tcb_dir(att_report: &AttestationReport, proc_gen: &ProcessorGeneration) -> String {
+    let tcb = &att_report.reported_tcb;
+    let mut tcb_dir = format!(
+        "bl{:02}_tee{:02}_snp{:02}_ucode{:02}",
+        tcb.bootloader, tcb.tee, tcb.snp, tcb.microcode,
+    );
+    if matches!(proc_gen, ProcessorGeneration::Turin) {
+        let fmc = tcb.fmc.unwrap_or(0);
+        use std::fmt::Write;
```
mkulke: we usually move imports to the top of the file

deps/verifier/src/snp/mod.rs
```diff
@@ -175,10 +191,19 @@ impl Snp {
...
-        let vcek_bytes = std::fs::read(&vcek_path)
-            .with_context(|| format!("Failed to read VCEK from offline store at {}", vcek_path))?;
-        Ok(vcek_bytes)
+        let default_vcek_path = format!("{path}/vcek/{hw_id}/vcek.der");
+        let tcb_vcek_path = format!(
+            "{path}/vcek/{hw_id}/{}/vcek.der",
+            format_tcb_dir(&att_report, proc_gen)
+        );
+
+        std::fs::read(&tcb_vcek_path)
```
mkulke: if this is supposed to fail for a reason, e.g. because it doesn't exist that we can check for existence. maybe we don't have the privileges to read the content of the folder, which should be an error and we shouldn't fall back to default path silently.

deps/verifier/src/snp/mod.rs
```diff
@@ -107,6 +107,22 @@ fn init_cache_manager() -> MokaManager {
...
 
+// convert TCB values into corresponding subdirectory name for offline cert
+// caching (only used for OfflineStore option)
+fn format_tcb_dir(att_report: &AttestationReport, proc_gen: &ProcessorGeneration) -> String {
+    let tcb = &att_report.reported_tcb;
+    let mut tcb_dir = format!(
+        "bl{:02}_tee{:02}_snp{:02}_ucode{:02}",
+        tcb.bootloader, tcb.tee, tcb.snp, tcb.microcode,
+    );
+    if matches!(proc_gen, ProcessorGeneration::Turin) {
```
fitzthum: Would it be more forward-compatible to just check if fmc is Some?

fitzthum (commented):
One note, but seems fine to me pending the other feedback.

deps/verifier/src/snp/mod.rs
amd-aliem: Combined with https://github.com/confidential-containers/trustee/pull/1498#discussion_r3629881012 - updated error handling and check for file existance first.

attestation-service/docs/amd-offline-certificate-cache.md
amd-aliem: true, added function header comments

deps/verifier/src/snp/mod.rs
amd-aliem: fixed

deps/verifier/src/snp/mod.rs
amd-aliem: good point, split the existence check out

deps/verifier/src/snp/mod.rs
amd-aliem: true, fixed (and verified that Turins will always fill with Some, prev versions always None)

amd-aliem:
Thanks all for the speedy reviews!!

After talking a bit with @lmilleri, found the subdir version is not compatible with how the trustee operator handles this config (mounts secrets onto the configured fs path, but k8s doesn't support mounting in nested dirs), so a flat layout works better.

I've tweaked the code (and docs) from `vcek/{hwid}/{tcb_dir}/vcek.der` -> `vcek/{hwid}/{tcb_prefix}_vcek.der` (where `tcb_prefix` is constructed the same way `tcb_dir` was).

I've also applied all review comments @fitzthum @Xynnn007 @mkulke @lmilleri

lmilleri (approved):
PR description needs updating for flat {tcb}_ filenames.
LGTM thank you

mkulke (approved):

fitzthum (approved):
LGTM

PR #1508, opened 2026-07-23
CI | set TEE requirement for helm e2e

### LINKED ISSUE #1500: Helm | Add TEE attestation assertion logic in e2e test
Now we run e2e helm test but never explicitly set the attestation policy. Thus it could using the default policy for KBS. Let's explicitly set the policy by asserting the requesting AS is HW-TEE.

Xynnn007:
close #1500

### DIFF
--- a/deployment/helm-chart/Makefile
+++ b/deployment/helm-chart/Makefile
@@ -9,6 +9,11 @@ REPO_ROOT := $(abspath ../..)
 CHART_DIR := $(abspath .)
 KBS_CLIENT ?= $(REPO_ROOT)/target/release/kbs-client
 
+# When set (e.g. EXPECTED_TEE=tdx), the e2e client additionally asserts that the
+# attestation evidence was produced by this TEE type, turning the leg into a real
+# hardware-path check. Empty by default so non-TEE / local runs are unaffected.
+EXPECTED_TEE ?=
+
 E2E_IMAGE_PREFIX ?= trustee-e2e
 E2E_IMAGE_TAG ?= e2e
 
@@ -35,6 +40,7 @@ HELM_E2E_IMAGE_ARGS := \
 	--set-string rvps.image.pullPolicy='Never'
 
 export KBS_CLIENT
+export EXPECTED_TEE
 
 .PHONY: e2e-test deploy undeploy test-client check-tools wait-ready \
 	check-docker helm-dependency-build helm-lint load-e2e-images-into-kind
--- a/deployment/helm-chart/e2e/test.sh
+++ b/deployment/helm-chart/e2e/test.sh
@@ -22,6 +22,12 @@ KBS_URL="http://127.0.0.1:8080"
 TEST_RESOURCE_FILE="${SCRIPT_DIR}/fixtures/test-resource.txt"
 RESOURCE_PATH="helm-e2e/test-repo/test-secret"
 
+# When set, additionally assert that the attestation evidence was produced by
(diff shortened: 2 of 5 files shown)


deployment/helm-chart/e2e/test.sh
```diff
@@ -52,6 +58,25 @@ kbs_client() {
...
+write_require_tee_policy() {
+	local tee="$1" out="$2"
+	cat >"${out}" <<EOF
+package policy
+import rego.v1
+
+default allow = false
+
+allow if {
+	some _, submod in input.submods
```
fitzthum: In the docs we just do `input["submods"]["cpu0"]["ear.veraison.annotated-evidence"]["<TEE>"]` which is a little clearer imo

deployment/helm-chart/e2e/test.sh
```diff
@@ -81,6 +106,45 @@ start_port_forward() {
...
+# the matching TEE type (resource must be released) and a negative check requires
+# a mismatched TEE type (resource must be denied), so we know the policy engine is
+# actually evaluating the evidence rather than releasing unconditionally.
+check_tee_type() {
+	local match_policy="${WORK_DIR}/require-tee.rego"
+	local mismatch_policy="${WORK_DIR}/require-wrong-tee.rego"
+	# Any TEE type the evidence cannot legitimately carry. Sample evidence never
+	# appears on real hardware; on the sample leg fall back to a real TEE name.
+	local mismatch_tee="sample"
+	[[ "${EXPECTED_TEE}" == "sample" ]] && mismatch_tee="tdx"
```
fitzthum: Maybe just set `mistmatch_tee` to be a nonexistent platform then you don't need this logic.

fitzthum (commented):
Nice. Two small comments.

deployment/helm-chart/e2e/test.sh
Xynnn007: The difference is the `cpu0` field, hinting that the logic is to assert "a cpu tee is <TEE>", while the current shape is "a tee is <TEE>".

Once we have device TEEs we still need such format, thus it's fine.

fitzthum: I don't anticipate having devices available in the Trustee CI anytime soon. Even when there are, we would then want the test to check both the cpu and the device.

Xynnn007: Ok. It works for both gpu and CPU. but anyway, let me change this to your idea since it looks like a blocker

deployment/helm-chart/Makefile
```diff
@@ -9,6 +9,11 @@ REPO_ROOT := $(abspath ../..)
 CHART_DIR := $(abspath .)
 KBS_CLIENT ?= $(REPO_ROOT)/target/release/kbs-client
 
+# When set (e.g. EXPECTED_TEE=tdx), the e2e client additionally asserts that the
+# attestation evidence was produced by this TEE type, turning the leg into a real
+# hardware-path check. Empty by default so non-TEE / local runs are unaffected.
```
mythi, line 14: any reason why `sample` is not used here so that the check could run unconditionally?

Xynnn007: Note that if this variable is not given, `check_tee_type` will not be executed. Thus by default it's empty

dcmiddle (approved):

fitzthum (approved):
LGTM

Xynnn007:
Let me get this merged as two approvals are got. We can fix it later if needed

PR #1509, opened 2026-07-23
docs(kbs): document protocol error types

### LINKED ISSUE #99: Document protocol errors

AnkushUjawane:
Resolves #99. Lists the actual error type values returned in the 'type' field of ErrorInformation responses, sourced from kbs/src/error.rs, and clarifies that the URL is an identifier string rather than a functioning hyperlink.

### DIFF
--- a/kbs/docs/kbs_attestation_protocol.md
+++ b/kbs/docs/kbs_attestation_protocol.md
@@ -727,6 +727,29 @@ format is used for that purpose:
   "detail": "$detail"
 }
 ```
+> **Note:** the `type` field is an identifier string, not a functioning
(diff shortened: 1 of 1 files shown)


fitzthum (approved):
LGTM

We can try to get keep this up-to-date

Xynnn007 (approved):

PR #1520, opened 2026-07-28
ci: switch zizmor to auditor persona and fix all pre-existing findings

supritai:
## Why

The zizmor workflow added in #1481 used the `regular` persona. While correct as a starting point, `regular` misses a class of findings that `auditor` catches — undocumented permissions, anonymous job definitions, missing concurrency limits, and template expansions from `matrix.*` / `runner.temp` in `run:` blocks.

`auditor` is the better long-term choice because it enforces stricter hygiene on every PR before anything merges, rather than silently allowing these patterns to accumulate.

## What

Runs zizmor with `persona: auditor` on all triggers (push, PR, weekly schedule). This is safe because all 52 pre-existing `auditor` findings are fixed in this commit — the repo now exits clean (`exit 0`) under the strictest persona.

Getting new findings going forward will be a rare scenario: they would only appear if someone introduces a new workflow with one of the five resolved finding types, which the PR scan would immediately catch and flag before merge.

Findings resolved across 15 workflow files:

- **template-injection** (21): moved `${{ matrix.* }}` and `${{ runner.temp }}` expansions in `run:` blocks into `env:` vars or `$RUNNER_TEMP`
- **undocumented-permissions** (16): added inline comments to all `packages: write`, `pages: write`, `id-token: write`, `security-events: write`, and `actions: read` grants
- **anonymous-definition** (14): added `name:` to all jobs identified only by their YAML key
- **ref-version-mismatch** (1): corrected stale OPA action comment `# v2.2.0 # v2` → `# v2.4.0` in `test-rust-ci.yml`
- **concurrency-limits** (1): added `concurrency:` block to `release-clis-to-ghcr.yml`

## Validation

```console
$ zizmor --persona auditor .
No findings to report. Good job!
```

> Assisted by [IBM Bob](https://github.com/IBM/bob)

### DIFF
--- a/.github/workflows/release-clis-to-ghcr.yml
+++ b/.github/workflows/release-clis-to-ghcr.yml
@@ -7,40 +7,51 @@ on:
 
 permissions: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
 jobs:
   build_and_push:
+    name: Build and push ${{ matrix.component }} on ${{ matrix.arch }}
     strategy:
       fail-fast: false
       matrix:
         include:
           # kbs-client
           - component: kbs-client
+            arch: x86_64
             instance: ubuntu-24.04
             dockerfile: kbs/docker/kbs-client/Dockerfile
             output_path: kbs-client
             repo: ghcr.io/confidential-containers/staged-images/kbs-client
           - component: kbs-client
+            arch: s390x
             instance: ubuntu-24.04-s390x
             dockerfile: kbs/docker/kbs-client/Dockerfile
             output_path: kbs-client
             repo: ghcr.io/confidential-containers/staged-images/kbs-client
           - component: kbs-client
+            arch: aarch64
             instance: ubuntu-24.04-arm
             dockerfile: kbs/docker/kbs-client/Dockerfile
             output_path: kbs-client
             repo: ghcr.io/confidential-containers/staged-images/kbs-client
           # trustee-cli
           - component: trustee-cli
+            arch: x86_64
             instance: ubuntu-24.04
             dockerfile: tools/trustee-cli/Dockerfile
             output_path: ./usr/local/bin/trustee
             repo: ghcr.io/confidential-containers/staged-images/trustee-cli
           - component: trustee-cli
+            arch: s390x
             instance: ubuntu-24.04-s390x
             dockerfile: tools/trustee-cli/Dockerfile
             output_path: ./usr/local/bin/trustee
             repo: ghcr.io/confidential-containers/staged-images/trustee-cli
           - component: trustee-cli
+            arch: aarch64
             instance: ubuntu-24.04-arm
             dockerfile: tools/trustee-cli/Dockerfile
             output_path: ./usr/local/bin/trustee
@@ -48,7 +59,7 @@ jobs:
     runs-on: ${{ matrix.instance }}
     permissions:
       contents: read
-      packages: write
+      packages: write  # push CLI artifacts to GHCR
 
     steps:
     - name: Install ORAS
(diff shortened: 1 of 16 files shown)


.github/workflows/release-clis-to-ghcr.yml
```diff
@@ -7,8 +7,13 @@ on:
 
 permissions: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
 jobs:
   build_and_push:
+    name: Build and push ${{ matrix.component }} (${{ matrix.instance }})
```
Xynnn007: This would look like
```
Build and push trustee-cli ubuntu-24.04
```

Maybe it should look like
```
Build and push trustee-cli on x86_64
```

and 
```
Build and push trustee-cli on s390x
```

.github/workflows/test-e2e-kbs.yml
```diff
@@ -132,7 +132,7 @@ jobs:
           path: ${{ runner.temp }}/docker-e2e-images
 
       - name: Load images
-        run: docker load --input "${{ runner.temp }}/docker-e2e-images/images.tar"
+        run: docker load --input "$RUNNER_TEMP/docker-e2e-images/images.tar"
```
Xynnn007: I see also line 132, `path: ${{ runner.temp }}/docker-e2e-images` refers `${{ runner.temp }}` also. Do we need to change it also?

.github/workflows/workflow-call-build-docker-e2e-materials.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           key: rust-${{ runner.arch }}-${{ env.OS_VERSION }}-kbs-client-${{ matrix.name }}-${{ hashFiles('./Cargo.lock') }}
 
       - name: Build kbs-client
-        run: cargo build -p kbs-client --release ${{ matrix.features }}
+        run: cargo build -p kbs-client --release ${{ matrix.features }} # zizmor: ignore[template-injection]
```
Xynnn007: Why not we use env templating here also?

.github/workflows/workflow-call-build-docker-e2e-materials.yml
```diff
@@ -135,17 +135,21 @@ jobs:
...
 
       - name: Save image
+        env:
+          MATRIX_TAG: ${{ matrix.tag }}
         run: |
           set -euo pipefail
-          mkdir -p "${{ runner.temp }}/docker-e2e-image"
-          docker save -o "${{ runner.temp }}/docker-e2e-image/${{ matrix.tag }}.tar" \
-            "${IMAGE_PREFIX}/${{ matrix.tag }}:latest"
+          mkdir -p "$RUNNER_TEMP/docker-e2e-image"
```
Xynnn007: I saw the original code and new codes are using like `$RUNNER_TEMP` and `${MATRIX_TAG}`. Do you think we should unify the embraces style? s.t. all use,

Xynnn007 (commented):
Thanks @supritai ! Some comments

.github/workflows/release-clis-to-ghcr.yml
supritai: Yes, that would look good.

.github/workflows/test-e2e-kbs.yml
supritai:
```
This is inside a with: block (an actions/download-artifact input), not a run: block. 
So, $RUNNER_TEMP cannot be used here — shell variables only work inside run: blocks. 
In with: blocks, only ${{ }} template expressions are evaluated by GitHub.

zizmor doesn't flag ${{ runner.temp }} in with: blocks — the template-injection rule only applies to 
run: blocks, where template expansion feeds into a shell. 
In with:block, the value goes directly to the action's input, not a shell.
```

.github/workflows/workflow-call-build-docker-e2e-materials.yml
supritai: Complexity to satisfy all checks..!

${{ matrix.features }} directly in run: | ✅ Works, ❌ zizmor flags it

${MATRIX_FEATURES} unquoted | ✅ zizmor clean, ✅ cargo works, ❌ shellcheck warns

${MATRIX_FEATURES} + # shellcheck disable=SC2086 | ✅ zizmor clean, ✅ cargo works, ✅ shellcheck clean

.github/workflows/workflow-call-build-docker-e2e-materials.yml
supritai: All variables across all workflow files consistently use ${VAR} style with braces.

Xynnn007 (approved):
LGTM. Thanks!

fitzthum (approved):

PR #1521, opened 2026-07-28
KBS: Add attestation policy selection

### LINKED ISSUE #282: KBS does not support multiple attestation policies
This issue is for discussion rather than any immediate fix. 

Recently we updated the AS API to allow callers to specify the ids of the policies that are evaluated. We [don't make use of this](https://github.com/confidential-containers/kbs/blob/f18a0e0ce9856360e7a7db495592ad7ad95a7cd4/kbs/src/api/src/attestation/coco/grpc.rs#L115) in our KBS should we? Perhaps we could allow the client to select which policies to evaluate?

Xynnn007:
This patch mainly introduces a static config that maps id field to a set of policy ids. By default, if no id is given from client side, the "default" policy will still be used.

For ITA, in previous version it has its own config to specify the policy ids. In this new version, if any id->policy ids is given by the request, it will use the mapped policy ids. If not, use the policy ids in the ITA config.

We do not bump a protocol version here as it's add-only feature and do not break.

Fixes #282

### DIFF
--- a/kbs/docs/kbs_attestation_protocol.md
+++ b/kbs/docs/kbs_attestation_protocol.md
@@ -117,6 +117,28 @@ transfer some specific information. For example, some attestations follow the
 Diffie–Hellman key exchange protocol to first build a secure channel and
 transfer secret messages (Such as AMD SEV(-ES) pre-attestation).
 
+`extra-params` may also carry a `policy-selector`, which selects the Attestation
+Service policies that evaluate this session's evidence:
+
+```json
+"extra-params": { "policy-selector": "alice" }
+```
+
+KBS maps each `policy-selector` to one or more attestation policies, so the
+accepted values are specific to a deployment and must be known to the KBC in
+advance. An unmapped `policy-selector` is rejected, and omitting the field
+selects a default policy.
+Note that not all backend attestation services support multiple policies, e.g.
+CoCo AS now only support one policy, while ITA supports multiple.
+
+Selecting a `policy-selector` does not by itself entitle the KBC to anything,
+since KBS still decides what each appraisal may release. It does change what the
+resulting token means, though, so a relying party should not treat every token
+alike. The [Attestation Results Token](#attestation-results-token) names the
+policy that was actually applied, which is the policy the `policy-selector`
+resolved to rather than the `policy-selector` itself, so a resource policy can
+require a particular appraisal before releasing a resource.
+
 ## `Challenge`
 
 If the KBC does not own any KBS generated HTTP Cookie, or if the Cookie validity
--- a/integration-tests/src/common.rs
+++ b/integration-tests/src/common.rs
@@ -31,6 +31,7 @@ use base64::{
 use jsonwebtoken::{encode, Algorithm, EncodingKey, Header};
 use openssl::pkey::PKey;
 use serde_json::json;
+use std::collections::HashMap;
 use std::sync::{Arc, Once};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use tempfile::TempDir;
@@ -332,6 +333,7 @@ impl TestHarness {
                 extra_teekey_paths: vec![],
             },
             attestation_service: AttestationConfig {
+                policy_id_map: HashMap::new(),
                 attestation_service: AttestationServiceConfig::CoCoASBuiltIn(
                     kbs::attestation::coco::builtin::Config {
                         rvps_config,
--- a/kbs/docs/config.md
+++ b/kbs/docs/config.md
@@ -154,6 +154,15 @@ Concrete attestation service can be set via `type` field. Supported attestation 
 
 Due to different `type` field, properties are different.
 
+`timeout` and `policy_id_map` apply to every type. The latter is a table mapping
+each policy-selector a client may select to one or more policy IDs:
+
+```toml
+[attestation_service.policy_id_map]
+alice = ["alice-strict"]
+bob = ["bob-cpu", "bob-gpu"]
+```
+
 #### Built-In CoCo AS
 
 When `type` is set to `coco_as_builtin`, the following properties can be set.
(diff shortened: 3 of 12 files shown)


kbs/docs/kbs_attestation_protocol.md
```diff
@@ -117,6 +117,21 @@ transfer some specific information. For example, some attestations follow the
...
 transfer secret messages (Such as AMD SEV(-ES) pre-attestation).
 
+`extra-params` may also carry an `id`, which selects the Attestation Service
+policies that evaluate this session's evidence:
+
+```json
+"extra-params": { "id": "alice" }
+```
+
+KBS maps each `id` to one or more attestation policies, so the accepted values
```
fitzthum: The number of policies is a bit tricky. The AS takes a list of policies but only accepts one. We should probably be more clear that we only support one policy for the coco AS. Ideally the interface would be for just one policy although ITA supports multiple.

kbs/docs/kbs_attestation_protocol.md
```diff
@@ -117,6 +117,21 @@ transfer some specific information. For example, some attestations follow the
 Diffie–Hellman key exchange protocol to first build a secure channel and
 transfer secret messages (Such as AMD SEV(-ES) pre-attestation).
 
+`extra-params` may also carry an `id`, which selects the Attestation Service
+policies that evaluate this session's evidence:
+
+```json
+"extra-params": { "id": "alice" }
```
fitzthum: `id` is way too broad of a name. This alludes to identity more generally, which this feature doesn't really provide. We should call this something more specific to the policy.

kbs/docs/kbs_attestation_protocol.md
```diff
@@ -117,6 +117,21 @@ transfer some specific information. For example, some attestations follow the
...
+```
+
+KBS maps each `id` to one or more attestation policies, so the accepted values
+are specific to a deployment and must be known to the KBC in advance. An
+unmapped `id` is rejected, and omitting the field selects a default policy.
+
+The KBC picks the `id` itself, so doing so is not an authorization decision. The
```
fitzthum: I don't really understand what is meant by authorization decision. The policy hint must exist in the map configured by the admin, but this is still a highly significant field. It can completely shift the meaning of the attestation result.

Should consumers of the attestation token now be checking the policy id to make sure it was the right one? Ofc they don't know the real policy id or content, just the hint.

fitzthum (commented):
Code looks fine. A couple design comments, but I think this is workable. At least it's disabled by default.

kbs/docs/kbs_attestation_protocol.md
Xynnn007: Let's state that **now** only one policy is supported if backend is CoCoAS. Only the first takes effort. 

To support multiple policy is on my todo list also.

wdyt?

kbs/docs/kbs_attestation_protocol.md
Xynnn007: I once thought that `policy-hint`, however - I have some thoughts in my mind to share

1. For a pod owner who is not the attestation service admin, suppose she only knows the Trustee and KBS abstraction, policy context would make her feel confused
2. For a group of pods with a same hint, in a way they have common identity for attestation
3. In a sense, if we use some name specific to policy, we will get back to something related to id in future, but we can change it later.

With my contexts, do you have any better names? and i'd like to change

kbs/docs/kbs_attestation_protocol.md
Xynnn007: Oh. That means selection of the id does not mean any authorization to access the KBS, just a hint. All authorization work is to filter the AS token based on kbs policy

Xynnn007: The user does not need to know. It does not matter. It matters for KBS to know it. The token provisioner hinted by `iss` and the policy id can be used to generally index a policy. The hint is enough for the user. But, if the user and the attestation admin/trustee admin, he needs to know it

kbs/docs/kbs_attestation_protocol.md
fitzthum: I guess. I don't think we should support multiple policies at any point really tho.

kbs/docs/kbs_attestation_protocol.md
fitzthum: Id is a product of many things (reference values, workload measurements, init-data, etc), not just the policies. After all the thinking we've done about identity, we can't just throw this in as an id field. Personally I think policy-hint or policy-selector is best. That is the most clear. We could do something like `attestation-context` but this will probably just confuse people. Best to call it what it is even if end users might not know the terminology at first.

kbs/docs/kbs_attestation_protocol.md
fitzthum: I am just a little worried about cases where we switched between two policies with different meanings. Keep in mind that this field is not measured so it could be modified in transit. In this case does the user need to inspect the attestation token to make sure the right policy was used?

kbs/docs/kbs_attestation_protocol.md
Xynnn007: We should I believe. But anyway, I use the words to objectively describe what we have now. I think it would be ok
```
KBS maps each `id` to one or more attestation policies, so the accepted values
are specific to a deployment and must be known to the KBC in advance. An
unmapped `id` is rejected, and omitting the field selects a default policy.
Note that not all backend attestation services support multiple policies, e.g.
CoCo AS now only support one policy, while ITA supports multiple.
```

kbs/docs/kbs_attestation_protocol.md
Xynnn007: Let me use `policy-selector`

kbs/docs/kbs_attestation_protocol.md
Xynnn007: It can be modified during transit. But note that AS will sincerely evaluate the evidences based on the "malwared" policy id(s). Then, it's responsibility of KBS, to check if the token has proper policy id(s).

fitzthum: i still think this is a little sketchy, but I guess we do not enable this feature by default so I can live with it

fitzthum (approved):
LGTM

like i said, i don't totally love this feature, but i guess it's ok to give people the option. 

maybe another review from @confidential-containers/trustee-maintainers would be good

Xynnn007:
cc @mythi I guess you will be interested in this multiple policy selection part.

Xynnn007:
Seem like I cannot get copilot to review this PR.

Xynnn007:
Let me get this merged.

fitzthum:
Would have been nice to get another review.

Xynnn007:
Yes - I hoped so too. 

But looks like nobody else cares about this and it hanged for a long time.  let's just proceed this. We can optimize/iterate it if there are strong reasons for this.

PR #1522, opened 2026-07-28
as: add /healthz endpoint to restful-as

AnkushUjawane:
KBS already exposes /healthz for liveness/readiness checks, and both RVPS and grpc-as implement the standard gRPC health checking protocol via tonic-health. restful-as had no equivalent, leaving it as the only Trustee component without a health check when deployed via Kubernetes.

Adds a /healthz route returning 200 OK, matching KBS's existing convention for actix-web based components.


fitzthum (approved):
LGTM

PR #1526, opened 2026-07-29
trustee-cli: add offline policy validate subcommand

AnkushUjawane:
Add 'trustee policy validate <file>' to check that a Rego policy file compiles and evaluates correctly, without needing a running Trustee instance.

Currently the only way to check a policy is valid is to deploy it to a live Attestation Service and test it against real or sample evidence. This exposes the same Regorus evaluation logic AS already uses internally (deps/policy-engine), so policy authors can catch syntax and logic errors locally before deploying.

Defaults to evaluating data.policy.allow against a built-in sample input; both --input and --rule can be overridden.

Manually verified against a known-good policy (compiles, evaluates) and a known-bad policy (fails with a clear Rego compiler error).

$ trustee policy validate my-policy.rego
Policy is valid. hash=5c6e0ec7...
data.policy.allow = true     # example — depends on the policy and input


Xynnn007:
Maybe we can just add document that tells a link to do so w/o adding a binary tool? This would bring extra maintaince effort while we have another easy way to use https://anakrish.github.io/regorus-playground/, in which we copy-paste the policy and input, then can get the same output.

AnkushUjawane:
That's a fair point, and I hadn't seen the playground — thanks for pointing it out. A couple of things it doesn't quite cover, though:

- Scriptability: this can be dropped into a CI step or pre-commit hook to fail a build automatically on an invalid policy, which a browser-based tool can't do.
- Offline use: Trustee is often deployed in restricted or air-gapped environments, and requiring a public website for a basic dev workflow step is friction for exactly that audience.
- Not having to paste policy content into a third-party site, which some operators may prefer to avoid even if the content isn't strictly secret.

If those don't feel compelling enough to justify the maintenance cost, I'm fine pivoting to a docs link instead — just wanted to lay out the tradeoff clearly.

fitzthum:
Thanks for the PR @AnkushUjawane 

One thing to think about is that we might deprecate the Trustee-CLI. We have both the KBS-Client and the trustee CLI. The Trustee CLI was supposed to be the shiny new version of the KBS-Client, but we haven't really had time to work on it and I'm not sure what the future is. 

Idk, do you think we should resurrect the tool or delete it? We could add a similar feature to the KBS Client or as @Xynnn007 suggests just tell people to go use the playground. I don't really have a strong opinion on it.

in the distant future I have pictured a Trustee GUI that would have policy validation built-in.

mkulke (commented):
do we validate any trustee-specifics (i.e. assert that a policy is doing something meaningful for coco) or is this essentially a generic policy linter for regorus?

In the latter case it might either already exist in regorus or would be a better fit there, no?

AnkushUjawane:
Good push on this — you're right that plain compile/eval is redundant with regorus eval and the playground. I've added an --ear flag that validates the actual Trustee/EAR-specific contract instead: that trust_claims uses only the 8 real AR4SI claim names (reusing ear::TrustClaim::try_from directly, so it can't drift from what the ear crate accepts) and values fit i8, plus that extensions entries have the right shape.

Concretely, this catches things like a typo'd claim key ("executible" instead of "executables") that Rego happily compiles
and evaluates, but which would silently produce a broken token in production — something neither regorus eval nor the playground can catch since they don't know what AR4SI is.

@fitzthum on the trustee-cli future question — I don't have a strong view there, this doesn't really weigh in on the deprecation question either way since it's a small, self-contained addition either to trustee-cli or wherever the CLI ends up living.

mythi:
> Concretely, this catches things like a typo'd claim key ("executible" instead of "executables") that Rego happily compiles
> and evaluates, but which would silently produce a broken token in production

I'm not sure this "fails in production" is fully accurate for the trust vector example at least because it looks rust ear errors if `k` is not valid

https://github.com/confidential-containers/trustee/blob/45d4b95389918ff0f2c8e0787d948b1d5ca524a9/attestation-service/src/ear_token/broker.rs#L247-L251

and

https://github.com/veraison/rust-ear/blob/c3a1073d7bca448be91bf52cb44ec9f7f6c348ee/src/trust/vector.rs#L83-L95

I like the intent of this PR but would it make more sense to focus on the case where you could also pass reference values in some format to evaluate `query_reference_values()`?

AnkushUjawane:
Thanks for catching the panic vs "silent" mistake — corrected that in my head at least, appreciate you actually tracing it into broker.rs.

On reference values: added a check-references subcommand that extracts every query_reference_value(...) call from a policy andcross-checks each key against a running RVPS, reusing the same client rvps-tool uses. It's lighter than fully evaluating with plugged-in reference data (doesn't stub the extension), but catches the specific failure mode docs/policy.md calls out an unregistered key silently evaluating to null rather than erroring. Happy to go further toward full evaluation with supplied reference data if that still seems more valuable, let me know.

AnkushUjawane:
Just following up in case this fell off the radar — I've addressed boththe failure-mode correction and the reference-value point since the last comments. Happy to make further changes if either of you get a chance to take another look. No rush if you're busy, just wanted to surface it.

fitzthum:
Yeah, if anything I would look into the admin endpoints where we add policies. They could potentially do more in-depth validation at that point.

AnkushUjawane:
That make sense and would mean the check happens exactly where it matters most, without relying on someone remembering to run a separate tool first.
before I am going build that: would this replace the CLI-side validation entirely, or complement it (e.g. CLI for local dev/CI before you even have a running AS, server-side as the final authoritative check)? And should it be a hard rejection on set_policy, or a warning that still allows setting an unusual policy?
Happy to go whichever direction you'd prefer, I just want to build the right thing once rather than guess.

AnkushUjawane:
Hey I want to build that, would you give me direction that you prefer

fitzthum:
server-side not client side. otherwise up to you

AnkushUjawane:
Opened #1567 with the server side validation on set_policy, per direction above.

mkulke (approved):

fitzthum:
This one is supplanted by https://github.com/confidential-containers/trustee/pull/1567?

AnkushUjawane:
Yes, #1567  covers this.

fitzthum:
Thanks

PR #1531, opened 2026-07-31
Nvidia Verifier: simplify the TEE claims and finish the claim documents

Xynnn007:
Currently some of the Nvidia verifier logic will postpone some check to AS policy, like certchain endorsement, signature verification, ... These work should be asserted.

This patch moves that parts into code logic, and extract those "variable" fields into parsed claims for AS policy to do filter.

Also, do some code refactoring and document complement work.


deps/verifier/src/nvidia/nras_response.rs
```diff
@@ -5,14 +5,471 @@
...
+    pub(crate) x_nvidia_gpu_switch_pdis: Option<Vec<String>>,
+}
+
+impl NrasIndividualClaims {
+    /// Verify individual GPU claims following the rules:
+    ///
+    /// - Present fields are validated
+    /// - Missing fields are logged at trace level and
+    /// skipped (they are not treated as failures).
+    pub fn verify(&self) -> Result<()> {
```
fitzthum: I think this should stay in the policy.

For one thing, this is too strict. It's fairly common for some of these fields to fail if you do not have the most recent firmware. We don't want the verifier itself to fail in that case. It's a policy decision if such a device can be used.

It also feels too brittle to hard code all of these claims into the source.

Finally, ik you discussed having a more firm API between verifiers and policy, but I don't agree with that. I think it's a good thing that we don't really define what a verifier does vs what the policy does. Implementers are free to do whatever they find most ergonomic. I see no reason to try formalizing that more.

fitzthum (commented):
First commits looks fine but I don't like moving everything out of the policy.

deps/verifier/src/nvidia/nras_response.rs
Xynnn007: We can add some options to the nvidia verifier to control the concrete behavior. But basically and at least, the signature, cert chain and the endorsement verification should be enforced. This is the verifier's responsibility.

I read the nvidia official document throughly, and get all the cert chain things into the logic. All other things that cannot be determined are brought into the policy. Note that the original default gpu policy defines to assert the endorsement chain. So I remove some of the claims of endorsements and put them into code logic.

I am not a master of GPU attestation, so I need your help. To throughly examine all the fields again, and give concrete reason for each field why it can be postponed to policy and what's the situation that they do not need to follow and what's the result.

Overall, if the **endorsement chain does not need to be cared** (please ensure what we are doing, it's endorsement chain. Ignorance means fake device can be alive), we can bring in a verifier side `mode` option (defaults to `strict`) to control whether to check the cert chain in code, or only check part of the fields, like `x-nvidia-overall-att-result` (may be `TrustNrasOverall` mode).

mythi:
> First commits looks fine but I don't like moving everything out of the policy.

FWIW I agree.

Do we know how `x-nvidia-overall-att-result` is defined?

Xynnn007:
> Do we know how `x-nvidia-overall-att-result` is defined?

Due to https://docs.nvidia.com/attestation/nv-attestation-sdk-cpp/latest/sdk-c/configuration.html

'''
    The following rego policy (part of the evidence policy) is used to evaluate the overall attestation result
'''

mythi:
I see this case somewhat similar to what DCAP QVL is doing: some terminal errors that are hard verification failures and then "supplemental data" and added claims for policies. So my thinking was whether the overall att result =false could be treated like that?

deps/verifier/src/nvidia/nras_response.rs
fitzthum: There are two problems here.

First, checks in the policy are non-fatal. Checks in the verifier are usually fatal. Since it is important to check the cert chain, it might seem like you would want to do this in the verifier, but this causes problems. If you do not have up-to-date firmware on your GPU, the check for the cert chain of the RIM (reference values) will fail. It's too brittle for this to create a fatal error. That is a big pain in development and testing environments. It is more practical to do this in the policy. That way, you can skip the check by changing the KBS policy.

There is something more subtle here btw. On most platforms we check the cert chain in the verifier. For NVIDIA/NRAS, there are two cert chains. One of them covers the platform itself via evidence signing. The other one covers the RIMs. We shouldn't be thinking about these two cert chains in the same way. I would entertain checking the platform cert chain in the verifier but the rim cert chain in the policy, but I think doing both in the policy is best.

The second factor, is that doing these checks in the verifier collapses the attestation result into a binary, which we should avoid. Notice in the current gpu policy we use these fields to calculate different parts of the trust vector. We should not lose that granularity.

Again, I don't think we should be too formal with the API between the verifier and the policy. It is a good thing that different verifiers can use this in different ways.

fitzthum:
The relying party policy in NVAT is not as robust as our existing policy mechanism. More significantly, we don't want to collapse the result into a binary. See above comment.

deps/verifier/src/nvidia/nras_response.rs
Xynnn007: > I would entertain checking the platform cert chain in the verifier but the rim cert chain in the policy, but I think doing both in the policy is best.

What's the reason for you to combine platform cert chain into policy also? Imo, this is the point I want to show in the PR. RIM things (covers firmware software) can somehow be postponed to AS policy, but the platform cert chain must be enforced in the verifier logic to assert the report and hardware is endorsed by Nvidia. Policy failures are fatal, but policy omissions are not. A fake hardware will not explicitly show any error unless KBS side policy explicitly shows it.

> The second factor, is that doing these checks in the verifier collapses the attestation result into a binary, which we should avoid. Notice in the current gpu policy we use these fields to calculate different parts of the trust vector. We should not lose that granularity.

I am not trying to turn it into a binary, but to simplify "what policy should check". The claims returned by NVIDIA are very detailed now - a situation that brings a mix of pleasure and frustration. We do not need to cover every single one of them within the policy; items that clearly require assertion or can be aggregated can be handled by the policy in a streamlined manner. Some thoughts from my side (partially showed in the PR). Now Nvidia GPU has following claims types, and nvswitch has similar.

1. GPU report endorsement, including `x-nvidia-gpu-attestation-report-parsed`, `x-nvidia-gpu-attestation-report-signature-verified`, `x-nvidia-gpu-attestation-report-cert-chain`, `x-nvidia-gpu-attestation-report-cert-chain-fwid-match`. They altogether shows "this evidence is from a legal nvidia hardware".
2. Driver endorsements, including `x-nvidia-gpu-driver-rim-fetched`, `x-nvidia-gpu-driver-rim-schema-validated`, `x-nvidia-gpu-driver-rim-signature-verified`, `x-nvidia-gpu-driver-rim-version-match`, `x-nvidia-gpu-driver-rim-measurements-available` and `x-nvidia-gpu-driver-rim-cert-chain`. They altogether shows if the driver is valid.
3. vbios endorsements, including `x-nvidia-gpu-vbios-rim-fetched`, `x-nvidia-gpu-vbios-rim-schema-validated`, `x-nvidia-gpu-vbios-rim-signature-verified`, `x-nvidia-gpu-vbios-rim-version-match`, `x-nvidia-gpu-vbios-rim-measurements-available`, `x-nvidia-gpu-vbios-rim-cert-chain` and `x-nvidia-gpu-vbios-index-no-conflict`. They altogether shows if vbios is valid. I guess the firmware issue happens here?
4. GPU runtime measurements: `measres`.
5. gpu arch check: `x-nvidia-gpu-arch-check`
6. an overall judgement from NRAS `x-nvidia-overall-att-result`
7. variable fields, e.g. `hwmodel`, `ueid`, `oemid`, `x-nvidia-gpu-driver-version`, `x-nvidia-gpu-vbios-version`

 Verifier should include part 1 and 5 into code logic. For other parts, verifier provides two modes via verifier config:
1. `strict` (default). check 2, 3, 4, 6 in code logic.
2. `debug`/ or `compatible`. check only 6 incode.

both mode return 7. also 1\~6 can be included into the claim, but in a simplified way. Note that 1~6 are all boolean values, so emit logs in code to show each boolean to help debug.

Also, note that current policy rule matching also needs to be updated, e.g. `vbios` things should be classified into `hardware`. But the proposal will get rid of vbios things from the policy, thus this fix is not needed in the scenario.

fitzthum: > What's the reason for you to combine platform cert chain into policy also? 

I am open to this. The downside is that it is maybe a little more confusing for users to have the checks in different places. I think it's kind of nice that today you can glance at the policy and see all of the logic.

For your list, I agree that 1 and 5 are the ones that would be the best fit for the verifier (although I still probably wouldn't move them, but I'm open to it). I don't really like having two modes. That seems like it will be more confusing. Ultimately that's really serving the same purpose of setting the required trust level in the KBS policy.

Btw vbios should not be part of hw imo.

Xynnn007: re: 2 modes. 

I see your concerns for two modes. To make it more clear, this is not related to KBS policy for two modes way. If users want to use an old version of GPU, they need to set `mode = "debug"` in verifier config. Note that by default it's set to `mode = "strict"`. Our document will explain in this way there is no drivers/vbios endorsement/validity check because NVIDIA only has valid RIMs/endorsements for certain versions of firmware. If `mode` is still `strict`, the verifier logic will fail and no token will be provisioned. Note that no claims of vbios/drivers/... validation things will be included in the claims in both modes, but the version info and mode info will be included (s.t. 7).

A compromised way w/o two modes, would be to combine claims, s.t. we will have `vbios_ok`, `driver_ok` as claims in policy. But the concrete reason like `x-nvidia-gpu-vbios-rim-measurements-available` can be printed in log. This would both help user to have easier understanding for policy and debugging. The default policy can require all the parts to be `true`. If a user use some old GPUs, he still need to change the policy, to explicitly allow `vbios_ok/...` to be false. He still needs to read the documents to understand Nvidia has a very strict strategy to update the RIMs and endorsements, we must ignore them for some specific firmware versions. In this way, always there will be one token to be provisioned.

Both ways need extra user comprehension work (which cannot be avoidable) and need to change something upon the AS deployment (verifier config vs policy). My preference is still the verifier-config approach, because it makes the security posture explicit at the verifier layer and harder to bypass accidentally. If you strongly prefer avoiding modes, I can go with the summary-claims approach with a strict default policy. Which direction do you prefer?

Re: vbios as hardware. https://www.ietf.org/archive/id/draft-ietf-rats-ar4si-10.html#section-2.3.4 groups firmware with hardware in the hardware trustworthiness claim. I think vbios is gpu firmware thus falling into the  model.

fitzthum: We should not be introducing extra flags to the verifier config that fulfill the same purpose as the policy. 

I also don't think it makes much sense to calculate things like `vbios_ok` or `driver_ok` in the verifier. This doesn't improve anything. The existing policy logic concisely reflects what needs to be checked. This is exactly the kind of thing that OPA is good for. Why would we use rust to calculate some intermediate values Furthermore, we already propagate any errors from NRAS into the TCB Claims and the attestation token, which is logged. It's easy to see if your device is not affirming and then find the error from NRAS.

> Both ways need extra user comprehension work (which cannot be avoidable) and need to change something upon the AS deployment (verifier config vs policy).

The current approach doesn't require users to change anything. For testing most people use the allow-all kbs policy. This allows old GPUs. When they go to the affirming level, they have to update their firmware. There is no need to change the attestation policy at all. 

Honestly, I don't see any of the changes to the policy here as improvements. There are a lot of other improvements that we can focus on here, but the policy is fine imo.

Xynnn007: I got the point that you mentioned multiple times - for testing and development, it's required to run e2e successfully with AS & KBS, even using a GPU with old firmware. When production, the KBS policy can be affirming and then update GPU firmware. This totally makes sense.

But I still hold the point that the verifier has responsibility to own whether the NRAS evidence is trustworthy at all. Things like RIM endorsement, OCSP status, or cert-chain validation are just how the verifier interprets the raw NVIDIA attestation response. Requiring AS policy to enumerate RIM failures is a **leaky abstraction**. The policy should only need to know whether the NVIDIA hardware, firmware, and driver are endorsed; Users shouldn't have to understand the internal mechanics/terms of RIM, OCSP, vbios, or cert-chain validation to make that decision.

What I'd like to achieve is: the verifier by default checks signature, cert chain, nonce, and overall attestation result. Optionally, for older hardware, a verifier config can relax the more volatile endorsement checks (driver/vbios RIM, etc.) without requiring AS policy to understand those details. The policy then only sees a small, stable set of claims about the software versions, etc.

When the mode is `strict`, error information will be detailed printed in log to show why verifier fails, thus can greatly help with debug/dev scenario. With `debug` mode, only helps to skip the things except gpu evidence endorsement to be checked and can quickly get the e2e running work.

Xynnn007:
> I see this case somewhat similar to what DCAP QVL is doing: some terminal errors that are hard verification failures and then "supplemental data" and added claims for policies. So my thinking was whether the overall att result =false could be treated like that?

[QVL claims](https://github.com/confidential-containers/trustee/blob/main/attestation-service/docs/tcb_claims.md#intel-tdx) like `platform_instance_id`, `root_ca_crl_num` things are about the endorsement materials. `tcb_status` and `tcb_date` things are an overall view of the endorsement check.

But for nvidia gpu [claims](https://docs.nvidia.com/attestation/advanced-documentation/latest/claims-guide/gpu_claims.html) returned by NRAS, most of them are like `x-nvidia-gpu-vbios-rim-schema-validated`, `x-nvidia-gpu-driver-rim-signature-verified`, `x-nvidia-gpu-attestation-report-parsed` - they are like a code path to show where the error happens. As a user, I might only want to know whether the current evidence endorsement and driver status are valid; I don't necessarily need to know the specific details—such as whether a network access error occurred or if the retrieved format was incorrect—as having that information logged is sufficient. Equivalently, the claims issued by the QVL won't specify boolean flags for whether the TDX quote failed to parse or whether the collateral could not be retrieved.

So there is a [discussion](https://github.com/confidential-containers/trustee/pull/1531#discussion_r3690746837) for this. Hope that you can give some suggestions and ideas based on the dialogue.

PR #1541, opened 2026-08-04
helm: add native KBS TLS support

### LINKED ISSUE #1532: Helm Chart | Add KBS TLS support
The Trustee Helm chart currently configures the KBS listener with `insecure_http = true` and does not provide a supported way to supply the KBS endpoint certificate and private key.

I would love to change the confidentialcontainers.org hosted example [NVIDIA confidential NIM deployment](https://confidentialcontainers.org/docs/examples/nvidia-nim-confidential-gpu-attestation/), which configures an HTTPS KBS URL, generates endpoint identity material, and sets `insecure_http = false`, to adopt using the Helm chart.

I have created a feature branch here, but I am not too familiar with this repository: https://github.com/confidential-containers/trustee/compare/main...manuelh-dev:trustee:mahuber/helm-native-kbs-tls?expand=1

My kind ask would be to assess whether you agree it is worth doing this, and to see if my feature branch mostly does the right thing - if I am on the right track, I would be happy to open a PR, but I am of course also happy for anyone else to open a PR.

manuelh-dev:
Allow deployments to opt into KBS HTTPS with a pre-created Kubernetes TLS Secret. Keep insecure_http=true as the default so
existing chart deployments remain unchanged.

Mount a Kubernetes TLS Secret and configure KBS to read the standard tls.crt and tls.key files. Switch health probes to HTTPS when native TLS is enabled, and exercise the HTTPS path in the Helm end-to-end test.

Fixes: #1532 

If we have this capability, I would like to add deployment with the helm chart to `Checkpoint 2: Deploy Trustee and Install kbs-client` in https://confidentialcontainers.org/docs/examples/nvidia-nim-confidential-gpu-attestation/

### DIFF
--- a/deployment/helm-chart/README.md
+++ b/deployment/helm-chart/README.md
@@ -153,6 +153,39 @@ helm upgrade --install trustee ./deployment/helm-chart \
 
 Or use **`scenarios/bring-your-own-keys.yaml`** (adjust Secret name / file paths in the comments there).
 
+### Native KBS HTTPS
+
+The default KBS listener uses plaintext HTTP. For a KBS endpoint that clients or
+confidential guests reach directly, enable native HTTPS and provide an existing
+Secret containing the endpoint private key and certificate chain. The chart does
+not generate this identity material because its certificate SAN must match the
+address used by KBS clients.
+
+```bash
+kubectl create namespace coco-trustee
+kubectl create secret tls trustee-kbs-tls \
+  --namespace coco-trustee \
+  --key ./kbs.key \
+  --cert ./kbs.crt
+
+helm upgrade --install trustee ./deployment/helm-chart \
+  --namespace coco-trustee \
+  -f ./deployment/helm-chart/scenarios/native-tls.yaml
+```
+
+With `kbs.tls.enabled=true`, the chart leaves `insecure_http` at its secure
+default (`false`), mounts the standard `tls.key` and `tls.crt` data keys from the
+selected Kubernetes TLS Secret, and changes the KBS health probes to HTTPS.
+
+KBS does not reload endpoint identity material dynamically, so restart the KBS
+Deployment after replacing the TLS Secret contents.
+
+Ingress TLS termination is a separate mode: leave native KBS TLS disabled and
+configure `ingress.tls` when the Ingress controller should serve HTTPS and
+forward plaintext HTTP to KBS. Do not combine native KBS TLS with Ingress unless
+the chosen Ingress controller is explicitly configured to use HTTPS for its
+backend connection.
+
 ### IBM Secure Execution (s390x)
 
 On **s390x**, the **IBM Secure Execution (SE)** verifier needs attestation materials at runtime. Because KBS talks to a **remote `coco_as_grpc` AS**, the verifier runs inside the **AS Pod**, so these materials must be mounted on **AS**, not KBS. (This differs from the builtin-AS kustomize overlay in `kbs/config/kubernetes/overlays/ibm-se`, which mounts them on KBS.)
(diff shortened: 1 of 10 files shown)


deployment/helm-chart/templates/_helpers.tpl
```diff
@@ -405,6 +405,19 @@ KBS admin JWT settings (bootstrap token + kbs-config.toml).
...
+attestation-token keys because it represents the externally visible KBS identity.
+*/}}
+{{- define "coco-trustee.kbs.tlsEnabled" -}}
+{{- $tls := ((.Values.kbs | default dict).tls | default dict) -}}
+{{- if ($tls.enabled | default false) }}true{{- end -}}
+{{- end }}
+{{- define "coco-trustee.kbs.tlsSecretName" -}}
+{{- $tls := ((.Values.kbs | default dict).tls | default dict) -}}
+{{- required "kbs.tls.existingSecretName must be set when kbs.tls.enabled is true" (trim (default "" $tls.existingSecretName)) -}}
+{{- end }}
```
Xynnn007, line 522: `coco-trustee.kbs.tlsSecretName` does not check `kbs.tls.enabled`; it relies on callers to only invoke it when TLS is enabled. That works for the current deployment template. Not a blocker, but if you think it an issue, can add some detect logic here also

deployment/helm-chart/README.md
```diff
@@ -109,6 +109,43 @@ helm upgrade --install trustee ./deployment/helm-chart \
...
+```
+
+With `kbs.tls.enabled=true`, the chart configures KBS with
+`insecure_http=false`, mounts the standard `tls.key` and `tls.crt` data keys
+from the selected Kubernetes TLS Secret, and changes the KBS health probes to
+HTTPS.
+
+Native KBS HTTPS is independent of the admin and attestation-token keys
+described in BYOK above. Use separate Secrets because these identities have
+different purposes, owners, and rotation policies. KBS does not reload endpoint
```
Xynnn007: we might need to mention the difference with attestation token keys? Just mention KBS does not reload .... here is ok?

Xynnn007 (approved):
LGTM. Thanks! @manuel-dev only some nits

Xynnn007:
TDX and non-tee helm test passed. SNP test failed because of a known issue not related to this PR.

deployment/helm-chart/templates/_helpers.tpl
manuelh-dev: @Xynnn007  updated via: https://github.com/confidential-containers/trustee/compare/8ff1aa54997b354520a0efaf23359f22cfd37bab..d330d2a1a59e64e40c70b8f6edd55f88876e0d21

deployment/helm-chart/README.md
manuelh-dev: updated - is https://github.com/confidential-containers/trustee/compare/8ff1aa54997b354520a0efaf23359f22cfd37bab..d330d2a1a59e64e40c70b8f6edd55f88876e0d21 - what you were looking for?

Xynnn007: Ah sorry. I mean we might **not** need but forgot the `not` : (

The reading feeling is that we can focus on the current topic, no need to mention other parts

deployment/helm-chart/templates/kbs-deployment.yaml
```diff
@@ -126,6 +137,16 @@ spec:
...
 {{- include "coco-trustee.userKeysSecretVolumeItems" . | nindent 12 }}
+        {{- if $tlsEnabled }}
+        - name: kbs-tls
+          secret:
+            secretName: {{ include "coco-trustee.kbs.tlsSecretName" . }}
+            items:
+              - key: tls.key
+                path: tls.key
+              - key: tls.crt
+                path: tls.crt
```
mythi: could be simplified by dropping `items`

deployment/helm-chart/files/kbs-config.toml.template
```diff
@@ -2,9 +2,14 @@
 session_storage_type = "{{ include "coco-trustee.kbs.sessionStorageType" . }}"
 {{- end }}
 
+{{- $tlsEnabled := eq (include "coco-trustee.kbs.tlsEnabled" . | trim) "true" }}
 [http_server]
 sockets = ["0.0.0.0:{{ include "coco-trustee.port.kbs" . }}"]
-insecure_http = true
+insecure_http = {{ not $tlsEnabled }}
+{{- if $tlsEnabled }}
```
mythi, line 8: `insecure_http` is an opt-in and false by default so perhaps add `insecure_http = true` only in the else branch.

deployment/helm-chart/values.yaml
```diff
@@ -103,6 +103,11 @@ kbs:
       poolSize: 200
       # -- Request timeout in seconds for the KBS -> gRPC AS client (`timeout` in `files/kbs-config.toml.template`).
       timeout: 30
+  tls:
+    # -- Enable native HTTPS on the KBS listener. Requires `existingSecretName`; the chart does not generate endpoint identity material.
+    enabled: false
+    # -- Existing Secret containing the KBS HTTPS private key and certificate chain. Required when `enabled=true`.
+    existingSecretName: ""
```
mythi:
```suggestion
    secretName: ""
```

"existing" might not always be case for instance if we look for adding `cert-manager` support later so perhaps simplify the value name a bit?

deployment/helm-chart/README.md
manuelh-dev: Updated - and yes, now I get it, this now totally makes sense ;)

deployment/helm-chart/files/kbs-config.toml.template
manuelh-dev: @mythi - thanks for the comments. updated all 3 items accordingly

Xynnn007 (approved):

mythi:
LGTM but needs rebase

PR #1542, opened 2026-08-04
Add instance identifiers to TCB Claiims

fitzthum:
In some cases, admin may want to ensure that evidence comes from a particular guest. Surface the platform instance id and chip id to enable this. Actually, we were already reporting the platform instance id for tdx, but it wasn't in the tcb_claims doc.

I think you could probably do something similar on SNP already by pre-provisioning VCEKs only for certain hosts. Nonetheless, this approach is a bit more explicit.

For now, don't update the attestation policy. If this use case becomes more salient we can update the attestation policy and add a reference value for this.

### DIFF
--- a/attestation-service/docs/tcb_claims.md
+++ b/attestation-service/docs/tcb_claims.md
@@ -225,6 +225,7 @@ along with the full set of detached claims.
 
 ## AMD SEV-SNP
 
+- `snp.chip_id`: Unique identifier for platform instance (field may be zero if masked)
 - `snp.measurement` Launch Digest covering initial guest memory
 - `snp.platform_smt_enabled`:  Whether Simultaneous Multithreading is enabled on the system
 - `snp.platform_tsme_enabled`: Whether Transparent SME is enabled on the system
--- a/deps/verifier/src/snp/mod.rs
+++ b/deps/verifier/src/snp/mod.rs
@@ -655,6 +655,9 @@ pub(crate) fn parse_tee_evidence(report: &AttestationReport) -> TeeEvidenceParse
         "measurement": hex::encode(report.measurement),
         "report_data": hex::encode(report.report_data),
         "init_data": hex::encode(report.host_data),
+
+        // platform identity
+        "chip_id": hex::encode(report.chip_id),
     });
 
     claims_map as TeeEvidenceParsedClaim
@@ -1066,10 +1069,11 @@ mod tests {
         // Generate parsed claims
         let claims = parse_tee_evidence(&attestation_report);
 
-        // Extract the three key fields that should be hex-encoded
+        // Extract the key fields that should be hex-encoded
         let measurement = claims.get("measurement").and_then(|v| v.as_str()).unwrap();
         let report_data = claims.get("report_data").and_then(|v| v.as_str()).unwrap();
         let init_data = claims.get("init_data").and_then(|v| v.as_str()).unwrap();
+        let chip_id = claims.get("chip_id").and_then(|v| v.as_str()).unwrap();
 
         // Verify they are valid hex strings by checking:
         // 1. All characters are valid hex digits (0-9, a-f)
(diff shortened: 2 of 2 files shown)


attestation-service/docs/tcb_claims.md
```diff
@@ -225,6 +225,7 @@ along with the full set of detached claims.
 
 ## AMD SEV-SNP
 
+- `snp.chip_id`: Unique identifier for platform instance
```
Xynnn007: Note that if `mask_chip_key` is `true`, this field will be zero

attestation-service/docs/tcb_claims.md
```diff
@@ -122,6 +122,7 @@ The following fields always exist.
 - `tdx.sgx_type`: The type of memory used in SGX. Can be one of (`Standard`, `Scalable`, `Scalable with Integrity`).
 - `tdx.tcb_date`: Date time value in RFC3339 format - Earliest date between tcbInfo and qeIdentity.
 - `tdx.tcb_eval_num`: Indication of the freshness of the reference values used.
+- `tdx.platform_instance_id`: Uniquely identifies the platform instance (optional)
```
Xynnn007: This field is coupled with `is_platform_ca`. If `is_platform_ca` is true, this field is not null.

I guess we need to tidy the claims documents with more details, like conditions and types.

Xynnn007 (commented):
For SNP parts, there is overlap work in https://github.com/confidential-containers/trustee/pull/1358/changes also. Without blockers, I'd suggest to consider #1358 as it provides more information about the fields and dependencies.

For TDX, as comment left. We need some similar work also.

attestation-service/docs/tcb_claims.md
fitzthum: Yeah I'm not sure how much detail we want to have in each of these fields. We also may want to add this to the website at some point.

fitzthum:
#1358 is still kind of blocked. Also, that has "breaking" changes whereas this just adds a field. I was looking at https://github.com/edgelesssys/contrast/security/advisories/GHSA-hjgc-jc5v-fw7h and figured that we should supply these fields.

Xynnn007:
> #1358 is still kind of blocked. Also, that has "breaking" changes whereas this just adds a field. I was looking at [GHSA-hjgc-jc5v-fw7h](https://github.com/edgelesssys/contrast/security/advisories/GHSA-hjgc-jc5v-fw7h) and figured that we should supply these fields.

It's a misunderstanding that #1358 brings in breaking changes - it's a add-only PR.

I got your point to migrate the CVE in a way of providing explicit way to bind the evidence with hardware same as constellation. I can give an approval for this one for security issue, but we **SHOULD**
1. pushing #1358 forward, as SNP claim fields have dependency which will be **clearly** declared in the PR
2. put a TDX fixup PR to update documents about the claims to make TDX claims better-documented

Xynnn007 (approved):

fitzthum:
> It's a misunderstanding that https://github.com/confidential-containers/trustee/pull/1358 brings in breaking changes - it's a add-only PR.

It changes the SNP claims structure. It puts the claims inside an `evidence` object, so the policy has to [change](https://github.com/confidential-containers/trustee/pull/1358/changes#diff-ca0c7248f27045f03ad32680a6ad890f59e252d6acba8f3aa2c34f2f72841087R113). Anyway, I guess we're still waiting for AMD on that one.

Let me drop the TDX doc change from this PR and we can use your PR for that. We will just do the SNP change here.

attestation-service/docs/tcb_claims.md
fitzthum: Added a note.

Xynnn007:
> It changes the SNP claims structure. It puts the claims inside an `evidence` object, so the policy has to [change](https://github.com/confidential-containers/trustee/pull/1358/changes#diff-ca0c7248f27045f03ad32680a6ad890f59e252d6acba8f3aa2c34f2f72841087R113). Anyway, I guess we're still waiting for AMD on that one.

It still [keeps the previous structure](https://github.com/confidential-containers/trustee/pull/1358/changes#diff-f396654b0e3e0b311c4049fdc3e98825b3f05dc44ecd46e777cfad2a4905e5beR273) but I'd like to totally change the format.

PR #1548, opened 2026-08-10
toolchain: update rust from 1.93.0 to 1.95.0

Xynnn007:
This commit also includes some lint fixes.

### DIFF
--- a/deps/verifier/src/nvidia/spdm_response.rs
+++ b/deps/verifier/src/nvidia/spdm_response.rs
@@ -311,12 +311,10 @@ impl OpaqueData {
             bail!("OpaqueDataType is not multiple of {}", MSR_COUNT_SIZE)
         }
         let mut values: Map<String, Value> = Map::new();
-        let mut index: u8 = 1;
-        for chunk in bytes.chunks(MSR_COUNT_SIZE) {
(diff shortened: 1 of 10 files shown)


deps/verifier/src/nvidia/spdm_response.rs
```diff
@@ -311,12 +311,10 @@ impl OpaqueData {
             bail!("OpaqueDataType is not multiple of {}", MSR_COUNT_SIZE)
         }
         let mut values: Map<String, Value> = Map::new();
-        let mut index: u8 = 1;
-        for chunk in bytes.chunks(MSR_COUNT_SIZE) {
+        for (index, chunk) in (1_u8..).zip(bytes.chunks(MSR_COUNT_SIZE)) {
```
fitzthum, line 314: Is this mandatory with new rust? I don't really love the style tbh

fitzthum (approved):
LGTM 

one nit

deps/verifier/src/nvidia/spdm_response.rs
Xynnn007: It's warned and suggested by clippy. The clippy error is to try to avoid defining counter but to use the loop itself, by using `zip`ping the counter range with chunks.

PR #1551, opened 2026-08-10
verifier: add local NVIDIA Blackwell GPU attestation

chalabi2:
## Why

Trustee can send Blackwell evidence to NRAS, but the local verifier rejects it.
That makes Blackwell attestation depend on NVIDIA's remote service.

Blackwell uses the same report signature format and NVIDIA device root as
Hopper. The difference that matters here is the certificate FWID: Blackwell
stores it in `fwids[0]` of a TCG DICE `DiceTcbInfoAlias` extension.

## Changes

- Accept Hopper and Blackwell in the local verifier. LS10 remains unsupported
  locally.
- Keep Hopper's existing FWID extraction and parse `fwids[0]` for Blackwell.
- Expose the verified leaf certificate serial as `ueid` and reject a repeated
  device report, even if its guest-supplied UUID changes.
- Decode NVIDIA's NVDEC status bytes as `enabled` or `disabled` instead of
  switching between UTF-8 and hex representations.
- Reject truncated or unknown final opaque-data items in a separate parser
  hardening commit.
- Add a `verifier` claim to identify local and remote NVIDIA verification, and
  use it to select the matching GPU policy rules.
- Evaluate local GPU claims against generic RVPS values.
- Add a report and certificate-chain fixture captured from an RTX PRO 6000.
  No RIM data is included.

## Validation

The captured report and chain passed NVIDIA's nvtrust local verifier
(`2026.06.04.001`) and NVAT `1.2.2`. Both returned the same FWID and 64
measurements as Trustee.

On Rust 1.95.0:

- NVIDIA verifier tests: 31 passed; 2 live NRAS tests ignored.
- Attestation Service tests with `nvidia-verifier`: 16 passed.
- Verifier and Attestation Service clippy checks passed with warnings denied.
- `cargo fmt --all -- --check` passed.
- OPA 1.19.0 accepted the GPU policy with the repository capabilities file.

## Scope

The hardware fixture covers RTX PRO 6000 only. B200, NVSwitch, and the
destructive CC-mode transition from nvtrust issue 150 have not been tested.
RIM fetching, OCSP, and automatic reference-value provisioning are not part of
this change.

The fixture contains public device certificates and a report for a fixed
nonce. It contains no private key, workload secret, host address, or captured
device UUID, and it cannot answer a fresh challenge.


attestation-service/tests/fixtures/nvidia/blackwell_reference_values.json
fitzthum, line 1: Keep in mind that licensing prevents copying RIMs from the NVIDIA rim service into this repo. So you might want to remove this or clarify where it came from.

attestation-service/src/ear_token/ear_default_policy_gpu.rego
```diff
@@ -33,6 +33,73 @@ trust_claims := {
 	"sourced-data": sourced_data,
 }
 
+# Locally verified Blackwell evidence. The Rust verifier has already checked
+# the report signature, certificate chain, FWID binding, and nonce before it
+# emits these claims.
+hardware := 2 if {
+	input.nvidia.arch == "Blackwell"
```
fitzthum: Need to make sure that the blocks for targeted for the tcb claims from the local verifier and the blocks targeted for the remote verifier are mutually exclusive and only run when the appropriate mode is being used.

attestation-service/docs/tcb_claims.md
```diff
@@ -201,29 +201,43 @@ CCA claims are grouped into `cca.realm` and `cca.platform`:
...
+- `attestation_key_id`: SHA-256 identifier of the DER SubjectPublicKeyInfo from
+  the certificate that signs the attestation report, formatted as
+  `sha256:<lowercase hex>`. Trustee rejects duplicate identifiers within one
+  NVIDIA evidence list. This identifies the signing key, not a permanent
+  vendor-defined physical GPU identity.
 - `measurements`: List of measurements and its respective index
-- `uuid`: Device UUID
+- `uuid`: Unverified identifier reported by the guest. Trustee retains it for
+  payload compatibility but does not use it for attestation or device
+  uniqueness.
```
fitzthum: We should just drop this claim.

fitzthum (commented):
A few initial comments. btw see our [AI Policy](https://github.com/confidential-containers/confidential-containers/blob/main/AI.md)

chalabi2:
> A few initial comments. btw see our [AI Policy](https://github.com/confidential-containers/confidential-containers/blob/main/AI.md)

Thank you for your response. I will go in an abide by the AI policy where relevant as it was used heavily in the development of this feature.

attestation-service/tests/fixtures/nvidia/blackwell_reference_values.json
chalabi2: Ok I had no idea about that, but, all attestation records / info are from local GPU's. NRAS was not used here.

fitzthum: Try to make that clear in the source somewhere just to avoid any snags.

zvonkok: If the RIMs are extracted from the driver package, this will not be available in the future and will break this flow.

attestation-service/docs/tcb_claims.md
fitzthum: Actually, this claim does seem to be validated at least by NRAS. Double check that it is validated by the local verifier. This comment is incorrect. The UUID is no longer transmitted explicitly by the attester, but it can still be extracted from the evidence.

deps/verifier/src/nvidia/mod.rs
```diff
@@ -310,37 +333,64 @@ impl Nvidia {
...
+        expected_nonce: &[u8],
     ) -> Result<(TeeEvidenceParsedClaim, String)> {
-        // Only Hopper GPU is supported for local verification.
-        if device.arch != Architecture::Hopper {
-            bail!("Device architecture not supported");
+        match device.arch {
+            Architecture::Hopper | Architecture::Blackwell => {}
+            Architecture::LS10 => {
+                bail!("Device architecture LS10 not supported for local verification")
+            }
```
Xynnn007: we can use if here for simplicity

deps/verifier/src/nvidia/report.rs
```diff
@@ -33,14 +34,35 @@ impl fmt::Display for NvidiaAttestationReport {
...
+            Architecture::Hopper,
+        )
+    }
+
+    pub fn try_new_for_architecture(
+        report_bytes: &[u8],
+        report_signature_length: usize,
+        cert_chain_bytes: &[u8],
+        expected_nonce: &[u8],
+        architecture: Architecture,
```
Xynnn007: We can use the original function name `try_new`. This crate is not published thus these changes can be acceptable.

deps/verifier/src/nvidia/report.rs
```diff
@@ -69,6 +91,20 @@ impl NvidiaAttestationReport {
...
     }
 
+    fn validate_report_consumption(
+        report_bytes: &[u8],
+        report_size: usize,
+        architecture: Architecture,
+    ) -> Result<()> {
+        let trailing = report_bytes
+            .get(report_size..)
+            .ok_or_else(|| anyhow!("Nvidia report size exceeds input"))?;
```
Xynnn007: I guess it should be "report_bytes has less length thanreport_size"?

deps/verifier/src/nvidia/report.rs
```diff
@@ -69,6 +91,20 @@ impl NvidiaAttestationReport {
...
+        match (architecture, trailing) {
+            (_, []) | (Architecture::Hopper, [0]) => Ok(()),
+            _ => bail!("Unexpected trailing data in NVIDIA report"),
+        }
```
Xynnn007: Just out of curiosity: where can we find this behavior / definition? This behavior seems very detailed

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -80,11 +104,27 @@ impl NvidiaCertificateChain {
...
     /// way up to the trusted NVIDIA root CA.
+    /// Verify a Hopper certificate chain.
+    ///
+    /// This is the original public API. Use [`Self::verify_for_architecture`]
+    /// for architecture-specific certificate formats.
     pub fn verify(&self, expected_fwid: &str) -> Result<()> {
+        self.verify_for_architecture(expected_fwid, Architecture::Hopper)
+    }
+
+    pub fn verify_for_architecture(
```
Xynnn007: Feel free to keep the original function with changing the member arguments

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -137,8 +177,24 @@ impl NvidiaCertificateChain {
...
+///
+/// This is the original public API. Use
+/// [`get_fwid_from_cert_for_architecture`] for architecture-specific
+/// certificate formats.
 pub fn get_fwid_from_cert(cert: &X509) -> Result<String> {
-    let fwid_oid = oid!(2.23.133 .5 .4 .1);
+    get_fwid_from_cert_for_architecture(cert, Architecture::Hopper)
+}
+
+pub fn get_fwid_from_cert_for_architecture(
```
Xynnn007: ditto

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -137,8 +177,24 @@ impl NvidiaCertificateChain {
...
+    cert: &X509,
+    architecture: Architecture,
+) -> Result<String> {
+    let fwid_oid = match architecture {
+        Architecture::Hopper => oid!(2.23.133 .5 .4 .1),
+        Architecture::Blackwell => oid!(2.23.133 .5 .4 .1 .1),
```
Xynnn007, line 160: btw, are there any public documents for these OIDs? It would be better to add link as inline notes here

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -150,12 +206,97 @@ pub fn get_fwid_from_cert(cert: &X509) -> Result<String> {
         .ok_or_else(|| anyhow!("FwId oid not found in the NVIDIA signing certificate"))?
         .value;
 
-    // The FwId data is the last 48 bytes
-    let fwid = value
-        .get(value.len() - 48..)
-        .ok_or(anyhow!("Unexpected fwid oid size"))?;
+    match architecture {
+        Architecture::Hopper => {
+            // Preserve the Hopper format: the FwId is the last 48 bytes.
```
Xynnn007: I did not quite understand this. Do you mean that the  fwid extension value has more information than fwid itself. The last 48 bytes (sha384 length) is the fwid?

still, do we have any public documents for this? a link inline would help maintaince.

deps/verifier/src/nvidia/cert_chain.rs
```diff
@@ -150,12 +206,97 @@ pub fn get_fwid_from_cert(cert: &X509) -> Result<String> {
...
+        Architecture::LS10 => unreachable!("LS10 was rejected before extension parsing"),
+    }
+}
+
+/// Parse `fwids[0]` from a TCG DICE `DiceTcbInfo` extension.
+///
+/// Blackwell certificates encode the FWID list as context-specific field `[6]`
+/// using IMPLICIT tagging. Its payload is therefore the DER content of
+/// `FWIDLIST`, a sequence of `FWID ::= SEQUENCE { hashAlg OID, digest OCTET STRING }`.
+fn parse_dice_tcb_info_fwid(value: &[u8]) -> Result<String> {
```
Xynnn007, line 199: This logic is somehow overlapped with what we already have in [Nvidia dpu verifier](https://github.com/confidential-containers/trustee/blob/main/deps/verifier/src/nvidia_dpu/dice.rs#L219).

We will need to combine them together into a common code. But looks like the two parsing logic is a slightly different, thus we need unit tests fixtures for DPU also to make sure the common code works.

The work described by this comment can be postponed in other PRs. I will keep an eye on this.

deps/verifier/src/nvidia/mod.rs
```diff
@@ -403,6 +477,42 @@ mod tests {
...
+
+        verifier
+            .evaluate(
+                evidence,
+                &ReportData::Value(&expected_nonce),
+                &InitDataHash::NotProvided,
+            )
+            .await
+    }
+
```
Xynnn007, line 492: Can we combine these into existing `test_evaluation` function as rstest cases?

deps/verifier/src/nvidia/mod.rs
```diff
@@ -372,19 +425,40 @@ impl Verifier for Nvidia {
...
 
-            all_devices_claims.push(claims);
+            // Only admit an identity after the local verifier or NRAS has
+            // successfully verified the corresponding evidence.
+            let attestation_key_id = device
+                .attestation_key_id()
+                .context("deriving NVIDIA attestation signing key ID")?;
+            if !seen_attestation_keys.insert(attestation_key_id) {
+                bail!("Duplicate NVIDIA device attestation signing key");
+            }
```
Xynnn007: The filter logic here assumes that we MUST not have two evidences in device_evidence_list to have a same leaf cert? I saw a logic in `validate_ppcie` to assume the number of the devices. So this mechanism is to prevent attacks like to reuse a same device evidence multiple times to pass the ppcie detection logic?

Let's add some more reason in the commits, than what we are doing

attestation-service/src/ear_token/broker.rs
```diff
@@ -505,14 +504,312 @@ mod tests {
...
+    #[case::cpu("cpu", 0, "cpu0")]
+    #[case::one_blackwell("gpu", 0, "gpu0")]
+    #[case::first_of_two_blackwell("gpu", 0, "gpu0")]
+    #[case::second_of_two_blackwell("gpu", 1, "gpu1")]
+    #[case::one_hopper("gpu", 0, "gpu0")]
+    #[case::one_ls10("switch", 0, "switch0")]
+    #[case::one_ppcie("ppcie", 0, "ppcie0")]
+    fn test_make_submod_name(#[case] tee_class: &str, #[case] index: u8, #[case] expected: &str) {
+        assert_eq!(make_submod_name(tee_class, index), expected);
+    }
```
Xynnn007: I'd prefer to avoid defining the new function `make_submod_name`, and just use that sentence in-code. BTW, the code refactoring is not related to the commit message very much. We'd better to add a separate commit with enough reason for the refactoring.

attestation-service/src/ear_token/broker.rs
```diff
@@ -505,14 +504,312 @@ mod tests {
...
+            evaluate_blackwell_policy(&policy_engine, claims.clone(), reference_values.clone())
+                .await;
+        assert_ne!(trust_claims["configuration"], 2);
+
+        claims["config"]["driver_version"] = json!("595.58.03");
+        claims["measurements"]["1"] = json!(7);
+        let trust_claims =
+            evaluate_blackwell_policy(&policy_engine, claims, reference_values).await;
+        assert_ne!(trust_claims["executables"], 3);
+    }
```
Xynnn007: Look like these unit tests are to test the regorus black box, rather than `EarAttestationTokenBroker`.

Usually the unit tests covers the API definitions in the same mod.

It brings in a moked `get_reference_value` extension to play the role as API for RVPS.

My idea is to either remove these test. Or, put them into `policy-engine` crate.

attestation-service/src/ear_token/broker.rs
```diff
@@ -563,8 +860,10 @@ mod tests {
             cert_path: None,
         };
 
-        let mut config = EarTokenConfiguration::default();
-        config.signer = Some(signer);
+        let config = EarTokenConfiguration {
+            signer: Some(signer),
+            ..Default::default()
+        };
```
Xynnn007, line 564: Still. These changes are not related to the commit. Maybe we can move them into a separate commit

attestation-service/src/ear_token/broker.rs
```diff
@@ -638,8 +937,10 @@ mod tests {
             cert_path: Some("tests/fixtures/jwk/cert-chain.pem".to_string()),
         };
 
-        let mut config = EarTokenConfiguration::default();
-        config.signer = Some(signer);
+        let config = EarTokenConfiguration {
+            signer: Some(signer),
+            ..Default::default()
+        };
```
Xynnn007, line 641: ditto

attestation-service/src/ear_token/ear_default_policy_gpu.rego
```diff
@@ -33,6 +33,73 @@ trust_claims := {
...
+		valid_measurement_reference(expected)
+		measurement_matches(input.nvidia.measurements[index], expected)
+	}
+}
+
+executables := 3 if {
+	input.nvidia.arch == "Blackwell"
+	reference_measurements_match("nvidia_blackwell_driver_measurements")
+	reference_measurements_match("nvidia_blackwell_vbios_measurements")
+}
```
Xynnn007, line 108: Note that if we put these into the default policy, we are assuming
1. all users should a same format of reference value in RVPS in advance for blackwell
2. the default policy will be large. As we can forecast that for hopper and future other devices, we will also have a bunch of logic like this.

I hold the point that we do not need these things to be included in the default policy. The default policy should keep simple, and we provide ways for users to define their own policies. A practical way from my side is to have a separate directory to give some example policies and together with documents.

Then, when we finish the policy setting/selecting/simplication work, a detailed official document would help users to define policies based on their needs.

Back to this PR, my suggestion is
1. create a dir `attestation-service/examples`
2. put the gpu policy, with specifical blackwell parts into the examples dir
3. add a `attestation-service/examples/README.md` to show that these are examples of policies, and the policy in step 2 is for blackwell. Togetherly, you can mention that to use the policy, there should be some setting for RVPS

attestation-service/docs/policy.md
```diff
@@ -73,11 +73,30 @@ For EAR token issuance, policy lookup is done by appending the device class to t
...
 
 - `default_cpu` for CPU-class evidence
-- `default_gpu` for GPU-class evidence
-- `default_switch` for switch-class evidence
-- `default_ppcie` for ppcie-class evidence
-
-AS ships these default policies: `default_cpu`, `default_gpu`, `default_switch`, and `default_ppcie`.
+- `default_gpu` for NVIDIA Hopper and Blackwell GPU evidence
+- `default_switch` for NVIDIA LS10 switch evidence
+- `default_ppcie` for NVIDIA protected PCIe topology evidence
```
Xynnn007: let's not change these. Currently they are general conceptions not only to nvidia. In future we will try to get rid of tee_class prefix-ed sub policies.

attestation-service/docs/policy.md
```diff
@@ -73,11 +73,30 @@ For EAR token issuance, policy lookup is done by appending the device class to t
...
+- `nvidia_blackwell_allowed_fwids`
+- `nvidia_blackwell_driver_measurements`
+- `nvidia_blackwell_vbios_measurements`
+
+Measurement reference values are objects keyed by Trustee's one-based SPDM
+measurement index. Each value can be a single digest string or an array of
+allowed digest alternatives. Allowed-version and FWID lists, and both
+measurement maps, must be nonempty and contain values of the expected type.
+Trustee does not fetch NVIDIA RIMs or populate these reference values
+automatically.
```
Xynnn007: See my comments on the previous commit. This document is an overview, might not be proper to tell too many nvidia-specific things

Xynnn007 (commented):
Thanks for the contribution. I left some comments. btw, when reviewing your code I was inspired with some points about the current arch/doc for the repo and I will take a look then.

attestation-service/src/ear_token/ear_default_policy_gpu.rego
fitzthum: I would keep the measurement here, but not make it specific to blackwell. Give the RVs a more generic name and rework the policy such that it checks them if that field is present in the tcb claims.

attestation-service/tests/fixtures/nvidia/blackwell_reference_values.json
chalabi2: Removed it. The PR no longer includes RIM-derived reference values or RIM names and hashes. `BLACKWELL_FIXTURE.md` now says the report and chain came directly from the two NVML calls, without NRAS or the RIM service.

attestation-service/src/ear_token/ear_default_policy_gpu.rego
chalabi2: Fixed. `local_nvidia_evidence` requires local config and measurements with no NRAS parsed marker. `remote_nvidia_evidence` requires the local fields to be absent. A mixed-claim test checks that neither hardware path affirms it.

attestation-service/docs/tcb_claims.md
chalabi2: Checked against the local implementations. NVAT does not authenticate the wrapper `uuid`; nvtrust uses the verified leaf certificate serial as `ueid`. Local claims now expose that serial and keep `uuid` only for payload compatibility. The claims doc warns against using `uuid` for identity or authorization.

deps/verifier/src/nvidia/mod.rs
chalabi2: Changed to an `if`; only LS10 is rejected locally.

deps/verifier/src/nvidia/report.rs
chalabi2: Done. The wrapper is gone and the original `try_new` now takes the architecture.

deps/verifier/src/nvidia/report.rs
chalabi2: I removed this trailing-report check. It was not needed for Blackwell FWID handling, so there is no new length error to word here.

deps/verifier/src/nvidia/report.rs
chalabi2: This followed NVIDIA’s [C++ parser](https://github.com/NVIDIA/attestation-sdk/blob/73c032ebff680ca6d2ba06f4006b511491b71ce9/nv-attestation-sdk-cpp/src/spdm/spdm_resp.cpp#L178-L188), which rejects bytes left after the SPDM response. The Hopper exception was for Trustee’s existing [`hopper_coco_report1` fixture](https://github.com/confidential-containers/trustee/blob/15ae6593c390d8c8d597e40758a877e6362a1aa0/deps/verifier/test_data/nvidia/hopper_coco_report1.txt) from a CoCo CI run. It has one trailing `0x00`, while the other Hopper fixture and this Blackwell capture do not.

I couldn’t find anything defining that byte as Hopper-specific framing, so I removed the check from this PR instead of keeping an architecture rule based on one fixture. We can handle strict framing separately once we know why that CoCo capture includes the byte.

deps/verifier/src/nvidia/cert_chain.rs
chalabi2: Done. This remains `verify`; architecture is an added argument.

deps/verifier/src/nvidia/cert_chain.rs
chalabi2: Done here too. This remains `get_fwid_from_cert` with architecture added.

deps/verifier/src/nvidia/cert_chain.rs
chalabi2: Added an inline link to sections 6.1.1 and 6.1.1.3 of the public TCG DICE Attestation Architecture v1.1 specification. Those sections define `2.23.133.5.4.1` and its `.1` alias.

deps/verifier/src/nvidia/cert_chain.rs
chalabi2: Yes. Hopper uses NVIDIA’s legacy value under the TcbInfo OID, and its verifier takes the final 48-byte SHA-384 digest. Blackwell uses the structured alias and `fwids[0]`. I added inline links to both the TCG specification and the tagged nvtrust implementation.

deps/verifier/src/nvidia/cert_chain.rs
chalabi2: Agreed; I left the GPU and DPU parsers separate here. Sharing them should wait until we have DPU fixtures and can prove both encodings through the same tests.

deps/verifier/src/nvidia/mod.rs
chalabi2: Done. Blackwell is now a case in the existing `test_evaluation` rstest, and Hopper and Blackwell claim construction share one rstest as well.

deps/verifier/src/nvidia/mod.rs
chalabi2: Yes, that is the reason. A repeated signed report must not count as another GPU for device-count or PPCIE policy. The check now uses the verified leaf-certificate serial, not guest UUID metadata, and a test repeats one report with two different UUID values. The reason is also in the commit body.

attestation-service/src/ear_token/broker.rs
chalabi2: I restored the useful part in its own commit, `1b7aaf1`. The formatting stays inline as requested. The map now uses the entry API and a checked `usize` counter, which removes the repeated lookup and avoids `u8` overflow.

attestation-service/src/ear_token/broker.rs
chalabi2: Moved these out of `broker.rs` into `ear_token/policy_tests.rs`, beside the policy they exercise. I kept them in Attestation Service rather than policy-engine because the default GPU policy is owned by AS, not by the generic engine.

attestation-service/src/ear_token/broker.rs
chalabi2: Restored in the separate test-only commit `c11b580`. This fixes the Rust 1.95 `field_reassign_with_default` lint, and Attestation Service all-target clippy now passes without an allowance.

attestation-service/src/ear_token/broker.rs
chalabi2: This second initializer is isolated in the same test-only commit, `c11b580`.

attestation-service/src/ear_token/ear_default_policy_gpu.rego
chalabi2: I reworked this after Fitzthum’s follow-up. There are no Blackwell-specific RV names or bundled measurement values now. One small local policy covers Hopper and Blackwell with generic `allowed_gpu_*` values, and it runs only when the local config and measurement claims are present.

attestation-service/docs/policy.md
chalabi2: Reverted. The policy overview is unchanged from main.

attestation-service/docs/policy.md
chalabi2: Reverted this section too. NVIDIA-specific details now stay in the NVIDIA verifier README and TCB claims page.

deps/verifier/src/nvidia/spdm_response.rs
```diff
@@ -194,7 +194,14 @@ impl OpaqueData {
...
 
-        while offset + 4 < bytes_len {
+        while offset < bytes_len {
+            if bytes_len - offset < 4 {
+                bail!(
+                    "OpaqueData item header overflow: {} trailing byte(s)",
+                    bytes_len - offset
+                );
+            }
+
```
Xynnn007, line 207: I think these logic are part of overflow detection. Can we move them into a separate commit?

deps/verifier/src/nvidia/spdm_response.rs
```diff
@@ -230,6 +239,14 @@ impl OpaqueData {
...
                         .to_string(),
                 ),
+                // Hopper fixtures encode the disabled value 0x55, which is
+                // printable ASCII and historically appeared in claims as "U".
+                // Blackwell encodes enabled as 0xaa, so preserve compatibility
+                // for valid UTF-8 while representing non-text bytes losslessly.
+                OpaqueDataType::Nvdec0Status => match std::str::from_utf8(data_bytes) {
+                    Ok(value) => Value::String(value.trim_end_matches('\0').to_string()),
+                    Err(_) => Value::String(hex::encode(data_bytes)),
+                },
```
Xynnn007: Is `U` always for hopper? and is there a const always for blackwell?

I mean, do we have a case where we fall in a scenario of ambiguity for a hex string here? Can not distinguish if it's from utf8 or hex.

attestation-service/src/ear_token/ear_default_policy_gpu.rego
```diff
@@ -33,9 +33,85 @@ trust_claims := {
 	"sourced-data": sourced_data,
 }
 
+local_nvidia_evidence if {
+	input.nvidia.arch in {"Hopper", "Blackwell"}
```
Xynnn007, line 38: can we add a claim for nvidia to show if it's remote/local verifier?

deps/verifier/src/nvidia/spdm_response.rs
chalabi2: Done. The opaque-data bounds checks and their regression tests are now isolated in `019e941` (`verifier: reject truncated NVIDIA opaque data`). The Blackwell commit no longer contains that general parser hardening.

deps/verifier/src/nvidia/spdm_response.rs
chalabi2: `U` was not Hopper-specific. It was raw `0x55` being decoded as ASCII by the old generic string path. NVIDIA defines `0x55` as disabled and `0xaa` as enabled for NVDEC0 in [nvtrust](https://github.com/NVIDIA/nvtrust/blob/0c5d627313037c1e577d05a232e79394a41b2c21/guest_tools/gpu_verifiers/local_gpu_verifier/src/verifier/config.py#L81) and its [C++ verifier](https://github.com/NVIDIA/attestation-sdk/blob/73c032ebff680ca6d2ba06f4006b511491b71ce9/nv-attestation-sdk-cpp/src/gpu/verify.cpp). I replaced the UTF-8/hex heuristic with those two values, reject other encodings, and emit `disabled` or `enabled`. Tests cover both constants and invalid values.

attestation-service/src/ear_token/ear_default_policy_gpu.rego
chalabi2: Added `verifier: "local" | "remote"` to each NVIDIA device claim after the configured verifier accepts the evidence. The GPU policy now selects its rule set from that claim instead of inferring the mode from claim shape. Tests flip the marker on both local and NRAS-shaped claims and confirm neither can satisfy the other mode.

deps/verifier/src/nvidia/spdm_response.rs
```diff
@@ -331,6 +334,18 @@ impl OpaqueData {
         Ok(Value::String(feature.to_string()))
     }
 
+    fn decode_nvdec_status(bytes: &[u8]) -> Result<Value> {
+        // This field is a status byte. Treating 0x55 as text produced the old
+        // "U" claim and made its representation depend on UTF-8 validity.
```
Xynnn007: Does the comment fix the code here? 

btw, I found that `protected_pcie_status` also uses `"U"`. I am not sure if that one has similar mechanism with this. if so, could you help to apply the same mechanism here? 

Also, please update `tcb_claims.md` for related document part with possible value and clarifies.

deps/verifier/src/nvidia/mod.rs
```diff
@@ -387,17 +387,26 @@ impl Verifier for Nvidia {
         let mut seen_ueids = HashSet::new();
 
         for device in devices.device_evidence_list {
-            let (claims, tee_class) = match &self.verifier_type {
+            let (mut claims, tee_class, verifier) = match &self.verifier_type {
```
Xynnn007: Maybe we can add a serde macro `#[serde(rename_all = "lowercase")]` upon `NvidiaVerifierTypeInternal` and directly insert it to the claims, instead of adding new hardcode strings here?

attestation-service/src/ear_token/ear_default_policy_gpu.rego
```diff
@@ -33,9 +33,83 @@ trust_claims := {
...
+	allowed := query_reference_value(reference_id)
+	nonempty_string_array(allowed)
+	actual in allowed
+}
+
+configuration := 2 if {
+	local_nvidia_evidence
+	reference_value_matches(input.nvidia.config.driver_version, "allowed_driver_versions")
+	reference_value_matches(input.nvidia.config.vbios_version, "allowed_vbios_versions")
+	reference_value_matches(input.nvidia.config.fwid, "allowed_gpu_fwids")
```
Xynnn007, line 72: Some questions from user perspective are 
1. what's the expected reference value shape
2. dowe have a way / where can we get reference values?

attestation-service/docs/tcb_claims.md
```diff
@@ -218,7 +221,8 @@ claims.
 - `config.fwid`: Firmware ID. Found in the report and the signing certificate
 - `config.gpu_info`: GPU information
 - `config.measurement_count`: One measurement_count for each entry in `measurements`. Each measurement_count indicates how many times the respective measurement was extended to get to its current value
-- `config.nvdec0_status`: NVIDIA decoder status
+- `config.nvdec0_status`: NVIDIA decoder status (`enabled` for `0xaa` or
```
Xynnn007: Some parts are in commit https://github.com/confidential-containers/trustee/pull/1551/changes/050b094c1f2ad6c677b7ca46e3ecec973b4baf53. I guess it's easy to move them all in this PR.

deps/verifier/src/nvidia/spdm_response.rs
chalabi2: I checked this against both NVIDIA implementations. `nvtrust` leaves this field as raw bytes, and the current [C++ SDK](https://github.com/NVIDIA/attestation-sdk/blob/73c032ebff680ca6d2ba06f4006b511491b71ce9/nv-attestation-sdk-cpp/include/nv_attestation/gpu/spdm/gpu_opaque_data_parser.hpp#L155-L166) types `protected_pcie_status` as a byte vector. The `0xaa` and `0x55` meanings are only defined for NVDEC. I kept `enabled` and `disabled` for NVDEC and now encode protected PCIe as lowercase hex, so the captured `0x55` becomes `"55"` instead of accidental UTF-8 `"U"`. The regression test and `tcb_claims.md` now cover that distinction.

deps/verifier/src/nvidia/mod.rs
chalabi2: I used the serde approach on a separate `NvidiaVerifierKind` unit enum. `NvidiaVerifierTypeInternal::Remote` also owns config and JWKS state, so serializing it directly would couple the claim shape to those internals. `kind()` maps to `Local` or `Remote`, and `#[serde(rename_all = "lowercase")]` now produces the claim value. There is a serialization test for both variants.

attestation-service/src/ear_token/ear_default_policy_gpu.rego
chalabi2: I expanded the NVIDIA README with the exact RVPS shapes and a sample decoded payload. It now points to both registration paths, explains that every configured measurement must match but not every report index needs a reference, and calls out that the RVPS sample extractor is demo-only. Production values need authenticated NVIDIA RIMs or a separate qualification process. Copying them from the evidence under evaluation would not establish trust.

attestation-service/docs/tcb_claims.md
chalabi2: Moved all `tcb_claims.md` edits out of the verifier commit and into `docs(nvidia): describe local Blackwell verification` (`efe0751`). The verifier commit no longer touches that file.

deps/verifier/src/nvidia/spdm_response.rs
Xynnn007: do we have some example values with what they mean for `protected_pcie_status`?

deps/verifier/src/nvidia/mod.rs
Xynnn007: Oh ok. Maybe a `as_ctype_string()->&str` function for `NvidiaVerifierTypeInternal` is better, than bringing in a new enum?

deps/verifier/src/nvidia/README.md
```diff
@@ -3,8 +3,57 @@
...
+    "<another-index>": [
+      "<96-character-sha384-hex>",
+      "<another-accepted-sha384-hex>"
+    ]
+  }
+}
+```
+
+The policy requires every listed measurement to match. It does not require a
+reference for every measurement in the report.
```
Xynnn007, line 44: Sry - what does this mean?

deps/verifier/src/nvidia/README.md
```diff
@@ -3,8 +3,57 @@
...
+
+Trustee does not fetch NVIDIA RIMs or populate these values. Operators must
+derive them from an authenticated source, such as signed NVIDIA driver and
+VBIOS RIMs, or from a separate hardware and software qualification process.
+Values copied from the evidence being evaluated are not a trust anchor.
```
Xynnn007, line 49: I guess we need to fill this up later to have a concrete guide. If possible, this PR can be good; if not, we can open an issue to record this.

Are there any license issues?

deps/verifier/src/nvidia/mod.rs
```diff
@@ -76,6 +77,22 @@ enum NvidiaVerifierTypeInternal {
...
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
+#[serde(rename_all = "lowercase")]
+enum NvidiaVerifierKind {
+    Local,
+    Remote,
+}
```
Xynnn007, line 94: Let's impl a `fn to_verifier_kind(&self) -> &str` for `NvidiaVerifierTypeInternal`, as `NvidiaVerifierKind` is only for serialization and this way we can save code.

PR #1554, opened 2026-08-11
fix(cca): add missing trust_claims object to default CCA OPA policy

shashank-fujitsu:
## Summary
The default CCA OPA policy (`deps/verifier/test_data/cca/conf/default.rego`) never assembles a `trust_claims` object. The attestation service queries `data.policy.trust_claims` in `attestation-service/src/ear_token/broker.rs` to build the EAR token - if that query returns undefined, the AS panics and every CCA attestation request fails.

The file already defines `hardware`, `instance_identity`, `runtime_opaque`, and `executables` individually, but never wraps them into `trust_claims`, and has no `configuration` claim at all.

## Changes
- Add `default configuration := 36` (EAR value for "unavailable to the Verifier")
- Add a `trust_claims` object assembling all five required claims with kebab-case EAR keys (`instance-identity`, not `instance_identity`) matches the pattern already used by `attestation-service/src/ear_token/ear_default_policy_cpu.rego`

## Testing
Verified by inspection against the working `ear_default_policy_cpu.rego` pattern and by running the real `opa fmt` over the change.

---

Thanks to @shefali-kamal for co-authoring and reviewing this work.

### DIFF
--- a/deps/verifier/test_data/cca/conf/default.rego
+++ b/deps/verifier/test_data/cca/conf/default.rego
@@ -33,10 +33,25 @@ default runtime_opaque := 2
 # This is the default, unless the RIM claim matches one of the configured
 # reference values.
 default executables := 33
+
 # The value 3 stands for
 # "Only a recognized genuine set of approved executables have
 #  been loaded during the boot process."
 # the RIM (realm initial measurement) must match
 executables := 3 if {
 	input.cca.realm["cca-realm-initial-measurement"] in data.reference["cca.realm.cca-realm-initial-measurement"]
+}
+
+# For the `configuration` trust claim, the value 36 stands for
+# "Elements of the configuration relevant to security are unavailable to the Verifier."
+default configuration := 36
+
+# trust_claims MUST be defined as an object with kebab-case EAR claim names.
+# Missing this causes the AS to panic at token issuance time.
+trust_claims := {
+	"hardware": hardware,
+	"instance-identity": instance_identity,
+	"runtime-opaque": runtime_opaque,
+	"executables": executables,
+	"configuration": configuration,
 }


deps/verifier/test_data/cca/conf/default.rego
```diff
@@ -33,10 +33,26 @@ default runtime_opaque := 2
...
 	input.cca.realm["cca-realm-initial-measurement"] in data.reference["cca.realm.cca-realm-initial-measurement"]
 }
+
+# For the `configuration` trust claim, the value 36 stands for
+# "Elements of the configuration relevant to security are unavailable to the Verifier."
+default configuration := 36
+
+# trust_claims MUST be defined as an object with kebab-case EAR claim names.
+# Missing this causes the AS to panic at token issuance time (data.policy.trust_claims
+# query in attestation-service/src/ear_token/broker.rs).
```
Xynnn007: I am hesitate about the comments here. Two options are ok to me
1. remove them all
2. remove only the code file path

deps/verifier/test_data/cca/conf/default.rego
```diff
@@ -33,10 +33,26 @@ default runtime_opaque := 2
...
 # "Only a recognized genuine set of approved executables have
 #  been loaded during the boot process."
 # the RIM (realm initial measurement) must match
 executables := 3 if {
 	input.cca.realm["cca-realm-initial-measurement"] in data.reference["cca.realm.cca-realm-initial-measurement"]
 }
+
+# For the `configuration` trust claim, the value 36 stands for
+# "Elements of the configuration relevant to security are unavailable to the Verifier."
+default configuration := 36
```
Xynnn007, line 47: Not related to this PR.

I think we need a fix upon the claims. Due to https://www.ietf.org/archive/id/draft-ietf-rats-ar4si-10.html#section-3.4 not all claims are required, but all are optional.

Xynnn007 (approved):
LGTM. Thanks @shashank-fujitsu ! Only one nit

deps/verifier/test_data/cca/conf/default.rego
fitzthum: btw, if we do include this claim here, let's put it above with the other defaults.

fitzthum (commented):
I think I would prefer if we added CCA to the main default policy.

deps/verifier/test_data/cca/conf/default.rego
shashank-fujitsu: Thanks @Xynnn007! 
Applied the file-path fix - dropped the reference to `broker.rs`

deps/verifier/test_data/cca/conf/default.rego
shashank-fujitsu: > Not related to this PR.
> 
> I think we need a fix upon the claims. Due to https://www.ietf.org/archive/id/draft-ietf-rats-ar4si-10.html#section-3.4 not all claims are required, but all are optional.

Agreed on the claims-optionality point!
Happy to follow up on this with a separate PR.

shashank-fujitsu:
> btw, if we do include this claim here, let's put it above with the other defaults.

Good catch... Will move the configuration up with the other defaults.
Thanks!

fitzthum:
> I think I would prefer if we added CCA to the main default policy.

Please see ^^

shashank-fujitsu:
> I think I would prefer if we added CCA to the main default policy.

Hi @fitzthum, 

Agreed, and it's the right direction long-term.
This PR is scoped to just the panic bug (missing `trust_claims` causing `data.policy.trust_claims` to be undefined at token issuance) in the existing test-data policy.

I think we can work on a separate PR to add an `input.cca` branch to `ear_default_policy_cpu.rego`, mirroring the SNP/TDX pattern, so CCA no longer needs the manual-copy step.

Happy to create a new follow-up PR for the same.

shashank-fujitsu:
Hi @fitzthum ,
Would like to know your thoughts on this.

fitzthum:
Well, it's pretty simple to move this to the main default policy, but I guess you can do that in a separate PR if you would prefer.

fitzthum (approved):

PR #1555, opened 2026-08-11
fix(docker-compose): make cert validity periods overridable, fix token cert's silent 30-day default

shashank-fujitsu:
## Summary
`kbs/config/docker-compose/setup.sh` generates the token-signing certificate using `openssl x509 -req` without a `-days` argument. OpenSSL defaults to 30 days.

Symptom: ~30 days after initial deployment, all attestation requests return 401 ("JWK cannot be validated by trust anchor"). There is no warning in logs that the certificate has expired - the failure looks identical to a misconfiguration.

## Changes
- Add `TOKEN_CERT_DAYS="${TOKEN_CERT_DAYS:-90}"` and pass it via `-days` to the `openssl x509 -req` call that signs `token-cert.pem`. Defaulting to a fixed 90-day period avoids silently inheriting openssl's own default while still forcing rotation. The env var lets users who want longer-lived dev/test certs opt in explicitly.
- Converted the root CA cert's already-hardcoded `-days 3650` to a `CA_CERT_DAYS="${CA_CERT_DAYS:-3650}"` variable (default unchanged), so both cert lifetimes are controlled the same way instead of one being a magic number.

---

Thanks to @shefali-kamal for co-authoring and reviewing this work.

### DIFF
--- a/kbs/config/docker-compose/setup.sh
+++ b/kbs/config/docker-compose/setup.sh
@@ -4,6 +4,12 @@ set -euxo pipefail
 KEY_DIR="/opt/confidential-containers/kbs/user-keys"
 cd "${KEY_DIR}"
 
+# Token-signing cert validity in days; defaults to 90 instead of openssl's silent 30.
+TOKEN_CERT_DAYS="${TOKEN_CERT_DAYS:-90}"
+
+# Root CA cert validity in days; default unchanged at 10 years.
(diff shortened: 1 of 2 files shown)


Xynnn007 (commented):
Please also add documents in `kbs/docs/cluster.md` to show the usage

fitzthum (approved):
LGTM pending Ding's comment

PR #1564, opened 2026-08-14
kbs: document verbose_token on AttestationTokenBroker

### LINKED ISSUE #1563: kbs: verbose EAR tokens exceed actix-http 128 KiB request-head limit
## Summary

CDH resource GETs fail after a successful multi-device attest with:

```
ERROR actix_http::h1::dispatcher: stream error: request parse error: message head is too large
```

The request never reaches a KBS handler.

## Why

1. Guest AA sends TDX primary evidence plus NVIDIA additional evidence (e.g. 8× Blackwell via NRAS).
2. AS verifies every device and, with the default `verbose_token = true`, embeds raw `additional-evidence` in the EAR (`transform_claims` in `attestation-service/src/ear_token/broker.rs`).
3. Guest CDH `cc_kbc` always uses the passport client (`Authorization: Bearer <EAR>`), not the session cookie.
4. actix-http 3.x hardcodes `MAX_BUFFER_SIZE = 131_072` in `h1/decoder.rs`. There is no `HttpServer::max_header_size` on actix-web 4.14. A ~170 KiB EAR JWT is rejected while parsing the request head.

`verbose_token` was added in 8632b02 specifically because additional evidence can overwhelm HTTP header limits, including KBS. The default remains `true`. `kbs/docs/config.md` does not list the field (it is documented under the AS config).

## Workaround

```toml
verbose_token = false
```

Verification still runs. The EAR omits raw additional-device evidence, so the Bearer token fits in the 128 KiB head. Session-cookie `GET /resource` would also avoid the header, but stock CDH does not send the cookie.

## Seen on

- Trustee / KBS `ghcr.io/confidential-containers/staged-images/kbs:b2442c22` and current `main`
- Kata 4.0 guest, `kata-qemu-nvidia-gpu-tdx-runtime-rs`, 8× B200, NRAS remote NVIDIA verifier
- CDH `GET /kbs/v0/resource/default/...` after `POST /kbs/v0/attest` returned 200

## Suggestions

- Document `verbose_token` on the KBS `AttestationTokenBroker` table (operators look there first).
- Consider defaulting `verbose_token` to `false`, or warning when the issued token is larger than 128 KiB.
- Longer term: make the actix request-head limit configurable, or have CDH use the session cookie for resource fetch.

pamanseau:
## Summary

- Document `verbose_token` on the KBS `AttestationTokenBroker` table. The field already exists on `EarTokenConfiguration` (default `true`) but was only listed in the AS config docs.
- Operators hit `message head is too large` when CDH sends a verbose multi-device EAR as `Authorization: Bearer` (actix-http 128 KiB request-head cap).

Fixes #1563

## Test plan

- [ ] Confirm the table matches `EarTokenConfiguration` in `attestation-service/src/ear_token/mod.rs`
- [ ] No runtime change; docs only

Made with [Cursor](https://cursor.com)

### DIFF
--- a/kbs/docs/config.md
+++ b/kbs/docs/config.md
@@ -184,7 +184,7 @@ When `type` is set to `coco_as_builtin`, the following properties can be set.
 | `build_name`     | String                 | The build name to be used as part of the Verifier ID in the EAR                | No       | Automatically generated from Cargo package and AS version             |
 | `profile_name`   | String                 | The Profile that describes the EAR token                                       | No       | tag:github.com,2024:confidential-containers/Trustee`                  |
 | `signer`         | [TokenSignerConfig][4] | Signing material of the attestation result token.                              | No       | None                                                                  |
-| `verbose_token`  | Boolean                | Include raw additional-device evidence in the EAR. Multi-GPU guests can produce tokens larger than actix-http's 128 KiB request-head limit (`message head is too large` on CDH `Authorization: Bearer` resource GETs). Set `false` to omit that evidence from the token; verification still runs. | No       | `true`                                                                |
+| `verbose_token`  | Boolean                | Include detailed information in the EAR, such as the raw additional-device evidence. Device evidence (for example a GPU) can produce tokens larger than actix-http's 128 KiB request-head limit (causing errors like `message head is too large` on CDH `Authorization: Bearer` resource GETs). Set `false` to omit that extra detail from the token; verification still runs. | No       | `true`                                                                |
 
 ##### TokenSignerConfig
 
--- a/kbs/docs/config.md
+++ b/kbs/docs/config.md
@@ -184,6 +184,7 @@ When `type` is set to `coco_as_builtin`, the following properties can be set.
 | `build_name`     | String                 | The build name to be used as part of the Verifier ID in the EAR                | No       | Automatically generated from Cargo package and AS version             |
 | `profile_name`   | String                 | The Profile that describes the EAR token                                       | No       | tag:github.com,2024:confidential-containers/Trustee`                  |
 | `signer`         | [TokenSignerConfig][4] | Signing material of the attestation result token.                              | No       | None                                                                  |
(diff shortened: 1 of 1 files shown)


kbs/docs/config.md
```diff
@@ -184,6 +184,7 @@ When `type` is set to `coco_as_builtin`, the following properties can be set.
 | `build_name`     | String                 | The build name to be used as part of the Verifier ID in the EAR                | No       | Automatically generated from Cargo package and AS version             |
 | `profile_name`   | String                 | The Profile that describes the EAR token                                       | No       | tag:github.com,2024:confidential-containers/Trustee`                  |
 | `signer`         | [TokenSignerConfig][4] | Signing material of the attestation result token.                              | No       | None                                                                  |
+| `verbose_token`  | Boolean                | Include raw additional-device evidence in the EAR. Multi-GPU guests can produce tokens larger than actix-http's 128 KiB request-head limit (`message head is too large` on CDH `Authorization: Bearer` resource GETs). Set `false` to omit that evidence from the token; verification still runs. | No       | `true`                                                                |
```
Xynnn007: We can make it more general

```suggestion
| `verbose_token`  | Boolean                | Include raw additional-device evidence in the EAR. Device evidences like GPU's can produce tokens larger than actix-http's 128 KiB request-head limit (causing errors like `message head is too large` on CDH `Authorization: Bearer` resource GETs). Set `false` to omit that evidence from the token; verification still runs. | No       | `true`                                                                |
```

Xynnn007 (approved):
Thanks - Only one comments

kbs/docs/config.md
fitzthum: This flag may be hooked into other fields in the future, so this should say something like "Include detailed informations, such as the raw additional-device evidence."

pamanseau: Updated: GPU is now an example of device evidence that can blow the 128 KiB request-head limit. The flag is described as detailed EAR information, not only additional-device evidence.

pamanseau: Updated to “include detailed information, such as the raw additional-device evidence,” so this is not pinned to one field if the flag grows later. The 128 KiB / `message head is too large` warning is still there.

fitzthum (approved):
LGTM

PR #1567, opened 2026-08-15
as: reject policies that fail to compile in set_policy

AnkushUjawane:
Currently AttestationService::set_policy accepts any policy string without validation, storing it as-is. A syntactically invalid policy is only discovered later, at actual attestation time, with no clear signal that the policy itself was the problem.

This validates that a policy compiles (via the same Regorus engine AS uses internally) before it's stored, returning a clear compiler error immediately if it doesn't. This is the single entry point both gRPC (grpc-as) and RESTful (restful-as) SetPolicy handlers funnel through, so both are covered by this one change.

Note: this is a behavioral change to the SetPolicy/set_policy API - a request that previously succeeded with an invalid policy will now be rejected with an error. Valid policies are unaffected.

Manually verified against a running restful-as instance: a syntactically invalid policy is now rejected with 500 and a specific compiler error (previously would have returned 200), and a valid policy still succeeds with 200 as before.

This follows from discussion on #1526, where @fitzthum suggested validation belongs at the admin endpoint rather than in a client-side CLI tool.

### DIFF
--- a/deps/policy-engine/src/lib.rs
+++ b/deps/policy-engine/src/lib.rs
@@ -23,6 +23,11 @@ pub trait EngineTrait {
     fn policy_suffix() -> &'static str {
         ""
     }
+
+    // Check that a policy is well-formed, before it is accepted and stored.
+    fn check_policy_format(_policy: &str) -> Result<()> {
+        Ok(())
+    }
 }
 
 #[derive(Clone)]
@@ -36,6 +41,7 @@ impl<T: Send + Sync + EngineTrait> PolicyEngine<T> {
     /// The policy is expected to be provided as string.
     /// Concrete policy engine backend may handle the policy in different ways.
     pub async fn set_policy(&self, policy_id: &str, policy: &str, overwrite: bool) -> Result<()> {
+        T::check_policy_format(policy)?;
         let params = SetParameters { overwrite };
         let policy_id = format!("{}{}", policy_id, T::policy_suffix());
         let _ = self
--- a/deps/policy-engine/src/policy/rego.rs
+++ b/deps/policy-engine/src/policy/rego.rs
@@ -28,6 +28,13 @@ impl EngineTrait for Regorus {
     fn policy_suffix() -> &'static str {
         ".rego"
     }
+    fn check_policy_format(policy: &str) -> Result<()> {
+        let mut engine = regorus::Engine::new();
+        engine
+            .add_policy("".to_string(), policy.to_string())
+            .map_err(PolicyError::LoadPolicyFailed)?;
+        Ok(())
+    }
 }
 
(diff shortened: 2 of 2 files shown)


attestation-service/src/lib.rs
```diff
@@ -173,6 +173,13 @@ impl AttestationService {
     pub async fn set_policy(&mut self, policy_id: String, policy: String) -> Result<()> {
         let policy_bytes = base64::engine::general_purpose::URL_SAFE_NO_PAD.decode(policy)?;
         let policy = String::from_utf8(policy_bytes)?;
+
```
Xynnn007: Can we move the regorus compliation check around [here](https://github.com/confidential-containers/trustee/blob/main/deps/policy-engine/src/lib.rs#L34-L47)?

s.t.
1. add a `fn check_policy_format(policy: &str) -> Result<()>;` func member to `pub trait EngineTrait`
2. impl the logic showed in this commit to `impl EngineTrait for Regorus` block with adding `fn check_policy_format(policy: &str) -> Result<()> {...}`
3. add `T::check_policy_format(policy)?;` to ` pub async fn set_policy(&self, policy_id: &str, policy: &str, overwrite: bool) -> Result<()>`

AnkushUjawane:
Done, moved it into EngineTrait::check_policy_format as you laid out, called from PolicyEngine::set_policy. this also
covers KBS's resource policy upload for free, since it goes through the same generic method — that's broader protection than my original AS-only version had.

Xynnn007 (approved):
LGTM. Thanks!

AnkushUjawane:
Thanks for review! Happy to have this merged whenever's convenient.

deps/policy-engine/src/lib.rs
```diff
@@ -23,6 +23,12 @@ pub trait EngineTrait {
     fn policy_suffix() -> &'static str {
         ""
     }
+
+    // Check that a policy is well-formed, before it is accepted and stored.
+    fn check_policy_format(policy: &str) -> Result<()> {
+        let _ = policy;
```
fitzthum: Is this necessary? Can you just do `_policy` in the signature?

fitzthum (commented):
Please squash your commits. Also, shorten the commit message. No need to mention previous discussion in the commit message.

deps/policy-engine/src/lib.rs
AnkushUjawane: Done, unnecessary code removed and making correction as you suggest

Xynnn007 (approved):

Xynnn007:
@AnkushUjawane please fix the sign-off(DCO) error

AnkushUjawane:
Done, Thank you!!

PR #1570, opened 2026-08-18
verifier: intel_dcap: normalize PCCS Root CA CRL to DER

mythi:
PCCS /rootcacrl responses contain the Root CA CRL as hex-ASCII, whereas Intel PCS retrieves it directly from the CRL distribution point as raw DER.

CollateralService::get promises raw DER for CRL bodies regardless of the backing source, so decode PCCS RootCaCrl responses in Pcs::get_from_http.

Although QVL also accepts other representations, QuoteCollateral version 3.1 specifies raw DER for CRL fields. Normalizing the PCCS response therefore enforces the CollateralService contract.


deps/verifier/src/intel_dcap/collateral_service.rs
```diff
@@ -295,6 +295,11 @@ pub enum CollateralType<'a> {
     QeIdentity(&'a IntelTee),
     /// CA type; the CRL is always returned as raw DER bytes.
     PckCrl(PckCaType),
+    /// CRL distribution point URL, if known; the CRL is always returned as
+    /// raw DER bytes. A `CollateralService` implementation is responsible
+    /// for normalizing to DER regardless of what its backing source returns
+    /// (e.g. some PCCS mirrors return the Root CA CRL as hex-ASCII text
+    /// rather than binary DER over HTTP).
```
Xynnn007: These parts can be included in the commit message.

deps/verifier/src/intel_dcap/collateral_service.rs
```diff
@@ -309,6 +314,10 @@ pub enum IntelTee {
 pub trait CollateralService: Send + Sync {
     type Error: std::error::Error + Send + Sync + 'static;
 
+    /// Fetches the requested collateral. CRL bodies ([`CollateralType::PckCrl`]
+    /// and [`CollateralType::RootCaCrl`]) must always be returned as raw DER
+    /// bytes — implementations own normalizing whatever encoding their
+    /// backing source actually used (PEM, hex-ASCII, or DER).
```
Xynnn007: These too

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -351,7 +352,18 @@ impl Pcs {
...
             .transpose()?;
 
-        let body = Vec::from(response.bytes().await?);
+        let response_bytes = response.bytes().await?;
+        // Intel's official PCS returns the Root CA CRL as raw DER, fetched
+        // directly from the CRL Distribution Point URL embedded in the
+        // certificate. Third-party caching services following Intel's spec,
+        // reached through the `/rootcacrl` endpoint (API v3.0) instead, always return it
+        // hex-encoded. As `CollateralService::get` promises raw DER, decode RootCaCrl
+        // here.
```
Xynnn007: Do we have any official document link? it can be add here

mythi: added

deps/verifier/src/intel_dcap/collateral_service.rs
mythi: included, same for below

Xynnn007: Looks like we have three places to tell the same truth. You can choose one place to tell that "pccs returns der, due to the spec". Other parts can be removed

mythi: the trait rustdoc is not about "pccs returns der"

deps/verifier/src/intel_dcap/pcs.rs
```diff
@@ -289,6 +289,7 @@ impl Pcs {
         let is_pcs = *is_pcs;
         let mut params: HashMap<&str, String> = HashMap::new();
         let chain_header = Self::cert_chain_header(&ct);
+        let needs_hex_decode = matches!(ct, CollateralType::RootCaCrl(_)) && !is_pcs;
```
Xynnn007, line 299: Looks like the comments can be moved here https://github.com/confidential-containers/trustee/pull/1570/changes#diff-c892af151c02636bf53777777fb6183abbfddb87760ef64f28d0ca7bf3ec6c8fR356-R362 to explain the behavior of `needs_hex_decode`?

mythi: I'd prefer to keep it as is because the comments also refer to what's in those `response_bytes`.

Xynnn007: The `response_bytes` needs hex decode because `needs_hex_decode` flag. line 292 shows explicit logic for `needs_hex_decode` flag is based on rootcacrl + non-pcs, fitting with the most former part of the comments. Logic + comment for it are located in different places would make some trouble for maintaince.

you may need to remove `As `CollateralService::get` promises raw DER, decode RootCaCrl here.` as-is or change the words a little

mythi: Updated.

Xynnn007 (approved):

PR #1571, opened 2026-08-18
verifier: intel_dcap: borrow quote certification data instead of copying

mythi:
Quote::cert_data() (QE authentication data and the PCK certificate chain) never outlives the quote_bin buffer it was parsed from, so give Quote an explicit lifetime and slice directly out of quote_bin instead of allocating two Vec<u8> copies per parse.

### DIFF
--- a/deps/verifier/src/intel_dcap/quote.rs
+++ b/deps/verifier/src/intel_dcap/quote.rs
@@ -78,17 +78,17 @@ struct QeReport {
 
 #[repr(C)]
 #[derive(Debug)]
-pub(crate) struct QeCertificationData {
+pub(crate) struct QeCertificationData<'a> {
     qe_report: QeReport,
-    qe_authentication: Vec<u8>,
-    pub(crate) certificates: Vec<u8>,
+    qe_authentication: &'a [u8],
+    pub(crate) certificates: &'a [u8],
 }
 
 #[repr(C)]
 #[derive(Debug)]
-pub(crate) struct QuoteCertificationData {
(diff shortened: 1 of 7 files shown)


deps/verifier/src/intel_dcap/quote.rs
```diff
@@ -380,13 +380,15 @@ impl fmt::Display for QuoteV5Body {
     }
 }
 
-/// A parsed Intel DCAP ECDSA quote.
-pub(crate) enum Quote {
+/// A parsed Intel DCAP ECDSA quote, borrowing its variable-length fields
+/// (QE authentication data, PCK certificate chain) directly out of the
+/// original `quote_bin` buffer rather than copying them.
```
Xynnn007: Let's keep the original comment as-is. The extra parts can be directly seen from the code - if needed add them to commit message.

deps/verifier/src/tdx/claims.rs
```diff
@@ -35,7 +35,7 @@ macro_rules! body_field {
     };
 }
 
-impl Quote {
+impl Quote<'_> {
```
Xynnn007, line 38: The reference will require the original data slice always be alive with the `Quote` object. Now `Quote` object is marked as `pub(crate)`, thus we can assert this. Thus ok to me.

Xynnn007 (commented):
LGTM. Only one comment about the inline comment.
btw the link ci is not related to this PR. we can ignore it .

Xynnn007 (approved):

Issue #1558, opened 2026-08-12
RFC: SPIRE Plugin


fitzthum:
Adding SPIFFE/SPIRE support to CoCo/Trustee is tricky, but I think we finally have a pretty good understanding of how to do this.

My basic proposal is to add a plugin to Trustee that can generate SVIDs and provide them to a workload. This will skip over the two-tiered SPIRE attestor design, which isn't a great fit for confidential computing (where the node is untrusted). Rather than generating SVIDs completely from scratch and managing workload state, the plugin will  use SPIRE's delegated identity API to connect to an existing SPIRE server. This approach will only provide pod-level selectors. Container-level selectors require additional logic inside the podvm, which is out of scope for now.

Taking a step back, we have had a couple previous proposals for SPIFFE support. First, there has been some work to put the SPIRE Agent itself into the podvm. This has some drawbacks. We end up with a lot of "nodes" and the k8s workload attestor doesn't work inside a podvm without major alterations. We have also discussed integrating SPIFFE directly into the KBS protocol. This is a much more involved effort and I think we are currently leaning more towards mTLS for this. In any case, adding a SPIRE plugin does not preclude future development.

**Delegated Identity API**

The delegated identity API is an interface of the SPIRE agent. It can take a list of selectors and return the corresponding SVID. This is perfect for us, since we can skip over the entire attestor flow and come up with the selectors our own way (described below), but we can still use the existing SPIRE server. The snag is that to use the delegated identity API, a client needs its own SVID and there needs to be a SPIRE agent running locally. This means that we will need to do some bootstrapping (described below). Although this is a bit inconvenient, it is actually fairly elegant. The real SPIRE agent is going to delegate its powers to the Trustee plugin. As such, Trustee should be attested by SPIRE. 

**Selectors**

The selectors will come from the init-data / validated-identifiers.  We have already laid the groundwork for this. Basically the selectors will stem from the measured configurations in the init-data (such as the agent policy and image-signature-policy). We can use the validated-identifiers extension to map these to more standard fields, which the plugin can use as selectors. We should be able to approximate typical k8s pod-level selectors, although not all fields will necessarily be measured. We will first focus on the mechanics of the plugin. The policies and mappings can be refined as we get more requirements.

**Bootstrap**

This proposal mainly targets users that already have some kind of SPIRE setup and want to add CoCo workloads to it without too many changes. For simplicity, let's assume that we will deploy Trustee with the Helm chart. Here are the steps for the first deployment.

[...]

Issue #987, opened 2025-10-02
RFC: Add identifiers extension to attestation token


fitzthum:
This is not a complete solution for identity, but it might be a step in the right direction.

Let's consider adding an extension to our attestation token called `identifiers`. This would be one place to store different ids. `identifiers` would have two subfields: `unvalidated` and `validated`.

* `unvalidated` identifiers could be things that are "self-reported" by a guest, such as a request id or a desired policy id. This could also include things like the physical location or ip address of a workload that can't be measured confidentially. Basically any id that isn't `validated` would go here.

* `validated` identifiers would be those that are backed by the measurement. More specifically validated identifiers would have two properties. First, the hardware measurement itself would reflect the id. Second, the guest runtime would be known to honor the values. For example, a validated identifier could be the issuer of a cert that signs a container image. If the container image signing policy and public key are included in the init-data, then the first property is met. If we know that we are using image-rs in the guest to pull the image, the second property is met. We can then expose this identifier as being validated. See below for information on how the AS would actually compute this.

Previously we have talked about allowing users to write policies that map TCB Claims to certain identities. This would require users to understand attestation at a fairly deep level and it creates a binding between hw-specific TCB Claims and identities. Instead, this proposal would allow the AS to extract platform-agnostic identifiers. This extension would crystallize all of the ways that a guest could be identified into one structure with well-defined properties.

**How to calculate**

I'm not totally sure how to implement this yet. One option would be to add another fundamental step to the AS. Currently we have a policy to take TCB Claims and calculate an AR4SI vector. We could add another policy that takes TCB Claims and extracts identifiers. We could also have an individual policy for each identifier. Either way, the policy would define per-platform how to extract/validate identifiers. Ideally a user would never need to touch these policies.

One snag is that the second property I listed above (knowing that the guest will use/honor the identifier), is hard to enforce without knowing what is happening inside of the guest. We may need some further discussion on how to handle this. It's also debatable if we need this property at all.

All of the identifiers would be optional. So if one platform does not support measuring a certain field, it could be skipped, or just set as an unvalidated identifier. 

**How to consume**

If we add this field to the AS Token, it makes it much easier to conceptualize identity in the KBS. First, a KBS policy can simply use the identifiers directly. If you want to write a policy that only releases a certain secret to images signed by a certain party, you can easily express that.

Perhaps more importantly, we could build an identity system on top of this. For instance, we could implement a SPIFFE plugin that has selectors over the identifiers.

[...]

PR #1581, opened 2026-08-20
SPIRE Delegated Identity Plugin

fitzthum:
Add a plugin that fulfills SPIFFE SVID requests using the delegated identity API. This bootstraps from an existing SPIRE agent/server and uses selectors generated from the attestation token.

For more information see the [RFC](https://github.com/confidential-containers/trustee/issues/1558), the README added in this PR, and the commit messages.

There is still work to do to get the right validated-identifiers / selectors in the attestation policy. I tweaked this slightly, but mostly left it for later. Of course, the validated-identifiers are very closely tied to the Kata Agent policy. We may need to rework the representation of the podspec in the policy/init-data to unlock the best possible set of identifiers/selectors.

Either way, it's good to have some concrete SPIRE support. We've theorized about this for a long time, but now we can finally have some real people try it out.


fitzthum:
CI needs a bit of work still. Seems like there is some intermittent failure with occlum.

Xynnn007:
is this ready for review?

fitzthum:
> is this ready for review?

Yes. Ready for review.

Keep in mind that there will be some future work to refine the selectors. I am trying to make the first version fairly flexible so people can try it out and see what they need.

tools/kbs-client/src/main.rs
```diff
@@ -38,11 +38,16 @@ enum Commands {
     /// Get confidential resource
     #[clap(arg_required_else_help = true)]
     GetResource {
-        /// KBS Resource path, e.g my_repo/resource_type/123abc
+        /// KBS Resource path, e.g my_repo/resource_type/123abc.
         /// Document: https://github.com/confidential-containers/attestation-agent/blob/main/docs/KBS_URI.md
         #[clap(long, value_parser)]
         path: String,
```
Xynnn007, line 44: I think we can gradually deprecate this field. Starting from raise a warning saying "it will be deprecated, use `plugin` instead". Then, we can get rid of it in the release after the next. Also, `plugin` can directly be a KBS URL, s.t. `kbs+x://xxx`/`kbs://xxx` style

kbs/src/plugins/implementations/provisioner/mod.rs
```diff
@@ -188,6 +188,7 @@ impl super::super::plugin_manager::ClientPlugin for Provisioner {
         path: &[&str],
         method: &Method,
         init_data: Option<&serde_json::Value>,
+        _validated_identifiers: Option<&serde_json::Value>,
```
Xynnn007: for general KBS plugins, `customized_claims` (behavior) or `verifier_claims` (the way ear calls) would be better. `validated_identifiers` could be a spiffe plugin level name than a general one.

kbs/src/plugins/implementations/spire/mod.rs
```diff
@@ -0,0 +1,513 @@
...
+                let response = self.fetch_svid(selectors).await?;
+                Ok(serde_json::to_vec(&response)?)
+            }
+            ["x509-bundles"] => {
+                let response = self.fetch_bundles().await?;
+                Ok(serde_json::to_vec(&response)?)
+            }
+            _ => {
+                warn!(?path, "SPIRE plugin: request to unsupported path");
+                bail!("Illegal path. Only x509 SVIDs and bundles are supported.")
```
Xynnn007, line 310:
```suggestion
                bail!("Illegal path. Only `x509-svid` and`x509-bundles` are supported.")
```

kbs/src/plugins/implementations/spire/mod.rs
```diff
@@ -0,0 +1,513 @@
...
+    } else {
+        trustee_prefix
+    };
+
+    // Parse as string without quotes
+    let value = match value {
+        Value::String(s) => s.clone(),
+        Value::Number(n) => n.to_string(),
+        Value::Bool(b) => b.to_string(),
+        Value::Null | Value::Array(_) | Value::Object(_) => return None,
```
Xynnn007, line 148: How to handle the nested struct values? Maybe a flatten mechanism and document is needed?

kbs/src/plugins/implementations/spire/mod.rs
```diff
@@ -0,0 +1,513 @@
...
+    "k8s".to_string()
+}
+
+fn default_trustee_selector_prefix() -> String {
+    "trustee".to_string()
+}
+
+// Pod-level selector key names that are provided by the SPIRE
+// workload attestor. Plugin handles these selectors specially.
+const KNOWN_K8S_SELECTOR_KEYS: &[&str] = &[
```
Xynnn007, line 48: Do we have a translation layer from pod yaml policy to the selectors?

kbs/Makefile
```diff
@@ -2,6 +2,7 @@ AS_TYPE ?= coco-as
 ALIYUN ?= false
 PQC_EXPERIMENTAL ?= false
 NEBULA_CA_PLUGIN ?= false
+SPIRE_PLUGIN ?= true
```
Xynnn007, line 5: I lean towards to make this false.

This enablement warns a starting for accepting plugins by default built-in. heavy dependencies is a very subjective saying. I must say that gRPC version are much more encouraged to avoid putting everything into the single binary.

Like composible guest image in kata. We would need to decouple plugins into separate released container than built-in. This could be a decoupled ecosystem than an all-in-one close-source on.

So, if you believes this need to be built-in. I am ok, but what I will do in future is to decouple things into separate plugin images. We can release them officially, but in plugin mode.

cc @bpradipt

kbs/docs/plugins/spire.md
```diff
@@ -0,0 +1,89 @@
...
+and returns the SVID to the VM as an encrypted resource.
+The plugin also provides trust bundles.
+
+At a high-level, the goal of confidential SPIRE is to provide an SVID that is backed by cryptographic
+guarantees.
+In other words, the properties represented by the SPIFF-ID, must reflect the workload that is
+actually running inside the enclave, and there should be cryptographic evidence of this.
+
+These guarantees are achieved through an abstraction called Validated Identifiers.
+Validated Identifiers are an extension of the Trustee attestation token that contain
```
Xynnn007, line 17: Emmm I think we should conceptual boundaries here. Some different levels
1. What AS includes inside the ear token, defined by policy, should be named to something like `userdata` or `customized claims`
2. validated identifiers are k8s/workload related conceptions for spiffe plugin. It can parse required parts from `userdata` or `customized claims` from evidences. The identifiers are something like `selector`.
3. Note that validated identifiers are an example in scope of user defined policies. It focus on policy-rego for CoCo. It's not a framework conception or  abstraction for AS. KBS spire plugin logic treats the user defined claims parts as the carrier of the validated identifiers. So it has responsibility to get `policy.rego` first and then parse it.

I am not sure the current code shape could fit very well with this, but I would like to push forwawrd TC abstraction to help this, also for expansion of `initdata_claims`/`validated_selectors` parameters.

deps/verifier/Cargo.toml
```diff
@@ -117,6 +117,7 @@ reqwest-middleware = "0.5.1"
 http-cache-reqwest = { version = "1.0.0-alpha.6", default-features = false, features = [
     "manager-moka",
     "url-standard",
+    "reqwest-middleware",
```
Xynnn007, line 121: Let's add more context in the commit message why and what we see. `Doesn't build without this tweak.` shows little useful.

Xynnn007 (commented):
Overall I think it's a right way to go and nice to see the integration with existing ecosystem.

Some ideas facing to future

tools/kbs-client/src/main.rs
fitzthum: Yeah, I almost changed that in this PR, but as of today seems most correct for this to be a resource path rather than a resource URI. Can rethink in the future.

kbs/src/plugins/implementations/provisioner/mod.rs
fitzthum: Yeah, I am open to having a more standardized interface here as long as we continue to surface the identifiers.

kbs/src/plugins/implementations/spire/mod.rs
fitzthum: Not supported atm. I can documented that more clearly.

kbs/src/plugins/implementations/spire/mod.rs
fitzthum: Not sure I follow exactly. This list is not dynamic. This is just all of the known SPIRE k8s pod selectors.

kbs/docs/plugins/spire.md
fitzthum: Well, the idea behind validated identifiers was that they are a first-class member of the attestation token. They definitely are not specific to the SPIRE plugin. Otherwise, they would just be called selectors. The purpose is to have generic workload identifiers in the attestation token. They are implemented in the policy so that if you use a different runtime class, you can implement your own validated identifiers, but they are still intended to be a feature of the attestation token itself.

kbs/Makefile
fitzthum: I don't have a strong opinion on weather this needs to be enabled by default. Sounds like RH is interested in using it out of the box, so I enabled it. cc @beraldoleal 

Definitely want this to be an internal plugin tho.

deps/verifier/Cargo.toml
fitzthum: Updated

beraldoleal:
@fitzthum , I tested this against the code before your latest changes and it works great!

Got a delegated X.509-SVID end-to-end and confirmed in the SPIRE agent logs too (Creating X509-SVID ... spiffe_id="...").  So except from @Xynnn007 comments, LGTM.

One question on `scenarios/spire.yaml`: I had to run KBS as root to get access to the admin socket. Is that expected/required, or could we avoid that?

kbs/Makefile
bpradipt: I don't know what's the criteria for making a plugin default. 
But personally, I'm ok with SPIRE being default as there are a good number of use cases relying on SPIRE. 
Having said that, we should start providing some guidance on the qualification criteria for a default plugin.

fitzthum:
> One question on scenarios/spire.yaml: I had to run KBS as root to get access to the admin socket. Is that expected/required, or could we avoid that?

I think the Helm chart currently runs the KBS as root (root user, but un-privileged), so this is internally consistent. I think we could run both the SPIFFE chart and the KBS as a different user and it would work, but not sure that's in the scope of this PR.

kbs/Makefile
beraldoleal: I'm not aware of a hard requirement on our side for this to be default, our build can just enable the flag if needed, so no strong opinion here either way. cc @lmilleri .

deployment/helm-chart/README.md
```diff
@@ -353,6 +353,10 @@ Default **`values.yaml`** is intentionally small. Fixed on-disk paths for **Loca
...
 | kbs.service.port | int | `8080` | Service port for KBS; used by both the internal `ClusterIP` Service and the optional external `LoadBalancer` Service. |
+| kbs.spire | object | `{"adminSocketPath":"/run/spire/sockets/admin.sock","attestedSelectorValue":"true","enabled":false}` | SPIRE plugin: delivers SPIFFE X.509-SVIDs to attested guests via an existing SPIRE deployment's Delegated Identity API. Requires a KBS image built with the `spire-plugin` cargo feature, and a cluster that already runs a SPIRE server + agent with the admin (Delegated Identity API) socket enabled and mounted on the host (this chart does not deploy SPIRE itself). See `scenarios/spire.yaml` and `kbs/docs/plugins/spire.md`. |
+| kbs.spire.adminSocketPath | string | `"/run/spire/sockets/admin.sock"` | Path (or `unix://` URI) to the SPIRE Agent admin (Delegated Identity API) socket, as seen inside the KBS container. Must be reachable, e.g. via a hostPath volume mount configured through `kbs.extraVolumes`/`kbs.extraVolumeMounts` (see `scenarios/spire.yaml`). |
```
bpradipt: I think this must be only Path as the code already appends "unix://) to the socket path

tools/kbs-client/src/main.rs
Xynnn007: Ok. We can have a routine to do the change. Ok to add deprecating info in a separate PR.

deployment/helm-chart/README.md
mythi: I believe the KBS pod also needs some pod affinity set to make sure it lands on the same node than the spire agent for this local socket sharing work.

What's the idea for this in the TiC case where the kata pod cannot see host sockets?

kbs/Makefile
Xynnn007: Good. I am ok to have it as default.

But for plugin contribution, as @bpradipt mentioned, the qualification criteria for a default plugin should be made. 

I also want to mention the accepting criteria for a plugin, this is related to that one.

Given that https://github.com/confidential-containers/trustee/pull/1593 helps gRPC support as default, I lean towards to making future plugins as non-built-in whenever possible if it's not really related to the core attestation logic.

lmilleri:
@fitzthum SPIRE Server/Agent + trustee must run in a trusted domain, right? This is because the admin socket can be tampered with by host admin

kbs/docs/plugins/spire.md
Xynnn007: Good - I think we are closer than the wording suggests, so let me separate three layers.

**Agree:** the attestation token can carry workload-related fields that more than the SPIRE plugin may consume. That is why those fields should not be called "selectors".

**Where I disagree** is treating Validated Identifiers as an AS framework abstraction. Today AS only has a generic policy `extensions` path. `ear.trustee.identifiers` is produced by the default CoCo CPU policy as a thin projection of `init_data_claims.agent_policy_claims` (e.g. image name / UID). The kata/initdata coupling does not go away - it is moved into the AS policy and then read again by the plugin. Until there are multiple consumers that need the same stable projection *and* the "guest enforces this" half of "validated" is actually enforced, that extra hop does not buy much. Also, Trustee and CoCo are decoupled: AS can attest that initdata is hash-bound and expose structured claims, but it cannot assume or verify that the guest runtime enforces those fields. So the “guest enforces this” part of “validated” is outside Trustee’s scope; treating identifiers as a first-class AS abstraction over-promises what attestation can guarantee.

Also, "implemented in policy so other runtimes can supply their own" is exactly why this is a **policy convention**, not a framework primitive: consumers still depend on a specific policy profile, not on an AS API guarantee.

**Preferred split for CoCo + SPIRE:** 
(1) AS: keep structured `init_data_claims` (hash-bound initdata → JSON; already mostly done). 
(2) KBS SPIRE plugin: translate the needed fields from `init_data_claims` / `agent_policy_claims` into k8s selectors (the translation layer I asked about elsewhere in this PR). 
(3) Optional convenience extensions in a policy are fine - but they should be documented as a CoCo default-policy convention, not as an AS core abstraction.

Practically for this PR I am fine landing and I like the design the SPIRE plugin mechanics. What I would like to avoid cementing is documenting Validated Identifiers as a first-class AS abstraction, and naming the **generic** plugin parameter `validated_identifiers` (prefer something like `customized_claims` / full claims; the SPIRE plugin can interpret an identifiers-shaped path internally).

If helpful, I can follow up with a small commit/PR that shows the SPIRE plugin reading init_data_claims directly (and treating identifiers as optional), so the shape is concrete.

deployment/helm-chart/README.md
fitzthum: Hmm. I guess in a real cluster there would be a spire agent on every worker node.

atm this use case is not supported with TiC (which is somewhat on the back burner for me). I think the agent admin API can only be used locally, so we could probably need to add an agent alongside a confidential trustee deployment, using a join token to connect it to an existing SPIRE server. should be possible but definitely for later.

fitzthum:
> SPIRE Server/Agent + trustee must run in a trusted domain, right? This is because the admin socket can be tampered with by host admin

Yes, Trustee plus SPIRE are part of a trusted cluster. Normally you might hook SPIRE into your worker nodes, but in the CoCo case we probably don't want to do that. Technically, the only requirement here is that there is a SPIRE agent running next to Trustee, but the bigger picture is that the plugin is bootstrapping from a SPIRE setup that is trusted.

Regardless of the admin socket (which does have some validation logic; the process that connects to it must have a workload entry), the trustworthiness of your SPIRE deployment depends on the correct behavior of the SPIRE agents. Normally the node attestor ensures that the node is safe and that the agents can therefore be trusted. With coco, we usually cannot make that assumption. Thus, you generally should not have SPIRE agents on the untrusted worker nodes. You bootstrap from an existing SPIRE environment, but this should flow through the trusted clusters, not untrusted ones.

deployment/helm-chart/README.md
fitzthum: Yes. Let me fix.

kbs/Makefile
fitzthum: Yes. Let's discuss this on another PR/issue. I think built-in plugins still make sense in a lot of cases.

kbs/docs/plugins/spire.md
fitzthum: Ok. Let's try not to get this PR too bogged down. I guess we already went back and forth on this a lot in the original validated-identifiers issue and PR. I don't remember exactly where we left things, but I think we tried to compromise.

In fact, the current state of the code is a compromise. The validated identifiers are generated by the policy (+1 for you), but they are added to the attestation token itself as a first class element (+1 for me). The way I see this is that the identifiers are a first-class element of the attestation token but they are totally optional. So yes, I know you want to totally decouple the identifiers from the token, but remember this is a compromise already. Anyway, this is a bit tied up with https://github.com/confidential-containers/trustee/issues/1584 so maybe we can discuss there.

One note, 

> So the “guest enforces this” part of “validated” is outside Trustee’s scope; treating identifiers as a first-class AS abstraction over-promises what attestation can guarantee.

Not exactly. The idea is that the attestation policy/token reflect the soundness of the validated identifiers in the trust vector. I guess we could change things such that the identifiers are only added when when certain aspects of the trust vector are met (like the binaries are affirming, suggesting that the guest is going to enforce the identifiers), but I think it's probably better this way. 

It's certainly true that the identifiers depend on a number of different components to be valid, but I think that's just the nature of identity. Btw, the current identifiers for the images names might need to be tweaked to make sure that the image signature policy is being evaluated or something like that. There is future work to refine the identifiers for coco and create more.

deployment/helm-chart/README.md
mythi: > Hmm. I guess in a real cluster there would be a spire agent on every worker node.

Right, maybe that's what we get by default from the helm chart (daemonset). Isn't only one necessary in our use case or do I get this setup wrong?

kbs/docs/plugins/spire.md
Xynnn007: > In fact, the current state of the code is a compromise. The validated identifiers are generated by the policy (+1 for you), but they are added to the attestation token itself as a first class element (+1 for me). The way I see this is that the identifiers are a first-class element of the attestation token but they are totally optional. So yes, I know you want to totally decouple the identifiers from the token, but remember this is a compromise already. Anyway, this is a bit tied up with #1584 so maybe we can discuss there.

Current code already have initdata in claims in `init_data_claims`. It is not directly related to #1584

- https://github.com/confidential-containers/guest-components/blob/main/attestation-agent/attestation-agent/src/bin/ttrpc-aa.rs#L149C30-L149C43
- https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/backend.rs#L427
- https://github.com/confidential-containers/trustee/blob/main/kbs/src/attestation/coco/grpc.rs#L139
- https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/bin/grpc/mod.rs#L186
- https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/lib.rs#L239-L240
- https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/lib.rs#L378
- https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/lib.rs#L393-L407
- https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/ear_token/broker.rs#L157

The default CPU policy `ear.trustee.identifiers` block only copies a few paths out of `init_data_claims.agent_policy_claims` again. So this makes me to think more towards projection belongs in the default policy is not needed. The default policy should focus on TCB appraisal, and the SPIRE plugin should read `init_data_claims` directly for selector translation.

We have some compromise: in #987 I agreed to the generic policy `extensions` mechanism and, pragmatically, to landing customized fields that way (#1103 approve). I did not agree to treating Validated Identifiers as an AS framework primitive, or to naming the generic KBS plugin parameter `validated_identifiers`. 

For this PR I am fine with two paths and either is OK. And I am willing to help to get this in **ASAP**:

But at the same time we should care about the architecture and we should not leave too much budget to the arch. 

**Way 1: I push a small commiton this branch that **
- keeps the SPIRE delegated-identity flow
- has the SPIRE plugin read selector inputs from `init_data_claims` / `agent_policy_claims`
- removes or stops relying on the default-policy `ear.trustee.identifiers` projection for SPIRE
- renames the generic plugin parameter to something neutral (`customized_claims` or equivalent), treating any remaining identifiers extension as legacy/provisional only if needed for transition.

**Way 2: merge the SPIRE mechanics here with an explicit tracked follow-up **
1. Almost the same goal with way 1. But a PR will do refactor on current shape to ensure that SPIRE reads `init_data_claims`, default policy drops the identifiers projection, generic plugin API naming, and doc wording ("first-class" -> attested initdata + plugin-local selectors). I can own that follow-up.

I think you can contribution more selector logic directly inside the spire plugin to map initdata claims or anything else to the selector to flowerish the ecosystem. Policy might not be a proper place.

deployment/helm-chart/README.md
fitzthum: Hmm yeah the daemonset will deploy the agent to every node, so this should work fine. The only question is if that exposes the admin api more broadly than is necessary. It's possible that you could have a big cluster with only one KBS pod. In that case you could pin the KBS and the SPIRE agent admin API to one node. That might be best left to the consumer, though. wdyt? Setting up SPIRE isn't entirely in scope of our Helm chart.

kbs/docs/plugins/spire.md
fitzthum: I don't have time to go all the way back through #987, but `init-data-claims` and `validated-identifiers` are very different. This PR should make it obvious why we need both. We do not want to pass all the init-data-claims to the SPIRE agent as selectors. That's too much information, not all of the claims can be trusted, the claim names don't match workload entries, etc.

We need some additional layer on top of `init-data-claims`. The best place to do this is the policy. I think you were the one who suggested that in the first place. Although there are only a couple simple identifiers specified now, more are on the way, and some of them will rely on information besides the init-data-claims (e.g. the event log), which is not exposed to plugins. I tried to be clear that this was one of the uses for `validated-identifiers` from the beginning so it feels like we're going backwards a bit here.

Neither of your suggestions really works for me. If you are worried about plugin signature taking `validated_identifiers`, we can change that to be more generic and take `extensions` instead. I think that is worse, but it would be a fine compromise. I am also open to creating some kind of attestation context (not in this PR) as long as it does not collapse the trust vector as we discussed on another issue somewhere.

To me it's not important if the validated-identifiers are an "AS framework primitive" or not. It is fine to just call them an optional extension. Obviously I think they are useful and should be used more broadly. I think this should be the bedrock of our identity abstractions across the entire project, but this is still somewhat experimental so we'll see.

Xynnn007: It ia not to expose all init data claims to spire agent. I meant to organize selectors based on initdata claims in the plugin code logic. If you do not like the hard coding, a more flexible way is to use some non-code logic like a policy in the spiffe plugin as plugin config.

The core idea is that - AS verification logic ca be less coupled with coco logic thus making it more generic.

What the attestation policy can decided for now is only ensure the claims of runtime data and initdata are binded to the evidence. Other parts about whether the integrity of runtime that enforces policy is not always ensured. idk what will also be included inside into the eventlog, but we can support it when we meet it.

I really appreciate your work and like the design as it's simple, also i'd like to help get it merged, though i have a different design from a vertical perspective and not overlapped with this. Historically I made a lot of effort to make the components decoupled, and what i really care is the architecture. I do not want us to go back to a coupled one and making troubles for future when we can see it.

deployment/helm-chart/README.md
mythi: Looks like we have kbs.affinity so my comment is addressed

kbs/docs/plugins/spire.md
fitzthum: Well, here is how I see the scope of things. Validated identifiers are broader than the SPIRE plugin and they're broader than CoCo. You can use them with any runtime; you just tweak the policy. That also means that the SPIRE plugin can work with any runtime that specifies some identifiers. It is not a CoCo-specific feature.

I've always said that the AS policy is scoped to the runtime, and that fits perfectly with the identifiers, which hinge on the runtime. They really have the same scope as the trust vector, which they are closely related to. The default AS policy that we have today is scoped to CoCo. It is specific to how the podvm is booted (although normal CVMs are usually fairly similar). We haven't been explicit about this in the past, but this is precisely what is happening. The runtime determines how the reference values / boot artifacts are combined with the attestation primitives. 

Now, I would not want to put workload-specific logic into the attestation policy. I have always said that the attestation policy should change as little as possible. The identifiers follow this. They are runtime-specific, not workload-specific. The policy never changes for one runtime (i.e. coco). We just tease out a few key values from the runtime and highlight them in the attestation token, which I think is very powerful.

Point is, I don't think we're shifting the role of the AS and KBS here. The SPIRE plugin does require you to use the CoCo AS. I think that is very reasonable, but if you want to make it more generic, we could change the plugin interface to take all of the selectors. Then if some other attestation token wants to work with the plugin, they could add their own logic to pull selectors from another extension.

Xynnn007: Now the attestation token should only carry evidence-bound material (init_data_claims, runtime_data_claims) and policy-authored custom fields. validated identifiers are just a sub-scope of the `init_data_claims` in CoCo scenario (where `policy.rego` occurs), thus heavily requiring the runtime setting so coupled with CoCo.

A specific AS policy can couple with a specific runtime - but the default one should not. This means that users can customize their own policy. Policy only influences the trust tier and customized fields of ear. runtime_data_claims and initdata_claims are always included.

From architecture perspective, KBS plugins should consume the generic context including `customized_claims/initdata_claims/runtime_data_claims/..`; SPIRE-specific selector translation belongs in the SPIRE plugin. We should not require every token/runtime to emit selectors. If useful, selector-shaped data may appear inside `customized_claims`, but that remains optional policy output, not a framework contract. but also, spire inputs of `customized_claims` are now all from `initdata_claims`, so it's practical to directly use `initdata_claims`.

fitzthum: Well, I feel I have already addressed all of these points earlier in this thread and in the previous issue. I think we may be stuck here. Let's think if there is a way we can compromise here. I thought we already had compromised on the previous issues and the extension implementation.

> validated identifiers are just a sub-scope of the init_data_claims in CoCo scenario (where policy.rego occurs), thus heavily requiring the runtime setting so coupled with CoCo.

This is not accurate. Like I said, you can create validated-identifiers for any runtime, you just need some runtime-specific logic. For CoCo, many of these will be based on the Kata Agent policy, but not all of them (as I mentioned with the eventlog). The validated identifier itself is simply an optional extension for any runtime to specify identifiers for the workload that are backed by hw evidence. This is a natural fit for the attestation token (look at the definition of verifier claims, for instance). It feels arbitrary to say that we can have init-data claims, but not validated-identifiers. Remember that the init-data claims themselves aren't part of the evidence. The verifier adds them to the token after doing some checks. Not really very different from validated identifiers. Validated identifiers are just more fine-grained.

> A specific AS policy can couple with a specific runtime - but the default one should not.

The default runtime policy already is the CoCo policy. We developed it to work with CoCo and we test it with CoCo. Granted, many of the checks could be similar for VMs, but the recommendation is that people with other runtimes (i.e. VMs) should use a different policy. Perhaps we should explicitly change the name from default to CoCo to make this more clear. In theory we could have multiple policies upstream (default, coco, vm), but I don't think we have the bandwidth for that at the moment.

Last paragraph I have already answered. Validated identifiers are not SPIRE specific. The idea is to have an identity abstraction that we can use across various plugins and components. One of the nicest things about the plugin is that it requires very little configuration. Introducing another policy for this is a non-starter from a user perspective and inelegant architecturally imo.

Xynnn007: Ok. On your first point about compromise: we did discuss identifiers and I agreed to land customized identifiers in the default policy via the generic `extensions` machanism. That was allowing a policy-level convenience, not agreeing that Validated Identifiers become part of the AS/Trustee framework design. I think that distinction got blurred later - optional policy output was read as a framework contract. I want to clear that up.

About Runtime-scoped AS policy. It's fine. A dedicated `validated identifiers` abstraction on top of it is not. Policy-authored fields should go through `customized_claims` / `extensions`; we do not need a separate identifiers contract. The machanism is enough, so it would look duplicated to me to make a new one. Do you have any concrete examples where `extensions` cannot cover validated identifiers?

I am inspired by your ideas that hard-coding CoCo paths in the SPIRE plugin would indeed lock it to CoCo. So claims->selector mapping should live in the SPIRE plugin as **config or a small policy**, with a default CoCo profile for zero-config, and other runtimes overriding the mapping. Plugin stays generic; AS stays free of identity specialization.

What we call validated identifiers is just one arrangement of already-attested material. Specialization belongs to the consumer (KBS plugins), not to a Trustee-wide identity sub-domain. AS should keep producing evidence-bound claims plus optional customized fields.

For this PR. Temporary optional identifiers in policy are OK; cementing them as the generic plugin API or as an AS framework abstraction is not. Renaming the plugin input to `extensions` / claims context and keeping selector mapping SPIRE-local (configurable) is enough compromise for me to help move this forward.

kbs/src/plugins/implementations/spire/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+                }
+                value => selectors.extend(parse_selector(key, value, k8s_prefix, trustee_prefix)),
+            }
+        }
+    }
+
+    // Add the default selector, which is always present.
+    selectors.push(Selector::Generic((
+        trustee_prefix.to_string(),
+        attested_selector_value.to_string(),
```
bpradipt, line 186: This effectively means svid is issued without validated identifiers, no?

kbs/src/plugins/plugin_manager.rs
```diff
@@ -40,6 +43,7 @@ pub trait ClientPlugin: Send + Sync {
         path: &[&str],
         method: &Method,
         init_data: Option<&Value>,
+        validated_identifiers: Option<&Value>,
```
bpradipt: May be there is a simplification scope in future to converge, like what @Xynnn007 had suggested in one of the earlier issues

kbs/src/plugins/implementations/spire/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+    pub admin_socket_path: String,
+
+    /// Value used for the always-added `trustee:<value>` marker selector.
+    #[serde(default = "default_attested_selector_value")]
+    pub attested_selector_value: String,
+
+    /// SPIRE selector type used for `validated_identifiers` keys that match a
+    /// selector name SPIRE's built-in k8s workload attestor actually produces
+    /// (see `KNOWN_K8S_SELECTOR_KEYS`). By default, this will produce some
+    /// selectors that are identifical to what the k8s attestor would produce.
```
bpradipt: nit: s/identifical/identical

kbs/src/plugins/implementations/spire/mod.rs
```diff
@@ -0,0 +1,514 @@
...
+        _body: &[u8],
+        _query: &HashMap<String, String>,
+        _path: &[&str],
+        _method: &Method,
+    ) -> Result<bool> {
+        Ok(true)
+    }
+}
+
+#[cfg(test)]
```
bpradipt, line 337: rstest ?

kbs/src/plugins/implementations/spire/mod.rs
fitzthum: Whether the SVID is issued really depends on the workload entries. The idea of the default selector is that we always mark that the selectors are coming from Trustee rather than the normal workload attestors.

This does also mean that if there are no validated-identifiers, you will still get one selector. By default, SPIRE isn't going to do anything with this, but you could register a generic workload entry that grants all workloads validated by Trustee a basic SPIFFE-ID.

kbs/src/plugins/plugin_manager.rs
fitzthum: Yeah, probably going to change this to Extensions.

kbs/docs/plugins/spire.md
fitzthum: I think you've got the wrong idea about validated identifiers. I know it might not seem that different from the init-data-claims, but having an explicit identity abstraction in the attestation token is extremely powerful. 

Anyway, I changed the plugin interface to take extensions. Unfortunately there isn't a great way to generically extract all the extensions from the token, so we do it manually.

Moving the policy into the plugin is a non-starter for me, but I would be open to creating a separate extensions policy in the attestation agent. Adding more policies is not good for users, but this would keep the identifiers out of the default policy. Is that a workable compromise?

Xynnn007: I think that up to now, I got to know familiar enough about the validated identifiers.
1. It's used for identity abstraction
2. It's for different TEE models, s.t. CoCo/CVM/..
3. The aim is to provide a common view for upper consumer.
4. It includes things about the target environment where the context comes from attestation. (Note that I do not mention those who are not covered by attestation because it's out of scope for trustee)

If I missed anything, please point out. 

The aim is good, but there are some duplicated design. Note that we now support AS policy to add **any** defined field into the token and identity is just one use case. In the abstract, validated identifier is a subset of extensions.

From the AS point of view, it SHOULD NOT care about what the extensions mean, and it only work as a black box to generate the expected EAR token. What the token fields mean should be translated by the consumer, s.t. KBS.

My idea is that, validated identifier can be a use case for Trustee to handle the identity issues. It cannot go into the trustee core arch and bring it as a first class member.

About your two ideas
1. extensions. I'd like to help to push forward TC abstraction, and get all of them into a parameter.
2. The policy idea basically, is to help this plan to handle different runtimes. We can start with no policy but only a fixed logic aiming for CoCo.
3. what do you mean by new policy in attestation-agent? If it's used to tell what should be included in the attestation token, it shoud be on AS side rather than GC side.


Also, another way is to define the contract of spire plugin in kbs for `customized_claims`. Then we can have different attestation policies (than default) for different runtimes to map attestation context into the defined validated identifier. The pros are that still validated identifier does not go into the framework and it's a typical use case for different attestation policies. The cons are that users should select the specific policy to generate such claims, and this is a whole solution with help of both attestation policy and kbs plugin. Also, the policy should be carefully defined to generate expected claims but I think it's practical with examples. This also matches I once mentioned in https://github.com/confidential-containers/trustee/pull/1551#discussion_r3755818866 - to have a place to give examples of different attestation policies togetherly with their use cases.

fitzthum: > extensions. I'd like to help to push forward TC abstraction, and get all of them into a parameter.

Like I said, I am fine with changing the interface to the plugins in the future as long as we keep providing the extensions and as long as we do not reduce the attestation result to a boolean, as we have discussed before. I think we can handle this in a later PR. 

> The policy idea basically, is to help this plan to handle different runtimes. We can start with no policy but only a fixed logic aiming for CoCo.

I don't quite know what you mean here. 

> what do you mean by new policy in attestation-agent? If it's used to tell what should be included in the attestation token, it shoud be on AS side rather than GC side.

Sorry, meant attestation service. If you don't like having the validated identifier logic in the AS policy, we could introduce separate policies for the trust vector and the extensions.

There are a few things I didn't quite get in your comment, but one thing I want to address is that you've mentioned the idea of "Trustee core" a few times lately, but this isn't something we've ever really agreed on (although the concept is reasonable). In any case, I think that an identity abstraction is in fact a very important part of Trustee. Not having one has held us back. For instance, think about the other plugins that are in progress or were recently merged. They all would have benefited from an interface that describes the guest. This goes way beyond SPIRE support (and we definitely don't want to generate the identifiers in the SPIRE plugin).

I also want to reiterate that I think the attestation token is the perfect scope for this information. The attestation token is scoped to the runtime. Validated identifiers are scoped to the runtime. They are the perfect companion to the trust vector.

Xynnn007: > > The policy idea basically, is to help this plan to handle different runtimes. We can start with no policy but only a fixed logic aiming for CoCo.
> 
> I don't quite know what you mean here.

Yes I mentioned that "So claims->selector mapping should live in the SPIRE plugin as config or a small policy, with a default CoCo profile for zero-config, and other runtimes overriding the mapping." in previous comment, and you said "Moving the policy into the plugin is a non-starter for me, but I would be open to creating a separate extensions policy in the attestation agent."

What I meant is that AS token provides things like `attester_claims` in another PR you saw, and it includes `policy.rego`. Then, for different runtime (let's call CoCo one runtime. Different runtime would have different claims in the ear token) would have different logic to map the claims into Spire selectors. We can define the mapping as a sub policy that is handled by spire plugin of kbs when we meet more cases besides CoCo.


> Sorry, meant attestation service. If you don't like having the validated identifier logic in the AS policy, we could introduce separate policies for the trust vector and the extensions.

What do you mean by policy for `trust vector`? Current policy does this. We have a separate thread for extensions vs claims. And it would be not good to bring in another policy in AS. Can this PR just move the parse-and-handle-claims logic into spire plugin, than relying on AS policy? It's doable.

> In any case, I think that an identity abstraction is in fact a very important part of Trustee. Not having one has held us back. For instance, think about the other plugins that are in progress or were recently merged. They all would have benefited from an interface that describes the guest. This goes way beyond SPIRE support (and we definitely don't want to generate the identifiers in the SPIRE plugin).
>
> I also want to reiterate that I think the attestation token is the perfect scope for this information. The attestation token is scoped to the runtime. Validated identifiers are scoped to the runtime. They are the perfect companion to the trust vector.

The identity things are important. But we should have a boundry of the TEE claims and the identity. Note that when we talk about the identity, it's usually tied with access control / IAM. Without an overall contract of representation of the identity, it's hard for us to make clear how the identity looks like. You proposed to use validated identifiers to do this. FWIW - as I mentioned earlier, the proposed identifiers all come from the evidence + claims + eventlog, and are just transformations. So my idea is to keep the AS semantics clean. Unless the entire ecosystem reaches a consensus on the definition and representation of "identity"—in which case an extension or some claim could be implemented. Or, it is better to provide the raw material as claims for downstream consumers to process and construct the identity materials - identifiers themselves.

fitzthum: Like I said, moving the identifiers to the SPIRE plugin is a non-starter. The identifiers are not SPIRE specific. They are useful to other plugins. They are useful in the KBS policy. They are useful when simply looking at the attestation token. We already agreed to put them in the AS policy.

Now, I was suggesting a potential compromise where we could introduce an additional policy in the AS just for extensions (keep the existing policy just for the trust vector). Sounds like you don't want to do this. I don't really introduce an extra policy either, but I was trying to compromise.

I have been trying to build consensus about identity for more than a year. I proposed validated identifiers in October 2025 after months of discussion. I think this is the optimal solution, but we can't be sure yet, which is why it is an optional extension. We are not going to get anywhere unless we can provide people an e2e solution that they can actually try. This is the last piece of that puzzle, but now we are getting stuck discussing things that I thought we had settled months ago.

I have already addressed your last point many times. There is a big difference between init-data claims and validated identifiers. The attestation service is the optimal place to generate validated identifiers. An extension in the attestation token is the perfect place to expose validated identifiers. What you are suggesting (keeping the AS the same), is what we are already doing and it doesn't work. People don't understand how identity works in CoCo. They don't think we have identity support. This is what triggered all those discussions in the first place. We need an abstraction to capture identity. This should not be left to downstream consumers or to plugins.

Xynnn007: When I refer to "Trustee core", I mean the code outside individual plugin implementations, including AS, KBS, and the generic plugin API. 
My original design preference is that to include all raw claims in the attestation token, and let the plugin to decide how to use them. The concern is about making one particular identity representation a hard-coded framework abstraction before its validation semantics and interoperability contract are clear. But let me take a step back.

I think #1614 gives us a practical compromise:

1. We keep policy-defined EAR extensions.
2. A CoCo-scoped policy may derive and emit an `ear.trustee.identifiers` extension from attested claims.
3. The AS/KBS framework transports policy-defined Trustee extensions generically, for example as a map from extension name to JSON value.
4. The SPIRE plugin consumes the identifiers extension and translates it into SPIRE selectors.

This keeps identifier derivation in the AS policy, rather than moving it into SPIRE, while avoiding a generic plugin API that is specifically coupled to validated identifiers.

However, this requires an extension naming and schema contract as I mentioned. Otherwise the generic code has to manually extract a specific path `"/submods/cpu0/ear.trustee.identifiers"`.


A consumer such as the SPIRE plugin will necessarily know the extension name, but the generic plugin manager should not. It should receive only policy-defined appraisal extensions from a reserved namespace such as `ear.trustee.*`, rather than arbitrary token claims such as `iss` or `aud`. The contract should define not only the namespace, but also each extension's name, JSON schema and semantics.

There are still two questions behind the "validated identifiers" abstraction:

1. What conditions make an identifier "validated" across different environments?
2. What stable representation allows different runtimes and downstream consumers to interoperate?

Until those are specified, identifiers are mostly a useful projection of raw attested claims rather than a stronger framework-level primitive.

Finally, #1614 deliberately leaves open whether policy-derived values should remain EAR extensions or eventually live under `ear_verifier_claims`. That decision will affect this design. Therefore, I suggest keeping the generic transport/API neutral for now and documenting `ear.trustee.identifiers` as a CoCo policy/profile contract, without treating its current token location as irreversible.

With this split, I am fine with identifiers as part of extensions/customized-claims being generated by a CoCo AS policy and consumed by SPIRE and other plugins.

PR #1585, opened 2026-08-21
verifier: snp: expose reported TCB FMC in TEE evidence

00xc:
Turin and newer processors expose a new field in the reported TCB. Expose it in the parsed TEE evidence so that policies can use it.


Xynnn007 (approved):
Thanks @00xc for this. 

Per AMD folks the SNP evidence would meet some change in future. FYI

cc @hgowda-amd

fitzthum:
Any preference regarding reporting a claim with a null value vs not reporting the claim at all?

hgowda-amd:
> Any preference regarding reporting a claim with a null value vs not reporting the claim at all?

I would prefer omitting the claim when not applicable.

Issue #1308, opened 2026-04-30
Improve CoRIM extractor


fitzthum:
We have a very basic CoRIM extractor. There are a few things to improve.

- Support more types of identifiers (such as uuid)
- Support more types of triples (such as endorsements)
- Support checking signature
- Switch to other CoRIM crate
- Refine RV naming strategy

ariel-adam:
@fitzthum could you provide more information on what is this needed for and why so AMD can look at it and decide on the priority? 
Also, is there a standard or reference you could refer us to?

fitzthum:
I think the important thing is that when we produce our reference values in the RV repo, we can sign them. We should then be able to ingest that signed manifest in Trustee and check the signature.

Currently none of our extractors check the manifest signature and the KBS API only supports the sample extractor. We have another interface (RVPS API) for loading other formats, but it's rarely used.

This format could be CoRIM, but doesn't need to be. I am not aware of any AMD-specific links to CoRIM, but this is an important gap to close with reference values.

PR #1586, opened 2026-08-21
rvps: add ListReferenceValues RPC and rvps-tool list command

AnkushUjawane:
- RVPS previously only supported registering a provenance and querying a single reference value by ID; there was no way to see what's currently registered without already knowing the exact IDs.

- Adds a new ListReferenceValues gRPC RPC, backed by the existing KeyValueStorage::list() the storage backen already provides (the same mechanism policy-engine's list_policies uses), and wires it through to a new 'rvps-tool list' subcommand.

- rvps/src/rvps_api/reference.rs is regenerated from protos/reference.proto via the rebuild-grpc-protos feature; the checked-in file still builds correctly without that feature enabled for normal contributors.

- Manually verified end-to-end: started a real rvps instance, registered two reference values via rvps-tool register, then confirmed rvps-tool list correctly returns both registered IDs.

### DIFF
--- a/attestation-service/src/bin/grpc/mod.rs
+++ b/attestation-service/src/bin/grpc/mod.rs
@@ -35,8 +35,8 @@ use crate::rvps_api::{
     reference_value_provider_service_server::{
         ReferenceValueProviderService, ReferenceValueProviderServiceServer,
     },
-    ReferenceValueQueryRequest, ReferenceValueQueryResponse, ReferenceValueRegisterRequest,
-    ReferenceValueRegisterResponse,
+    ReferenceValueListRequest, ReferenceValueListResponse, ReferenceValueQueryRequest,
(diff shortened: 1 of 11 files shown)


fitzthum (commented):
CI unhappy. This seems fine generally.

The RVPS API isn't really secure at the moment (none of the extractors check the signature of the manifest, so anyone could add RVs), so it is typically not exposed. This is something we could work on.

AnkushUjawane:
Good context, thanks. This PR only adds read access to reference value IDs — it doesn't touch the signature verification gap on the write side. Happy to open a separate issue for that if useful, but keeping it out of scope here.

Xynnn007 (approved):
LGTM Thanks. The reading keys and the validate-insert keys are decoupled parts. We can track them regardless this one. Related issue https://github.com/confidential-containers/trustee/issues/1308

PR #1590, opened 2026-08-24
verifier: add hardware_type claim to parsed evidence

Xynnn007:
Add a normalized hardware_type claim in each verifier so policy engines can identify the attested platform without inferring it from Tee enum values or EAR submodule labels. Hoist the claim to annotated-evidence top level in the AS EAR broker alongside init_data and report_data.

### DIFF
--- a/deps/verifier/src/hardware_type.rs
+++ b/deps/verifier/src/hardware_type.rs
@@ -1,0 +1,43 @@
+// Copyright (c) 2026 by Alibaba.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Normalized `hardware_type` claim for verifier parsed evidence.
+//!
+//! Prefer `"hardware_type": types::...` in local `json!` builders.
+//! Use [`insert`] when claims come from serde or other helpers.
+
+use serde_json::Value;
+
+/// Claim key for the normalized hardware platform identifier.
+pub const CLAIM: &str = "hardware_type";
+
+pub mod types {
+    pub const INTEL_SGX: &str = "intel-sgx";
+    pub const INTEL_TDX: &str = "intel-tdx";
+    pub const AMD_SNP: &str = "amd-snp";
+    pub const AZURE_SNP_VTPM: &str = "azure-snp-vtpm";
+    pub const AZURE_TDX_VTPM: &str = "azure-tdx-vtpm";
+    pub const ARM_CCA: &str = "arm-cca";
+    pub const HYGON_CSV: &str = "hygon-csv";
+    pub const HYGON_DCU: &str = "hygon-dcu";
+    pub const IBM_SE: &str = "ibm-se";
+    pub const TPM: &str = "tpm";
+    pub const NVIDIA_HOPPER: &str = "nvidia-hopper";
+    pub const NVIDIA_BLACKWELL: &str = "nvidia-blackwell";
+    pub const NVIDIA_LS10: &str = "nvidia-ls10";
+    pub const NVIDIA_PPCIE: &str = "nvidia-ppcie";
+    pub const NVIDIA_DPU: &str = "nvidia-dpu";
+    pub const SAMPLE: &str = "sample";
+    pub const SAMPLE_DEVICE: &str = "sample-device";
+}
+
+/// Insert `hardware_type` into claims assembled outside a local `json!` builder.
+pub fn insert(claims: &mut Value, hardware_type: &str) {
+    if let Some(obj) = claims.as_object_mut() {
+        obj.insert(
+            CLAIM.to_string(),
+            Value::String(hardware_type.to_string()),
+        );
+    }
+}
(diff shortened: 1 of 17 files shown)


mkulke (commented):
how is this different from the "tee" identifier that we use elsewhere? if the current location of "tee" in the claim is impractical for consumers, then we can move that around or duplicate it, but adding a new definition "hardware_type" might be confusing.

fitzthum:
I thought the idea was to introduce a new tag that was more specific than the existing Tee class and Tee enum. This  hardware type seems redundant. 

What I suggested is a field called `device_name` that would point to a specific device..

Also, I am strongly against introducing a centralized enum to keep track of this.

Xynnn007:
> how is this different from the "tee" identifier that we use elsewhere? if the current location of "tee" in the claim is impractical for consumers, then we can move that around or duplicate it, but adding a new definition "hardware_type" might be confusing.

verifiers do not emit a standalone `tee` field today.  Attestation Token will contain nested remaining claims under the serde variant name (`Tdx`, `Nvidia`, …). `device_name` is the explicit, stable string policies can match on.

There are some contexts https://github.com/confidential-containers/kbs-types/issues/90#issuecomment-5091097204 and https://github.com/confidential-containers/trustee/issues/1364.

`Tee` in kbs-types identifies the attestation **method** (attester/verifier pair), not the physical device.

What policy authors still need is a way to tell platforms apart *within* a `Tee`, e.g. `Tee::Nvidia` covers Hopper, Blackwell, and LS10;
Today they infer this from `tee_class` + EAR submod labels (`cpu0`, `gpu0`), which is imprecise and, for same-class multi-device attestations, non-deterministic e.g. a hopper + a blackwell.

This PR adds a claim field so policies can target `intel-tdx`, `nvidia-hopper`, etc. without guessing from submod names.

@fitzthum 

1. removed the global registry and turn it into an alias name for `String`, thus verifier can choose the actual value locally.
2. changed name to `device_name`.

deps/verifier/src/hardware_type.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use serde_json::Value;
+
+/// Claim key for the normalized hardware platform identifier.
+pub const CLAIM: &str = "hardware_type";
+
+pub mod types {
+    pub const INTEL_SGX: &str = "intel-sgx";
+    pub const INTEL_TDX: &str = "intel-tdx";
+    pub const AMD_SNP: &str = "amd-snp";
```
mkulke, line 18: AMD_SEV_SNP might be more canonical

deps/verifier/src/hardware_type.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub const AZURE_SNP_VTPM: &str = "azure-snp-vtpm";
```
mkulke, line 19: That should not exist as device_name since it is an attestation/verificatiom method. It should just be snp/tdx, correct?

mkulke (commented):
Thanks for the explanation, the distinction of att/ver method vs device name makes sense, some nits remaining

fitzthum:
> What policy authors still need is a way to tell platforms apart within a Tee, e.g. Tee::Nvidia covers Hopper, Blackwell, and LS10;

Yes, I agree with this. We should introduce something more fine-grained than the Tee enum. That didn't seem like what the first iteration of this PR was doing. The first version was basically just combining the Tee enum with the Tee class. We already expose that information. The tee class can already differentiate different kinds of devices covered by the same attester/verifier. If we make a new field, it should be more specific than that, like what is the actual name of the device.

Note that figuring out the device_name is a bit tricky. A number of the examples you have given don't seem like device names. For instance, `amd-snp` isn't a device. That's a technology. The device is something like `AMD Milan 7663`. It's tricky to figure out all the best ways to name things, though. Do you include the generation or not? idk. That's why I'm saying that it should be up to the individual verifiers.

mythi:
> That's why I'm saying that it should be up to the individual verifiers.

I also think this is better but independent of where it's set, we'd need a better definition of what to have in it and the expected use-cases for this field (could not find any from the conversation above). I can see at least NRAS puts quite detailed device information in the claims so what do we miss?

For Intel, this can also be ambiguous (e.g., `fmspc` does not say if it's SGX, TDX, or Azure TDX) which we want to avoid.

Xynnn007:
emm. Ok, let's get a step back. The idea - current labels of EAR, e.g. `cpu0`, is not quite a good abstraction. also for `gpu` or `switch`. different platforms have different attestation technique. For example `TDX` is different from `Azure-vtpm-tdx`, and `azure-vtpm-tdx` is also different from `vtpm`. So I want that we have a field, to tell the technique.

My example regarding Nvidia GPUs is a bit narrow in scope, only shows that distinguish Hopper and Blackwell hardware. (Hopper and Blackwell have different attestation technique. The ROTs of them are slightly different.) Thus fall into the case where term "hardware" is used everywhere. The aim is to distinguish "technique" broader than hardware. The technique reflects what is protected/attested and how.


To me, the most expected represention is that submodule name changed to `tdx` (or `intel-tdx`), `nvidia-hopper`, `vtpm`, `amd-snp`, ... Note that they now are `cpu0` or `gpuX`. This will have a exact place and we give some documents about the possible values and what do they mean.

Sorry I put the PR but looks like it's making things worse. Please skip the codes, and take a look if my worry makes sense and also how can we handle them.

PR #1592, opened 2026-08-25
kbs: add resource listing to the resource plugin

AnkushUjawane:
GET /kbs/v0/resource (no repository/type/tag) now lists all stored resource paths instead of failing to parse an empty path as a ResourceDesc.

Adds list_secret_resources to the StorageBackend trait with a default implementation that reports the operation as unsupported, and a concrete implementation for KvStorage backed by the existing KeyValueStorage::list(). The four KMS-based backends (AWS, GCP, Aliyun, Vault) are left unimplemented for now rather than guessed at, since I have no way to verify against real cloud accounts.

Listing requires admin authentication (validate_auth returns true for GET with an empty path), unlike retrieving a single resource, which is authorized separately via attestation - listing exposes which resources exist, which is a different trust boundary than releasing one specific resource's contents to an attested guest. The response is not JWE-encrypted, since it isn't attested-guest secret material.

Manually verified end-to-end against a running kbs + grpc-as: empty list, registering a resource, listing again to see it, and confirming an unauthenticated request to the same endpoint is rejected with 401.

### DIFF
--- a/kbs/src/plugins/implementations/resource/backend.rs
+++ b/kbs/src/plugins/implementations/resource/backend.rs
@@ -39,6 +39,11 @@ pub trait StorageBackend: Send + Sync {
 
     /// Delete secret resource from repository
     async fn delete_secret_resource(&self, resource_desc: ResourceDesc) -> Result<()>;
+
+    /// List all resource paths ("repository/type/tag") stored in this backend, across every repository.
+    async fn list_secret_resources(&self) -> Result<Vec<String>> {
+        bail!("list_secret_resources is not supported for this backend")
+    }
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -185,6 +190,10 @@ impl ResourceStorage {
             .with_label_values(&[&format!("{}", resource_desc)])
             .inc();
         self.backend.read_secret_resource(resource_desc).await
+    }
+
+    pub(crate) async fn list_secret_resources(&self) -> Result<Vec<String>> {
+        self.backend.list_secret_resources().await
     }
 }
 
--- a/kbs/src/plugins/implementations/resource/mod.rs
+++ b/kbs/src/plugins/implementations/resource/mod.rs
@@ -43,6 +43,10 @@ impl ClientPlugin for ResourceStorage {
                 self.set_secret_resource(resource_description, body).await?;
                 Ok(vec![])
             }
+            "GET" if path.is_empty() => {
+                let resources = self.list_secret_resources().await?;
+                Ok(serde_json::to_vec(&resources)?)
+            }
             "GET" => {
                 let resource_description = ResourceDesc::try_from(&resource_desc[..])?;
                 let resource = self.get_secret_resource(resource_description).await?;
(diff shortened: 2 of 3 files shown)


kbs/src/plugins/implementations/resource/backend.rs
```diff
@@ -39,6 +39,11 @@ pub trait StorageBackend: Send + Sync {
 
     /// Delete secret resource from repository
     async fn delete_secret_resource(&self, resource_desc: ResourceDesc) -> Result<()>;
+
+    /// List all resources stored in this repository
```
fitzthum: nit: `repository` -> `backend `. We sometimes use the term `repository` to refer to the first parameter of the resource path. That is not being invoked here, though. This would give you everything across all repositories. May want to fix the comment above as well.

kbs/src/plugins/implementations/resource/mod.rs
```diff
@@ -62,24 +66,26 @@ impl ClientPlugin for ResourceStorage {
...
+        path: &[&str],
         method: &Method,
     ) -> Result<bool> {
         if method.as_str() == "POST" || method.as_str() == "DELETE" {
             return Ok(true);
         }
-
+        if method.as_str() == "GET" && path.is_empty() {
+            return Ok(true);
+        }
```
fitzthum, line 77: It might be confusing that getting a specific resource is guest operation but the same request with no path is an admin operation. Any way to clarify this?

fitzthum (commented):
Seems ok. A couple comments.

kbs/src/plugins/implementations/resource/backend.rs
AnkushUjawane: fixed - clarified it means "backend" and spans all repositories, not just one.

Xynnn007 (approved):

kbs/src/plugins/implementations/resource/mod.rs
AnkushUjawane: reusing the same path with implicit semantics based on whether it's empty is easy to misread. Rather than just documenting around it, I could give listing its own top-level route (e.g. GET /kbs/v0/resources, plural) handled explicitly in api_server.rs's outer match, the same way resource-policy is - structurally distinct from the per-resource plugin dispatch instead of an empty-path special case inside it. Want me to go that direction, or would stronger comments in the current shape be enough?

kbs/src/plugins/implementations/resource/backend.rs
```diff
@@ -39,6 +39,13 @@ pub trait StorageBackend: Send + Sync {
 
     /// Delete secret resource from repository
     async fn delete_secret_resource(&self, resource_desc: ResourceDesc) -> Result<()>;
+
+    /// List all resource paths ("repository/type/tag") stored in this backend, across every repository.
+    // Not every backend supports listing;
+    // the default implementation reports this explicitly rather than silently returning nothing.
```
fitzthum: Probably want three slashes for all the lines of this comment. This comment isn't really necessary anyway, though. Code is self-explanatory.

AnkushUjawane: Trimmed it down

fitzthum:
Ok. Please fix your commit. The title and description got overwritten. Otherwise looks fine.

AnkushUjawane:
Fixed, back to the original title and description now.

fitzthum (approved):
LGTM

PR #1593, opened 2026-08-25
Enable external-plugin by default

bpradipt:


bpradipt:
The plugin config is not yet exposed via helm chart. I'll look into it in a separate PR.

kbs/docs/config.md
```diff
@@ -468,11 +455,13 @@ with `name = "external"` owns all backends via a `backends` inline array:
...
 backends = [
-  { name = "my-plugin", endpoint = "https://localhost:50051", tls_mode = "tls", ca_cert_path = "/etc/kbs/plugin-ca.pem" },
+  { name = "my-plugin", endpoint = "https://localhost:50051", ca_cert_path = "/etc/kbs/plugin-ca.pem" },
 ]
 ```
 
-Each backend is reachable at `/kbs/v0/external/<name>/...`.
+Each backend is reachable at `/kbs/v0/external/<name>/...`. The endpoint scheme
+selects the transport: `http://` is plaintext, `https://` is TLS. A `tls_mode`
+key is accepted for backwards compatibility but ignored.
```
Xynnn007: I am hesitating here. No code will handle `tls_mode` anymore, so whether we need to mention it here?

kbs/src/plugins/implementations/external_plugin.rs
```diff
@@ -111,9 +111,11 @@ pub struct BackendConfig {
...
 /// ```
 ///
-/// For each backend, if `tls_mode` is omitted, TLS is implied by an `https://` endpoint
-/// (and `ca_cert_path` is then required) and plaintext by `http://`. The explicit
-/// `tls_mode` + `ca_cert_path` form remains supported.
+/// The endpoint scheme decides the transport for each backend: `https://` means
+/// TLS (and `ca_cert_path` is then required), `http://` means plaintext. A
+/// `tls_mode` key is still accepted so older configs keep parsing, but it is
+/// ignored — a backend naming `tls_mode = "tls"` on a `http://` endpoint
+/// connects in plaintext.
```
Xynnn007: same.

kbs/docs/ext_plugin.md
```diff
@@ -236,15 +235,19 @@ running and the endpoint in `kbs-config.toml` matches its bind address.
...
-`ca_cert_path` must be present in the same backend table as `tls_mode`. Verify
-cert paths exist and are readable.
+The endpoint scheme is what selects the transport: `http://` is plaintext,
+`https://` is TLS. A `https://` endpoint requires `ca_cert_path` in the same
+backend table; verify the path exists and is readable.
+
+> A `tls_mode` key is accepted for backwards compatibility but is **ignored** —
+> the scheme decides. A backend that says `tls_mode = "tls"` while pointing at
+> a `http://` endpoint connects in plaintext without complaint. Set the scheme,
+> not `tls_mode`.
```
Xynnn007: same

Xynnn007 (approved):
The code and direction LGTM. Only one nit about the words - I have no strong preference upon this.

kbs/docs/config.md
bpradipt: Yeah, I was also in two minds, but this key exists and hence I thought it's better to at least mention it here

bpradipt: May be I can drop this tls_mode completely in a separate commit or PR. Wdyt?

Xynnn007: it sounds good to me. Could you help to add a commit in this PR? I guess we can give an end to the story in place : )

Xynnn007 (approved):

fitzthum (approved):
LGTM

PR #1594, opened 2026-08-25
policies: make allow all policy more permissive

### LINKED ISSUE #1575: Allow All policy does not allow access to plugins
Arguably this is a good thing, since it's a bit more restrictive, but personally I find it confusing that the allow-all policy (which you can add via the kbs-client) does not allow everything. It only allows requests to the resource backend. See [here](https://github.com/confidential-containers/trustee/blob/main/kbs/sample_policies/allow_all.rego)

fitzthum:
Make the allow-all policy truly allow all requests. Previously this policy allowed all resource requests but did not permit any plugin requests.

It seems more clear to have this truly allow everything.

Arguably, we could introduce a second allow-all policy to differentiate these two cases, but let's keep it simple.

Fixes: #1575

### DIFF
--- a/kbs/sample_policies/allow_all.rego
+++ b/kbs/sample_policies/allow_all.rego
@@ -1,10 +1,3 @@
 package policy
 
-default allow = false
-
-plugin = data.plugin
-
-allow if {
-	plugin == "resource"
-}
-
+default allow = true


bpradipt (approved):
/lgtm
Makes sense.

Xynnn007 (approved):

PR #1595, opened 2026-08-25
kbs-client: move init-data to be an option/flag

### LINKED ISSUE #1574: KBS Client takes init-data as an argument rather than an option
Help looks like 

```
Usage: kbs-client get-resource [OPTIONS] --path <PATH> [INIT_DATA]                                                                 
                                                                                                                                   
Arguments:
  [INIT_DATA]  InitData plaintext

Options:
      --path <PATH>
          KBS Resource path, e.g my_repo/resource_type/123abc (routed to the default "resource" plugin), or a full `kbs://` / `kbs+<plugin>://` URI to target a specific plugin, e.g. `kbs+spire:///x509-svid`. Document: https://github.com/confidential-containers/attestation-agent/blob/main/docs/KBS_URI.md
```

That's not really ideal. The init-data should be an option/flag rather than an argument imo

fitzthum:
Move init-data to an optional flag rather than a parameter. This is more
consistent with the other options.

Fixes: #1574

### DIFF
--- a/tools/kbs-client/src/main.rs
+++ b/tools/kbs-client/src/main.rs
@@ -67,6 +67,7 @@ enum Commands {
         /// token, allowing KBS policies to access raw init-data claims.
         /// The per the InitData specification, the plaintext should be toml or json
         /// depending on what the runtime expects.
+        #[clap(long)]
         init_data: Option<String>,
     },
 
@@ -81,6 +82,7 @@ enum Commands {
         tee_key_file: Option<PathBuf>,
 
         /// InitData plaintext. See GetResource
+        #[clap(long)]
         init_data: Option<String>,
     },
 }
--- a/tools/kbs-client/src/main.rs
+++ b/tools/kbs-client/src/main.rs
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-//! A simple KBS client for test.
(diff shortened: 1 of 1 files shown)


Xynnn007:
looks like `cargo fmt` is needed

fitzthum:
~~GitHub broken. Hopefully CI will run again soon.~~

tools/kbs-client/src/main.rs
```diff
@@ -67,6 +67,7 @@ enum Commands {
         /// token, allowing KBS policies to access raw init-data claims.
         /// The per the InitData specification, the plaintext should be toml or json
         /// depending on what the runtime expects.
+        #[clap(long, value_parser)]
```
pawelpros:
```suggestion
        #[clap(long)]
```

value_parser is my opinion is reduntant here -> It will use standard validation for String so in my opinion not needed

tools/kbs-client/src/main.rs
```diff
@@ -81,6 +82,7 @@ enum Commands {
         tee_key_file: Option<PathBuf>,
 
         /// InitData plaintext. See GetResource
+        #[clap(long, value_parser)]
```
pawelpros:
```suggestion
        #[clap(long)]
```

pawelpros (approved):

tools/kbs-client/src/main.rs
fitzthum: Good point. Dropped.

mythi (approved):

Issue #1412, opened 2026-06-10
Versionalize for Evidence Formats


Xynnn007:
We usually encounter updates to evidence formats across different attesters/verifiers. For example, IBM SE added a "report data" field, and TDX historically added an "eventlog" sub-item. To ensure synchronization between attesters/verifiers, I suggest introducing a version number for each platform's evidence. The initial ideas are:

1. Version numbers are not directly related to backward compatibility.

2. Incompatible version numbers will result in errors and failures, reported through error logs to reduce debugging difficulty.

3. Document version iteration changes for different pieces of evidence.

fitzthum:
I agree, but rather than introducing these all at once, we should try to enforce that the next time any change is made to a particular evidence struct, that platform must adopt versioning. This limits breaking changes to PRs that would have ahd them anyway.

mythi:
`az_*_vtpm` did something like this already.

PR #1596, opened 2026-08-27
feat(verifier): version TEE evidence with an explicit version field

lmilleri:
## What

Give TEE evidence an explicit format version. `TeeEvidence` changes from a
bare `serde_json::Value` to:

```rust
pub struct TeeEvidence {
    pub version: u8,             // top-level `version` field, 0 = legacy
    pub data: serde_json::Value, // TEE-specific payload, verbatim
}
```

Why

Today a changed evidence format just fails to deserialize with a cryptic
error. An explicit version lets the verifier reject evidence it's too old to
understand with a clear message, and gives attesters/verifiers a stable
contract to evolve against.

How

- Version is read from an optional top-level version field: absent → 0
  (legacy), present but not a u8 → hard error. The payload is passed to the
  TEE verifier untouched.
- Version support lives on the verifier: Verifier::max_supported_version()
  (default 0) plus a generic check_evidence_version() guard run before
  evaluate(). az_snp_vtpm / az_tdx_vtpm opt in to version 2.

Fully backward compatible — existing unversioned evidence is treated as
version 0; no attester changes required.

Refs: #1412

### DIFF
--- a/deps/verifier/src/lib.rs
+++ b/deps/verifier/src/lib.rs
@@ -214,8 +214,117 @@ pub async fn to_verifier(
 }
 
 pub type TeeEvidenceParsedClaim = serde_json::Value;
-pub type TeeEvidence = serde_json::Value;
 pub type TeeClass = String;
+
+/// Evidence as produced by an attester: an explicit format version alongside
+/// the opaque, TEE-specific payload.
+///
+/// # Wire contract
+///
+/// Evidence is a single JSON object. Its format version is carried by an
+/// optional top-level `version` field; everything else is the TEE-specific
+/// payload:
+///
+/// ```json
+/// { "version": 1, "tpm_quote": { /* ... */ } }   // explicit version 1
+/// { "attestation_report": { /* ... */ } }        // legacy, implicit version 0
+/// ```
+///
+/// The `version` must be a top-level field named exactly `version` and must
+/// be representable as a `u8`. If `version` is not provided, it defaults to `0`
+/// (legacy).
+///
+/// [`data`](Self::data) holds the whole payload verbatim (including the
+/// `version` field, if any). TEE verifiers deserialize [`data`](Self::data) into their
+/// own concrete evidence type.
+///
+/// Which versions are *supported* is a separate concern that belongs to the
+/// [`Verifier`] (see [`Verifier::max_supported_version`]).
+#[derive(Clone, Debug, PartialEq)]
+pub struct TeeEvidence {
+    /// The evidence format version, read from the top-level `version` field at
+    /// parse time. Legacy evidence without an explicit version is `0`.
+    pub version: u8,
+    /// The raw TEE-specific evidence payload, verbatim.
+    pub data: serde_json::Value,
+}
+
+/// Read a top-level `version` value as a `u8`. Returns `Ok(None)` when absent,
+/// `Err` when present but not representable as a `u8`.
+fn parse_version_field(value: &serde_json::Value) -> Result<Option<u8>> {
+    let Some(v) = value.get("version") else {
+        return Ok(None);
+    };
+
+    let Some(version) = v.as_u64().and_then(|n| u8::try_from(n).ok()) else {
+        bail!("invalid evidence version: {v}");
+    };
+
+    Ok(Some(version))
+}
+
+impl From<serde_json::Value> for TeeEvidence {
+    fn from(value: serde_json::Value) -> Self {
+        let version = parse_version_field(&value).unwrap_or(None).unwrap_or(0);
+        TeeEvidence {
+            version,
+            data: value,
+        }
+    }
+}
+
+impl<'de> Deserialize<'de> for TeeEvidence {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        // Read the payload as an opaque value once, then validate the version
+        // contract up front. This happens before any TEE-specific variant
+        // matching, so a malformed version yields a clear error rather than an
+        // opaque "failed to deserialize" later on.
+        let value = serde_json::Value::deserialize(deserializer)?;
+        let version = parse_version_field(&value)
+            .map_err(<D::Error as serde::de::Error>::custom)?
+            .unwrap_or(0);
+        // Qualified: `use anyhow::*` shadows `Ok` with `anyhow::Ok`.
+        std::result::Result::Ok(TeeEvidence {
+            version,
+            data: value,
+        })
+    }
+}
+
+impl serde::Serialize for TeeEvidence {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        // The payload is the wire form: it already carries `version`, so emitting
+        // it verbatim round-trips back to the same evidence.
+        self.data.serialize(serializer)
+    }
+}
+
+/// Reject evidence whose version exceeds what `verifier` supports, with a
+/// uniform, user-facing error message.
+///
+/// This is the generic version guard: it belongs to the dispatch layer, not to
+/// any individual TEE, and should be called before [`Verifier::evaluate`].
+pub fn check_evidence_version(
+    verifier: &dyn Verifier,
+    evidence: &TeeEvidence,
+    tee: &Tee,
+) -> Result<()> {
+    let max = verifier.max_supported_version();
+    let version = evidence.version;
+    if version > max {
+        bail!(
+            "Unsupported {tee:?} evidence version {version}. This verifier \
+             supports evidence versions 0-{max}",
+        );
+    }
+    Ok(())
+}
 
 /// Insert a verified eventlog as a `uefi_event_logs` claim -- a no-op if
 /// `ccel` is `None`.
@@ -265,6 +374,16 @@ impl ToHex for ReportData<'_> {
 
 #[async_trait]
 pub trait Verifier {
+    /// The highest evidence format version this verifier understands.
+    ///
+    /// Support for evidence versions is a property of the verifier, not of the
+    /// evidence itself. The default is `0` (legacy evidence only); a verifier
+    /// opts in to newer formats by overriding this and handling the additional
+    /// versions in [`evaluate`](Verifier::evaluate).
+    fn max_supported_version(&self) -> u8 {
+        0
+    }
+
     /// Verify the hardware signature.
     ///
     ///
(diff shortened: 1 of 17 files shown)


Xynnn007:
This PR relies on the `serde_json::from_slice` succeeding. This means if a format-changed evidence comes, there can be error like "failed to deserialize".

My original idea in the issue was to considering add an extra field in either kbs-protocol level, or evidence level to directly tell the evidence version. But surely, if a request comes w/o this field, it will fail too.

There must be some more ideas cc @confidential-containers/guest-components-maintainers

lmilleri:
Thanks @Xynnn007 you're right, this is a genuine gap in the current approach.
As it stands, validate_version() can only run after the evidence deserializes into one of the untagged variants. With the untagged pattern, the variant is chosen by which structure parses, so a format-changed evidence (new required field, renamed/removed field, changed type) matches neither V1 nor V0, serde_json::from_slice fails first, and the caller sees "failed to deserialize" instead of the clear version error.

What about implementing a simple two-phase parse, for example:

```
// Phase 1: read the version directly off the raw JSON, no variant matching.
let version = value.get("version").and_then(Value::as_u64).unwrap_or(0);
Self::check_version(version as u32)?;   // clear error if out of range

// Phase 2: full typed parse (untagged, newest variant first).
serde_json::from_value(value).with_context(|| format!("Failed to deserialize {} evidence", Self::tee_name()))
```
Of course the limitation will remain if the "version" is not in the json payload. In that case a parse error message will be returned if no evidence variant is parsed properly.

fitzthum:
cc @mkulke this sort of seems like a generalization of your compat approach (which I like)

mkulke (commented):
I think this it's a good idea to somehow make evidence versions more explicit, however I'm not sure a trait is the right approach here. A trait is a set of capabilities that you add to a struct, so an Evidence struct will have a e.g. function "max_supported_version()", which is a bit twisted. The verifier supports versions of evidence. The evidence itself doesn't support a version, it has a version.

Currently trustee's `Evidence` isn't really typed:

```rust
pub type TeeEvidence = serde_json::Value;
```

We could turn the alias into a struct:

```rust
pub struct TeeEvidence {
    version: u8,
    data: serde_json::Value,
}
```

A generic serializer could then extract the version (default to 0) and pass it to the TEE's evaluate() implementations. The implicit contract for evidence would be: if you have a top-level property called "version", that we will deserialize it into an u8 and declare this the version of the evidence. has to be top-level, cannot be another name ("v"), has to be u8 (not "2026-05-24").

If the goal is to have a reliable, explicit version contract between attesters and verifiers it might make sense to bump the kbs_protocol and make a version tag a requirement, though.

lmilleri:
@mkulke thanks, this is a fair point, I like the TeeEvidence struct idea.
I'll try to rework the PR along these lines.

lmilleri:
The plan is to have a three-phase approach. Please confirm if this aligns with your expectations:

**phase 1**: Explicit version on `TeeEvidence` (this PR)

**phase 2**: Mandatory version, out-of-band from the evidence blob
Move the version out of the JSON payload and make it a first-class sibling of the
evidence, filled by the attester directly. The `TeeEvidence.version` field is
populated from that out-of-band value instead of being peeked out of `data`.

**phase 3**: Bump `kbs_protocol` to guarantee the version
Bumping it converts the version from an optional convention into a **mandatory,
typed, negotiated** field

mkulke:
> The plan is to have a three-phase approach. Please confirm if this aligns with your expectations:
> 
> **phase 1**: Explicit version on `TeeEvidence` (this PR)
> 
> **phase 2**: Mandatory version, out-of-band from the evidence blob Move the version out of the JSON payload and make it a first-class sibling of the evidence, filled by the attester directly. The `TeeEvidence.version` field is populated from that out-of-band value instead of being peeked out of `data`.
> 
> **phase 3**: Bump `kbs_protocol` to guarantee the version Bumping it converts the version from an optional convention into a **mandatory, typed, negotiated** field

sounds good, albeit I would do phase 2 and 3 in one go.

deps/verifier/src/lib.rs
```diff
@@ -214,9 +214,119 @@ pub async fn to_verifier(
...
+    /// The evidence format version, read from the top-level `version` field at
+    /// parse time. Legacy evidence without an explicit version is `0`.
+    pub version: u8,
+    /// The raw TEE-specific evidence payload, verbatim.
+    pub data: serde_json::Value,
+}
+
+/// Read a top-level `version` value as a `u8`. Returns `Ok(None)` when absent,
+/// `Err` when present but not representable as a `u8`.
+fn parse_version_field(value: &serde_json::Value) -> std::result::Result<Option<u8>, String> {
```
mkulke: we're using anyhow Results elsewhere in this file, any reason why we don't use them here?
```suggestion
fn parse_version_field(value: &serde_json::Value) -> Result<Option<u8>> {
```

deps/verifier/src/lib.rs
```diff
@@ -214,9 +214,119 @@ pub async fn to_verifier(
...
+    match value.get("version") {
+        None => std::result::Result::Ok(None),
+        Some(v) => v
+            .as_u64()
+            .and_then(|n| u8::try_from(n).ok())
+            .map(Some)
+            .ok_or_else(|| format!("invalid evidence version: {v}")),
+    }
```
mkulke: this is a bit type-golfy, would this be more more readable if the chained expressions were split?


```suggestion
    let Some(v) = value.get("version") else {
        return Ok(None);
    };
    
    let Some(version) = v.as_u64().and_then(|n| u8::try_from(n).ok()) else {
        return Err(format!("invalid evidence version: {v}"));
    };
    
    Ok(Some(version))
```

deps/verifier/src/lib.rs
```diff
@@ -214,9 +214,119 @@ pub async fn to_verifier(
...
+    verifier: &dyn Verifier,
+    evidence: &TeeEvidence,
+    tee: &Tee,
+) -> Result<()> {
+    let max = verifier.max_supported_version();
+    let version = evidence.version;
+    if version > max {
+        bail!(
+            "Unsupported {tee:?} evidence version {version}. This verifier \
+             supports evidence versions 0-{max}. Evidence version {version} may \
```
mkulke: I'd make the error message more succinct, so they will read nice in a logfile. We can drop everything but the first 2 sentences IMO.

mkulke (commented):
with this check we will reject any evidence that introduces a top-level `version: "v1"` or  something in their evidence payload (maybe by bumping a TEE-related crate), Ironically this kind of breakage we wanted to avoid, by the version tag. but for this to be solid, we need a kbs_protocol bump

mkulke:
needs a rebase

mkulke (approved):
lgtm, I think this is a fair compromise. I hope the unit test cover the current evidence shapes of the all TEEs properly, so we don't introduce breakage.

Xynnn007 (approved):
LGTM. Thanks @lmilleri 

I am not quite sure if we need documents about the breakingchanges for the evidence format to track what was changed. Anyway it can be a future work

fitzthum:
Previously, we've tried to handle evidence compatibility at the evidence level (i.e. in the verifier). Here you are doing it at a higher level. I am not sure what the right approach is.

I guess it seems a little more general to do it this way, but really this PR does not provide any compatibility, it just flags if a version is not within some range. The actual compatibility logic would still need to be implemented in the verifier. How does the version field you are introducing here interact with a compatibility layer in the verifier?

Interestingly, your approach kinda makes things less compatible in the short term, since it requires us to keep track of a version, which we don't currently do. Of course, it might seem impossible to have backwards compatibility if most of our evidence structs don't even report a version today, but that's not the case. We can provide backwards compatibility from today onwards by simply requiring all verifiers to implement a compatibility layer the next time that they change the evidence struct. This layer can resolve unversioned evidence to the current layout. In other words, we don't necessarily need to introduce a version until we make the first change we want to be compatible with.

What it boils down to is why not just implement this at the verifier level, since we will need to make changes to the verifiers anyway, and since doing it at a higher level seems to introduce a breaking change across the whole protocol.

Xynnn007:
The current change is not protocol level also and no breaking change. Just a alias deserialization magic to bring `version` field from the evidence if any, while keeping the whole original evidence in `.data` field. Originally the verifier directly handle the whole `evidence`, but now it handles `evidence.data` which is the same as previous `evidence`, with extra `evidence.version` as a hint.

This 
1. Defines a contract for all TEE evidences. If more version of a same TEE evidence comes, an explicit `version` field should be inside the evidence structure to distinguish. 
2. Would help to reduce duplicated code. In future if different TEE implement new versions while keeping backward compability, they would all have deserialization and branching logic for `version` field.
3. if no `version`, a default `0` will be given and current verifier do the same as before. Thus no "since it requires us to keep track of a version" - the design is lazy, meaning that we can bring in more version if evidence actually changes.

Surely, if we use attester/verifier local ways to admin the versions, they would have their own ways to mark the version.

I guess both way work. From my side, this one brings out a contract (`version`) for future direction, which would help to have a common sense for different TEEs and easier to maintain.

mkulke:
I think @fitzthum has a point. Concretely this merely encodes a convention but doesn't change anything unless both attesters/verifier tuples adapt this, until then it's mostly dead code, that in the worst case could be a nuisance, because maybe some evidence wants to put a version prop that doesn't map to an uint in their evidence. So, it's probably not harmful, but the benefit over attesters/verifiers implementing a private versioning contract is arguable.

To really make versioning reliable and predictable, we need to change the protocol, and I would decouple this from an ambitious 1.0 conversation, I think we can iterate more or on the protocol before that.

fitzthum:
Yeah, this PR in itself isn't too drastic, but the next steps are a little unclear to me. In particular, how will this interact with the actual compatibility logic in the verifiers that we will eventually need to add.

For instance, we already have the compat code for the az verifiers. It has its own versioning scheme for this. Are we going to try to mirror those version numbers using this new abstraction? That seems kind of confusing.

If you think about it, we also have some compatibility code in the SNP verifier, but it works in a totally different way. There, we don't have compatibility logic for parsing the evidence struct, but we do already support different report versions, which are the main component of the evidence struct. Should we try to capture that in this version number? How?

My feeling is that maybe evidence compatibility should be handled at the verifier level. In fact, it seems inevitable that it will be handled by the verifier. Where else would the compatibility code live? In that case, why would we introduce a versioning scheme at a different layer? I guess it might be nice to have a more standardized and explicit versioning scheme, but if we're going to go that way, I think we need a clearer picture of how it will work.

mythi:
> Concretely this merely encodes a convention but doesn't change anything unless both attesters/verifier tuples adapt this, until then it's mostly dead code,

FWIW, this is what I was also commenting in the Trustee call on Monday. For the overall (KBS protocol) backwards compatiblity problem, we could prioritize different topics at least until we have an evidence format change proposed.

PR #1600, opened 2026-08-28
rvps: add DeleteReferenceValue RPC and rvps-tool delete command

AnkushUjawane:
RVPS supported registering and querying reference values, and (as of a previous PR) listing them, but had no way to remove one. A stale or mistakenly registered reference value could only be overwritten via register, never actually deleted.

Adds a DeleteReferenceValue gRPC RPC, backed by the existing KeyValueStorage::delete() the storage backend already provides (same mechanism KBS's resource plugin already uses for delete), and wires it through to a new 'rvps-tool delete' subcommand. Returns whether something was actually deleted, distinguishing 'deleted' from 'nothing to delete' rather than treating both as success uniformly.

attestation-service also implements ReferenceValueProviderService for its built-in and remote-gRPC RVPS modes; updated RvpsApi and both implementors (builtin, grpc) so this doesn't break AS's build, having learned that lesson from the ListReferenceValues PR.

Manually verified end-to-end: registered a reference value, listed it, deleted it, confirmed it no longer appears in the list, and confirmed deleting a nonexistent key reports 'not found' rather than erroring.

### DIFF
--- a/attestation-service/src/bin/grpc/mod.rs
+++ b/attestation-service/src/bin/grpc/mod.rs
@@ -35,8 +35,9 @@ use crate::rvps_api::{
     reference_value_provider_service_server::{
         ReferenceValueProviderService, ReferenceValueProviderServiceServer,
     },
-    ReferenceValueListRequest, ReferenceValueListResponse, ReferenceValueQueryRequest,
-    ReferenceValueQueryResponse, ReferenceValueRegisterRequest, ReferenceValueRegisterResponse,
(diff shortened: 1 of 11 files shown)


fitzthum (approved):
LGTM 

I do have some ideas for a more sophisticated flow for revoking RVs where we actually keep the RV storage but mark it as revoked (so we have a better record of what happened), but this seems fine for now.

Xynnn007 (approved):

PR #1601, opened 2026-08-29
kbs: make KV resource deletion idempotent

noeljackson:
Deleting an already-absent resource from the KV backend currently returns an error. Repeated deletes are normal during retryable cleanup, so absence should already satisfy the request.

This makes KV deletion idempotent and adds regression coverage for two consecutive deletes.

Tested with:
- cargo test -p kbs --no-default-features --test resource_kv_storage

### DIFF
--- a/kbs/src/plugins/implementations/resource/kv_storage.rs
+++ b/kbs/src/plugins/implementations/resource/kv_storage.rs
@@ -46,7 +46,10 @@ impl StorageBackend for KvStorage {
 
         let deleted = self.storage.delete(&ref_resource_path).await?;
         if deleted.is_none() {
-            bail!("resource not found: {}", ref_resource_path);
+            tracing::warn!(
+                resource = %ref_resource_path,
+                "resource was already absent during deletion"
+            );
         }
 
         Ok(())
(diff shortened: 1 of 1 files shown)


kbs/tests/resource_kv_storage.rs
fitzthum, line 1: Don't think we need a new file for this

fitzthum (commented):
I guess REST would prefer this to be idempotent. Might be good to log something tho.

kbs/src/plugins/implementations/resource/kv_storage.rs
```diff
@@ -44,10 +44,7 @@ impl StorageBackend for KvStorage {
             resource_desc.repository_name, resource_desc.resource_type, resource_desc.resource_tag
         );
 
-        let deleted = self.storage.delete(&ref_resource_path).await?;
-        if deleted.is_none() {
-            bail!("resource not found: {}", ref_resource_path);
-        }
+        self.storage.delete(&ref_resource_path).await?;
```
Xynnn007: let's at least raise a `warn!` here if not found

fitzthum (approved):
LGTM

kbs/tests/resource_kv_storage.rs
noeljackson: Resolved in `f4c23a6`: I moved the idempotency regression into the existing `kv_storage.rs` unit-test module. The focused test and the combined KBS library suite pass.

kbs/src/plugins/implementations/resource/kv_storage.rs
noeljackson: Resolved in `f4c23a6`: deleting an already-absent resource remains idempotent and now emits a structured warning containing the resource path. The focused regression passes.

Xynnn007 (approved):

PR #1602, opened 2026-08-29
kbs: match regex ACLs against request paths

noeljackson:
Regex ACL rules describe /kbs/... paths, but the authorizer currently compares them with the whole URI. An absolute-form URI therefore misses an otherwise valid rule.

This matches on request.uri().path() and adds regression coverage for an absolute URI.

Tested with:
- cargo test -p kbs --no-default-features --test admin_regex_acl

### DIFF
--- a/kbs/src/admin/authorization/regex_acl.rs
+++ b/kbs/src/admin/authorization/regex_acl.rs
@@ -63,7 +63,7 @@ impl AuthorizationTrait for RegexAclAuthorizer {
                 continue;
             }
 
-            if acl.regex.is_match(&request.uri().to_string()) {
+            if acl.regex.is_match(request.uri().path()) {
                 return Ok(AuthorizationDecision {
                     allowed: true,
                     reason: "Subject allowed".to_string(),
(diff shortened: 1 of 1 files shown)


kbs/src/admin/authorization/regex_acl.rs
```diff
@@ -63,7 +63,12 @@ impl AuthorizationTrait for RegexAclAuthorizer {
                 continue;
             }
 
-            if acl.regex.is_match(&request.uri().to_string()) {
+            // ACLs are configured and validated as absolute request paths
+            // beginning with `/kbs`. HTTP/2 may preserve an absolute-form URI
+            // on the server request, so matching the URI string would reject
+            // an otherwise identical request solely because it includes the
+            // scheme and authority. Match only the path contract.
```
fitzthum: Please make human comment.

kbs/tests/admin_regex_acl.rs
fitzthum: Probably don't need a new file for this.

kbs/tests/admin_regex_acl.rs
```diff
@@ -0,0 +1,39 @@
...
+        .authorize(
+            Claims {
+                role: "admin".to_string(),
+            },
+            &request,
+        )
+        .expect("authorization decision");
+
+    assert!(decision.allowed);
+}
```
Xynnn007: could we move this as a unit test inside `regex_acl.rs`?

kbs/src/admin/authorization/regex_acl.rs
Xynnn007: We can move the comments into commit message, and just leave the code as-is.

kbs/tests/admin_regex_acl.rs
noeljackson: Resolved in `1e8cb96`: I moved the absolute-form URI regression into the unit-test module in `regex_acl.rs`; the focused test and combined KBS library suite pass.

kbs/src/admin/authorization/regex_acl.rs
noeljackson: Resolved in `1e8cb96`: I removed the implementation comment and moved the contract rationale into the commit message.

Xynnn007 (approved):
LGTM. Thanks for this!

PR #1614, opened 2026-09-11
Update EAR token format to 04 Version

### LINKED ISSUE #1584: AS: upgrade EAR attestation token to the published EAR claim set
### Summary

Now EAR Draft has published 04 version with some good new features. I am working on upstream to make it rust-ify and is near to finished. Before applying to this repo, this issue gives an overview for the influences of the updation.

### Influence

Today AS puts parsed evidence under the Veraison extension `ear.veraison.annotated-evidence`, and policy-authored fields mostly under custom extensions. New `rust-ear` introduce first-class:

- `ear_attester_claims` — Attester-authority claims extracted from Evidence
- `ear_verifier_claims` — Verifier-authority claims from appraisal policy

That is a cleaner authority split and removes dependence on the annotated-evidence extension for our main payload.

It's going to split today's flat `ear.veraison.annotated-evidence` by authority. Keep raw evidence extracts in attester claims; put Verifier-validated / policy-authored material in verifier claims.

| Today (under annotated-evidence) | After |
| --- | --- |
| `<tee>` object (hardware / TCB parse) | `ear_attester_claims.<tee>` |
| `report_data` (hex/base64 string from evidence) | `ear_attester_claims.report_data` |
| `init_data` (hash string from evidence) | `ear_attester_claims.init_data` |
| `runtime_data_claims` (optional parsed JSON after report-data check) | `ear_verifier_claims.report_data` (omit if absent) |
| `init_data_claims` (optional parsed JSON after init-data check) | `ear_verifier_claims.init_data` (omit if absent) |
| Policy-custom fields (e.g. identifiers) | `ear_verifier_claims.custom` |

Sketch:

```json
"ear_attester_claims": {
  "tdx": { },
  "report_data": "<raw>",
  "init_data": "<raw>"
},
"ear_verifier_claims": {
  "report_data": { },
  "init_data": { },
  "custom": { }
}
```

### Breaking changes

Some claim names are renamed or removed; KBS / AS policies and other token consumers need to be updated accordingly.

Xynnn007:
Close #1584 

This is an updation of EAR token format implementation and also the consumer logic on kbs side.

After this PR in each appraisal we will have claims like
```json
"ear_attester_claims": {
  "tee": "tdx",
  "claims": { },
  "report_data": "<raw>",
  "init_data": "<raw>"
},
"ear_verifier_claims": {
  "report_data": { },
  "init_data": { },
}
```

A previous EAR
```json
{
  "eat_profile": "tag:github.com,2024:confidential-containers/Trustee",
  "iat": 1789109810,
  "exp": 1789110110,
  "ear.verifier-id": {
    "developer": "https://confidentialcontainers.org",
    "build": "attestation-service 0.1.0"
  },
  "submods": {
    "cpu0": {
      "ear.status": "contraindicated",
      "ear.trustworthiness-vector": {
        "instance-identity": 0, "configuration": 2, "executables": 33,
        "file-system": 0, "hardware": 97, "runtime-opaque": 0,
        "storage-opaque": 0, "sourced-data": 0
      },
      "ear.appraisal-policy-id": "default",
      "ear.trustee.identifiers": {
        "validated": {
          "container_images": ["bitnami/nginx:latest"],
          "container_uids": []
        }
      },
      "ear.veraison.annotated-evidence": {
        "init_data": "DezN7Q2H7ERwLHMw+sPrk2to3Mpcn2td8fs0gMC1Jaw=",
        "init_data_claims": {
          "agent_policy_claims": { "common": {}, "containers": ["bitnami/nginx:latest"], "request_defaults": {}, "sandbox": {} }
        },
        "report_data": "phJI3dPyGl2CydINNZWY5c5bOxyX9HPf8hmqCY9iny2KA8FB8Ln6Y1vt4bZttKYC",
        "runtime_data_claims": {
          "additional-evidence": "",
          "nonce": "cpv9Mow93KqTHLLMroh8R7X6xlSPxDmWVdJg1Qwh+BA=",
          "tee-pubkey": {
            "kty": "EC", "crv": "P-256", "alg": "ECDH-ES+A256KW",
            "x": "k01-fqjVb-bqtk9dznq7MMwMzmyVvw9DtlmTXk19MuI",
            "y": "6pBGBekzPZ-52EACJUdsqzhUEe3nVgMtomuIXpIlWWQ"
          }
        },
        "sample": {
          "debug": false, "launch_digest": "abcde", "svn": "1",
          "platform_version": { "major": 1, "minor": 4 }
        }
      }
    }
  }
}
```

and a new one
```json
{
  "eat_profile": "tag:github.com,2024:confidential-containers/Trustee",
  "iat": 1789109692,
  "exp": 1789109992,
  "ear_status": "contraindicated",
  "ear_verifier_id": {
    "developer": "https://confidentialcontainers.org",
    "build": "attestation-service 0.1.0"
  },
  "submods": {
    "cpu0": {
      "ear_status": "contraindicated",
      "ear_trustworthiness_vector": {
        "instance-identity": 0, "configuration": 2, "executables": 33,
        "file-system": 0, "hardware": 97, "runtime-opaque": 0,
        "storage-opaque": 0, "sourced-data": 0
      },
      "ear_appraisal_policy_ids": ["default"],
      "ear_attester_claims": {
        "init_data": "DezN7Q2H7ERwLHMw+sPrk2to3Mpcn2td8fs0gMC1Jaw=",
        "report_data": "UQEgam5r5JUnmKHXAlgiRvOcGEBcD+f0F9TP9aOuJQqF9assOQqc8rjWQ5RUaNRF",
        "tee": "sample",
        "claims": {
          "debug": false, "launch_digest": "abcde", "svn": "1",
          "platform_version": { "major": 1, "minor": 4 }
        }
      },
      "ear_verifier_claims": {
        "init_data": {
          "agent_policy_claims": { "common": {}, "containers": ["bitnami/nginx:latest"], "request_defaults": {}, "sandbox": {} }
        },
        "runtime_data": {
          "additional-evidence": "",
          "nonce": "CoE7ro/ze2tTD9oKWR9RA0a8FOiaPwFvC0dSsg/9nLE=",
          "tee-pubkey": {
            "kty": "EC", "crv": "P-256", "alg": "ECDH-ES+A256KW",
            "x": "_96Cp-fF8xuKdli6rt6HeKootgi_l93NoBo6NVlbbS0",
            "y": "D54qqxF8CLoGalrnemVOwKd7p3-Mv5ihBpcBZZfu2-A"
          }
        }
      },
      "ear.trustee.identifiers": {
        "validated": {
          "container_images": ["bitnami/nginx:latest"],
          "container_uids": []
        }
      }
    }
  }
}
```


fitzthum:
Adopting verifier vs attester claims looks fine, but it also looks like you are moving identifiers from being an extension to being part of some "custom claims," which I do not see in the spec. Please move that into a separate PR. It is a different discussion (and one that I am unlikely to approve tbh).

Xynnn007:
Sure. Let me just put the previous part and leave out the identifier part changes.


This is [a danling issue](https://github.com/confidential-containers/trustee/pull/1079#discussion_r2567821754) that we still not have a clear answer. To re-explaine, two design ways here
1. The policy could define extensions. This is what we have in code for now.
2. To have a more structured place to put all defined claims. During time of https://github.com/confidential-containers/trustee/pull/1079, ear token does not have a standard `verifier_claims` field, thus I proposed to make a new extension. But now we have it, then we can embed it.

The pros and cons still hold.
- For 1: the flexibility. it helps rego policy level to define any extensions of ear token. At the same time, the consumer (like KBS) should have an out-of-bind contract of the extension names to know what extension to be consumed. Another point is about the implementation now that we are defining a [key](https://github.com/confidential-containers/trustee/blob/main/attestation-service/src/ear_token/ear_default_policy_cpu.rego#L357) field explicitly in the policy. Actually it's used for CBOR encoding while now we only uses JSON ear token. This is somehow of no use to users but provide extra considerations about "does this number have conflict".
- For 2: packed semantics. Now all inputs are from attestation evidence (hardware evidence, eventlog, initdata, runtime data, ..). Therefore, the derived content is not actually sufficient to fit the semantics of an extension (usually issure time like controlling fields); it is more appropriate to place it under `verifier_claims` thus natural contract for consumers. At the same time, we do not have flexibility to add extensions.

If we still choose to go down the road of way 1 and see what will happen, we should resolve the contract engineering issue. One way from my side is to have a common prefix like `ear.trustee.*` or `trustee.*` and have a profile document to explicitly record this.

Xynnn007:
Updated description of the PR.

fitzthum:
Ok, we can get deeper into the extension discussion elsewhere, but my basic feeling is that we should use extensions. Fundamentally, the thing we are trying to do (add extra info to the token), seems to be exactly what extensions are intended for.

There are a few small wrinkles. First, as you mention the scope of the extensions seems to clash a little bit with the new attester/verifier claims. Maybe it would have made more sense to be able to register extensions as attester or verifier claims, but I don't think this is a big issue. 

Second, adding extensions dynamically is a little bit at odds with the profile feature of the token. We've never really used this feature anyway.

I don't think either of these issues is significant, and introducing our own, unstructured field seems to be going the wrong direction.

deps/policy-engine/test_data/policy_1.rego
```diff
@@ -5,5 +5,5 @@ default allow = false
 allow if {
     count(data["resource-path"]) == 3
     data.plugin == "resource"
-    input["submods"]["cpu0"]["ear.veraison.annotated-evidence"]["sample"]["productId"] == data["resource-path"][1]
+    input["submods"]["cpu0"]["ear_attester_claims"]["claims"]["productId"] == data["resource-path"][1]
```
fitzthum, line 9: These aren't equivalent. The old policy requires that the sample attester is used. Maybe fine to change, but worth flagging. Maybe we should add an explicit type check. Up to you.

deps/policy-engine/test_data/policy_5.rego
```diff
@@ -11,14 +11,14 @@ resources := {"secret1": 2, "secret2": 3}
...
+    input["submods"]["cpu0"]["ear_attester_claims"]["tee"] == "sample"
 
     # check repository_name and resource_type
     data.plugin == "resource"
 
     data["resource-path"][0] == "myrepo"
     data["resource-path"][1] == "secret"
     # check that the secret name exists and that the minimum svn is met
-    resources[data["resource-path"][2]] <= input["submods"]["cpu0"]["ear.veraison.annotated-evidence"]["sample"]["svn"]
+    resources[data["resource-path"][2]] <= input["submods"]["cpu0"]["ear_attester_claims"]["claims"]["svn"]
```
fitzthum, line 22: Similar to above, but it's worth noting that now the `claims` field can have totally different content depending on the attester. Is that field part of the spec? In some ways it might be more clear to have the Tee still be the key name here (or maybe claims->{tee}). Not sure what is optimal.

kbs/src/token/mod.rs
```diff
@@ -16,7 +16,7 @@ pub const TOKEN_TEE_PUBKEY_PATH_ITA: &str = "/tdx/attester_runtime_data/tee-pubk
 pub const TOKEN_TEE_PUBKEY_PATH_ITA_VTPM: &str = "/tdx/attester_user_data/tee-pubkey";
 pub const TOKEN_TEE_PUBKEY_PATH_COCO: &str = "/customized_claims/runtime_data/tee-pubkey";
 pub const TOKEN_TEE_PUBKEY_PATH_EAR: &str =
-    "/submods/cpu0/ear.veraison.annotated-evidence/runtime_data_claims/tee-pubkey";
+    "/submods/cpu0/ear_verifier_claims/runtime_data/tee-pubkey";
```
fitzthum, line 19: Maybe we should keep the original entry here for backwards compatibility. Tokens are short-lived, so it's not a big deal, but can't hurt.

fitzthum (commented):
PR looks pretty good. A couple of comments. 

Biggest thing is that this will be a large breaking change in terms of the policy, so we'll need to communicate this clearly (and update the website).

deps/policy-engine/test_data/policy_1.rego
Xynnn007: Ah yes. I should have added the TEE assertion here but with negligence.

deps/policy-engine/test_data/policy_5.rego
Xynnn007: For this case we already have a TEE assertion in the policy in line 14.

---

if `tee` out, users can distinguish/know the Tee and claims more straightforward. Corner case is that users forget to care about Tee type but directly use claims.
if `tee` nested, in a way users must know that they are targeting a specific TEE thus caring about the body claims.

I lean more towards bring `tee` out.

kbs/src/token/mod.rs
Xynnn007: tbo I think we should have an code-defined extension to take on the TEE public key, thus having a contract that is not easy to change. At the same time, keep `"/submods/cpu0/ear_verifier_claims/runtime_data/tee-pubkey"` as it comes from the runtime data of the TEE.

wdyt?

Xynnn007:
Updated codes.

---

Yeah. We should have a separate issue for the unclear parts. From my point of view, what's needed to discuss is 

- whether policy defined rules should be mapped to `attester/verifier` claims, or the extensions.

Does this one makes sense? is this one complete?

> Second, adding extensions dynamically is a little bit at odds with the profile feature of the token. We've never really used this feature anyway.

I do not quite get the point. what do you mean here?

deps/policy-engine/test_data/policy_5.rego
fitzthum: Ok, either way is fine with me.

kbs/src/token/mod.rs
fitzthum: Yeah, an extension could the tee public key could be nice. I think there is some elegance to using the runtime-data, since this is where the information is actually captured, but I am open to adding an extension. In any case, we should probably still include the old paths here at least for a release or two.

fitzthum:
We should continue to generate extensions from the policy imo. This is more powerful and more standardized than introducing our own custom claims abstraction. These policy-defined extensions are all optional.

Regarding profiles, the EAR crate uses these to describe the token for deserialization. When we add extensions, they should technically go into this profile. When we set extensions via the policy, we can't publish one standardized profile. Since we validate the token as a JWT and then parse it as JSON, we never use any profile. Maybe in the longterm we should create one, but I don't think it matters for now.

kbs/src/token/mod.rs
Xynnn007: oh well. now i got the point that you just want kbs consumer side to keep this rather than producer side.

Makes sense.

PR #1615, opened 2026-09-14
verifier/se: Firmware Attestation support for coco

annu0905:
**IBM SE Firmware Attestation (z17)**

This PR extends the IBM Secure Execution attestation flow to support firmware integrity verification on z17 systems.

1. **Challenge Generation**
When the AS receives an RCAR request, it parses the machine type from tee_parameters. For z17, it sets the firmware state flag (bit 5) in AttestationFlags. This instructs the Ultravisor to include a 320-byte firmware hash in the attestation evidence. For z16 and older, the flag is not set and the existing flow is unchanged.

2. **Evidence Verification**
The verifier detects firmware hash presence by checking whether additional_data has the exact expected size (64 bytes for phkh fields + 320 bytes for firmware hash = 384 bytes). If present, the firmware hash is extracted.
Firmware verification against the IBM API is opt-in: if enable_firmware_verification = true is configured but no firmware hash is present in the evidence, attestation is rejected.

The expected outcome of the firmware attestation is as follows:

> z17+firmware_hash present + enable_firmware_verification=true  → Firmware Verification, Attestation success
> z17+firmware_hash absent  + enable_firmware_verification=true  → Attestation failure
> z17+firmware_hash present + enable_firmware_verification=false → No Firmware Verification, Attestation success
> z16/none+firmware_hash absent  + enable_firmware_verification=true  → No Firmware Verification, Attestation success
> z16/none+firmware_hash absent  + enable_firmware_verification=false → No Firmware Verification, Attestation success

**Configuration**
Two new fields in kbs.toml under [attestation_service.se_verifier]:
enable_firmware_verification = false   # opt-in, default disabled
firmware_verify_url = "https://esupport.ibm.com/eccedge/ent/z/hmrs/firmware/attestation/v1/verify"

No changes to the attestation flow for z16 or other architectures.

E2E Test Results for Rust and Go runtimes:
https://github.com/kata-containers/kata-containers/actions/runs/34808071688/job/103927121651
https://github.com/kata-containers/kata-containers/actions/runs/34808071688/job/103916872324

Signed-off-by: Annu Sharma annu-sharma@ibm.com
Assisted-by: IBM-bob


deps/verifier/src/se/ibmse.rs
```diff
@@ -211,10 +241,110 @@ impl SeVerifierImpl {
...
+            payload: hash_b64.clone(),
+        };
+
+        // Return immediately if this hash was already verified successfully.
+        if self.firmware_cache.read().await.contains(&hash_b64) {
+            debug!("Firmware hash found in cache, skipping API call");
+            return Ok(());
+        }
+
+        let client = reqwest::Client::builder()
```
mkulke, line 286: the client could be created and re-used in the verifier constructor.

deps/verifier/src/se/ibmse.rs
```diff
@@ -211,10 +241,110 @@ impl SeVerifierImpl {
...
+            .user_agent("s390-tools-pvattest")
+            .build()
+            .context("Failed to create firmware verification HTTP client")?;
+
+        for attempt in 1..=MAX_RETRIES {
+            debug!("POST firmware verification to {url} (attempt {attempt}/{MAX_RETRIES})");
+
+            let resp = client
+                .post(url)
+                .header("x-client-id", "X")
```
mkulke, line 297: what is "X"? if it's required value let's make this a const.

deps/verifier/src/se/ibmse.rs
```diff
@@ -211,10 +241,110 @@ impl SeVerifierImpl {
...
+            return Ok(());
+        }
+
+        let client = reqwest::Client::builder()
+            .timeout(FIRMWARE_VERIFY_TIMEOUT)
+            .user_agent("s390-tools-pvattest")
+            .build()
+            .context("Failed to create firmware verification HTTP client")?;
+
+        for attempt in 1..=MAX_RETRIES {
```
mkulke, line 292: I think reqwest has retry functionality that uses backoff, parses Retry-After headers etc. if possible we should use it, instead of handrolling a retry loop.

deps/verifier/src/se/ibmse.rs
```diff
@@ -211,10 +241,110 @@ impl SeVerifierImpl {
...
+
+        let url = &self.config.firmware_verify_url;
+        let hash_b64 = BASE64.encode(firmware_hash);
+        let body = FwRequest {
+            version: "1.0".to_string(),
+            payload: hash_b64.clone(),
+        };
+
+        // Return immediately if this hash was already verified successfully.
+        if self.firmware_cache.read().await.contains(&hash_b64) {
```
mkulke, line 281: should there be a limit, revocation list or a TTL for that cache? we also have cache primitives in trustee for this purpose, but they are probably overkill.

deps/verifier/src/se/ibmse.rs
```diff
@@ -284,6 +415,51 @@ impl SeVerifierImpl {
             .attestation_public_host_key_hash()
             .ok_or(SeError::MissingAttestationPublicHostKeyHash)?;
 
+        let firmware_hash: Option<Vec<u8>> = match add_data.firmware_state() {
+            Some(fw_slice) if fw_slice.len() == FIRMWARE_STATE_SIZE => {
+                debug!("Firmware hash present: {} bytes from UV", fw_slice.len());
+                Some(fw_slice.to_vec())
+            }
+            Some(fw_slice) => {
+                warn!(
```
mkulke, line 424: malformed data should not produce a warning but an error, IMO.

mkulke (commented):
it would make sense to scope the PR name to "verifier/se:"

attestation-service/docs/config.md
```diff
@@ -338,6 +360,27 @@ Configuration with AMD SEV-SNP verifier using KDS:
...
+    },
+    "verifier_config": {
+        "se_verifier": {
+            "enable_firmware_verification": true,
+            "firmware_verify_url": "https://esupport.ibm.com/eccedge/ent/z/hmrs/firmware/attestation/v1/verify"
+        }
+    }
+}
+```
+Optional: override the firmware verification endpoint (firmware_verify_url) to your custom endpoint for verification.
```
Xynnn007, line 381:
```suggestion
Optional: override the firmware verification endpoint (`firmware_verify_url`) to your custom endpoint for verification.
```

deployment/helm-chart/templates/_helpers.tpl
```diff
@@ -190,6 +193,14 @@ AS verifier config JSON fragment
     "vcek_sources": {{ $snpVcekSources | toJson }}
 }
 {{- end -}}
+{{- if or $nv $dcap $snpVcekSources }},
```
Xynnn007, line 196: Maybe we should use
```suggestion
{{- if and $se (or $nv $dcap $snpVcekSources) }},
```

Or if we have any of `nvidia/dcap/snp`w/o `se`, it will be rendered with a dangling `,` as
```yaml
"nvidia_verifier": {
                "type": "Local"
            },
            "dcap_verifier": {
                "collateral_service": "https://api.trustedservices.intel.com/sgx/certification/v4/",
                "tcb_update_type": "early"
            }
            ,
```

deps/verifier/src/se/mod.rs
```diff
@@ -15,9 +16,44 @@ pub mod ibmse;
...
+
 #[derive(Debug, Default)]
 pub struct SeVerifier;
 
+impl SeVerifier {
+    pub fn new(config: Option<SeVerifierConfig>) -> Result<Self> {
+        // Populate the static VERIFIER with config at construction time, before any
+        // request arrives. get_or_try_init() in evaluate() will retrieve this value
+        // directly
+        let _ = VERIFIER.set(SeVerifierImpl::new(config)?);
```
Xynnn007, line 52: I wonder if we could get rid of the static variable here. This can be a future goal and not related to this PR.

deps/verifier/Cargo.toml
```diff
@@ -136,3 +136,4 @@ serial_test.workspace = true
 tempfile.workspace = true
 tokio.workspace = true
 tracing-subscriber.workspace = true
+wiremock = "0.6"
```
Xynnn007, line 139: looks like we do not import this dep anywhere?

deps/verifier/src/se/ibmse.rs
annu0905: Ah, reqwest retries only look at HTTP status codes (like 5xx or network drops). Because this API returns a 200 OK with the actual failure buried in the response body, reqwest sees it as a success.
That’s why we have to check the response payload manually and do the manual retries.

deps/verifier/Cargo.toml
annu0905: We are importing this dep in the unit test.

mythi:
what happens if you do `attestation_flags.set_firmware_state();` unconditionally?

annu0905:
> what happens if you do `attestation_flags.set_firmware_state();` unconditionally (in `generate_supplemental_challenge()`)?

On z16 and older machines, The Ultravisor rejects the attestation UVC because bit 5 is an unsupported flag on that hardware. The Ultravisor returns hardware error ("Unsupported plaintext attestation flag set").
The Attestation Agent on the guest fails to generate evidence, breaking attestation entirely for all z16 workloads.

PR #1620, opened 2026-09-19
kbs: bound remote fetches with timeouts; document rate limiting for /kbs/v0/auth

### LINKED ISSUE #1536: Rate-limit `/kbs/v0/auth` and add timeouts to remote key fetches
The `/kbs/v0/auth` endpoint is unauthenticated and currently has no throttling. It's not free to service (it generates a challenge), so a scripted client could hammer it and tie up server resources. Since auth is the entry point for every attestation flow, it'd be nice to throttle it per client IP.

Separately, when KBS verifies attestation tokens using remote JWKS/PEM keys, the requests to those URLs (and the remote attestation call) have no HTTP timeout. An unreachable or unresponsive endpoint can hang the request for a long time.

Proposed:
- Add a configurable per-IP token-bucket rate limit on `/kbs/v0/auth`, returning HTTP 429 when exceeded (set to 0 to disable).
- Add a timeout to outbound HTTP requests for JWKS/PEM fetches and the remote attestation call.

david-long1:
## Why

Two ways for a KBS deployment to be tied up by a remote party, both raised in #1536:

- KBS fetches remote key material (JWK sets, OpenID discovery documents, PEM public keys) with bare `reqwest::get` calls and sends Intel Trust Authority attestation requests through a `reqwest::Client` built without a timeout. An endpoint that accepts the TCP connection and never answers hangs KBS startup, or a client's attestation, indefinitely.
- `POST /kbs/v0/auth` is unauthenticated by design (it opens the RCAR handshake) and every call allocates a session record and a challenge, so a single peer can fill the session store for as long as it likes.

Per the discussion on the issue, the first belongs in KBS and the second belongs in front of it: KBS only sees the address of whatever fronts it, so an in-process per-client limiter cannot tell clients apart behind a proxy and would put every client in one bucket, while the proxy sees the real client address and keeps its counters across replicas.

## How

Two commits, each standing on its own.

**`kbs: bound remote key fetches and ITA attest requests with timeouts`**

Every remote call gets a total request timeout, which in reqwest also bounds the connect phase. Key material fetches are small static documents made when a verifier is initialized, so they use a fixed 30s. ITA attestation requests use a fixed 60s, matching the existing `AS_REQUEST_TIMEOUT` shape on the gRPC AS client, since the server does real verification work. The JWKS loader builds one client and reuses it across the direct fetch and the OpenID discovery fallback. The ITA backend builds its client once in `IntelTrustAuthority::new` and keeps it on the struct instead of constructing a new one per request.

**`helm-chart: document per-client rate limiting in front of KBS`**

A README section explaining why the limit belongs on whatever fronts KBS and pointing at the chart's existing `ingress.annotations` passthrough. It names no controller: rate limiting is expressed by neither the Ingress API nor Gateway API, so each implementation configures it its own way, and the rest of this chart names no third-party controller either. No KBS code changes.

## Downsides

- The key-fetch timeout is a constant, not config: it only guards startup against a wedged endpoint. The ITA timeout is also a constant now, per review; making it configurable can follow if a deployment needs it.
- The rate-limit guidance is prose rather than a worked example, so an operator still has to consult their controller's documentation.

## Alternatives considered

- **A per-IP limiter inside KBS** (an earlier revision of this PR, via actix-governor). Dropped after the boundary discussion on #1536: behind a proxy it sees one address for all clients, and it duplicates what edge proxies already do with better information.
- **A worked example against one controller** (an earlier revision shipped a `scenarios/ingress-rate-limit.yaml` using ingress-nginx annotations). Dropped: ingress-nginx reached end of life in March 2026, and naming any single controller is out of step with a chart whose scenarios and values name none.
- **Reusing `attestation_service.timeout` for the ITA request.** That key already means the session TTL in minutes, so overloading it would be confusing.

## Testing

Timeouts commit, with the toolchain pinned by `rust-toolchain.toml` (rustc 1.95.0) and the CI feature set for `make test-kbs-unit`:

```
$ cargo fmt -p kbs -p kbs-client -- --check --config format_code_in_doc_comments=true
(clean)
$ cargo clippy -p kbs -p kbs-client --no-default-features --features coco-as-builtin,coco-as-grpc,intel-trust-authority-as -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 13s
$ cargo test -p kbs --no-default-features --features coco-as-builtin,coco-as-grpc,intel-trust-authority-as
test result: ok. 89 passed; 0 failed; 1 ignored
```

An earlier revision carried a test that bound a listener which accepted and never replied, and asserted an ITA request failed within the timeout. It went with the configurable knob, since it only stayed fast by setting that knob to 1s; happy to add a `cfg(test)` hook if reviewers want the coverage back.

Chart commit:

```
$ cd deployment/helm-chart
$ helm lint .
1 chart(s) linted, 0 chart(s) failed
$ helm-docs   # README.md regenerated from README.md.gotmpl; diff is the new section only
```

The rate-limiting behaviour itself was verified on an earlier revision of this PR, on a throwaway `kind` cluster (Kubernetes v1.36.1) with a controller in front of the chart's Ingress: 100 concurrent `POST /kbs/v0/auth` requests from one client were allowed up to the configured allowance and the rest rejected at the proxy, a single request after a short pause succeeded, and a second client was unaffected during the first client's burst. The rejected requests never reached KBS. That configuration is now left to the operator rather than shipped as a scenario.

Fixes: #1536


david-long1:
@Xynnn007 reshaped per the discussion on #1536: the timeouts commit is unchanged, and the in-process limiter is gone in favour of a chart scenario plus README section for Ingress-level rate limiting, verified end to end on a kind cluster (numbers in the PR description). Could you please take a look at the PR and let me know if there is anything else you'd like me to change, or if it looks good?

kbs/src/attestation/intel_trust_authority/mod.rs
```diff
@@ -449,6 +460,14 @@ impl Attest for IntelTrustAuthority {
...
 impl IntelTrustAuthority {
     pub async fn new(config: IntelTrustAuthorityConfig) -> Result<Self> {
+        Self::new_with_timeouts(config, ITA_CONNECT_TIMEOUT, ITA_ATTEST_TIMEOUT).await
+    }
+
+    async fn new_with_timeouts(
+        config: IntelTrustAuthorityConfig,
+        connect_timeout: Duration,
+        timeout: Duration,
+    ) -> Result<Self> {
```
Xynnn007: Maybe we can add the `timeout` to the `IntelTrustAuthorityConfig` with a default value, and also update configuration document. Also, due to https://docs.rs/reqwest/0.13.5/reqwest/struct.ClientBuilder.html#method.timeout and https://docs.rs/reqwest/0.13.5/reqwest/struct.ClientBuilder.html#method.connect_timeout, `timeout` includes `connect_timeout`? So maybe we can only have `timeout` alone.

kbs/src/crypto/jwt.rs
```diff
@@ -49,10 +52,23 @@ fn normalize_jwk_set_source(source: &str) -> Result<String> {
...
 pub(crate) async fn read_pem_public_key_from_uri(uri: &str) -> Result<DecodingKey> {
+    read_pem_public_key_from_uri_with_timeouts(uri, KEY_FETCH_CONNECT_TIMEOUT, KEY_FETCH_TIMEOUT)
+        .await
+}
+
+async fn read_pem_public_key_from_uri_with_timeouts(
+    uri: &str,
+    connect_timeout: Duration,
+    timeout: Duration,
+) -> Result<DecodingKey> {
```
Xynnn007: We do not need nested structure here. Also for other parts. The timeout here is only used during the initilization time, thus it's ok to directly refer to the const

deployment/helm-chart/scenarios/ingress-rate-limit.yaml
```diff
@@ -0,0 +1,41 @@
...
+#     --namespace coco-trustee --create-namespace \
+#     -f ./deployment/helm-chart/scenarios/ingress-rate-limit.yaml \
+#     --set ingress.host=kbs.example.com
+#
+# Combine with scenarios/valkey-sessions.yaml when running several KBS
+# replicas behind the same Ingress.
+
+ingress:
+  enabled: true
+  className: nginx
```
Xynnn007: Does this require to install anything ahead of time? Also the comments look a little drift-ed.

kbs/src/attestation/intel_trust_authority/mod.rs
david-long1: Done. Added `request_timeout` (seconds, default 60) to the ITA config + config.md, and dropped `connect_timeout`. Called it `request_timeout` because `[attestation_service]` already has a `timeout` (the RCAR window, in minutes) in the same table.

kbs/src/crypto/jwt.rs
david-long1: Done, removed the `_with_timeouts` layers everywhere and use the consts directly. The jwk/pem timeout tests went with them; the ITA one just sets `request_timeout: 1` now.

deployment/helm-chart/scenarios/ingress-rate-limit.yaml
david-long1: Yes, ingress-nginx. Added that to the scenario and the README, and trimmed the header down.

Xynnn007:
cc @mythi about Intel TA parts

david-long1:
The two CodeQL "cleartext transmission" hits are on the JWKS and PEM fetches that this PR refactors from `reqwest::get(uri)` to a shared client with timeouts; the scanner now sees the config-to-request flow it did not model before. Both call sites are inside the `"https"` arm of the scheme match, so the transport is always TLS, and the values are URLs of public key material. No change in what is sent or where. Happy to add a `// codeql[rust/cleartext-transmission]` suppression if the project prefers that over dismissing the alerts.

kbs/src/attestation/intel_trust_authority/mod.rs
mythi: What's the benefit of having it configurable? It's not configurable on AS side for verifiers that make external calls either.

david-long1: Fair question. I originally had this as a fixed 60s timeout, matching the NRAS verifier from #1398, and changed it to `request_timeout` based on Xynnn007's earlier review.

The main case I can see for making it configurable is that ITA is an external hosted service, so deployments with slower egress may want to raise the timeout without rebuilding KBS. On the other hand, keeping it fixed would be more consistent with the existing AS-side verifiers and avoids adding another config option.

I’m fine either way — it’s a small change. @Xynnn007, do you have a preference here?

Xynnn007: Either approach works for me with proper reasons, and not a blocker to me. For the time being, making the KBS backend connection timeout configurable does no harm or breakingchange. This timeout works for every `/attest` call to verify evidence. In some cases users want to set to a tighter or looser timeout. Note that the default value is defined, meaning that if no value is set explicitly, the default works. As for the AS, the timeout setting can be made configurable later on an as-needed basis. 


At the same time, if not adding, the timeout issue is still resolved with less flexiblity and easier code.

david-long1: Thanks, both. I'm fine with either direction. The branch currently has the configurable `request_timeout` with the 60s default; if that works for you, @mythi, I'll leave it as is, and if you'd rather have a fixed timeout I'll drop the option and the config.md row. Just let me know which way you'd like to go.

mythi: My preference is to drop it and later we can think how to handle the configuration across the stack. `AS_REQUEST_TIMEOUT` is not configurable either so dropping it makes this consistent.

david-long1: Dropped `request_timeout` and the manual `Default` impl. The ITA client now uses a fixed 60s `ITA_REQUEST_TIMEOUT`, matching the existing `AS_REQUEST_TIMEOUT` pattern.

I removed the timeout test as well, since it only stayed fast by setting the timeout to 1s. Happy to add a `cfg(test)` hook if you’d prefer to keep coverage there.

david-long1:
Thanks again for the reviews here. Unrelated to this PR, I've picked up cloud-api-adaptor #2604, the RTMR3 initdata thread @Xynnn007 you were discussing with @matdibu and Tobin in June, and posted a design there. Would appreciate your thoughts when you get a chance, especially on the AAEL side.

deployment/helm-chart/README.md
```diff
@@ -186,6 +186,28 @@ forward plaintext HTTP to KBS. Do not combine native KBS TLS with Ingress unless
...
+helm upgrade --install trustee ./deployment/helm-chart \
+  --namespace coco-trustee \
+  -f ./deployment/helm-chart/scenarios/ingress-rate-limit.yaml \
+  --set ingress.host=kbs.example.com
+```
+
+The scenario sets ingress-nginx's `limit-rps` and `limit-burst-multiplier`
+annotations through `ingress.annotations`, so an
+[ingress-nginx](https://kubernetes.github.io/ingress-nginx/) controller must
+already be installed; other controllers expose the same capability under their
```
mythi: `ingress-nginx` is EOL so I think it's not a good choice here. Can we come up with a better alternative?

mythi:
> Thanks again for the reviews here. Unrelated to this PR, I've picked up cloud-api-adaptor #2604, the RTMR3 initdata thread @Xynnn007 you were discussing with @matdibu and Tobin in June, and posted a design there. Would appreciate your thoughts when you get a chance, especially on the AAEL side.

Do you have a pointer to this topic?

matdibu:
> Do you have a pointer to this topic?

https://github.com/confidential-containers/cloud-api-adaptor/issues/2604

deployment/helm-chart/README.md
david-long1: @mythi Agreed, it was archived in March, so we shouldn't be pointing users at it as-is.

I checked the alternatives. F5's maintained NGINX Ingress Controller (`nginx/kubernetes-ingress`) is a separate project from the retired one and is a fairly close replacement for this example: its rate-limit settings are standard Ingress annotations, so it wouldn't require changing the chart shape.

For now I've taken the smaller route and dropped the controller-specific scenario, leaving the README guidance generic through the existing `ingress.annotations`. That also matches the rest of this chart, which doesn't name a controller.

Happy to switch this to a concrete F5 example instead if you'd prefer operators to have something that works out of the box.

david-long1:
One thing this PR no longer addresses now that rate limiting is documentation-only: KBS still has no bound on live sessions.

Each `/kbs/v0/auth` call stores a session, and both the number and size are caller-influenced. The periodic cleanup then scales with the store: on Valkey it lists the keyspace and fetches each value, and Postgres similarly walks the stored sessions.

@Xynnn007, since you mentioned "bounded concurrency or an optional local safeguard" in #1536, I see three possible shapes:

- a per-replica semaphore/counter — simple, but only bounds each KBS process;
- a storage-level atomic admission operation such as `try_create_session(..., limit)` — gives a real bound across replicas sharing Valkey/Postgres, but needs backend-specific work;
- a session-specific storage abstraction, if TTL/capacity/admission semantics don't belong in the generic key-value trait.

I'd avoid `list().len()`, since that repeats the scan we're trying to contain, and a plain `count()`, since it adds a round trip and leaves a check-then-insert race across replicas.

Would you prefer a per-replica safeguard or a hard bound on the shared session store? Happy to move this into its own issue if that's easier to track.

deployment/helm-chart/README.md
```diff
@@ -186,6 +186,19 @@ forward plaintext HTTP to KBS. Do not combine native KBS TLS with Ingress unless
 the chosen Ingress controller is explicitly configured to use HTTPS for its
 backend connection.
 
+### Per-client rate limiting
```
mythi, line 189: only one small nit: the problem described here is not Helm specific but applies to any deployment model, including trustee-operator. I wonder if we can find a place for the common problem statement.

mythi (commented):
198ef95b9d LGTM

deployment/helm-chart/README.md
david-long1: Good point. Moved the general problem statement to `kbs/docs/rate-limiting.md` in 7ea286d and linked it from the KBS README, so the Helm section now only covers the chart-specific setup and links there. Happy to move it if there's a better home.

